PR #32, opened 2016-11-29
README: use kubectl create -f http

philips:
Instead of getting curl involved just use kubectl


soamvasani:
Ha, I didn't know about this kubectl feature, thanks!

PR #36, opened 2016-11-29
Updated README to include protocol scheme for FISSION_URL prefix

efexen:
Hey 👋 

I was going through the README to test this out and noticed that if you don't prefix the FISSION_URL with http:// you get the following error:

```
Failed to create environment: Post xxx.xxx.xxx.xxx/v1/environments: unsupported protocol scheme ""
```

I've added a bit on the README to prompt for this.

Hope this is ok.

Thanks,
Ville

ps. Awesome project! I'll definitely be playing around with this more soon! The cold start response time over what the AWS API Gateway latency is makes this very feasible!

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ minikube, change all instances of LoadBalancer services to NodePort.
 ```
 
 Save the external IP addresses of controller and router services in
-FISSION_URL and FISSION_ROUTER, respectively.  FISSION_URL is used by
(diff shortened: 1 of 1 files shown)


soamvasani:
Thanks @efexen! Yeah, the change makes sense.  (though maybe I should fix the code to just accept an IP address too..)

> Awesome project! I'll definitely be playing around with this more soon! The cold start response time over what the AWS API Gateway latency is makes this very feasible!

Thanks! As you play around with Fission I'd love to hear about any feedback you have -- soamvasani@platform9.com.

Issue #25, opened 2016-11-11
Set up continuous testing


soamvasani:
Set up travis CI, or something like it, for testing.

soamvasani:
Let's keep it open till we actually set up travis -- @jgavinray is working on that.

soamvasani:
We can close this now that we have that badge.  I'll file another issue to improve test coverage.

PR #42, opened 2016-12-04
#25 Continuous Testing

jgavinray:
1. Added all 1.7 versions of go to test against
2. Added Travis badge for master branch
3. Travis Builds Activated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,13 @@ dist: trusty
 
 language: go
 go:
+    - 1.7.0
+    - 1.7.1
+    - 1.7.2
+    - 1.7.3
     - 1.7.4
+    - release
+    - tip
 
 install:
    - go get github.com/Masterminds/glide
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,10 @@ dist: trusty
 
 language: go
 go:
-    - 1.7.0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,7 +3,11 @@ dist: trusty
 
 language: go
 go:
+    - 1.7.1
```
soamvasani: Do we need these? Is it common to find problems with one minor version but not another?

README.md
```diff
@@ -1,6 +1,7 @@
 Fission: Serverless Functions for Kubernetes
 ============================================
 [fission.io](http://fission.io)  [@fissionio](http://twitter.com/fissionio)
+[![Build Status](https://travis-ci.org/platform9/fission.svg?branch=master)](https://travis-ci.org/platform9/fission)
```
soamvasani: Hmm, we expect a lot of the tests to fail right now: controller tests assume a running etcd on localhost; poolmgr will require a running kubernetes cluster to test.  I'd suggest we hold off on enabling the badge until we get some passing tests.

soamvasani:
Very cool to have this automated testing.  It's already finding bugs -- I'd neglected to update the cache test after changing the cache interface.  Thanks for taking on this work!

PR #43, opened 2016-12-04
Fix cache test

soamvasani:
cache.Set() returns existing value; update the test accordingly.


soamvasani:
^ cache test passed; others are expected to fail, right now.

PR #48, opened 2016-12-08
Setup app.logger for python environment

### LINKED ISSUE #47: Python flask environment should set up app.logger
The python environment should set app.logger, so that the user's function can simply call app.log() to log.  The logs should be set up to go to stdout, so that "kubectl logs" etc. can work.

soamvasani:
Fixes #47


kaustubhvp:
👍

PR #56, opened 2016-12-16
Kubewatcher: trigger functions from Kubernetes Watch callbacks

soamvasani:
This is the start of an implementation to get Kubernetes watches to trigger Fission functions.  It uses HTTP to call functions, so perhaps the Kubewatcher service can be made a more general part of K8s.

(Using Kubernetes watches/events to launch functions was a use case suggested by @kelseyhightower at Kubecon. Thanks!!)

## Contents of this PR:

`fission.Watch` object in the API; its implementation in server and CLI.  A `fission.Watch` contains a namespace, object type and label and field selectors; together these can be used to create a Kubernetes watch.  It also contains the metadata of the function to call.  Label and field selectors aren't implemented yet, and for now only Pods and Services are supported as object types.

A new `kubewatcher` service which:

1. Syncs with the controller API to find out what it needs to watch on the K8s API

2. Creates/destroys K8s watches 

3. Does a POST to the function's URL when the watch fires.  When a K8s watch callback fires, it gives us a Kubernetes `watch.Event`, containing an event type and an object.  We serialize the object into JSON (using the JSON serialization code from `kubectl`) and place the event type in a `X-Kubernetes-Watch-Event-Type` header.

There is currently no reliability implemented; we drop events if there's any kind of error from the endpoint.  

Separately, we also send the events serially, so if one event is processed slowly, they all slow down.  I'm not sure what the right approach is here; making events concurrent might be a confusing interface for people implementing the handlers.

## Further work:

1. Separate out the Kubewatcher service from Fission entirely; it could be potentially a good thing to add to K8s itself?  In other words some sort general Kube Watch -> webhook connector.

2.  Kubewatcher event reliability.  i.e. what happens when you get an event but the function had an error?  Or kubewatcher itself was down? There are several possible approaches here:
    a. No reliability, i.e. it's the user's problem.  Practically this would mean the user has to write some sort of poll on the objects they're watching to reconcile their view of the state.  This may be hard to get right for every webhook implementer.
    b. Kubewatcher persists (in etcd?) the `event.Object.Metadata.ResourceVersion` of the last successfully acknowledged watch event.  Assumes `resourceVersion` is monotonically increasing; it looks like it is but I haven't verified it.
    c. Kubewatcher takes events and sends them into real message queue: AMQP/NATS.io Streaming/Kafka/etc.  Some of these queues are a fair amount of operational complexity for the user, so their usage should be optional in some way.


soamvasani:
With the latest changes the publisher is an interface.  Currently there's only a webhook publisher satisfying this interface.  WebhookPublisher does best-effort delivery but drops events if it can't reach the endpoint.  #64 will track the work of using a message queue, probably nats-streaming.

#63 will track separating out the kubewatcher service from fission.

PR #77, opened 2017-01-13
fission-bundle: allow setting the namespace

frodenas:
Poolmgr allows setting the `namespace`in which to run function containers, but `fission-bundle` does not allow that parameter.


soamvasani:
Thanks for the PR, @frodenas !

PR #84, opened 2017-01-21
Added support for running C# code in a dotnet core environment

ktrance:
@soamvasani, as we discussed about over slack here is the dotnet environment implementation.

Be aware that the examples uses a container located at dockerhub under fission/dotnet-env which namespace I of course do not own, but I figured you guys might want build and add it there next to the other environments.

Please check README.dm for further description.


ktrance:
Any clue as to why the check fails? 

To me it looks gofmt (I guess this is a Go linter) fails because it also looks into glide's (package manager?) cache as well but Im not really that well travelled in Go.

soamvasani:
Thanks for the PR!  Excited to have C#/.NET support.  Taking a look now.

The travis CI failure is indeed unrelated to your change, I will file another issue to track that.

environments/dotnet/ExecutorModule.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        private object Run()
+        {
+            if (_userFunc == null)
+            {
+                var response = (Response)"Generic container: no requests supported";
+                response.StatusCode = HttpStatusCode.InternalServerError;
+                return response;
+            }
+            var args = ((DynamicDictionary)Request.Query).ToDictionary();
```
soamvasani: I suggest creating a "fission context" class and passing in an instance of that class.  One property of that class can be the request object, plus it can have a method to get the querystring as a dictionary.  That way we can add more stuff to the context later, without breaking compatibility with any functions.

environments/dotnet/README.md
```diff
@@ -0,0 +1,158 @@
+# Fission: dotnet C# Environment
```
soamvasani: Really appreciate the detailed documentation!

environments/dotnet/ExecutorModule.cs
ktrance: Agreed that a context class is a better solution but I really do not like bleeding the Nancy API into the core implementation, it could be a problem later down the road if either we want to change hosting technology or if the Nancy guys decides to change their API.

Even if it might seem like a lot of extra plumbing I think the best way to proceed would be to create our own wrapper extracting the items from the request as we go along.

Thoughts?

environments/dotnet/README.md
ktrance: :+1:

environments/dotnet/ExecutorModule.cs
ktrance: I added the changes I wrote about above as I figured it would be easier to discuss if we had some code to look at.

soamvasani: Thanks for adding the context class.  

With the other environments (nodejs, python) we did not abstract away HTTP.  So functions can access the full request object in both the python and nodejs environments.  This was a conscious decision not just to reduce the complexity of the environments, but also to make it easier for function programmers to gather whatever info they want from the request.  The web frameworks being used in those cases have stable and well documented APIs.

But, both those are dynamic/duck-typed languages so some of the concerns you're raising about declarations haven't come up before.  I think one way to go would be to have the query string args easily parsed out as a dictionary, and the rest of the request object available to the function through some method.  That way the most common use cases simply use the context.args dictionary and don't need any using declarations for nancy.  What do you think?

For logging, stdout is good.  The plan is to aggregate all logs from stdout into one of log aggregation systems.  (stdout is also what shows up in `kubectl logs`, so it's good for users too.)

soamvasani: I forgot to answer the question about use cases for http headers.  I think they're needed often.  Here's a random example, if you're implementing a github webhook on fission, you may need access to [these headers](https://developer.github.com/webhooks/#delivery-headers).

ktrance: Fair enough, I would still really like to wrap the Nancy objects in something completely under our control but looking at the Nancy code its a larger tasks. One thing that should be considered is that most frameworks targeting dotnet core are still rather new or at least recently ported. 
The Nancy API seem fairly stable but there are guarantees, sorry if I seem a little sceptic but I have been bitten too many times by breaking API changes and the problem is that once that genie is out of the bottle it can be very hard to put back in ;)

soamvasani: Ok, this is a good discussion, I think we can keep the context free of nancy APIs for the moment and revisit the decision later if there is a use case that this context doesn't work for.  As you say it's much easier to expand the API than shrink it.  So I'm good with the context/request stuff as-is.

environments/dotnet/FissionContext.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (logger == null) throw new ArgumentNullException(nameof(logger));
+            Arguments = args;
+            Logger = logger;
+        }
+
+        public Dictionary<string, object> Arguments { get; private set; }
+        public Logger Logger { get; private set; }
+    }
+
+    public class Logger
```
soamvasani: Is there any existing lightweight logging library that we can use instead?

ktrance: I'll look into it and get back to you. I just needed to know if you already had something so I should go look for a .net port of that. Give me a few days and I'll figure it out.

soamvasani: Sounds good, thanks. (Sorry for bringing up this issue late.)

haf: I've built a lightweight single-file logging implementation for C# called the Logary Facade. https://github.com/logary/logary#using-logary-in-a-library (scroll down for the C# example) You can then position an arbitrary logging framework behind the facade, should you wish to (like Logary itself)

ktrance: @haf : Thanks, I'll have a look hopefully tomorrow just a little bit booked right now.

ktrance: @haf , the link to the .cs file is broken

haf: @ktrance Thanks, I've fixed that link now. Thanks for telling me.

PR #94, opened 2017-01-31
bugfix (cli) Update the URL check to work with https

andrewstuart:
I've added a bit of logic to allow users to specify `https` for their FISSION_URL, which is critical when exposing the services via ingress and automatic TLS certificates from vault or LetsEncrypt.


soamvasani:
Excellent, thank you @andrewstuart

PR #97, opened 2017-01-31
Add initial support for PHP7

dgoujard:
To test the environment : 
```
fission env create --name php7 --image damien/fission-php-env
echo '<?php echo "Hello World"; ?>' > myfunction.php
fission function create --name hellophp --code monscript.php --env php7
fission route create --method GET --url /hellophp --function hellophp
curl http://$FISSION_ROUTER/hellophp
```


soamvasani:
Thanks for this, @dgoujard! Taking a look.

environments/php7/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM php:7.1-cli
+
+RUN apt-get update && apt-cache search libjpeg && apt-get install -y libjpeg-dev && apt-get install -y \
```
soamvasani: what's the apt-cache search for?

environments/php7/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM php:7.1-cli
```
soamvasani: Would the php7.1-alpine image work?  If all else is equal it would be better to use a smaller image.

soamvasani:
@dgoujard, would you mind also adding an example in the examples directly under examples/php?  Preferably something that shows how to get a POST request body, how to log, etc.

environments/php7/Dockerfile
dgoujard: apt-cache is for searching package. it's not necessary. I will remove this command.

environments/php7/Dockerfile
dgoujard: I updated the base image to alpine. Before : 195 MB After : 149 MB

dgoujard:
@smadhura I updated my PR with examples and switched to alpine base image. I also handle "handler" function with Request, Response and Logger in parameter. Previous scripts without "handler" function still works

soamvasani:
Thanks @dgoujard and sorry for the delay in replying.  Taking a look now.

environments/php7/src/Server.php
```diff
@@ -0,0 +1,64 @@
...
+                        $response->getBody()->write("Generic container: no requests supported");
+                        return $response;
+                    }
+
+                    ob_start();
+                    include($codepath);
+                    //If the function as an handler class it will be called with request, response and logger
+                    if(function_exists("handler")){
+                        ob_end_clean();
+                        return handler($request, $response,$logger);
```
soamvasani: So in other environments, we wrap up the request, response, logger, and perhaps other stuff into one "context" class.  This helps because if we want to add anything to the context at some point (e.g. a handle to a DB, a Kubernetes connection, etc.), we can do so without breaking compatibility with existing functions.  

Would this pattern be doable here as well?

examples/php7/stock.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
soamvasani: Very nice to see a non-hello world example 👍 :)

PR #98, opened 2017-02-01
Error when env name/image not provided

lcrisci:
I also renamed image to envImg for consistency


soamvasani:
Thanks so much, @lcrisci!  Appreciate the improvement.

PR #100, opened 2017-02-02
function code download using HTTP URL

kaustubhvp:
Download, if URL is provided otherwise use standard file fetch method.

Testing Done:
Download and Update of function using URL and file path.

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -17,10 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"os/exec"
+	"regexp"
 	"text/tabwriter"
 
 	"github.com/satori/go.uuid"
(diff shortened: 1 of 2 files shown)


soamvasani:
Not sure why that travis build failed. 🤔

fission/function.go
```diff
@@ -29,6 +32,37 @@ import (
 	"github.com/fission/fission"
 )
 
+func _fnFetchCode(filePath string) []byte {
+	var code []byte
+	var err error
+
+	isURL, _ := regexp.MatchString("\\Ahttp(s|)://", filePath)
```
soamvasani: Don't ignore errors, please

fission/function.go
```diff
@@ -29,6 +32,37 @@ import (
 	"github.com/fission/fission"
 )
 
+func _fnFetchCode(filePath string) []byte {
```
soamvasani: No _ necessary, lowercase function name already means this function is not exported.

soamvasani:
Thanks, @kaustubhvp !  Nice addition. Just a couple of minor comments above.

PR #109, opened 2017-02-06
Make build an actual sh script

tobias:
This allows it to be executed from alternative shells (fish, for
example) as `./build.sh`.


soamvasani:
Whoops. Thanks.

PR #110, opened 2017-02-06
Minor improvements to build instructions in README

tobias:
These are things that tripped me up a bit when trying to build, so I'm passing them along in case they help someone else.


soamvasani:
Thanks @tobias !

Issue #112, opened 2017-02-08
Trying to create an existing function results in a 500 error


tobias:
If you create a function:
    
    fission fn create --name foo --env nodejs --code foo.js

then try to use the same name again:

    fission fn create --name foo --env nodejs --code foo2.js

You get:

    ERRO[0000] Failed to create function                     err="(Error 0) HTTP error 500" name=hello
    Failed to create function: (Error 0) HTTP error 500
    
I think the controller API could be improved a bit to give a better status code and message in this case (and probably others). I'm happy to take a look at doing that if there is interest.

soamvasani:
> I think the controller API could be improved a bit to give a better status code and message in this case (and probably others). I'm happy to take a look at doing that if there is interest.

Yes, agreed and thanks for offering to help!

The function API is implemented in controller/functionApi.go; that's a good place for this particular check.

PR #116, opened 2017-02-10
Better convey duplicate name errors to client

tobias:
Motivation: Creating duplicate resources currently results in a 500
error being displayed by the client, with no information about the true
nature of the failure.

Modifications:
* ResourceStore now converts any errors from the etcd client into
  fission errors, capturing the reason for the error in the case of a
  duplicate key (as ErrorNameExists)
* ErrorNameExists errors are signaled to the client with a
  422 (Unprocessable Entity) HTTP status
* MakeErrorFromHTTP() now reads the body of the error response to
  retrieve the actual error message instead of using the HTTP status
  message
* the controller client now uses MakeErrorFromHTTP() to centralize
  status code -> error code mapping
* tests for all of the resources now check that duplicate resources are
  reported properly

Result: The client can now provide more context when a duplicate name is
given

related to #112 

This doesn't address improving the output from the client to actually tell you _which_ name was duplicated, that would be the next step.

I'm also still learning go, so feel free to hit me with any feedback if these changes aren't up to snuff.

### DIFF
--- a/controller/resourceStore.go
+++ b/controller/resourceStore.go
@@ -25,6 +25,9 @@ import (
 	"github.com/coreos/etcd/client"
 	"github.com/satori/go.uuid"
 	"golang.org/x/net/context"
+
+	"fmt"
+	"github.com/fission/fission"
 )
 
 type (
@@ -93,7 +96,7 @@ func (rs *ResourceStore) create(r resource) error {
 
 	_, err = rs.KeysAPI.Set(context.Background(), key, string(serialized),
 		&client.SetOptions{PrevExist: client.PrevNoExist})
-	return err
+	return handleEtcdError(err)
 }
 
 func (rs *ResourceStore) read(rkey string, res resource) error {
@@ -105,7 +108,7 @@ func (rs *ResourceStore) read(rkey string, res resource) error {
 
 	resp, err := rs.KeysAPI.Get(context.Background(), key, nil)
 	if err != nil {
-		return err
+		return handleEtcdError(err)
 	}
 	return rs.serializer.deserialize([]byte(resp.Node.Value), res)
 }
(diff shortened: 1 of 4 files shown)


soamvasani:
Thanks for the much needed improvement @tobias!  Looking at this now.

BTW the check failed because `gofmt` wasn't run on these files:
./error.go
./controller/api_test.go
./controller/resourceStore.go

controller/resourceStore.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/coreos/etcd/client"
 	"github.com/satori/go.uuid"
 	"golang.org/x/net/context"
+	"github.com/fission/fission"
```
soamvasani: nit: mind adding a blank line above this import? (we separate out in-project imports from the rest)

controller/resourceStore.go
```diff
@@ -246,9 +247,26 @@ func (rs *ResourceStore) deleteAllFiles(key string) error {
...
-			return err
+			return handleEtcdError(err)
 		}
 	}
 	_, err = rs.KeysAPI.Delete(context.Background(), key, &client.DeleteOptions{Dir: true})
-	return err
+	return handleEtcdError(err)
+}
+
+func handleEtcdError(e error) error {
```
soamvasani: Nice! 👍

error.go
```diff
@@ -36,29 +38,39 @@ func MakeErrorFromHTTP(resp *http.Response) error {
...
 	switch resp.StatusCode {
+	case 400:
+		errCode = ErrorInvalidArgument
 	case 403:
 		errCode = ErrorNotAuthorized
 	case 404:
 		errCode = ErrorNotFound
-	case 400:
-		errCode = ErrorInvalidArgument
+	case 422:
```
soamvasani: Hmm, I'd think 409 Conflict fits better for duplicate names.  Thoughts?  Is 422 commonly used for this?

tobias: Yes, agreed - 409 is a better fit. I've used 422 in the past for more complex errors where I had to return more context as headers, but if we're just communicating a name collision, 409 is better.

tobias:
I pushed a new branch with fixes to all of the above - thanks for the review!

soamvasani:
Thanks @tobias, looks great!

Issue #108, opened 2017-02-06
controller error handling on failure to stat/create filestore


soamvasani:
The controller handles the case that `/filestore` doesn't exist, but not the case where `stat` failed for any other reason.  And `fission-bundle` doesn't handle `controller.MakeFilestore` errors at all.

PR #118, opened 2017-02-10
Fixed failed to delete function when function's file is not exist

life1347:
This PR fixed the problem that user may encounter when trying to delete a function without function file in filestore. Now, the controller checks the file existence before removing it. (https://github.com/fission/fission/issues/108)

```
$ fission fn list
NAME  UID                                  ENV
hello bdecc45c-a528-4f1f-9fca-728bdc3d86f6 nodejs
$ fission fn delete --name hello
Failed to delete function 'hello': Delete failed: remove /filestore/bdecc45c-a528-4f1f-9fca-728bdc3d86f6: no such file or directory
```

### DIFF
--- a/controller/fileStore.go
+++ b/controller/fileStore.go
@@ -85,7 +85,11 @@ func (fs *FileStore) fileStoreService() {
 		case WRITE:
 			response.error = ioutil.WriteFile(path.Join(fs.root, req.fileName), req.fileContents, 0600)
 		case DELETE:
-			response.error = os.Remove(path.Join(fs.root, req.fileName))
+			response.error = nil
+			filePath := path.Join(fs.root, req.fileName)
+			if _, err := os.Stat(filePath); os.IsExist(err) {
+				response.error = os.Remove(filePath)
+			}
 		default:
 			log.Panic("bad request")
 		}
--- a/controller/fileStore.go
+++ b/controller/fileStore.go
@@ -85,10 +85,9 @@ func (fs *FileStore) fileStoreService() {
 		case WRITE:
 			response.error = ioutil.WriteFile(path.Join(fs.root, req.fileName), req.fileContents, 0600)
 		case DELETE:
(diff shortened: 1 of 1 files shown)


life1347:
Also, we should return more meaningful error message back to user when trying to get function file that not exist. (https://github.com/fission/fission/issues/112)

```
$ fission function get --name hello
Failed to get function: (Error 0) HTTP error 500
```

tobias:
With #116, that message gets a little better:

    $ fission fn get --name foo
    Failed to get function: (Error 0) 100: Key not found (/Function/foo) [17]

but #116 doesn't map the etcd 100 error to fission's ErrorNotFound - I'll do that in a later PR if #116 is accepted. Thanks for pointing it out!

controller/fileStore.go
```diff
@@ -85,7 +85,11 @@ func (fs *FileStore) fileStoreService() {
 		case WRITE:
 			response.error = ioutil.WriteFile(path.Join(fs.root, req.fileName), req.fileContents, 0600)
 		case DELETE:
-			response.error = os.Remove(path.Join(fs.root, req.fileName))
+			response.error = nil
+			filePath := path.Join(fs.root, req.fileName)
+			if _, err := os.Stat(filePath); os.IsExist(err) {
```
soamvasani: Won't err just be nil if the stat succeeds?  I would suggest unconditionally doing the `os.Remove` but checking for and ignoring err when `os.IsNotExist(err)`.

life1347: @soamvasani thanks for your suggestion! I just updated the commit. By the way, should I squash the commits into one?

soamvasani: Thanks for the update! No need, I can squash the commits.

soamvasani:
BTW, #108 is about error handing in creating the filestore path itself; that's a bit different from the bug addressed here.

PR #120, opened 2017-02-13
Use latest for stable release of minikube

r2d4:
The latest stable release for minikube is published at
storage.googleapis.com/minikube/releases/latest


soamvasani:
Yeah... I admit I'm a little nervous about automatically pulling in the latest version, even though minikube is pretty stable.  @r2d4 would you mind changing this to '0.16.0' instead of latest?

r2d4:
Makes sense.  Changed it to the latest version.

r2d4:
It's worth it to note that minikube is actually compatible with a bunch of versions of kubernetes.  If you do ever run into issues with compatibility, you can always 'sticky' a version of kubernetes with 

`minikube config set kubernetes-version v1.5.2`

soamvasani:
Thanks! And good to know about `minikube config set`.

PR #121, opened 2017-02-13
Report KeyNotFound from etcd as a 404

tobias:
This allows for proper classification of the error on the client side.


soamvasani:
Looks great, thanks!

PR #122, opened 2017-02-14
Improve command-line client error output

tobias:
Motivation: The output when an error occurs shows an integer error code,
which isn't very helpful. And in the case of an error when creating a
resource, you also get a redundant log message from the controller
client.

Modifications:
* the controller client now has a flag to suppress error logging
* the CLI client uses a logger with that flag set
* the error codes enum is now typed to allow `stringer` to generate a
  String() for it
* a go:generate directive was added to run stringer automatically when
  `go generate` is called
* the generated String() allows for the enum name to be displayed in the
  error output
* the type used for an error code is now public (fission.ErrorCode) so
  it can be used as an argument to MakeError(), since the enum is now
  typed.
* the format of the error message from the CLI client was modified to
  display the error message before the error type

Related to #112 

I don't think this is quite ready to merge, but wanted to get it out there for discussion. My biggest concern with this change is the introduction of `stringer` and `go generate` into the build. This PR includes the output of the generation `errorcode_string.go`, but probably shouldn't. Including it allows other devs to build without having to install `stringer` or run `go generate`, with the chance of the generated code getting out of date. Not including it complicates the build, since it would need to be generated by each dev. This would be alleviated somewhat by having a Makefile, but it's kind of nice that one hasn't yet been needed.

Another concern is making `ErrorCode` public - was it intentionally private? Are there drawbacks to making it public? 

Any other feedback is also welcome.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -26,8 +26,6 @@ import (
 	"net/http"
 	"strings"
 
-	log "github.com/Sirupsen/logrus"
-
 	"github.com/fission/fission"
 )
 
@@ -104,10 +102,6 @@ func (c *Client) FunctionCreate(f *fission.Function) (*fission.Metadata, error) 
 
 	body, err := c.handleResponse(resp)
 	if err != nil {
-		log.WithFields(log.Fields{
-			"name": f.Metadata.Name,
-			"err":  err,
-		}).Error("Failed to create function")
 		return nil, err
 	}
 
@@ -240,10 +234,6 @@ func (c *Client) HTTPTriggerCreate(t *fission.HTTPTrigger) (*fission.Metadata, e
 
 	body, err := c.handleResponse(resp)
 	if err != nil {
-		log.WithFields(log.Fields{
-			"name": t.Metadata.Name,
-			"err":  err,
-		}).Error("Failed to create http trigger")
 		return nil, err
 	}
 
@@ -351,10 +341,6 @@ func (c *Client) EnvironmentCreate(env *fission.Environment) (*fission.Metadata,
 
 	body, err := c.handleResponse(resp)
 	if err != nil {
-		log.WithFields(log.Fields{
-			"name": env.Metadata.Name,
-			"err":  err,
-		}).Error("Failed to create environment")
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


Marak:
@tobias - 

Seems like a good start!

I think in most cases it's going to be better to return the full error code or error stack when possible ( making them public )

My thoughts are that it is that it's easier to munge or filter extra data before sending it to the client, versus not having the additional data to begin with. Would be interested to hear what the project leads think.

soamvasani:
Thanks @tobias! I'd suggest going further and simply removing all logging from the client library.  It was a mistake on my part to include it.  It should just return errors.  The CLI should be the one turning the errors into human-readable strings.

I see what you're saying about the use of stringer.  How about just hard coding a literal map of error codes to strings, in error.go, instead of generating this?  I know it's repetitive but the simplicity seems worth it.  Also if you do that, CLI errors could have slightly more human readable error output, e.g. "Name already exists" instead of "ErrorNameExists".

tobias:
@Marak:
 Thanks! My concern about returning the original error code and stack to the client is it wouldn't be very useful to users of the `fission` tool, since they can't really do anything with an etcd error code and a stack trace. If the error was triggered by user error, we should just tell them that. If it was triggered by a true internal failure, we should make sure those types of errors are in the controller log, not being shipped to the user. But, if others decide we should ship more information, it wouldn't be difficult to add.

@soamvasani:
 I assumed the logging was in the client to handle error output when used by other non-cli components (kubewatcher, poolmgr, router, etc). If we rip logging out of the controller client, we should make sure that all consumers of the client log the errors. 

WRT the error names, that makes sense. I like having the opportunity to provide a better message there. I'll work on that now.

soamvasani:
@tobias Yes, client errors already make it into the logs for the callers in poolmgr, router, and kubewatcher.

tobias:
I've pushed commit with the requested changes. The output now looks like:

```
$ fission env get --name asdf
Failed to get environment: Resource not found - environment 'asdf' does not exist
$ fission fn get --name asdf
Failed to get function: Resource not found - function 'asdf' does not exist
$ fission fn create --name hello --env nodejs --code hello.js
Failed to create function: Resource exists - function 'hello' already exists
```

soamvasani:
This looks great, thanks!

Issue #101, opened 2017-02-02
Internal server error (fission)


jedipunkz:
I got this error "Internal server error (fission)".

I ran these command as follows.

```
docker-machine start default
minikube start
export FISSION_URL=http://$(minikube ip):31313
export FISSION_ROUTER=$(minikube ip):31314
fission env create --name nodejs --image fission/node-env
echo 'module.exports = function(context, callback) { callback(200, "Hello, world!\n"); }' > hello.js
fission function create --name hello --env nodejs --code hello.js
fission route create --method GET --url /hello --function hello
curl http://$FISSION_ROUTER/hello
Internal server error (fission)
```

Any idea?

nouney:
Same here

nouney:
@jedipunkz I finally found the solution. In my case, `echo 'module.exports = function(context, callback) { callback(200, "Hello, world!\n"); }' > hello.js` creates a malformed file. I had to manually write the function. Use `node hello.js` to check the file.

Marak:
Service errors should be piped directly back to the client with an actual
error stack.

Getting back "Internal Server Error" message is unacceptable.

soamvasani:
@jedipunkz Would you mind looking at poolmgr logs (`kubectl --namespace fission get pod poolmgr` to find the poolmgr pod.)?  There was a known issue #89 which you _may_ have hit. Or it could be the same thing that @nouney saw.

So in general, we have work to do in this area. I agree that users shouldn't be seeing hard-to-diagnose "Internal server errors".

#53 (Function logs) is tracking an easier way for users to get logs from functions.

More significantly, the environments v2 work (Documentation/wip/environments-v2.md) is tracking a function "compile"-time check to surface syntax errors while a function is registered, instead of letting syntax errors show up at runtime.

Also I'd like to update the README if these instructions are resulting in a malformed file.

soamvasani:
I should point out that I disagree with @Marak's comment, though -- error stacks usually should _not_ be send to clients, unless it's in some sort of "development mode".  The HTTP error itself is right, in this case.

(The programmer of the function is the one interested in stack traces, not the user of the app the programmer is writing)

jaxxstorm:
So I have this same problem, with a standard deploy. The error looks like like this:

router:
```
172.29.132.132 - - [06/Feb/2017:22:52:32 +0000] "GET /hello HTTP/1.1" 500 32
[...]

PR #123, opened 2017-02-15
Bash command generate an invalid hello.js

gonrial:
Resolves the problem explained in #101.

The command:
 `echo 'module.exports = function(context, callback) { callback(200, "Hello, world!\n"); }' > hello.js` 
generates a `hello.js` file with this structure:
```
echo 'module.exports = function(context, callback) { callback(200, "Hello, world!
"); }' > hello.js
``` 
as the string is split in several lines, nodejs fails thus returning an error.

When using linux a `\\n` must be used instead of `\n`

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -101,7 +101,7 @@ Finally, you're ready to use Fission!
 ```
   $ fission env create --name nodejs --image fission/node-env
 
-  $ echo 'module.exports = function(context, callback) { callback(200, "Hello, world!\\n"); }' > hello.js  
+  $ echo -E 'module.exports = function(context, callback) { callback(200, "Hello, world!\n"); }' > hello.js  
 
   $ fission function create --name hello --env nodejs --code hello.js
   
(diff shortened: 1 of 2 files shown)


soamvasani:
Thanks for this!  I've been a bit puzzled by those bug reports.  But your example actually results in an extra \ for me (bash on mac os).  I guess it's dependent on the shell being used.  

I'm thinking that example should just be changed to something like this, instead of using echo:

```
curl https://raw.githubusercontent.com/fission/fission/master/examples/nodejs/hello.js > hello.js
```

gonrial:
Found another way ... `-E` is the default option when `echo` is used which disables the interpretation of backslash escapes. Some linux distros enables the interpretation, thus making this file unusable.

soamvasani:
Unfortunately -E isn't a standard `echo` option. E.g. when I run `sh`, `echo -E foo` literally echoes `-E foo`

ayzwah:
I also spent a bunch of time debugging the malformed js file.
I'd also suggest something like
`curl https://raw.githubusercontent.com/fission/fission/master/examples/nodejs/hello.js > hello.js`

soamvasani:
Sorry to hear that, @ayzwah. I'll change it to the curl command.

soamvasani:
Fixed in a separate commit, see #128.  Thanks for diagnosing this issue, @gonrial!

PR #124, opened 2017-02-15
Modify the stock example to show how to change the Content-Type

gonrial:
The example returns a JSON object, but headers are not properly set.

Also no checks or validations are performed on processing the body. This is probably a good opportunity  to show how to deal with errors.

### DIFF
--- a/examples/nodejs/stock.js
+++ b/examples/nodejs/stock.js
@@ -26,6 +26,9 @@ module.exports = async function (context) {
             status: 200,
             body: {
                 text: `${symbol} last traded at ${lastTrade}`
+            },
+            headers: {
+                'Content-Type': 'application/json'
             }
         };
     } catch (e) {


examples/nodejs/stock.js
```diff
@@ -20,10 +20,11 @@ module.exports = function (context, callback) {
 
             try {
                 var parsed = JSON.parse(resp.slice(3));
-                var lastTrade = parsed[0]['l_cur']
-                callback(200, `{ "text": "${symbol} last traded at ${lastTrade}" }`);
+                var lastTrade = parsed[0]['l_cur'];
+                var headers = {'Content-Type': 'application/json'};
```
soamvasani: Thanks, good idea to add this!  But since it's used in the catch block, it should be moved above the `try` on line 21, I think.

Issue #93, opened 2017-01-31
Buffered context means no streaming support


Marak:
`fission` appears to use a buffered context for both receiving and sending data.

This prevents the ability to create streaming transformations, or process any data coming into the function as a stream. Do you have any plans to add streaming support? Seems like a pretty big thing to not have

`microcule` has streaming support ( as well as a plugin system and support for 14+ languages ) https://github.com/stackvana/microcule

ktrance:
:+1:

soamvasani:
Yes, we're interested in the streaming use case.  Changing the title to be a bit more general.  I *think* this will only need changes in the environments, since the router is just a proxy; but we'll have to test it out.

Also, I see you've linked microcule a couple times in the issues -- do you work on that project?  Is your goal here some sort of collaboration?  Happy to discuss more.

Marak:
`microcule` is a project I built over the past two years to support https://hook.io

I have strong experience and knowledge in this domain.

My main interest is making sure that this project ( `fission` ) is adequately covering what I feel is the Minimal Viable Product for an open-source functions as a service platform.

schmurfy:
has anything change ?
We are considering a usecase where we would need to transfer a large volume of data between two functions, http should work well for this if supported.

sanketsudake:
This has been inactive for 1653 days and is being marked stale. It will be closed in a future scrub if there's no further activity. Comment to keep it open.

sanketsudake:
Streaming function invocation is supported natively now — set `streaming` on a Function's spec (with `streamingProtocol` to choose the upstream mode) and the router proxies the response as a stream.

Closing as completed; please reopen with details if your use case isn't covered on a current release.

Issue #24, opened 2016-11-11
Support for compiled languages


soamvasani:
To support Go, Rust, and other compiled languages, we'll need a separate build step.  This should be triggered at function create/update.

Environments will need a link to an image that will have the compiler and other tools (such as dependency managers).

We will need to figure out a way to specify dependencies for functions.  Maybe a new fission resource type for a dependency spec, and a way to associate that resource with functions.

isubuz:
When you say "a separate build step", do you mean a step which creates the executables and then load these executables when the function is _triggered_? 
I guess for Java which supports dynamic loading, the functions could be loaded from source, similar to what is done for Python? I guess the performance hit happens only the first time the class is loaded into memory.

soamvasani:
> When you say "a separate build step", do you mean a step which creates the executables and then load these executables when the function is triggered?

The functions would be compiled when they are created (or updated); the result of the compilation would be cached. That way we don't add the compilation time to the request path. 

Then, on request, the compiled function (.so/go plugin file/jar file) would be copied into the generic pod, loaded, and run.  We haven't yet measured the performance of loading a Java class dynamically, but yes, the performance hit of loading will only affect the "cold start" of a function.

GiorgioG:
How about some documentation as to how to add language support?  I'm particularly keen on adding C# / .NET Core support and would happy to contribute if only I had an inkling into what it involved.

soamvasani:
Hi @GiorgioG, there is an early version of a C# .NET environment contributed by @ktrance.  If you're interested in contributing to that, please take a look at [environments/dotnet](https://github.com/fission/fission/tree/master/environments/dotnet).

That environment compiles code on cold-start, which is a short term solution until this issue is resolved.

There isn't yet documentation for this, because we don't yet have compiled language support ready.  That work is being designed and specified in [Documentation/wip/environments-v2.md](https://github.com/fission/fission/blob/master/Documentation/wip/environments-v2.md).

Marak:
Compiled languages are tricky for faas. We've done some work on implementing these in hook.io. Have been considering adding support to https://github.com/stackvana/microcule, but it's not been a heavily requested feature.

The service should build fresh the first time and then you'll need a cache to store the compiled source code and then a smart way to know when to invalidate the cache. Compilation errors need to be piped back to the client in real-time.

ktrance:
[...]

PR #125, opened 2017-02-15
[WIP] Golang runtime

nouney:
**Work in progress**

_UPDATE 20/07/2017_: I focus the PR on the implementation of the golang runtime. The implementation of the builder will be the subject of another PR.

---

I started to work on a golang environment. It uses the [plugin](https://tip.golang.org/pkg/plugin/) feature of go 1.8 which should be released very soon.
This is the very first step to implement the [env v2 spec](https://github.com/fission/fission/blob/master/Documentation/wip/environments-v2.md).

This runtime use 3 entrypoints:
 - `func Handler func(w http.ResponseWriter, r *http.Request)`
 - `var Handler http.HandlerFunc`
 - `var Handler http.Handler`

You'll need to create a package that exports a symbol `Handler`  and then build it as a plugin (with [`go1.8rc3`](https://golang.org/dl/#unstable)).

```go
$ cat hello.go
package main

import (
	"net/http"
)

func Handler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("Hello, world!"))
}

$ go1.8rc3 build -buildmode=plugin -o hello.so hello.go
$ fission fn create --name hello --code hello.so --env go-runtime
$ curl http://$FISSION_ROUTER/hello
Hello, world!
```
TODO:
- Runtime
  - [x] Add entrypoints: 
    - [x] `func Handler func(w http.ResponseWriter, r *http.Request)`
    - [x] `var Handler http.HandlerFunc`
    - [x] `var Handler http.Handler`
  - [x] Reduce image size
- ~~Builder~~
 - ~~Build a plugin~~
 - ~~Should handle golang dependencies tools (glide, dep, etc.)~~


soamvasani:
Nice, I'm excited about this!  Will take a look.

The test failure is expected btw, since it's compiling your code with go 1.7.

soamvasani:
Hey, the change looks great!

But we shouldn't put binary packages in the --code field.  If we do that then fission won't be able to distinguish between source and binaries when we actually have the builder.

So maybe we should add a field for buildPackage just like we have one for code. 

So `fission function ... --code foo.go` would get sent to the builder (once this exists) and `fission function --buildPackage foo.so` would go straight to this runtime environment.

BTW, what's the size of hello.so in your example?

nouney:
> So fission function ... --code foo.go would get sent to the builder (once this exists) and fission function --buildPackage foo.so would go straight to this runtime environment.

I agree, I'll update the code. However `--buildPackage` is quite "heavy", IMHO. What do you think about `--package`?

> BTW, what's the size of hello.so in your example?
~10MB, which is big for just a "Hello, World!".

Marak:
I'm doing testing in another go branch and for `hello.go` and `echo.go` examples with `go version 1.6` on `darwin/amd64` I am seeing `2.1mb` size binary when not using plugins API.

Does it make sense to use the go plugins API for this?

soamvasani:
> I agree, I'll update the code. However --buildPackage is quite "heavy", IMHO. What do you think about --package?

Thanks. We're going to have source packages too, to support a set of source files.  So I wanted to distinguish build packages from that.  I suppose we could have `fission function ... --package` for binaries and `fission function ... --source-package` for source packages.

Marak:
Just released support for `golang` over in the [microcule](https://github.com/stackvana/microcule) repo.

Usage is something like:

**echo.go**

```go
package main

import (
    "os"
    "fmt"
)

func main() {
  for _,element := range os.Args {
    fmt.Println(element);
  }
  os.Exit(1);
}
```

```bash
microcule echo.go
```

Will start listening HTTP server. HTTP params are automatically mapped to incoming `argv` scope. Could probably use some improvement.

soamvasani:
@nouney 

Could you add the hello.go example into a new examples/go/ directory?  Also, since the user has to compile the plugin in a certain way, could you please add instructions for that to the README?

BTW, in other language envs, we have support for passing optional params in a "context" struct.
The use case for that is we want to be able to provide new/optional stuff to the function without breaking compatibility.  For example, in the case of watching kubernetes events we could have a customized environment that makes a connection to k8s API and passes the client into the function.  

One way to do that is to create a "fission context" type and add a reference to that to the function signature (or we could use the golang http request context, but it's not really supposed to be used for optional params).

nouney:
> We're going to have source packages too, to support a set of source files. So I wanted to distinguish build packages from that. I suppose we could have fission function ... --package for binaries and fission function ... --source-package for source packages.

Don't you think we could use `--code` for source package too?  By checking the file extension we can know if it's a simple file or a zip or whatever:

```
$ fission fn create --name hello --code hello.go --env go-runtime
$ fission fn create --name app --code app.zip --env go-runtime
```

I think it will make things easier to understand for the user.

> Could you add the hello.go example into a new examples/go/ directory? Also, since the user has to compile the plugin in a certain way, could you please add instructions for that to the README?

Done!

>BTW, in other language envs, we have support for passing optional params in a "context" struct.
> [...]
> One way to do that is to create a "fission context" type and add a reference to that to the function signature (or we could use the golang http request context, but it's not really supposed to be used for optional params).

Something like that? I think it's a good solution.

```
package main

import (
	"net/http"

	"github.com/fission/fission/go/runtime" // whatever
)

func Handler(ctx runtime.Context, w http.ResponseWriter, r *http.Request) {
         // gorilla/mux params, by example
	params := runtime.GetParams(ctx)
	msg := "Hello " + params.Name
	w.Write([]byte(msg))
}
```

soamvasani:
Thanks for the updates, @nouney !

>  Don't you think we could use --code for source package too? By checking the file extension we can know if it's a simple file or a zip or whatever:

I was thinking about this, but... both source and binary packages can be zip. And the source package could even contain a mixture of source and binaries.  It's up to the environment builder whether to support that.  I mean, it's _may_ be possible to distinguish source/binary automatically and in a language-independent way, but it seems complicated.

> Something like that? I think it's a good solution.

Yes, that looks great!

narg95:
am I wrong, but it looks like microclue and go plugins are two competitive solutions for this feature, right?

Marak:
@narg95 -

https://github.com/stackvana/microcule is a separate project entirely.

We also already have landed go support as of a couple of weeks ago.

soamvasani:
Microcule is a totally separate project, unaffiliated in any way with Fission.

(@Marak, I'm really not sure why you keep bringing up microcule in various issues on our issue tracker.  It's just confusing to people.  Github issues is not a good venue for generic discussions about the various frameworks in this space.  That said, please feel free to comment if you have something tangible to contribute to fission.  Thanks.)

Marak:
I've contributed [signifiant](https://github.com/fission/fission/issues/24#issuecomment-280504978) new information across multiple issues. I've also opened [several](https://github.com/fission/fission/issues/93) issues for missing [features](https://github.com/fission/fission/issues/126)

If you don't want my input I'm glad to not bother you. `microcule` is ahead of `fission` for a lot of functionality and we currently have integrated support and examples for 20+ languages ( including Go ). 

Developers looking for these missing functionalities should know they already exist elsewhere, and more importantly....that this information could be re-used here...

soamvasani:
> Developers looking for these missing functionalities should know they already exist elsewhere

This is really all I'm talking about; please don't use fission's Github issues as a marketing channel.  There's plenty of frameworks out there and many are ahead of fission on various aspects of functionality.  Github issues is not a good place to discuss them.

For everything else, exercise your judgement as to relevance.  Thanks :)

Marak:
>  and more importantly....that this information could be re-used here...

Marak:
Okay, I'll stop responding to your issues and adding information.

nouney:
@soamvasani 

> We're going to have source packages too, to support a set of source files. So I wanted to distinguish build packages from that. I suppose we could have fission function ... --package for binaries and fission function ... --source-package for source packages.

I added a `--package` flag, which is for now just an alias of `--code`. 

> One way to do that is to create a "fission context" type and add a reference to that to the function signature (or we could use the golang http request context, but it's not really supposed to be used for optional params).

I had to create 2 packages to not break compatibility between the runtime server and the plugin:
- context: contains only the `Context` type (which is just a `map[string]interface{}`). This package is used by the runtime server and the plugin. **To never break compatibility, this package should not be changed**.
- runtime: empty for now. This package will be used only by the plugin and will contains helper functions to manipulate the `Context`.

Example:

```
package main

import (
  // I think it's too long to type/remember, where would be the best place
  // for these packages?
  "github.com/fission/fission/environments/go/context"
  "github.com/fission/fission/environments/go/runtime"
)

func Handler(ctx context.Context, w http.ResponseWriter, r *http.Request) {
  whatever := runtime.GetWhatever(ctx)
  w.Write([]byte(whatever))
}
```

Unfortunately, to be able to use the `context` package in the plugin, the `GOPATH` used to build the plugin must be the same as the one used to build the runtime server. I added a shell script as an example.

soamvasani:
Thanks @nouney, I tested this out and it works great!

I think we should just remove the environments/go/runtime package for now, and add it later if we actually need it.  Other than that this is good for merge.

The build process needs improvement.  Maybe a self-contained docker image that knows how to build the plugin.  But let's take that on in another PR.

>  // I think it's too long to type/remember, where would be the best place

I've been thinking about splitting out the environments into their own github repositories, maybe it's time to do that.  That way user functions don't have to pull in the whole fission repo, just the small context library.  Again, let's tackle that in another PR.

yqf3139:
Hi. I just have a test on my minikube but get errors on the go-env pod
```
2017-03-22T04:09:58.760104246Z Listening on 8888 ...
2017-03-22T04:20:27.292661611Z Specializing ...
2017-03-22T04:20:27.334664502Z 2017/03/22 04:20:27 http: panic serving 172.17.0.4:53032: plugin.Open: /userfunc/user: cannot read file data
2017-03-22T04:20:27.334709503Z goroutine 18 [running]:
2017-03-22T04:20:27.334761453Z net/http.(*conn).serve.func1(0xc4200d8000)
2017-03-22T04:20:27.334772200Z 	/usr/local/go/src/net/http/server.go:1721 +0xd0
2017-03-22T04:20:27.334782224Z panic(0x7e3480, 0xc4200e00d0)
2017-03-22T04:20:27.334791191Z 	/usr/local/go/src/runtime/panic.go:489 +0x2cf
2017-03-22T04:20:27.334800577Z main.loadPlugin(0xc420032c88)
2017-03-22T04:20:27.334809454Z 	/usr/src/github.com/fission/fission/environments/go/server.go:21 +0x45e
2017-03-22T04:20:27.334818566Z main.specializeHandler(0xb7f240, 0xc4200ee000, 0xc4200e8000)
2017-03-22T04:20:27.334827601Z 	/usr/src/github.com/fission/fission/environments/go/server.go:64 +0x1c0
2017-03-22T04:20:27.334836791Z net/http.HandlerFunc.ServeHTTP(0x848e38, 0xb7f240, 0xc4200ee000, 0xc4200e8000)
2017-03-22T04:20:27.334845936Z 	/usr/local/go/src/net/http/server.go:1942 +0x44
2017-03-22T04:20:27.334854986Z net/http.(*ServeMux).ServeHTTP(0xba2640, 0xb7f240, 0xc4200ee000, 0xc4200e8000)
2017-03-22T04:20:27.334863555Z 	/usr/local/go/src/net/http/server.go:2238 +0x130
2017-03-22T04:20:27.334872385Z net/http.serverHandler.ServeHTTP(0xc42008a2c0, 0xb7f240, 0xc4200ee000, 0xc4200e8000)
2017-03-22T04:20:27.334880985Z 	/usr/local/go/src/net/http/server.go:2568 +0x92
2017-03-22T04:20:27.334889592Z net/http.(*conn).serve(0xc4200d8000, 0xb7f740, 0xc4200d00c0)
2017-03-22T04:20:27.334898117Z 	/usr/local/go/src/net/http/server.go:1825 +0x612
2017-03-22T04:20:27.334906666Z created by net/http.(*Server).Serve
2017-03-22T04:20:27.334915190Z 	/usr/local/go/src/net/http/server.go:2668 +0x2ce
```

fetcher seems worked
```
2017-03-22T04:20:26.629851737Z 2017/03/22 04:20:26 fetcher request: {http://controller.fission/v1/functions/hello?uid=dc9c2e1b-4b8c-4e27-a209-7ab7aec72812&raw=1 user}
```

The raw .so file exists
```
rm hello.so && fission fn get --name hello > hello.so && file hello.so                                                                
hello.so: ELF 64-bit LSB  shared object, x86-64, version 1 (SYSV), dynamically linked, stripped
```

soamvasani:
@yqf3139 could you please file a new issue for that? thanks.

yqf3139:
@soamvasani can you reproduce the same error? I was wondering if I missed something in the setup

soamvasani:
No, it worked for me.  But please, let's take this discussion to a new github issue.

yqf3139:
Thanks @soamvasani . I just tried a fresh minikube and it worked.
Sorry for the inconvenience.

Issue #53, opened 2016-12-10
Function logs


soamvasani:
Today a user can access function logs by looking for the pod and using `kubectl logs`:

```
kubectl --namespace fission-function get pods -l functionName=foo
```
```
kubectl --namespace fission-function logs podname
```

This has two problems:
(a) It doesn't work after the pod is deleted
(b) It's cumbersome, the user has to learn how to do this

We need to enable something like
```
fission function logs --name <function>
```
And have this solve both problems (a) and (b) above.  Of course (a) will require persisting logs somewhere.  We need to investigate if it's easy to integrate with one of the existing solutions, or whether we need to roll our own solution for that.

life1347:
For the log persistence, I think we can use influxDB to store and read logs from it. 
1. add extra log filter with influxDB query
2. influxDB provide cluster (HA)
3. influxDB is also written in go :D

soamvasani:
@life1347 sounds interesting! Have you used influxDB for log storage before?

Marak:
I'd recommend considering implementing a generic log provider which exposes a simple logging API. I've not used Influx DB before, but I'd imagine a lot of DBs could be used for storing logging entries.

The idea would be providing a decent logger by default, and allow developers to easily extend with their own logging solution. 

Here is an example of what a multi-transport logger API might look like https://github.com/winstonjs/winston

life1347:
@soamvasani yes, but couple months ago. I'll try to figure out how to pipeline pod's logs to db in an efficient way and make a simple PoC.

@Marak you're right! A generic log provider is needed for supporting different db. After the PoC, I will think it through. Thanks!

soamvasani:
@life1347 Great! Let me know if you have any questions on the way.

Here's a link to the [K8s logs API](https://godoc.org/k8s.io/client-go/kubernetes/typed/core/v1#PodExpansion).  For a first cut, that might be the simplest way.

life1347:
After some test today, I created a sidecar container (logger) inside the pod of function to pipe log to external db service. Following is a simple diagram to show how it works.
 
![logger](https://cloud.githubusercontent.com/assets/202578/22854653/8dc44734-f0ae-11e6-9db1-bfd41e6f78fe.png)

1. logger ask k8s api for the pod log
2. k8s reply with log stream
3. logger receive and parse the log stream, then send a HTTP POST to influxdb.

Video Demo:  https://youtu.be/6PeBQxVHx2M
[...]

PR #131, opened 2017-02-22
Add function logs support (#53)

life1347:
This PR add function logs support for issue #53.

Display function logs
```
  # Currently only influxDB is supported, use --help to see more options
  $ fission function logs --name hello --dbhost http://<ip>:8086
```

Display function logs from specific pod

```
  # Currently only influxDB is supported, use --help to see more options
  $ fission function pods --name hello --dbhost http://<ip>:8086
  $ fission function logs --name hello --dbhost http://<ip>:8086 --pod <pod>
```

Also, this PR implement the log database interface and use fluentd to pipe logs. Developer can implement their own log database solution for storing/searching function logs.


soamvasani:
Awesome stuff, @life1347 !!  Excited to have this feature and thanks for thinking it through and talking about it in the issue.  I'll review and make comments.

soamvasani:
Oh I forgot to say: could you add the diagram from the issue into the Documentation/ directory in this PR?  It's helpful to anyone trying to understand the architecture.

Also, we're trying to keep the environments/ directory for languages (I know fetcher is there, it should be moved out too, but later).  Could you move `logger` to the top level?

environments/logger/daemonset-logger.yaml
```diff
@@ -0,0 +1,95 @@
...
+  namespace: fission
+spec:
+  ports:
+  - name: "8086"
+    port: 8086
+    targetPort: 8086
+    protocol: TCP
+
+---
+apiVersion: v1
```
life1347: I'll duplicate the service files, one is for internal and the other is for external, cause user may want to use their own existing external db service.

fission/main.go
```diff
@@ -40,6 +40,13 @@ func main() {
 	fnEnvNameFlag := cli.StringFlag{Name: "env", Usage: "environment name for function"}
 	fnCodeFlag := cli.StringFlag{Name: "code", Usage: "local path or URL for source code"}
 	fnUidFlag := cli.StringFlag{Name: "uid", Usage: "function uid, optional (use latest if unspecified)"}
+	fnPodFlag := cli.StringFlag{Name: "pod", Usage: "function pod name, optional (use latest if unspecified)"}
+	fnFollowFlag := cli.BoolFlag{Name: "follow, f", Usage: "specify if the logs should be streamed"}
+	fnDetailFlag := cli.BoolFlag{Name: "detail, d", Usage: "display detailed information"}
+	fnLogDBHostFlag := cli.StringFlag{Name: "dbhost", Usage: "log database host to connect to"}
```
life1347: Yes, for sure. That would definitely help.

life1347: I think we can use Kubernetes Secret to store sensitive data, and then mount into the controller pod. Once the user try to show function log, fission-cli make a request to controller and then receiving log stream from controller, just like how kubernetes-api and kubectl worked.

atiwari:
Wondering when this feature will be merged?

life1347:
Hi @soamvasani, I've pushed the commit with following modification.
1. Move logger to top-level
2. Integrate logger with fission-bundle
3. Add influxdb deployment
4. Fix minor issues
And thanks for the code review.  :D

life1347:
Don't merge it right now, seems the logger failed to create symlink with latest commit.

life1347:
With minikube, pod may fail to resolve node name (https://github.com/kubernetes/minikube/issues/757). In the latest commit, poolmgr use host ip instead of node name to prevent dns problem.
And thanks for @yqf3139's help/test. :)

README.md
```diff
@@ -182,6 +182,39 @@ Finally, you're ready to use Fission!
...
+  $ cd $GOPATH/src/github.com/fission/fission/fission-logger.yaml
+  $ vim fission-logger.yaml # replace information in yaml file
+  $ kubectl apply -f fission-logger.yaml
+```
+
+Display function logs
+
+```
+  # Currently only influxDB is supported, use --help to see more options
+  $ export FISSION_LOGDB=http://$(kubectl --namespace fission get svc influxdb -o=jsonpath='{..clusterIP}'):8086
```
soamvasani: I don't understand why clusterIP would work here?  The clusterIP of a service is usually not accessible from outside the cluster.

fission/function.go
```diff
@@ -276,3 +279,134 @@ func fnEdit(c *cli.Context) error {
...
+
+func fnLogs(c *cli.Context) error {
+	client := getClient(c.GlobalString("server"))
+
+	fnName := c.String("name")
+	if len(fnName) == 0 {
+		fatal("Need name of function, use --name")
+	}
+
+	dbHost := c.String("dbhost")
```
soamvasani: We should also all users to use env var here, instead of specifying this every time (also ok to do in a later change)

poolmgr/gp.go
```diff
@@ -487,6 +514,23 @@ func (gp *GenericPool) CleanupFunctionService(podName string) error {
 		return nil
 	}
 
+	pod, err := gp.kubernetesClient.Core().Pods(gp.namespace).Get(podName)
+	if err != nil {
+		return err
+	}
+
+	loggerUrl := fmt.Sprintf("http://%s:1234/v1/log/%s", pod.Spec.NodeName, pod.Name)
```
soamvasani: This port should be configurable (as an env var); but we can do that in a later change.

poolmgr/gp.go
```diff
@@ -262,6 +264,31 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *fission.Metadata) er
 		return errors.New(fmt.Sprintf("Error from fetcher: %v", resp.Status))
 	}
 
+	logReq := logger.LogRequest{
```
soamvasani: can you move this whole thing into a separate function?  So we can just have `gp.setupLogging(pod)` here

poolmgr/gp.go
life1347: I'll add the env support in later PR with changes that cli retrieves log stream from controller instead of connecting to database directly.

poolmgr/gp.go
soamvasani: Sorry I should have been more clear -- I meant the whole functionality of setting up the logger, not just the one statement (upto line 285)

soamvasani:
So I'm trying to set this up locally to test it out, and I'm not sure I understand the instructions -- after the influxdb username and password are created, how does one retrieve them?

life1347:
Yes, that's the drawback of the current design. A user needs to know the username/password in advance when tries to access function logs. 
I'm working on the feature that controller helps to aggregate logs and return log stream to the user, so that users do not need to know the sensitive info when they try to access logs.

soamvasani:
Yes, I understand the drawback and I'm ok with it, I'm just trying to understand how to get it working on my setup :)  Once I run the db, where do I get the credentials from?

soamvasani:
(I rebased life1347/function-logs-support to fission/master)

PR #136, opened 2017-03-02
Fix Markdown table

RobertHerhold:
Before:

image

After:

image


soamvasani:
Thanks! (I broke this while trying to use emacs org-mode to edit the markdown table.)

PR #140, opened 2017-03-04
fix(kubeEventsSlack): typo and wrong variable name

Pindar:
- change comma to semicolon
- change variable name `path` to `slackWebhookPath`

example is not working without these changes.


soamvasani:
Thanks!

Issue #139, opened 2017-03-04
poolmgr fails while creating, invoking and deleting function continually


yqf3139:
Use the following script to create a temp function, invoke it's default http trigger and then delete it continually.

```
COUNTER=1
while true;
do
fission fn create --name tmp$COUNTER --env python --code hello.py
sleep 4
curl 127.0.0.1:31314/fission-function/tmp$COUNTER
fission fn delete --name tmp$COUNTER
let "COUNTER++"
done

```
In a fresh minikube, with fresh poolmgr and router, the first 12 test is successful, but errors occurred from the 13th test.

```
function 'tmp8' deleted
function 'tmp9' created
Hi
function 'tmp9' deleted
function 'tmp10' created
Hi
function 'tmp10' deleted
function 'tmp11' created
Hi
function 'tmp11' deleted
function 'tmp12' created
Hi
function 'tmp12' deleted
function 'tmp13' created
Internal server error (fission)
function 'tmp13' deleted
function 'tmp14' created
Internal server error (fission)
function 'tmp14' deleted
function 'tmp15' created
Internal server error (fission)
function 'tmp15' deleted
function 'tmp16' created
```

router logs:
```
2017-03-04T06:59:02.308183246Z 2017/03/04 06:59:02 Proxying request for /fission-function/tmp8 to 172.17.0.25:8888
2017-03-04T06:59:02.309984150Z 172.17.0.1 - - [04/Mar/2017:06:59:02 +0000] "GET /fission-function/tmp8 HTTP/1.1" 200 3
2017-03-04T06:59:06.339363047Z 2017/03/04 06:59:06 Not cached, getting new service for {tmp9 26c00d74-7338-4e49-bf00-fd9587128b44}
2017-03-04T06:59:06.384710543Z 2017/03/04 06:59:06 Proxying request for /fission-function/tmp9 to 172.17.0.27:8888
2017-03-04T06:59:06.387306939Z 172.17.0.1 - - [04/Mar/2017:06:59:06 +0000] "GET /fission-function/tmp9 HTTP/1.1" 200 3
2017-03-04T06:59:10.449102423Z 2017/03/04 06:59:10 Not cached, getting new service for {tmp10 ccfaac47-5bdf-4880-8647-defad6b41b23}
2017-03-04T06:59:10.538835410Z 2017/03/04 06:59:10 Proxying request for /fission-function/tmp10 to 172.17.0.28:8888
[...]

PR #141, opened 2017-03-05
Fixed pod has no ip (#139)

life1347:
This PR aims to fixed the issue (https://github.com/fission/fission/issues/139) when fission is busy at creating functions. 
Here is the explanation of the root cause. (https://www.goinggo.net/2013/09/iterating-over-slices-in-go.html)


soamvasani:
Excellent find, @life1347.  Now I need to check the rest of the code to see if I made this mistake anywhere else...

PR #142, opened 2017-03-07
Fix function delete with uid

yqf3139:
The current version may delete the function record and left some function code behind if an uid is provided.

If no uid provided, all versions of function files and record will be deleted.
If an uid provided, delete the specific version code, update the function with the latest code.

### DIFF
--- a/controller/api_test.go
+++ b/controller/api_test.go
@@ -126,6 +126,61 @@ func TestFunctionApi(t *testing.T) {
 	panicIf(err)
 }
 
+func TestFunctionVersionApi(t *testing.T) {
+	testFunc := &fission.Function{
+		Metadata: fission.Metadata{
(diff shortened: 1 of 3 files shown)


yqf3139:
Hi @soamvasani, will this be affected by TPR investigation?

soamvasani:
Thanks @yqf3139 , no that won't affect this.  I'm just a bit behind on reviews, sorry.  Should be merged in a day or so.

yqf3139:
Thanks, I am not sure whether the action taken in this commit is correct.

soamvasani:
@yqf3139, would you mind adding a test to api_test.go?  Both for the case of deleting something that has an old version, and something that doesn't?

yqf3139:
@soamvasani Ok.

Issue #144, opened 2017-03-07
text/plain response instead of application/json


RobertHerhold:
Many of the endpoints, like `/v1/triggers/http` return JSON data yet the `Content-Type` header is `text/plain` instead of `application/json`.

PR #147, opened 2017-03-10
Set correct Content-Type in the http response

LingxianKong:
When the request is handled successfully, json data is returned, this
patch changes the response Content-Type header from text/plain to
application/json

Fixes: #144

### DIFF
--- a/controller/api_test.go
+++ b/controller/api_test.go
@@ -120,6 +120,20 @@ func TestFunctionApi(t *testing.T) {
 	assert(len(funcs) == 2,
 		"created two functions, but didn't find them")
 
+	funcs_url := g.client.Url + "/v1/functions"
+	resp, err := http.Get(funcs_url)
+	panicIf(err)
+	defer resp.Body.Close()
+	assert(resp.StatusCode == 200, "http get status code on /v1/functions")
+
+	var found bool = false
+	for _, b := range resp.Header["Content-Type"] {
+		if b == "application/json; charset=utf-8" {
+			found = true
+		}
+	}
+	assert(found, "incorrect response content type")
+
 	err = g.client.FunctionDelete(&fission.Metadata{Name: "foo"})
 	panicIf(err)
 	err = g.client.FunctionDelete(&fission.Metadata{Name: "bar"})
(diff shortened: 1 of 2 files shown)


RobertHerhold:
I was literally just about to make this pull request! Looks like you beat me to it 😂

LingxianKong:
@RobertHerhold hi, sorry to do this without letting you know, I don't know how to do this in Github (I am using Launchpad and Gerrit for the other projects I am contributing, Github is totally a different thing for me). Could you please tell me how to make myself an assignee before coding to avoid many people are doing the same thing at the same time?

RobertHerhold:
Collisions like this rarely happen, so it's usually not an issue. Only people with write access to the repository can assign people to an issue, but it's usually good enough to just drop a comment saying you're working on it.

LingxianKong:
@RobertHerhold Thanks, I will leave something before coding for my next bugfix :)

controller/api_test.go
```diff
@@ -282,6 +282,15 @@ func TestMain(m *testing.M) {
 	resp, err := http.Get("http://localhost:8888/")
 	panicIf(err)
 	assert(resp.StatusCode == 200, "http get status code on root")
+
```
soamvasani: This is only testing the homehandler; could you also test one of the other endpoints?

soamvasani:
Thanks, @LingxianKong!  (and @RobertHerhold too :) )

controller/api_test.go
LingxianKong: @soamvasani done :)

LingxianKong:
hi, @soamvasani , is there anything I still need to improve?

controller/api_test.go
```diff
@@ -120,6 +120,19 @@ func TestFunctionApi(t *testing.T) {
 	assert(len(funcs) == 2,
 		"created two functions, but didn't find them")
 
+	funcs_url := g.client.Url + "/v1/functions"
+	resp, err := http.Get(funcs_url)
+	panicIf(err)
+	assert(resp.StatusCode == 200, "http get status code on /v1/functions")
```
soamvasani: Thanks for adding this test! You need a `defer resp.Body.Close()` somewhere here (it's only a test so it doesn't really matter, but still, good to add it).

PR #148, opened 2017-03-10
Return 201 for created resources

### LINKED ISSUE #143: Resource create should return a Location header
image

Deleting a non-existant resource throws a 500, but I would think it would make more sense as a 404. There are also other status code discrepancies; creating a resource returns 200 instead of 201

RobertHerhold:
Partially resolves #143 

One weird issue:

I was trying to set the `Location` header with 

```go
w.Header().Set("Location", "/v1/functions/"+f.Name)
```

but it always ended up being stripped out of the response for some reason. Any ideas?

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -84,6 +84,14 @@ func (c *Client) handleResponse(resp *http.Response) ([]byte, error) {
 	return body, err
 }
 
+func (c *Client) handleCreateResponse(resp *http.Response) ([]byte, error) {
+	if resp.StatusCode != 201 {
+		return nil, fission.MakeErrorFromHTTP(resp)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	return body, err
+}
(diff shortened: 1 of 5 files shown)


soamvasani:
Thanks, @RobertHerhold !

Would you mind looking at the test failure?  You can find logs by clicking on details and then the Go version on that page.  Here they are: https://travis-ci.org/fission/fission/jobs/209716921

RobertHerhold:
@soamvasani All fixed!

soamvasani:
Thanks @RobertHerhold .  Before I merge this, can the fission-ui folks @dgoujard @yqf3139 confirm this doesn't break the ui code?  This changes HTTP status from 200 to 201 for create requests, compliant to HTTP.

yqf3139:
@soamvasani @RobertHerhold Hi, thanks for the reminder. It is ok to change 200 to 201.

PR #149, opened 2017-03-12
use fmt.Errorf instead of error.New()

maxwell92:
use fmt.Errorf() instead of error.New() in line 49


soamvasani:
@maxwell92 this doesn't compile.  Please take a look at the travis failure.

RobertHerhold:
Looks like there's an extra parenthesis

PR #151, opened 2017-03-12
Upgrade node environment to Node.js 7.6.0+

RobertHerhold:
This PR upgrades the Node.js environment to Node.js 7.6.0+. Node 7.6.0 introduced async/await without the harmony flag, which allows you to write cleaner (asynchronous) functions without the mess that is callbacks. With this change, Node environment functions now look like this:

```javascript
module.exports = async function(context) {
    const response = await request({ url: 'http://google.com'});
    const anotherResponse = await request({ url: 'http://google.com'});
    return {
        status: 200,
        body: "Hello, world!\n"
    };
}
```

Changelog:
 * Upgrade Node.js to 7.6.0+
 * Add support for `async` functions
 * Replace `request-promise` with `request-promise-native`

### DIFF
--- a/environments/nodejs/server.js
+++ b/environments/nodejs/server.js
@@ -81,7 +81,7 @@ app.all('/', function (req, res) {
         response: res
         // TODO: context should also have: URL template params, query string
     };
-    
+
     function callback(status, body, headers) {
         if (!status)
             return;
@@ -100,8 +100,21 @@ app.all('/', function (req, res) {
     // you can do that here by adding properties to the context.
     //
 
-    // Make sure their function returns a promise
-    let functionProm = Promise.resolve(userFunction(context));
+    let functionProm;
+    if (userFunction.length === 1) { // One argument (context)
+        // Make sure their function returns a promise
+        functionProm = Promise.resolve(userFunction(context));
+    } else { // 2 arguments (context, callback)
+        functionProm = new Promise(function(resolve, reject) {
+            userFunction(context, function(status, body, headers) {
+                return resolve({
+                    status,
+                    body,
+                    headers
+                });
+            });
+        });
+    }
 
     functionProm.then(function(result) {
         callback(result.status, result.body, result.headers);
--- a/environments/nodejs/server.js
+++ b/environments/nodejs/server.js
@@ -103,25 +103,20 @@ app.all('/', function (req, res) {
     let functionProm;
     if (userFunction.length === 1) { // One argument (context)
         // Make sure their function returns a promise
-        functionProm = Promise.resolve(userFunction(context));
-    } else { // 2 arguments (context, callback)
-        functionProm = new Promise(function(resolve, reject) {
-            userFunction(context, function(status, body, headers) {
-                return resolve({
-                    status,
-                    body,
-                    headers
-                });
-            });
+        Promise.resolve(userFunction(context)).then(function(result) {
+            callback(result.status, result.body, result.headers);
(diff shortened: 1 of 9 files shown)


soamvasani:
Thanks @RobertHerhold!  This is really great but we can't change the function interface so easily -- user functions would break -- we have to preserve compatibility.

Options:
1. We could turn this into a new separate "nodejs7" environment.
2. Or we could check at runtime whether the user's function wants two args or one, and accordingly pass in a callback or not.

1 is simpler but the user has to decide which to use.  2 is transparent to users but a little more work to implement.  I'm a little bit in favor of 2 but haven't thought about it enough.  What are your thoughts?

RobertHerhold:
@soamvasani 
Ideally, I think we could just version the Node 7 image as v2 and deprecate v1, but there is only a `latest` tag that everyone is currently using. Checking for a callback argument is pretty simple, though. I just added it in my latest commit

environments/nodejs/server.js
```diff
@@ -90,18 +92,37 @@ app.all('/', function (req, res) {
...
+    //
+    // If you want to modify the context to add anything to it,
+    // you can do that here by adding properties to the context.
+    //
+
+    let functionProm;
+    if (userFunction.length === 1) { // One argument (context)
+        // Make sure their function returns a promise
+        functionProm = Promise.resolve(userFunction(context));
+    } else { // 2 arguments (context, callback)
```
soamvasani: Thanks for adding this and preserving compatibility.  Could you keep one example in the old style so it can be tested?  Maybe just make a copy of the old `hello.js` as `hello-callback.js`?

examples/nodejs/stock.js
```diff
@@ -2,29 +2,36 @@
...
-
-            try {
-                var parsed = JSON.parse(resp.slice(3));
-                var lastTrade = parsed[0]['l_cur']
-                callback(200, `{ "text": "${symbol} last traded at ${lastTrade}" }`);
-            } catch (e) {
-                callback(200, `{ "text": "Error (invalid NYSE symbol?)" }`);
-            }
+    return new Promise(function(resolve, reject) {
+        http.get({
```
soamvasani: So can this be changed to an `await request(...)`, like in your commit message?  That would make this example simpler, right?  

Also this library [node-fetch](https://github.com/bitinn/node-fetch) seems to make the response fetching stuff simpler too.  Would be good to use that and make this even simpler.

RobertHerhold: @soamvasani There are a number of Promise-based HTTP clients available for Node, which would indeed be much cleaner. However, wouldn't adding additional dependencies require a custom fission environment? I assumed you wanted to use the native `http` module to avoid adding dependencies.

soamvasani: I don't mind using a dependency if it is very widely used and makes the code easier to read and write. (But we can totally do this later, I don't want to block this PR for this example.)

In your PR description above, you seem to indicate you can use `await request` without adding a dependency; maybe I misunderstood it?

RobertHerhold: Yes, that was just a generic example to demonstrate async/await. Do you want me to just add a request library to the standard node environment?

soamvasani:
This is great, thanks for updating the examples too!  Will merge it now and push the docker image.

soamvasani:
These changes are now in the fission/node-env docker images: `fission/node-env:latest` and `fission/node-env:node7.6.0`

Issue #153, opened 2017-03-13
Add timer trigger API and client


yqf3139:
Timer trigger is quite useful. It invokes the target function like cron.

Related actions:
- List: return the timer uid, function metadata, cron variables and status
- Create: create a stopped timer trigger by sending a timer trigger metadata (uid, name optional for description ), a function metadata, cron variables in a string and status(running, stopped)
- Delete: by sending a timer metadata
- Update: by sending a timer metadata, cron variables and status

Features nice to have: parameter binding, history

tobilarscheid:
How would that interface with kubernetes cron jobs?

yqf3139:
@tobilarscheid It is a separated implementation and does not interfere with kubernetes cron jobs.

soamvasani:
Fixed by #161

PR #161, opened 2017-03-21
Add Time Trigger API and client (#153)

yqf3139:
Use [robfig/cron](https://godoc.org/github.com/robfig/cron) for cron implementation.
fission.yaml should be updated because of new docker image is needed.

Usage:
```bash
fission tt list
# use a new uuid to name a time trigger
fission tt add --cron '@every 1m' --function foo
# specify a name and more infomation
fission tt add --name timer1 --cron '@hourly' --function foo --desc 'some desc'
# update trigger, the related function is not updatable
fission tt update --name timer1 --cron '0 30 * * *' --desc 'some desc 2'
fission tt delete --name timer1
```


soamvasani:
Thanks @yqf3139 !

I don't think we should model the timer as a crontab string in the API.  We should model it explicitly as a struct with fields for second/minute/hour/day etc.

yqf3139:
robfig/cron provides the underlying heavy lifting to parse crontab string and set the clock.
The crontab string is powerful for describing timers. If we split the fields into fine granularity, we may end up doing the same thing.
Or the ideas is we split the '0 30 * * *' into second:0, minute:30, hour: *, day of month: *, month: *, day of week:* ? And we reduce them into one crontab string and pass it to the underlying library. 
Also robfig/cron provides '@every 2m' '@hourly' which I think is quite useful, we can keep that.

soamvasani:
It's easy to add to the API but hard to remove functionality.  What if we start with something very simple like, 3 fields:

  1. When should this timer first run? (time in UTC)
  2. Should this timer repeat? (bool) 
  3. How often should this timer repeat? (msec)

And see if that satisfies the use cases?

If it's not good enough and people want cron compatibility, we could always come back and add this.

yqf3139:
Thanks @soamvasani to think through this.

I started to use cron because I thought cron is the timer standard for developers. And also I see AWS uses cron string and schedule for timers.
The new design is simple and powerful because it can describe some timer cron can not do easily. For example, run some job at a target time.
But for some common cases like, run at 12:00, 9:00 in working days needs 2*5 timers in our format.

If we still use the new design, I think it will be done in another pr.

LingxianKong:
I agree with @yqf3139, cron pattern is well known to developer. In another openstack project i am contributing to, we use cron string to create a cron-trigger resource.

soamvasani:
Okay, I will get over my fear of cron strings :)

[Sorry for the delays so far and next couple of weeks, I'm traveling right now.]

LingxianKong:
enjoy your travel :)

soamvasani:
Ok, finally looking at this. Sorry again for the delay @yqf3139. Nice work overall!

Couple of questions:

1. What is the use case for the "status" flag?  The user could simply delete the trigger when they don't want it to run.

2. Let's figure out a way to avoid the copy pasted webhookPublisher. I'll move the publisher interface + webhookpublisher implementation to the top level, and we can reuse it here (there's some kubernetes-specific stuff in the kubewatcher one, which I'll need to change).

yqf3139:
Thanks @soamvasani for reviewing.
Yes, the `status` makes things complicated and not easy to understand. I have not found another use case for this field, so `status` can be removed.

yqf3139:
Hi @soamvasani, I made some changes in eeb85a1 to share the webhook Publisher and remove the `status` field.

soamvasani:
Thanks for taking on this refactor! 

Can you move the publisher interface and implementation into a subdirectory called `publisher/` instead of the top level?

publisher.go
```diff
@@ -14,17 +14,15 @@ See the License for the specific language governing permissions and
...
+import "net/http"
 
 type (
 	Publisher interface {
-		// Publish an event to a "target".  Target's meaning depends on the
+		// Publish an request to a "target".  Target's meaning depends on the
 		// publisher: it's a URL in the case of a webhook publisher, or a queue
 		// name in a queue-based publisher such as NATS.
-		Publish(event watch.Event, target string)
+		Publish(req *http.Request)
```
soamvasani: The intent of this interface is to abstract over HTTP calls as well as async messages via a message queue.  So we don't want it to have an http request in the interface.  

How about: `Publish(body string, headers map[string]string, target string)` (we'd still need to serialize that into a byte stream later, but we can discuss that in another change).

yqf3139: @soamvasani 

Yes. I think of that for a moment. But we have to pass almost all the params to build a http request to the channel (body, headers, target, method), and in the service we have only the http request as the protocol. So I used the request instead.
To make things simpler, we may restrict the protocol to a subset of http request. For example, only post method with string as body and custom headers as you suggest. And then your interface is universal for all the trigger events.

soamvasani: It'sn not a good idea to have an http request object in there when it may not always be published over http.  (I know the only implementation right now is http, but it won't always be the case.)

I don't understand the last part of your comment -- what do you mean by "restrict the protocol"?  Are you suggesting a different publisher interface?

yqf3139: I will use the interface you suggest. 
For the timer, I use the `GET` method, switching to `POST` is fine for me. So we restrict the method to `POST` or we add a method field ?

soamvasani: Ah, I missed that. Yes, we should use `POST`. `GET` is supposed to be side-effect-free.

soamvasani:
Nice work, @yqf3139 !  This look almost ready to merge.

I renamed 'watchSync' to 'timerSync' since it actually syncs timers.

I'm also going to remove the `description` field.  While there is nothing wrong with it, it doesn't fit the pattern of the rest of the fission or k8s APIs.  (With future API revisions, the user will be able to add information to the object using labels or annotations.)

yqf3139:
Thanks for reviewing and making the changes.

PR #163, opened 2017-03-24
Go build helper script

soamvasani:
Make it a little easier to compile fission functions in go.

cc @nouney


soamvasani:
Going to merge this, @nouney if you have any comments I can address them later.

PR #170, opened 2017-04-05
Make the chart work with helm 2.2

apenney:
Primarily this involved removing the release namespace, which is
controlled by helm.  If you want resources in a namespace you set
--namespace= with helm install, anything else leaves helm unable
to properly manage the resources.

I think the functions namespace won't work either on a delete.

### DIFF
--- a/charts/fission/templates/deployment.yaml
+++ b/charts/fission/templates/deployment.yaml
@@ -1,11 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: {{ .Release.Namespace }}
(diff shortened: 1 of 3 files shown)


soamvasani:
> removing the release namespace

Thanks! This makes sense.

> I think the functions namespace won't work either on a delete.

Could you elaborate?  Do you mean `helm delete` won't be able to delete the fission-function namespace?

soamvasani:
Thanks for the PR @apenney. I've filed #180 to track the issue with the functions namespace.

PR #171, opened 2017-04-07
Lighten up the python3 base image (alpine)

syassami:
- Saves 241 MB from the base example image, bringing things down to 242MB
  in total.


soamvasani:
Very nice, thanks @syassami !

PR #175, opened 2017-04-20
Fix FISSION_URL & FISSION_ROUTER export instructions

kunalparmar:


soamvasani:
Hi, Kunal!  Thanks for the PR!  

But ClusterIP of a service isn't usually routable from outside the cluster.  The "..ip" jsonpath returns pod.status.loadBalancer.ingress.ip (this is under the "gke or other cloud" section).

kunalparmar:
@soamvasani I had a AWS cluster that "..ip" didn't work for. ClusterIP also did not work. There was a hostname somewhere that eventually did the trick. I will go find it and update here what that was.

Feel free to close this ticket.

soamvasani:
Ok, closing. A bit puzzled about the cause though, if you find out later please let me know. Thanks.

Issue #50, opened 2016-12-09
pod name hits limit of 63 chars if env name is > 17 chars


kphatak:
pod name hits limit of 63 chars if env name is greater than 17 chars
may want to drop adding entire uid

`2016/12/09 06:42:23 eager-create pool failed: Deployment.extensions "fission-alpha9-mod-b20e9823-72a8-4cbd-8ecc-d6fbfef8dadb-hxyrtehs" is invalid: [spec.selector.matchLabels: Invalid value: "fission-alpha9-mod-b20e9823-72a8-4cbd-8ecc-d6fbfef8dadb-hxyrtehs": must be no more than 63 characters, spec.selector: Invalid value: {"matchLabels":{"pool":"fission-alpha9-mod-b20e9823-72a8-4cbd-8ecc-d6fbfef8dadb-hxyrtehs"}}: invalid label selector.]`

life1347:
Solved by #975

PR #182, opened 2017-05-01
Shorten name given to environment deployments

erwinvaneyk:
This small PR is a proposal to shorten the name of the generic pods to something a bit more readable. There does not seem to be reason to add the identifiers to this name, as these are also present in the labels. It might be that I am overlooking a detail, that would invalidate this solution. Anyhow, this would also mitigate the issue #50 of pod names hitting the 64 character limit of Kubernetes. Finally, I added the `env-` as a prefix to avoid conflicts with existing components and to avoid unclear naming. 

Before:
`nodejs-1f348359-5f29-435e-b4e1-0f86b8398d64-hv6j3tek-1519022795`
After:
`env-nodejs-3126632610`

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/182)

### DIFF
--- a/poolmgr/gp.go
+++ b/poolmgr/gp.go
@@ -26,6 +26,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/dchest/uniuri"
@@ -309,13 +310,13 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *fission.Metadata) er
 // A pool is a deployment of generic containers for an env.  This
 // creates the pool but doesn't wait for any pods to be ready.
 func (gp *GenericPool) createPool() error {
-	poolDeploymentName := fmt.Sprintf("env-%v", gp.env.Metadata.Name)
+	poolDeploymentName := fmt.Sprintf("env-%v-", strings.ToLower(gp.env.Metadata.Name))
 
 	sharedMountPath := "/userfunc"
 	deployment := &v1beta1.Deployment{
 		ObjectMeta: v1.ObjectMeta{
-			Name:   poolDeploymentName,
-			Labels: gp.labelsForPool,
+			GenerateName: poolDeploymentName,
+			Labels:       gp.labelsForPool,
 		},
 		Spec: v1beta1.DeploymentSpec{
 			Replicas: &gp.replicas,
--- a/poolmgr/gp.go
+++ b/poolmgr/gp.go
@@ -26,7 +26,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


soamvasani:
Thanks for this PR, @erwinvaneyk ! 

If poolmgr restarts, we create a new deployment and delete the old one. If they were named the same, they'd conflict. That's the reason for instanceId in the name. So we might have to keep that one.

I think the env- prefix is unnecessary? Are there conflicts with existing components? We keep all this in the separate fission-function namespace.  Also, the name gets copied into pods etc., so pods will become env-nodejs-xxx etc.

erwinvaneyk:
@soamvasani thanks for the comments! 

Keeping the entire instanceId in the name might still result in the lengthy deployment names that we try to avoid. Of course, if there is no other way, it should be there. What about putting the instanceId and other identifiers in the metadata/annotations of the deployment? 

The `env-` prefix might indeed be a bit unnecessary. I did not consider the difference in namespaces. I'll remove that.

soamvasani:
Yeah, unfortunately the deployment name has to be unique, otherwise a restart of the poolmgr could fail.  That's why we either need the instanceId or some other way to make it unique. We could use the GenerateName property instead of Name.  That would work too, k8s will add a unique suffix.

(Here's the metadata api docs if you need them https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta )

erwinvaneyk:
Almost forgot about this pr. Nice idea with using the GenerateName @soamvasani! That results in the following names, which are of reasonable length:

```
NAME                                READY     STATUS    RESTARTS   AGE
env-nodejs-6h5ln-2309684998-gh8t6   2/2       Running   0          2m
env-nodejs-6h5ln-2309684998-jw964   2/2       Running   0          2m
env-nodejs-6h5ln-2309684998-zs4x7   2/2       Running   0          2m
```

erwinvaneyk:
@soamvasani what is your plan with this PR? If we are not planning on addressing this one, let's close it for now - we can keep the branch

poolmgr/gp.go
```diff
@@ -310,14 +310,13 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *fission.Metadata) er
 // A pool is a deployment of generic containers for an env.  This
 // creates the pool but doesn't wait for any pods to be ready.
 func (gp *GenericPool) createPool() error {
-	poolDeploymentName := fmt.Sprintf("%v-%v-%v",
-		gp.env.Metadata.Name, gp.env.Metadata.Uid, strings.ToLower(gp.poolInstanceId))
+	poolDeploymentName := fmt.Sprintf("env-%v-", strings.ToLower(gp.env.Metadata.Name))
```
life1347: remove the last `-` in `env-%v-`

Issue #160, opened 2017-03-19
RBAC rules 


soamvasani:
Kubernetes will soon (1.6?) have RBAC (Role Based Access Control) enabled by default.  Our helm chart as well as fission.yaml will need some rules added so that the `fission` namespace service account is allowed to create resources in the `fission-function` namespace.

gamefiend:
I just cooked up the RBAC code to get it working in my 1.6 cluster.  Mind if I submit a PR with the changes?

soamvasani:
Nice! Yes, please do!

soamvasani:
This is in pretty good shape now; it's in the helm chart and tested in the e2e tests, so I'm going to close this issue. Thanks, @gamefiend !

PR #183, opened 2017-05-04
Adding fission-rbac.yml for 

gamefiend:
To deal with the issue of lack of RBAC permission noted in fission/fission#160 , created 'fission-rbac.yml'. Made it a separate file so that users have choice (in case they use older k8s) on whether to install with RBAC permissions or not.

If changes are approved, happy to help update documentation to reflect changes.

### DIFF
--- a/fission-rbac.yaml
+++ b/fission-rbac.yaml
@@ -1,0 +1,180 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fission-svc
+  namespace: fission
+  labels:
+    kubernetes.io/cluster-service: "true"
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: fission-admin
+  labels:
+    kubernetes.io/cluster-service: "true"
+subjects:
+  - kind: ServiceAccount
+    name: fission-svc
+    namespace: fission
+roleRef:
+  kind: ClusterRole
+  name: admin
+  apiGroup: rbac.authorization.k8s.io
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
+  name: fission-function-admin
+  namespace: fission-function
+  labels:
+    kubernetes.io/cluster-service: "true"
+subjects:
+  - kind: ServiceAccount
+    name: fission-svc
+    namespace: fission
+roleRef:
+  kind: ClusterRole
+  name: admin
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: fission
+  labels:
+    name: fission
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: fission-function
+  labels:
+    name: fission-function
+---
+apiVersion: extensions/v1beta1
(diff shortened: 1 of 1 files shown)


fission-rbac.yaml
```diff
@@ -0,0 +1,180 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fission-svc
+  namespace: fission
+  labels:
+    kubernetes.io/cluster-service: "true"
```
soamvasani: What's this label for? As I understand it, the cluster-service label is used for cluster addons... but i can't find some definitive documentation either way.

fission-rbac.yaml
```diff
@@ -0,0 +1,180 @@
...
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+metadata:
```
soamvasani: Should the RoleBinding be namespaced too?

fission-rbac.yaml
gamefiend: doh, that is something used out of habit. Will change.

fission-rbac.yaml
gamefiend: Yep! Will fix.

fission-rbac.yaml
```diff
@@ -0,0 +1,175 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: fission-svc
+    namespace: fission
+roleRef:
+  kind: ClusterRole
+  name: admin
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: v1
```
soamvasani: Sorry, one more comment: can you remove the contents of fission.yaml from this file?  Less duplication that way.

gamefiend: Makes sense! I had originally done it this way for my personal  convenience.

Slahser:
i am new for fission

this pr maybe not a clear permission for fission i think

is this project need 'admin' role to operate?

fission-rbac.yaml
```diff
@@ -0,0 +1,175 @@
...
+kind: ServiceAccount
+metadata:
+  name: fission-svc
+  namespace: fission
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1alpha1
```
Slahser: in kubernetes v1.6.2 it should be `apiVersion: rbac.authorization.k8s.io/v1beta1`

fission-rbac.yaml
```diff
@@ -0,0 +1,175 @@
...
+  name: fission
+  labels:
+    name: fission
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: fission-function
+  labels:
+    name: fission-function
```
Slahser: maybe we should change the 'Namespace ' position to top

soamvasani:
Hi @gamefiend, I think I'll just merge this, since it's working (I tested it on 1.6.0) and useful to other people.  Other improvements can be made in subsequent changes.  Thanks a lot for doing this work, much appreciated!

@Slahser thanks for your comments too, perhaps we can investigate binding some role more specific than `admin` in a future change.

gamefiend:
Thanks @soamvasani ! Catching up after OSCON and have quite the backlog :)

PR #185, opened 2017-05-13
Collect and stream function logs to client at controller

life1347:
This PR tries to let controller aggregates function logs and streaming logs to the client side. Now, the client sends a RESTful HTTP request to the controller and receive logs from streaming HTTP response. So we don't need to expose database user/password to end user. Also, it's very helpful to integrate with another service with simple HTTP call.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -131,25 +131,11 @@ Finally, you're ready to use Fission!
 Fission uses InfluxDB to store logs and fluentd to forward them from
 function pods into InfluxDB.  To setup both InfluxDB and fluentd:
 
-Edit `fission-logger.yaml` to add a username and password for the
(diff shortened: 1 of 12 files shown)


soamvasani:
Hi @life1347, I'm wondering if we really need to do this.  Logs can be quite a bit of data and we should think carefully about adding more complexity to the log data path.

What if the controller simply provided a proxy server to influxdb at a certain URL?  Then the client would use that as the url instead of influxdb, and influxdb wouldn't be exposed as an externally visible service.

We'd have a single point of access + simpler data path.  What do you think?

life1347:
Close this PR. Please refer https://github.com/fission/fission/pull/207 instead.

Issue #187, opened 2017-05-17
404 getting fission-openshift.yaml


methadata:
Hi!

When deploying Fission following the steps in the README, there's a 404 when trying to execute:
```
oc create -f http://fission.io/fission-openshift.yaml
```

Could you upload that file from the repository to the fission.io site? This way, the procedure is consistent with other plattforms (i.e. k8s), and there's no need to change the README file.

Thanks!

soamvasani:
Whoops, fixed now.  Thanks!

Issue #177, opened 2017-04-23
Failed to create environment: unexpected end of JSON input


jonathan-kosgei:
I've exported the following svc urls
```
jonathan@sc-test-cluster:~$ export FISSION_URL=http://10.91.98.28:31313
jonathan@sc-test-cluster:~$ export FISSION_ROUTER=10.91.98.28:31314
```

I'm on azure which doesn't support type Loadbalancer so I've customised the fission-cloud.yml file as follows
```
apiVersion: v1
kind: Service
metadata:
  name: router
  namespace: fission
  labels:
    svc: router
spec:
  type: NodePort
  ports:
  - port: 80
    targetPort: 8888
    nodePort: 31314
  selector:
    svc: router

---
apiVersion: v1
kind: Service
metadata:
  name: controller
  namespace: fission
  labels:
    svc: controller
spec:
  type: NodePort
  ports:
  - port: 80
    targetPort: 8888
    nodePort: 31313
  selector:
    svc: controller
```

I'm trying to use type `nodePort`. My node's ip is `10.91.98.28`.

I get the following error when trying to add an environment
```
Failed to create environment: unexpected end of JSON input
```

soamvasani:
Apologies, I pushed an incompatible client/server combination and broke this. It is fixed now, you can do this to fix an existing setup:

```
kubectl apply -f http://fission.io/fission.yaml
```

(and, we're working on better release automation and testing so this doesn't happen again)

jonathan-kosgei:
Hi no prob, I've been able to get it working with nodeport, however an issue a bit more specific to my usecase, can't I use fission to monitor thirdpartyresources?

I have resources of type CronTab, and I create a function to monitor them, however I get tonnes of the following error on the kubewatcher pod

```
2017/04/25 10:03:50 (re)starting watch {0f90bf09-86ea-4cc1-8a1f-428141b8fe0e a041cddd-a3eb-4de5-9238-82d13e3af034} (ns:default type:repos) at rv:
2017/04/25 10:03:50 Error: unknown obj type 'crontab'
```

jonathan-kosgei:
I can `kubectl get crontab` and see the resources I've created

soamvasani:
@jonathan-kosgei kubewatcher is missing support for TPRs. Let's use a new issue to track this -- I've created #178.

PR #188, opened 2017-05-18
Bump version of fission-bundle to testing20170328

panchoh:
Using the old version produces an error, just as described on #177.

Also, I kindly request this version to be uploaded to the web site, so that it can be obtained via http://fission.io/fission-openshift.yaml.
Note that the version that was placed there yesterday (as requested on #187) was missing the first paragraph.

Thanks in advance!

### DIFF
--- a/fission-openshift.yaml
+++ b/fission-openshift.yaml
@@ -66,7 +66,7 @@ spec:
         emptyDir: {}
       containers:
       - name: controller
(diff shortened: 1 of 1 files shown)


soamvasani:
Thanks for the PR and sorry for the delay @panchoh. Fixed fission-openshift.yaml and uploaded to fission.io (the image version should be alpha20170328, not testing*)

PR #207, opened 2017-05-24
Retrieve function logs from controller

life1347:
This PR tries to let controller establishes a reverse proxy to log database, and redirect query command send from client to database and then proxy back the db responses. So that we dont need to expose database to the public network.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -151,18 +151,6 @@ particular pod:
   $ fission function logs --name hello --pod <pod name>
 ```
 
-* (Optional) Edit `fission-logger.yaml` to add Influxdb username and password and following env to controller deployment.  Then apply `fission-logger.yaml` and new controller deployment.
-
-```
-env:
-- name: INFLUXDB_URL
-  value: "http://influxdb:8086"
-- name: INFLUXDB_USERNAME
-  value: "<username>"
-- name: INFLUXDB_PASSWORD
-  value: "<password>"
-```
-
 ### Use the web based Fission-ui (Optional)
 
 [Fission-ui](https://github.com/fission/fission-ui) is the ui for fission maintained by the community.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -129,30 +129,13 @@ Finally, you're ready to use Fission!
 ### Enable Persistent Function Logs (Optional)
 
 Fission uses InfluxDB to store logs and fluentd to forward them from
-function pods into InfluxDB.  To setup both InfluxDB and fluentd:
-
-Edit `fission-logger.yaml` to add a username and password for the
-Influxdb deployment.  Then:
+function pods into InfluxDB.  
 
 ```
   $ kubectl create -f fission-logger.yaml
 ```
 
-On the client side,
-
-If you're using minikube or a local cluster:
-
-```
-$ export FISSION_LOGDB=http://$(minikube ip):31315
-```
-
-If you're using GKE or other cloud:
-
-```
-$ export FISSION_LOGDB=http://$(kubectl --namespace fission get svc influxdb -o=jsonpath='{..ip}'):8086
-```
-
-That's it for setup.  You can now use this to view function logs:
+That's it for the basic setup.  You can now use following command to view function logs:
 
 ```
   $ fission function logs --name hello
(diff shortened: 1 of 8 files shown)


controller/functionApi.go
```diff
@@ -171,3 +175,35 @@ func (api *API) FunctionApiDelete(w http.ResponseWriter, r *http.Request) {
...
+	director := func(req *http.Request) {
+		// only replace url Scheme and Host to remote influxDB
+		// and leave query string intact
+		req.URL.Scheme = svcUrl.Scheme
+		req.URL.Host = svcUrl.Host
+		req.URL.Path = svcUrl.Path
+		// set up http basic auth for database authentication
+		req.SetBasicAuth(dbCnf.username, dbCnf.password)
+	}
+	rrt := router.NewRetryingRoundTripper(10, 50*time.Millisecond)
```
soamvasani: Why do we need this? When does the default roundtripper not work?

INSTALL.md
```diff
@@ -168,6 +151,18 @@ particular pod:
   $ fission function logs --name hello --pod <pod name>
 ```
 
+* (Optional) Edit `fission-logger.yaml` to add Influxdb username and password and following env to controller deployment.  Then apply `fission-logger.yaml` and new controller deployment.
```
soamvasani: I suggest just removing this section. We should generate a random password in the helm chart (the plan is to move to that as the main recommended installation method).

controller/api.go
```diff
@@ -78,6 +105,7 @@ func (api *API) Serve(port int) {
 	r.HandleFunc("/v1/functions/{function}", api.FunctionApiGet).Methods("GET")
 	r.HandleFunc("/v1/functions/{function}", api.FunctionApiUpdate).Methods("PUT")
 	r.HandleFunc("/v1/functions/{function}", api.FunctionApiDelete).Methods("DELETE")
+	r.HandleFunc("/v1/functions/{function}/logs/{dbType}", api.FunctionLogsApiPost).Methods("POST")
```
soamvasani: Thinking about this path: It seems the `{function}` variable is not really being used? The query from the CLI has the function name. 

Also this isn't really part of the versioned API. The API is still influxdb's, this is just a proxy.

How about simply `/proxy/{dbType}`?

controller/functionApi.go
life1347: In fact, just want to follow the proxy setting in the router, not really need it. I'll remove it.

life1347:
The test process:

1. Follow the step described in INSTALL.md to set up persistent Function Logs
2. Compile and deploy the new version of controller
3. Compile fission CLI
4. Create a function and route for function`hello1`
5. Try to curl the function
6. Once the curl request is success, try with the following commands to retrieve logs from controller
```
fission fn logs --name hello1 -f -d
fission fn pods --name hello1
```

fission/logdb/influxdb.go
```diff
@@ -17,47 +17,35 @@ limitations under the License.
...
-	if err != nil {
-		log.Fatal(err)
-		return InfluxDB{}, err
-	}
-
-	return InfluxDB{
-		dbClient: dbClient,
-	}, nil
+func NewInfluxDB(serverURL string) (InfluxDB, error) {
+	return InfluxDB{endpoint: serverURL}, nil
```
soamvasani: Can't we use the influxdb client, but specify the new URL in the HTTPConfig Addr field?  That way we won't have to implement the query() function too.

(Looking at the code https://github.com/influxdata/influxdb/blob/master/client/v2/client.go#L82, it looks like it should just work.)

fission/logdb/influxdb.go
```diff
@@ -17,47 +17,35 @@ limitations under the License.
...
 }
 
 type InfluxDB struct {
-	dbClient influxdbClient.Client
+	endpoint string
 }
 
 func (influx InfluxDB) GetPods(filter LogFilter) ([]string, error) {
 	query := fmt.Sprintf("select * from \"log\" where \"funcuid\" = '%s' group by \"pod\"", filter.FuncUid)
```
soamvasani: Can we remove this Sprintf, which is vulnerable to "sql" injection, and instead use `influxdb.NewQueryWithParameters`?

fission/logdb/influxdb.go
life1347: Influxdb client will replace the Addr url path before sending the query string to remote host.(https://github.com/influxdata/influxdb/blob/master/client/v2/client.go#L475). So we need to write our own read-only db client. (In fact, this issue took some time to debug when I tried to use proxy server solution. lol)

fission/logdb/influxdb.go
life1347: oh, you are right! I'll take a look at it.

soamvasani:
Thanks, @life1347 ! The final changes look great.

PR #208, opened 2017-05-27
Fix unstoppable kubewatcher

life1347:
This PR aims to fix the following issues
1. watchSubscription is unstoppable even the watch is removed by the user.
2. add more checks to make sure that watchSubscription is stopped by the user or the unknow issues

### DIFF
--- a/kubewatcher/kubewatcher.go
+++ b/kubewatcher/kubewatcher.go
@@ -191,8 +191,7 @@ func (kw *KubeWatcher) removeWatch(w *fission.Watch) error {
 			fmt.Sprintf("watch doesn't exist: %v", w.Metadata))
 	}
 	delete(kw.watches, w.Metadata.Uid)
-	atomic.StoreInt32(ws.stopped, 1)
-	ws.kubeWatch.Stop()
+	ws.stop()
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


soamvasani:
Would you mind explaining the bug a bit?  Makes it faster to review if I know what problem this is fixing. Thanks :)

life1347:
ResultChan of watcher is closed if the kubwatcher is deleted, so watchSubscription should print out "stopped" message and stop.

```
for {
    for {
        ev, more := <-ws.kubeWatch.ResultChan()
        if !more {
            log.Println("Watch stopped", ws.Watch.Metadata.Name)
            break
        }
    ...
    if atomic.LoadInt32(ws.stopped) == 0 {
        err := ws.restartWatch()
        if err != nil {
            log.Panicf("Failed to restart watch: %v", err)
        }
    }
}
```

However, the detection of ResultChan close is inside a nested for loops in current implementation, and the break statement only breaks internal for loop. So watchSubscription will keep trying to wait for events but failed and print stopped message endlessly.

```
for {
    if ws.isStopped() {
        ws.kubeWatch.Stop()
        break
    }
    ev, more := <-ws.kubeWatch.ResultChan()
    if !more {
        if ws.isStopped() {
            // watch is removed by user.
            log.Println("Watch stopped", ws.Watch.Metadata.Name)
            return
        } else {
            // watch closed due to unknown issues, restart it.
            log.Println("Watch stopped unexpectedly", ws.Watch.Metadata.Name)
            err := ws.restartWatch()
            if err != nil {
                log.Panicf("Failed to restart watch: %v", err)
            }
            continue
        }
    }
}
```

This PR removes the outer for loop, and check the close of ResultChan is because of trigger deletion or caused by unknown issues. If a watcher is stopped by the user, then stop it. Otherwise, restart the watcher.

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,71 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
-				break
-			}
-			rv, err := getResourceVersion(ev.Object)
-			if err != nil {
-				log.Printf("Error getting resourceVersion from object: %v", err)
+				return
 			} else {
-				log.Printf("rv=%v", rv)
-				ws.lastResourceVersion = rv
+				// watch closed due to unknown issues, restart it.
```
soamvasani: Nitpick -- it's not an unknown issue, I just didn't know why when I first wrote this code :) The watch has a timeout, after which it returns and needs to be restarted. Would you mind changing this comment and log statement to something like "watch timed out, restarting"?

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,71 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
-				log.Printf("Failed to serialize object: %v", err)
-				// TODO send a POST request indicating error
-			}
+		if ev.Type == watch.Error {
+			e := apierrs.FromObject(ev.Object)
+			log.Println("Watch error, retrying in a second: %v", e)
+			// Start from the beginning to get around "too old resource version"
+			ws.lastResourceVersion = ""
+			time.Sleep(time.Second)
+			continue
```
soamvasani: Hmm, the old code restarts the watch when this happens, but the new code tries to read from the watch again. I think you need to call `ws.restartWatch()` here.

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,71 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
-				log.Panicf("Failed to restart watch: %v", err)
-			}
+
+		// Event and object type aren't in the serialized object
+		headers := map[string]string{
+			"Content-Type":             "application/json",
+			"X-Kubernetes-Event-Type":  string(ev.Type),
+			"X-Kubernetes-Object-Type": reflect.TypeOf(ev.Object).Elem().Name(),
 		}
+		// Event and object type aren't in the serialized object
```
soamvasani: duplicated comment

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,71 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
 	for {
-		for {
-			ev, more := <-ws.kubeWatch.ResultChan()
-			if !more {
+		// check watchSubscription is stopped or not before waiting for event
+		// comes from the kubeWatch.ResultChan(). This fix the edge case that
+		// new kubewatch is created in the restartWatch() while the old kubewatch
+		// is being used in watchSubscription.stop().
+		if ws.isStopped() {
+			ws.kubeWatch.Stop()
```
soamvasani: (It's not wrong to have this call here, I'm just trying to make sure I understand the code correctly.)

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,74 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
 	for {
-		for {
-			ev, more := <-ws.kubeWatch.ResultChan()
-			if !more {
+		// check watchSubscription is stopped or not before waiting for event
+		// comes from the kubeWatch.ResultChan(). This fix the edge case that
+		// new kubewatch is created in the restartWatch() while the old kubewatch
+		// is being used in watchSubscription.stop().
+		if ws.isStopped() {
+			ws.stop()
```
soamvasani: I don't understand this either. If `ws.isStopped()` is true why do you need to call `ws.stop()`?

kubewatcher/kubewatcher.go
```diff
@@ -258,50 +257,74 @@ func getResourceVersion(obj runtime.Object) (string, error) {
...
 			if err != nil {
 				log.Panicf("Failed to restart watch: %v", err)
 			}
+			continue
+		}
+		rv, err := getResourceVersion(ev.Object)
+		if err != nil {
+			log.Printf("Error getting resourceVersion from object: %v", err)
+		} else {
+			log.Printf("rv=%v", rv)
```
soamvasani: could you remove this log, it's too verbose, and was just here for debugging

PR #210, opened 2017-05-28
Fix http response body not closed correctly & return immediately when error occurred

life1347:
This PR aims to fix bugs in the current codebase. So far, I’ve checked all over the project and there is no more similar issue exists after this PR.

### DIFF
--- a/controller/environmentApi.go
+++ b/controller/environmentApi.go
@@ -47,6 +47,7 @@ func (api *API) EnvironmentApiCreate(w http.ResponseWriter, r *http.Request) {
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		api.respondWithError(w, err)
+		return
 	}
 
 	var env fission.Environment
@@ -103,6 +104,7 @@ func (api *API) EnvironmentApiUpdate(w http.ResponseWriter, r *http.Request) {
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		api.respondWithError(w, err)
+		return
 	}
 
 	var env fission.Environment
--- a/controller/environmentApi.go
+++ b/controller/environmentApi.go
@@ -48,7 +48,6 @@ func (api *API) EnvironmentApiCreate(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		api.respondWithError(w, err)
 	}
-	defer r.Body.Close()
 
 	var env fission.Environment
 	err = json.Unmarshal(body, &env)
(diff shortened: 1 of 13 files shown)


yqf3139:
Thanks. @life1347  What tools do you use to sort import? Can we integrate it into some auto checking script.

life1347:
@yqf3139 goimports does the jobs. It will be very helpful if we integrate goimports into auto checking script, however, under some situations, we still need to organize the imports manually.

Currently, I sort imports with the following rule.

```
import (
    <native package>

    <third-party package>

    <fission package>
)
```

For example:

```
import (
	"log"
	"time"

	"k8s.io/client-go/1.5/kubernetes"

	"github.com/fission/fission"
	"github.com/fission/fission/controller/client"
)
```

controller/environmentApi.go
```diff
@@ -48,7 +48,6 @@ func (api *API) EnvironmentApiCreate(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		api.respondWithError(w, err)
 	}
-	defer r.Body.Close()
```
life1347: According to the document (https://golang.org/pkg/net/http/#Request) 

```
// Body is the request's body.
//
// For client requests a nil body means the request has no
// body, such as a GET request. The HTTP Client's Transport
// is responsible for calling the Close method.
//
// For server requests the Request Body is always non-nil
// but will return EOF immediately when no body is present.
// The Server will close the request body. The ServeHTTP
// Handler does not need to.
Body io.ReadCloser
```

server http handler does not need to close request body, so I removed this line for consistency.

controller/environmentApi.go
```diff
@@ -47,8 +47,8 @@ func (api *API) EnvironmentApiCreate(w http.ResponseWriter, r *http.Request) {
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		api.respondWithError(w, err)
+		return
 	}
-	defer r.Body.Close()
```
life1347: According to the document (https://golang.org/pkg/net/http/#Request) 

```
// Body is the request's body.
//
// For client requests a nil body means the request has no
// body, such as a GET request. The HTTP Client's Transport
// is responsible for calling the Close method.
//
// For server requests the Request Body is always non-nil
// but will return EOF immediately when no body is present.
// The Server will close the request body. The ServeHTTP
// Handler does not need to.
Body io.ReadCloser
```

server does not need to close the request body, so I removed this line for consistency.

soamvasani:
Thanks @life1347 !

PR #213, opened 2017-05-30
Fix two links in Roadmap doc

markpeek:
Correct two incorrect links in the roadmap doc due to typo (cut/paste from previous header).


soamvasani:
Thanks @markpeek !

Issue #200, opened 2017-05-18
Opentracing integration


soamvasani:
Integrate opentracing into fission.  Expose the function call itself, but also fission-internal stuff like cold-start, routing overhead, etc.

This will allow users to trace fission functions as part of a bigger solution, as well as troubleshoot performance problems with fission itself.

yqf3139:
Open Tracing seems quite easy to use. Will make a quick demo integration. 

Some discussions:
- When we instrumenting the code, we should provide a switch to collect trace or not. How to make the switch by the image version or the function calling params?
- Tracing has some overlap with metrics. I think metrics is more general statistics about the system, it is always collected and all the time-series data are reduced to provide insights. And in tracing, we choose some special case and have a detailed look at the workflow. If we want to summary the trace data, maybe we have to export them to other database to make the analysis.
- We can connect several fission services into one trace. For example, we can trace what fission does in a function call workflow (router get the request, choose or start a user function endpoint, proxy the request to the user function, user function does it's job and then return the response, router finish this request).

vishal-biyani:
Thank you so much again https://github.com/ajbouh for your work on this. This is now part of Fission 🎉

PR #215, opened 2017-05-31
[WIP] Opentracing with Zipkin as backend

yqf3139:
This PR for #200 is a preview of integrating Open-Tracing compatible services.
It can provide tracing for fission developer and user.

Currently only the function calling path in router, poolmgr and nodejs image is instrumented.
Below is a demo of a cold function call and a hot function call.

image
image

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/215)

### DIFF
--- a/environments/nodejs/package.json
+++ b/environments/nodejs/package.json
@@ -19,8 +19,13 @@
     "minimist": "",
     "morgan": "",
     "mz": "~2.1.0",
+    "opentracing": "^0.14.1",
     "request": "^2.81.0",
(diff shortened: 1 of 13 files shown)


soamvasani:
Awesome stuff, @yqf3139 !  I'll dive into this soon.

vishal-biyani:
Hi @yqf3139 We now have opentracing using OpenCensus which supports both Jaeger as well as Zipkin as backends: https://github.com/fission/fission/pull/1079

We will close this in favor of the other change, thanks a lot for your effort!

Issue #194, opened 2017-05-18
NATS Streaming triggers


soamvasani:
Launch functions from events via NATS streaming server.

soamvasani:
cc @life1347

life1347:
![httptrigger](https://cloud.githubusercontent.com/assets/202578/26530311/54ce6284-4405-11e7-8012-976521eb41b4.png)

Above is the blueprint of message queue trigger proposal.

Currently, I am writing a general `MessageQueueTrigger` for multiple message queues (#194 #195 #196 #197 #198) to extend. After that, we only need to implement the different part of multiple message queues separately without too much redundant code. And I am wandering does trigger needs to send back the function response to origin event publisher? or just ignore it?

erwinvaneyk:
Nice informative diagram @life1347 ! What is not entirely clear from it though is the location of the generic `MessageQueueTrigger` in this context. Is it correct that the `Nats, CloudPubSub Trigger ...etc` are the implementations of this interface that you describe?

Given the generic nature of message queue usage, it might be best to emit the response of the function as separate event. This would allow users to determine for themselves whether they care about the response. (if the response is irrelevant to them, they can just ignore it)

life1347:
@erwinvaneyk Yes, correct. But since it's still under the implementation, the design may be slightly changed in the end. I'll update a new diagram once the design is confirmed.

soamvasani:
@life1347 @erwinvaneyk I think it makes sense to emit the response back into the queue.  The event trigger for the function could include an optional "response topic" that the function's response would be sent into.

javierbq:
In the case of AMQP it would be nice if the subscriber gets the response so it can ack or nack the message depending on the result of the function.

soamvasani:
MessageQueueTrigger supports NATS streaming, #218.

PR #217, opened 2017-06-02
Refactor publisher interface

life1347:
If we want to reply the function response back to event sender mentioned at #194, the publisher needs a way to send the response back to the trigger. 

In this PR, triggers pass a response channel to publish interface so that webhookPublisher can send function response through the channel to trigger. The trigger can decide to send the response back to event sender or just ignore it.

For demonstration, timer passes a respChan to publisher and prints out received function response. The code will be removed before merged.

How to test:

1. Create a function
```
fission fn create --name hello1 --code ~/Desktop/Fission/hello.js --env nodejs
```
2. Create a time trigger for the function
```
fission timer create --name hello1 --function hello1 --cron '@every 5s'
```
3. Check timer log
```
kubectl logs -f timer-3198201534-0nfzd
```

### DIFF
--- a/publisher/webhookPublisher.go
+++ b/publisher/webhookPublisher.go
@@ -41,6 +41,7 @@ type (
 		target     string
 		retries    int
 		retryDelay time.Duration
+		respChan   chan string
 	}
 )
 
@@ -56,13 +57,14 @@ func MakeWebhookPublisher(baseUrl string) *WebhookPublisher {
 	return p
 }
 
-func (p *WebhookPublisher) Publish(body string, headers map[string]string, target string) {
+func (p *WebhookPublisher) Publish(body string, headers map[string]string, target string, respChan chan string) {
 	p.requestChannel <- &publishRequest{
 		body:       body,
 		headers:    headers,
 		target:     target,
 		retries:    p.maxRetries,
 		retryDelay: p.retryDelay,
+		respChan:   respChan,
 	}
 }
 
@@ -91,17 +93,24 @@ func (p *WebhookPublisher) makeHttpRequest(r *publishRequest) {
 
 	// All done if the request succeeded with 200 OK.
 	if err == nil && resp.StatusCode == 200 {
-		resp.Body.Close()
-		return
+		defer resp.Body.Close()
+		if r.respChan == nil {
+			return
+		}
+		body, err := ioutil.ReadAll(resp.Body)
+		if err == nil {
+			r.respChan <- string(body)
+			return
+		}
 	}
 
 	// Log errors
 	if err != nil {
 		log.Printf("Request failed: %v", r)
 	} else if resp.StatusCode != 200 {
+		defer resp.Body.Close()
 		log.Printf("Request returned failure: %v", resp.StatusCode)
 		body, err := ioutil.ReadAll(resp.Body)
-		resp.Body.Close()
 		if err == nil {
 			log.Printf("request error: %v", string(body))
 		}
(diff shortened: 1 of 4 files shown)


publisher/webhookPublisher.go
```diff
@@ -116,6 +125,9 @@ func (p *WebhookPublisher) makeHttpRequest(r *publishRequest) {
 		})
 	} else {
 		log.Printf("Final retry failed, giving up on %v", url)
+		if r.respChan != nil {
```
soamvasani: Hmm, we should not designate an arbitrary string as the error value.

We should have a Response struct, with an error code in it and a []byte for the response.  Then the error field can be used for indicating success or failure.

publisher/webhookPublisher.go
```diff
@@ -91,17 +93,24 @@ func (p *WebhookPublisher) makeHttpRequest(r *publishRequest) {
...
 	if err == nil && resp.StatusCode == 200 {
-		resp.Body.Close()
-		return
+		defer resp.Body.Close()
+		if r.respChan == nil {
+			return
+		}
+		body, err := ioutil.ReadAll(resp.Body)
+		if err == nil {
+			r.respChan <- string(body)
```
soamvasani: do you need an `if respChan != nil`?

publisher/webhookPublisher.go
```diff
@@ -116,6 +125,9 @@ func (p *WebhookPublisher) makeHttpRequest(r *publishRequest) {
 		})
 	} else {
 		log.Printf("Final retry failed, giving up on %v", url)
+		if r.respChan != nil {
+			r.respChan <- StatusFunctionFailure
+		}
 		// Event dropped
 	}
```
soamvasani: it's not super clear but do we guarantee publishing a response into the response channel?

timer/timer.go
```diff
@@ -136,7 +137,10 @@ func (timer *Timer) newCron(t fission.TimeTrigger) *cron.Cron {
 		headers := map[string]string{
 			"X-Fission-Timer-Name": t.Name,
 		}
-		(*timer.publisher).Publish("", headers, fission.UrlForFunction(&t.Function))
+		respChan := make(chan string)
+		(*timer.publisher).Publish("", headers, fission.UrlForFunction(&t.Function), respChan)
+		fmt.Printf("Trigger %s get response from function: %s\n", t.Metadata.Name, <-respChan)
```
soamvasani: This will block forever if there's no response. Also, it's better to be explicit and put the channel read on a separate line from the logging.

Also please use log.Printf not fmt.Printf

soamvasani:
Actually, a higher-level question: Why do this at all? Why does the publisher need a response in the interface?

The message-queue-based publisher implementation could send the function response directly into a queue topic without dealing with caller of the publisher.

life1347:
@soamvasani that make sense. Or how about let triggers defined a reply function to deal with the response, and pass the function to the Publisher interface. 

This design has the following advantages:
1. A publisher could use the replyFunc to send the response directly.
2. We don't need to add too much complexity or changed at the publisher side, even a trigger changed the way to deal with the response.

```
type (
    Publisher interface {
        Publish(body string, headers map[string]string, target string, replyFunc Reply)
    }

    Reply func(msg *Response)

    Response struct {
        ErrCode int
        Body    string
    }
)
```

soamvasani:
@life1347 The point of the publisher interface is to be an abstraction around sending events into webhooks or message queues. When does the caller of the Publisher want to implement a response function? Can you provide an example? With the existing interface, the implementation of the publisher would be able to send a response into the message queue.

soamvasani:
@life1347 What's the status of this? Do you still want to do something like this?  If not, feel free to close the pr.

life1347:
New publisher will check in with message queue trigger PR

PR #218, opened 2017-06-04
Add message queue trigger support

life1347:
This PR aims to add message queue support for #194 and it's not ready to merge yet. 

For people who want to try this feature
1. build new fission-bundle image & fission cli
2. apply `fission.yaml` to create new deployment (mqtrigger, nats-streaming) and service (nats-streaming)
3. follow fission cli instruction to create a message queue trigger 

Here is the demonstration video: https://youtu.be/mzjf_5fOrD8

Currently, message queue trigger only supports nats-streaming. CloudPubSub is in planning. The feature of sending function response back to event publisher will be implemented when the design of #217 is confirmed and merged.

Since it's still in development, any suggestion is welcome :)


mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,140 @@
...
+			} else {
+				log.Infof("MessageQueue trigger %s deleted", trigger.Name)
+			}
+		}
+	}
+}
+
+func isTopicValidForNats(topic string) bool {
+	// nats-streaming does not support wildcard channl.
+	return nsUtil.IsSubjectValid(topic, false)
```
yqf3139: Compile failure. I think the `nsUtil.IsSubjectValid` just changed its arguments.

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,140 @@
...
+func (nats *Nats) add(trigger fission.MessageQueueTrigger) error {
+	subj := trigger.Topic
+
+	if !isTopicValidForNats(subj) {
+		return errors.New(fmt.Sprintf("Not a valid topic: %s", trigger.Topic))
+	}
+
+	handler := func(msg *ns.Msg) {
+		headers := map[string]string{
+			"X-Fission-Timer-Name": trigger.Function.Name,
```
life1347: yes, I noticed this before, it will be fixed in the later commit. thanks :)

life1347:
Hi @soamvasani, this PR is ready to merge. Also, I wrote a simple test script under test/mqtrigger to test the functionality of message queue trigger.

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,182 @@
...
+		// Create a durable subscription to nats, so that triggers could retrieve last unack message.
+		// https://github.com/nats-io/go-nats-streaming#durable-subscriptions
+		ns.DurableName(trigger.Uid),
+
+		// Nats-streaming server is auto-ack mode by default. Since we want nats-streaming server to
+		// resend a message if the trigger does not ack it, we need to enable the manual ack mode, so that
+		// trigger could choose to ack message or simply drop it depend on the response of function pod.
+		ns.SetManualAckMode(),
+	}
+	sub, err := (*nats.nsConn).Subscribe(subj, handler, opts...)
```
soamvasani: Do you need the `(*nats.nsConn)`? Why not just `nats.nsConn`?

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,182 @@
...
+				log.Warnf("MessageQueue trigger %s deleted failed: %v", trigger.Name, err)
+			} else {
+				log.Infof("MessageQueue trigger %s deleted", trigger.Name)
+			}
+		}
+	}
+}
+
+func isTopicValidForNats(topic string) bool {
+	// nats-streaming does not support wildcard channl.
```
soamvasani: channel*

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,182 @@
...
+		if resp.StatusCode != 200 {
+			log.Printf("Request returned failure: %v", resp.StatusCode)
+			return
+		}
+		// trigger acks message only if a request done successfully
+		err = msg.Ack()
+		if err != nil {
+			log.Warningf("Failed to ack message: %v", err)
+		}
+		if len(trigger.ResponseTopic) > 0 {
```
life1347: I think we should keep the response topic for the following reasons:
1. Nats-Streaming does not provide the request/reply as Nats does. (https://godoc.org/github.com/nats-io/go-nats-streaming#Msg)
2. For the other message queue service, they may not provide the reply topic such as CloudPubSub (https://godoc.org/cloud.google.com/go/pubsub#Message).

test/mqtrigger/test.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+fi
+
+cd $FISSIONDIR"/fission/"
+go build
+mv fission $FISSIONDIR"/test/mqtrigger"
+cd $FISSIONDIR"/test/mqtrigger"
+
+./fission env create --name nodejs --image fission/node-env
+./fission fn create --name hello1 --env nodejs --code ~/Desktop/main.js --method GET
```
life1347: oops

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,188 @@
...
+const (
+	natsClusterID  = "fissionMQTrigger"
+	natsProtocol   = "nats://"
+	natsClientID   = "fission"
+	natsQueueGroup = "fission-messageQueueNatsTrigger"
+)
+
+type (
+	Nats struct {
+		MessageQueueTriggerManager
```
soamvasani: I don't think you need to wrap the trigger manager into this object. 

You should keep the functionality of managing triggers in the trigger manager, and the functionality of talking to NATS in this file.

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,188 @@
...
+	if !ok {
+		return errors.New("Trigger not exists")
+	}
+	defer nats.delTrigger(triggerUid)
+	nsSub := triggerSub.subscription.(ns.Subscription)
+	err := nsSub.Close()
+	return err
+}
+
+func (nats *Nats) sync() {
```
soamvasani: Is there anything NATS specific here?

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type (
+	Nats struct {
+		MessageQueueTriggerManager
+		nsConn ns.Conn
+	}
+)
+
+func makeNatsTriggerManager(mqTriggerMgr MessageQueueTriggerManager) (MessageQueueTriggerManagerInterface, error) {
```
soamvasani: This should not return a trigger manager interface, but rather a new interface, let's call it `MessageQueue`.

The `MessageQueue` interface would just have simple `subscribe(trigger)` and `unsubscribe(trigger)` methods.  Nats can then satisfy that interface.

mqtrigger/messageQueue/nats.go
life1347: Originally, I treated Nats as a trigger manager and itself has the functionality of connectint to NATS. But as you said, it's better to keep the functionality in right place instead of putting them together.
After the refactoring, now it's more clear what message queue exactly do and easier to add new message queue in future.
Thanks for the comments, that really helps a lot.

mqtrigger/messageQueue/messageQueue.go
```diff
@@ -0,0 +1,175 @@
...
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/fission/fission"
+	controllerClient "github.com/fission/fission/controller/client"
+)
+
+type MessageQueueConfig struct {
```
soamvasani: can you keep this in the `type` block below?

mqtrigger/messageQueue/messageQueue.go
```diff
@@ -0,0 +1,175 @@
...
+	MQType string
+	Url    string
+}
+
+const (
+	NATS string = "nats-streaming"
+)
+
+type (
+	MessageQueue interface {
```
soamvasani: Suggestion -- no need for `triggerSubscription` in this interface. Just return an `interface{}` from `subscribe` and pass the same to `unsubscribe`. The implementations wouldn't use any field other than `triggerSubscription.subscription` anyway.

mqtrigger/messageQueue/messageQueue.go
```diff
@@ -0,0 +1,175 @@
...
+		subscribe(trigger fission.MessageQueueTrigger) (*triggerSubscription, error)
+		unsubscribe(triggerSub *triggerSubscription) error
+	}
+
+	MessageQueueTriggerManager struct {
+		sync.RWMutex
```
soamvasani: We've so far avoided mutexes for this kind of use case.  Can you change this to use a request and a requestChannel?  See `poolmgr/gpm.go` for an example.

mqtrigger/messageQueue/messageQueue.go
```diff
@@ -0,0 +1,175 @@
...
+	var err error
+
+	mqTriggerMgr := MessageQueueTriggerManager{
+		triggerMap: make(map[string]*triggerSubscription),
+		controller: ctrlClient,
+	}
+	switch mqConfig.MQType {
+	case NATS:
+		fallthrough
+	default:
```
soamvasani: Why not error on default? Let's keep the choice of message queue explicit.

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	triggerSub := triggerSubscription{
+		Metadata: fission.Metadata{
+			Name: trigger.Name,
+			Uid:  trigger.Uid,
+		},
+		funcMeta:     trigger.Function,
+		subscription: sub,
+	}
+	return &triggerSub, nil
```
soamvasani: could just return `sub` here

mqtrigger/messageQueue/nats.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (nats Nats) unsubscribe(triggerSub *triggerSubscription) error {
```
soamvasani: suggest changing to `subscription interface{}`

publisher/natsPublisher.go
```diff
@@ -0,0 +1,93 @@
+/*
```
soamvasani: is NatsPublisher used anywhere now? If not, it's still useful, we could put it in a separate pull request.

PR #221, opened 2017-06-08
Fix pool contains wrong environment metadata

life1347:
The generic pool manager creates a generic pool with the reference of environment metadata. Since the reference is the address of loop variable, the content of metadata will be replaced with the last element of env list. So pool manager sets up logging with the wrong env.

### DIFF
--- a/poolmgr/gpm.go
+++ b/poolmgr/gpm.go
@@ -157,8 +157,8 @@ func (gpm *GenericPoolManager) eagerPoolCreator() {
 		// creating pools for envs that are actually used by functions.  Also we might want
 		// to keep these eagerly created pools smaller than the ones created when there are
 		// actual function calls.
-		for _, env := range envs {
-			_, err := gpm.GetPool(&env)
+		for i := range envs {
+			_, err := gpm.GetPool(&envs[i])
 			if err != nil {
 				log.Printf("eager-create pool failed: %v", err)
 			}


poolmgr/gpm.go
```diff
@@ -157,8 +157,8 @@ func (gpm *GenericPoolManager) eagerPoolCreator() {
 		// creating pools for envs that are actually used by functions.  Also we might want
 		// to keep these eagerly created pools smaller than the ones created when there are
 		// actual function calls.
-		for _, env := range envs {
-			_, err := gpm.GetPool(&env)
```
yqf3139: @life1347 
`Get Pool` seems to be synchronized function call, so in the loop, every envs item is copied to `env`, get the address and then send to the function, so it should work fine although the address of `env` is the same?

life1347: If GenericPool store the value of the `env`, then its find. 
 
```
GenericPool struct {
    env              *fission.Environment
    ...
}
```
But since GenericPool store the address of the `env`, all generic pools contain the same address of the loop variable which's value is the last element of the env slice. 
Its the same issue with PR https://github.com/fission/fission/pull/141

yqf3139: Yes. I haven't digged into the generic poll methods, getting address of a temp variable is risky.

soamvasani:
😭 Argh.  

Thanks for the fix @life1347 .

PR #223, opened 2017-06-08
add ruby-env

sbfaulkner:
attn: @soamvasani 

this PR adds a ruby environment... this initial version just uses webrick

the rack request object (and the full env) are (optionally) passed to the handler in a simple context object

the handler can return either a simple string response (for a 200) or a standard rack response array

includes examples and readme

### DIFF
--- a/environments/ruby/server.rb
+++ b/environments/ruby/server.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+require 'rack'
+
+CODEPATH = '/userfunc/user'
+
+module Fission
+  class Request < Rack::Request
+    def headers
+      Hash[
+        *env.select { |k,v| k.start_with?('HTTP_') }
+             .map { |k,v| [k.sub(/\AHTTP_/, '').split('_').map(&:capitalize).join('-'), v] }
+             .sort
+             .flatten
+      ]
+    end
+  end
+
+  class Context
+    attr_reader :env
+
+    def initialize(env)
+      @env = env
+    end
+
+    def request
+      @request ||= Request.new(env)
+    end
+  end
+
+  module Specializer
+    def self.call(env)
+      load CODEPATH
+      Rack::Response.new([], 201).finish
+    rescue
+      Rack::Response.new(['500 Internal Server Error'], 500, {}).finish
+    end
+  end
+
+  module Handler
+    def self.call(env)
+      response = if method(:handler).arity > 0
+        handler(Context.new(env))
+      else
+        handler
+      end
+
+      response.is_a?(Array) ? response : Rack::Response.new([response]).finish
+    rescue
+      Rack::Response.new(['500 Internal Server Error'], 500, {}).finish
+    end
+  end
+end
+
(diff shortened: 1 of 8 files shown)


environments/ruby/server.rb
```diff
@@ -0,0 +1,66 @@
...
+    def request
+      @request ||= Request.new(env)
+    end
+  end
+
+  module Specializer
+    def self.call(env)
+      load CODEPATH
+      Rack::Response.new([], 201).finish
+    rescue
```
soamvasani: Can you log code loading errors to stdout?  (Otherwise things will just fail mysteriously if there's an error in the user's code.)

environments/ruby/server.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+require 'rack'
+
+CODEPATH = '/userfunc/user'
+
+module Fission
+  class Request < Rack::Request
+    def headers
+      Hash[
+        *env.select { |k,v| k.start_with?('HTTP_') }
```
soamvasani: What's this for?

environments/ruby/server.rb
```diff
@@ -0,0 +1,66 @@
...
+  module Handler
+    def self.call(env)
+      response = if method(:handler).arity > 0
+        handler(Context.new(env))
+      else
+        handler
+      end
+
+      response.is_a?(Array) ? response : Rack::Response.new([response]).finish
+    rescue
```
soamvasani: If there's an exception can we log it?

examples/ruby/request_data.rb
```diff
@@ -0,0 +1,19 @@
...
+    #{request.headers.map { |h| h.join(': ') }.join("\n") }
+
+    ---PARAMS---
+    #{request.params.map { |h| h.join('=') }.join("\n") }
+
+    --BODY--
+    #{request.body.read}
+  MSG
+
+  Rack::Response.new([msg]).finish
```
soamvasani: Is this equivalent to simply returning `msg`?

sbfaulkner:
thanks for the feedback on this @soamvasani ... will address, but wanted to get this up here in your line of sight asap

environments/ruby/server.rb
sbfaulkner: the Rack env contains more than just request/http headers... so the headers method is a convenience method to get just the actual http headers (which all begin with HTTP_ in rack)

examples/ruby/request_data.rb
sbfaulkner: yes... the thought was that by explicitly building the Rack::Response here, the example would illustrate that... some comments would probably help here...

environments/ruby/server.rb
soamvasani: Makes sense, thanks for the explanation.

sbfaulkner:
will get back to this later this week

soamvasani:
Thanks @sbfaulkner -- given that this is in working shape, I'm going to merge this PR and open an issue for the remaining comments.

PR #224, opened 2017-06-08
Fix s/Sirupsen/sirupsen/ for logrus

n1koo:
Because of https://github.com/sirupsen/logrus/issues/543 `glide install` fails to install `logrus` right now. This PR simply bumps us to the newest version with correct references

Used latest SHA on master for locking which obviously isn't ideal but until a new release is cut ( https://github.com/sirupsen/logrus/issues/555 ) probably good enough

### DIFF
--- a/controller/api.go
+++ b/controller/api.go
@@ -23,7 +23,7 @@ import (
 	"runtime/debug"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


soamvasani:
Thanks @n1koo, appreciate the fix!

Issue #158, opened 2017-03-17
Retrieve URL params in functions


nouney:
We can use `gorilla/mux` URL params when defining a route but we can't retrieve them in functions. 

Maybe could we add them in the query string or HTTP headers?

yqf3139:
I think add them in the headers is a good idea.
For example, X-Fission-Params-${key}: ${value}.

sbfaulkner:
alternate approaches could be...

(all examples below use a route of `/path/{x}/{y}` and path of `/path/1/2`)

1. url encoding the key value pairs into a single X-Fission-Params header

    eg.  the resulting header could be `X-Fission-Params: x=1&y=2`

2. adding params to the query string with a prefix

    eg.  the resulting request_uri could be `/?_fission_x=1&_fission_y=2`

3. adding params to the query string using "nested" params

    eg.  the resulting request_uri could be `/?_fission[x]=1&_fission[y]=2`

sbfaulkner:
thoughts @soamvasani ?

soamvasani:
I think the first option of using one header per param is the most obvious thing to do.  (The only catch is, since the path can contain `:`s, we'd need to escape them in the header.)

Looks like @yqf3139 already has a fix in progress.

yqf3139:
@soamvasani 
I just made some test for the `:`. The original fix works well.

PR #226, opened 2017-06-09
Retrieve URL params in functions (#158)

yqf3139:
Retrieve gorilla/mux URL params in functions
Add them in HTTP headers, such as X-Fission-Params-${key}: ${value}

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/fission/fission"
 	poolmgrClient "github.com/fission/fission/poolmgr/client"
+	"github.com/gorilla/mux"
 )
 
 type functionHandler struct {
@@ -92,6 +93,12 @@ func (fh *functionHandler) tapService(serviceUrl *url.URL) {
 
 func (fh *functionHandler) handler(responseWriter http.ResponseWriter, request *http.Request) {
 	reqStartTime := time.Now()
+
+	// retrieve url params and add them to request header
+	vars := mux.Vars(request)
+	for k, v := range vars {
+		request.Header.Add(fmt.Sprintf("X-Fission-Params-%v", k), v)
+	}
 
 	// cache lookup
 	serviceUrl, err := fh.fmap.lookup(&fh.Function)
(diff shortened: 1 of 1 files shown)


router/functionHandler.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	"github.com/fission/fission"
 	poolmgrClient "github.com/fission/fission/poolmgr/client"
+	"github.com/gorilla/mux"
```
life1347: just a reminder, we should sort the imports

yqf3139: @life1347 I tried goimports, but it did not sort the imports. Could you provide your toolchain configs or add it to the `hack` folder to automatically make the check ?

router/functionHandler.go
```diff
@@ -93,6 +95,12 @@ func (fh *functionHandler) tapService(serviceUrl *url.URL) {
 func (fh *functionHandler) handler(responseWriter http.ResponseWriter, request *http.Request) {
 	reqStartTime := time.Now()
 
+	// retrieve url params and add them to request header
+	vars := mux.Vars(request)
+	for k, v := range vars {
+		request.Header.Add(fmt.Sprintf("X-Fission-Params-%v", k), v)
```
soamvasani: What happens if there's a url like "http://example.com/foo: bar/"?  Both ":" and " " are valid chars in URLs.

soamvasani:
Thanks for working on this @yqf3139 !

soamvasani:
@yqf3139 in case you missed it, I had a question above: does this break when URLs have a ": " string in them?  Do you need to escape these chars?

yqf3139:
Thanks, I will have a double check. It worked fine in my last test with ":".

router/functionHandler.go
yqf3139: I double checked. Both are fine.

soamvasani: Thanks for checking. I finally realized that the ": " would show up in the value, not the key, and the key itself can never have a ": ".

PR #229, opened 2017-06-11
Aggregate tap service request in interval

yqf3139:
For every function call, the router has to make a tap-service request to poolmgr.
We can aggregate the requests in a interval (5 seconds).

### DIFF
--- a/poolmgr/client/client.go
+++ b/poolmgr/client/client.go
@@ -84,9 +84,7 @@ func (c *Client) service() {
 					for u := range c.tappedByUrl {
 						c._tapService(u)
 					}
-					if len(urls) > 0 {
-						log.Printf("Tapped %v services in batch", len(urls))
-					}
+					log.Printf("Tapped %v services in batch", len(urls))
 				}()
 				log.Printf("Tapped %v services in batch", len(urls))
 			}
--- a/poolmgr/client/client.go
+++ b/poolmgr/client/client.go
@@ -29,33 +29,19 @@ import (
 	"github.com/fission/fission"
 )
 
-const (
-	TOUCH int = iota
-	SYNC
-)
-
 type Client struct {
 	poolmgrUrl  string
 	tappedByUrl map[string]bool
-	requestChan chan clientRequest
-	tapUrlChan  chan string
-}
-
-type clientRequest struct {
-	requestType int
-	serviceUrl  string
+	requestChan chan string
 }
 
 func MakeClient(poolmgrUrl string) *Client {
 	c := &Client{
 		poolmgrUrl:  strings.TrimSuffix(poolmgrUrl, "/"),
 		tappedByUrl: make(map[string]bool),
-		requestChan: make(chan clientRequest, 100),
-		tapUrlChan:  make(chan string, 100),
+		requestChan: make(chan string),
 	}
-	go c.tapWorker()
-	go c.timer()
-	go c.tapService()
+	go c.service()
 	return c
 }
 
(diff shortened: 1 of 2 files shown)


soamvasani:
@yqf3139 Thanks for working on this. We do need this or something like it.

We've avoided locks pretty much everywhere in fission. While the usage of locks here looks correct, I'd still like to preserve the property of no locks, especially in a client library. The pattern followed in other places is to have anything related to a shared data structure work in one goroutine only, and have that goroutine listen on a channel.  This makes it much easier to reason about what's going on.

yqf3139:
OK. I will make a commit to use the goroutine-service pattern.

poolmgr/client/client.go
```diff
@@ -60,12 +84,49 @@ func (c *Client) GetServiceForFunction(metadata *fission.Metadata) (string, erro
...
-func (c *Client) TapService(serviceUrl *url.URL) error {
-	url := c.poolmgrUrl + "/v1/tapService"
+func (c *Client) tapService() {
+	for {
+		req := <-c.requestChan
+		switch req.requestType {
+		case TOUCH:
+			c.tappedByUrl[req.serviceUrl] = true
+		case SYNC:
+			for u := range c.tappedByUrl {
```
soamvasani: You don't really need a channel and a worker for this part.  You can do this:

```
urls := c.tappedByUrl
c.tappedByUrl = make(map[string]bool)
go func() {
   for u := range urls {
       c._tapService(u)
    } 
}()
```

poolmgr/client/client.go
```diff
@@ -17,32 +17,56 @@ limitations under the License.
...
+	serviceUrl  string
 }
 
 func MakeClient(poolmgrUrl string) *Client {
-	return &Client{poolmgrUrl: strings.TrimSuffix(poolmgrUrl, "/")}
+	c := &Client{
+		poolmgrUrl:  strings.TrimSuffix(poolmgrUrl, "/"),
+		tappedByUrl: make(map[string]bool),
+		requestChan: make(chan clientRequest, 100),
+		tapUrlChan:  make(chan string, 100),
```
soamvasani: See below comment, I think we can remove `tapUrlChan` and `tapWorker`

poolmgr/client/client.go
```diff
@@ -60,12 +84,49 @@ func (c *Client) GetServiceForFunction(metadata *fission.Metadata) (string, erro
...
+}
+
+func (c *Client) tapWorker() {
+	for {
+		u := <-c.tapUrlChan
+		c._tapService(u)
+	}
+}
+
+func (c *Client) timer() {
```
yqf3139: Thanks. I was worried about if there are too many requests and then the ticker channel may not be selected. I ran an experiment and it worked fine. Will make the changes you proposed.

poolmgr/client/client.go
```diff
@@ -60,12 +70,38 @@ func (c *Client) GetServiceForFunction(metadata *fission.Metadata) (string, erro
...
+			c.tappedByUrl[serviceUrl] = true
+		case <-ticker.C:
+			urls := c.tappedByUrl
+			c.tappedByUrl = make(map[string]bool)
+			if len(urls) > 0 {
+				go func() {
+					for u := range c.tappedByUrl {
+						c._tapService(u)
+					}
+					if len(urls) > 0 {
```
soamvasani: Minor nitpick: you already checked `len(urls) > 0`, before launching this goroutine, so it shouldn't be needed here.

yqf3139: Thanks. I will fix this.

Issue #193, opened 2017-05-18
Execution strategies for functions


soamvasani:
Fission today uses a pool of running environment pods and loads the function into a running pod on-demand.  This is great for cold-start times but it isn't the only strategy.  An alternative is to launch a a new pod on each cold start. This is slower, but acceptable when cold-start times aren't a priority (most async use cases).

There are trade-offs of each approach and neither one fits all use cases.  The pool buys us very fast cold start (especially with preloading common libraries, cold start can be < 10ms), which is important for any use case where the latency is "visible" to a user.  Mostly this the REST API use case.  But, using a pool doesn't allow us to mount volumes for a function, or set CPU/memory requests and limits.

We should implement multiple strategies for executing functions, and choose an appropriate one automatically based on what the function needs.  For example, if a function needs a volume mounted, it should use the new-pod strategy.

The downside of this approach is that a user can't have both volume access and fast cold-starts.  In the future we can investigate working on k8s changes to allow dynamic volume mounting in running pods (though I have no idea how the k8s maintainers feel about that).  But short of that, there's no way to get both fast cold-start and volumes in the same function.

Hybrid approaches are also possible: the first instance of a function can be drawn from the pool (to make cold start fast), and subsequent instances can be be created in new pods (to make autoscaling easier).  The beginnings of this implementation already exist on the `autoscaling` branch.

erwinvaneyk:
Interesting idea about the strategy-based approach @soamvasani. There is definitely something to gain by deploying the functions based on the priority/requirements of a function. 

Regarding the volume mounting, what are the use cases? If it is for caching/temporary storage, it might be an idea to look at how AWS Lambda does it. There it provides by default a `tmp` storage (~512MB) that persists (to some extend) across invocations.

soamvasani:
The other use case for this is if users want to specify functions as docker images. Then we'd need to start the pod on request.

soamvasani:
So, we keep coming with use cases for doing this.

This approach will allow more fine-grained tuning of resource utilization and auto-scaling. It would allow functions to have Kubernetes [`ResourceRequirements`](https://godoc.org/k8s.io/kubernetes/pkg/api#ResourceRequirements). Functions will also be able to specify some characteristics of their Deployment, like the initial scale, and some attributes for the autoscaler, such as min and max scale. (We'll have to add all these properties to the FunctionSpec.)

[...]

PR #230, opened 2017-06-13
Use horizontal autoscaler to autoscale

yqf3139:
These two commits are based on @soamvasani previous commits.
- Use horizontal autoscaler to scale the function instance, currently deployment cpu usage is the only metric
- Make hpa related params adjustable
- Logger watch k8s pod events to setup/unset logging for function automatically

I use the function below to generate loads, on a two core k8s in GCP
```
module.exports = function(context, callback) {
    for (i=0; i<1000;i++) {
        x = Math.random();
    }
    callback(200, "Hello, world!\n");
}
```

### DIFF
--- a/poolmgr/gp.go
+++ b/poolmgr/gp.go
@@ -352,15 +352,26 @@ func (gp *GenericPool) createFunctionDeployment(
 		return err
 	}
 
-	rs, err := gp.kubernetesClient.Extensions().ReplicaSets(gp.namespace).List(api.ListOptions{
-		LabelSelector: labels.Set(funcLabels).AsSelector(),
-	})
-	if err != nil || len(rs.Items) == 0 {
-		fmt.Println("replicasets is nil or empty")
+	// try to fetch the corresponding replica set for 5 times
+	var rs *v1beta1.ReplicaSet = nil
+	for range [5]struct{}{} {
+		rsList, err := gp.kubernetesClient.Extensions().ReplicaSets(gp.namespace).List(api.ListOptions{
+			LabelSelector: labels.Set(funcLabels).AsSelector(),
+		})
+		if err != nil || len(rsList.Items) == 0 {
+			fmt.Println("replicasets is nil or empty, retry later")
+			time.Sleep(500 * time.Microsecond)
+			continue
+		}
+		rs = &rsList.Items[0]
+		break
+	}
+	if rs == nil {
+		fmt.Printf("replicaset not found, label template-hash to pod [%v] failed", pod.ObjectMeta.Name)
 		return nil
 	}
 
-	pod.Labels["pod-template-hash"] = rs.Items[0].Labels["pod-template-hash"]
+	pod.Labels["pod-template-hash"] = rs.Labels["pod-template-hash"]
 	_, err = gp.kubernetesClient.Core().Pods(gp.namespace).Update(pod)
 	if err != nil {
 		log.Printf("failed to add template hash to pod [%v]: %v", pod.ObjectMeta.Name, err)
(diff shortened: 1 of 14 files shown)


router/functionHandler.go
```diff
@@ -52,11 +64,12 @@ func (fh *functionHandler) getServiceForFunction() (*url.URL, error) {
 type RetryingRoundTripper struct {
 	maxRetries    int
 	initalTimeout time.Duration
+	transport     http.Transport
 }
 
 func (rrt RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	timeout := rrt.initalTimeout
-	transport := http.DefaultTransport.(*http.Transport)
+	transport := &rrt.transport
```
yqf3139: The default transport will send all the traffic to the same backend pod via the service url, though there are several pods behind the service.

yqf3139:
To make sure the HPA is working, follow the instructions in [here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/). 

Also notice the default schedule period for HPA is 30 seconds:

> The Horizontal Pod Autoscaler is implemented as a control loop, with a period controlled by the controller manager’s --horizontal-pod-autoscaler-sync-period flag (with a default value of 30 seconds).

yqf3139:
In #193, a hybrid approach is proposed and implemented in this branch.
> the first instance of a function can be drawn from the pool (to make cold start fast), and subsequent instances can be be created in new pods (to make autoscaling easier).

I add resource restriction to the environment to make HPA working. Setting request/limit on the environment loses some flexibility, an alternative way is that user can create multiple environments to suit their need (node-high-cpu, node-low-cpu), which I think is practicable for common usage scenarios.

router/functionHandler.go
soamvasani: Could you elaborate? Why would this be related to what transport is used?

yqf3139: @soamvasani 
I can use a experiment to reproduce but I cannot fully explain it yet. I will dig into the reason later. Please use this [yaml](https://gist.github.com/yqf3139/e7a39dace9335ac28fb8d3552d291410). The code is [here](https://gist.github.com/yqf3139/e9a9ec3bf33452c55e07aa135eb56619).

I create one proxy pod and two server pod. When we visit the proxy pod, it visits the server service and proxy the result back. We should see different uuid of the server when curling the node_ip:31320/proxy.

If we use the `use-pointer` tag image, the response is the same among all the response. `use-copy` is the correct one, which uses a copy of transport instead of a pointer.
I tried to set the `keep-alive` to 0 in the default transport but it is not working.

yqf3139: I just verified that setting `DisableKeepAlives` to `true` is working. But seems we are using the `RetryingRoundTripper`, it modifies the timeout field of the transport. If the transport is shared by all the router's proxy requests, it may cause problems?

poolmgr/gp.go
```diff
@@ -327,9 +358,46 @@ func (gp *GenericPool) createFunctionDeployment(metadata *fission.Metadata, labe
...
+
+	rs, err := gp.kubernetesClient.Extensions().ReplicaSets(gp.namespace).List(api.ListOptions{
+		LabelSelector: labels.Set(funcLabels).AsSelector(),
+	})
+	if err != nil || len(rs.Items) == 0 {
+		fmt.Println("replicasets is nil or empty")
+		return nil
+	}
+
+	pod.Labels["pod-template-hash"] = rs.Items[0].Labels["pod-template-hash"]
```
yqf3139: Relabeling the pod so that it can be adopted by the new replica set could be problematic.
There is a chance that the new replica set set is not created when we relabel the pod.
Async listing the new replica set several times may help. Need to find a better way to transfer ownership of pod.

soamvasani:
closing in favor of #387 etc.

PR #232, opened 2017-06-16
Fix creating redundant pods on heavyload coldstart

yqf3139:
This is a proposal for fixing creating redundant pods on heavy load cold start.

When there are many requests on a cold function and the func service is not ready, each of the request will trigger poolmgr to specialize a pod. If the env pool size is smaller than the incoming concurrent requests, the request is hung up until a pod is available in the env pool. Then a pod is specialized and deleted if a func service already exists (beated by other requests). In `poormgr.go`, there is a already TODO showing we need a fix for this.

In this commit, the first request will create the func service as usual and create a channel for other requests to block on. Once the func service is ready, it will notify all the wait requests to fetch the func request from the cache again. There is no lock to visit the func service waiting channel, so there is a tiny chance that multi goroutines step into creating the same func service.

I use `ab -n 10 -c 10` to test the commit on a two vCPU, single node GKE. Time per request is 1842.712 ms (current master) vs 174.420 ms (this commit).

### DIFF
--- a/poolmgr/api.go
+++ b/poolmgr/api.go
@@ -44,24 +44,79 @@ type funcSvc struct {
 	atime time.Time
 }
 
+type createFuncServiceRequest struct {
+	funcMeta *fission.Metadata
+	respChan chan *createFuncServiceResponse
+}
+type createFuncServiceResponse struct {
+	address string
+	err     error
+}
+
 type API struct {
-	poolMgr      *GenericPoolManager
-	functionEnv  *cache.Cache // map[fission.Metadata]fission.Environment
-	fsCache      *functionServiceCache
-	controller   *controllerclient.Client
-	fsCreateChan map[fission.Metadata]chan struct{}
+	poolMgr          *GenericPoolManager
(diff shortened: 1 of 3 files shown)


poolmgr/api.go
```diff
@@ -135,6 +137,30 @@ func (api *API) getServiceForFunction(m *fission.Metadata) (string, error) {
 		return fsvc.address, nil
 	}
 
+	// Cache miss -- is this first one to request the func?
+	createCh, found := api.fsCreateChan[*m]
```
soamvasani: Since the API is called in multiple goroutines, this access to the map is unsafe, as is the one below that adds to the map.

yqf3139: Yes, but the chance is tiny. We can add access control to it if necessary.

soamvasani:
Thanks, these changes look great :) Just to verify -- did you test the final version of the change?

yqf3139:
@soamvasani Yes. It is tested.

Issue #203, opened 2017-05-18
Add a CLI command to launch the UI


soamvasani:
`fission ui` should open fission-ui in a browser, if it is installed. If the UI isn't installed, it should output directions on how to install the ui.

soamvasani:
@yqf3139 @dgoujard

yqf3139:
To fulfill this issue, there are some problems:
- To determine fission-ui is installed, I saw #190 and will try to read the fission-ui service from k8s api
- To open an url on different platform, I think we can find a package like [this] (https://github.com/skratchdot/open-golang), or we have to wrap it for ourselves

life1347:
Fission UI is deprecated. Close this until we have newer version.

Issue #190, opened 2017-05-18
Get rid of FISSION_URL


soamvasani:
The fission cli should use the kubernetes api to find the fission service and contact it directly, instead of requiring the user to set FISSION_URL.

soamvasani:
Done in 0.6

PR #242, opened 2017-06-25
add a CLI command to get fission url and launch UI

yqf3139:
- issue #190  instead of use the url directly, the url is set to env variables
  manually
- issue #203

### DIFF
--- a/fission/common.go
+++ b/fission/common.go
@@ -17,12 +17,21 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
+
+	"k8s.io/client-go/1.5/kubernetes"
+	"k8s.io/client-go/1.5/pkg/api"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+	"k8s.io/client-go/1.5/tools/clientcmd"
 
 	"github.com/fission/fission/controller/client"
 )
+
+const UI_INSTALL_URL = "https://github.com/fission/fission/blob/master/INSTALL.md#use-the-web-based-fission-ui-optional"
 
 func fatal(msg string) {
 	os.Stderr.WriteString(msg + "\n")
@@ -32,7 +41,11 @@ func fatal(msg string) {
 func getClient(serverUrl string) *client.Client {
 
 	if len(serverUrl) == 0 {
-		fatal("Need --server or FISSION_URL set to your fission server. Or try source <(fission get-env-variables)")
+		serverUrl = getFissionEnvVariable("controller")
+	}
+
+	if len(serverUrl) == 0 {
+		fatal("Need --server or FISSION_URL set to your fission server. Or check kubernetes config.")
 	}
 
 	isHTTPS := strings.Index(serverUrl, "https://") == 0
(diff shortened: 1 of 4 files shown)


soamvasani:
Thanks @yqf3139 ! 

[@life1347 was also going to work on #190, so you two might want to talk... in the future it will help to comment on an issue when you start working on it, to avoid duplication of efforts.]

If we use the [apiserver proxy](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls), we don't need to expose the controller service at all, and we can take advantage of apiserver authentication too.  This will be more reliable and secure than our current set up.

fission/common.go
```diff
@@ -32,7 +32,7 @@ func fatal(msg string) {
 func getClient(serverUrl string) *client.Client {
 
 	if len(serverUrl) == 0 {
-		fatal("Need --server or FISSION_URL set to your fission server.")
+		fatal("Need --server or FISSION_URL set to your fission server. Or try source <(fission get-env-variables)")
```
soamvasani: We should eliminate this fatal error, and get the url from the API. Why make the user set a var that we know how to get automatically?

fission/main.go
```diff
@@ -17,11 +17,105 @@ limitations under the License.
...
 
 import (
+	"errors"
+	"fmt"
 	"os"
+	"path/filepath"
 
+	"github.com/skratchdot/open-golang/open"
 	"github.com/urfave/cli"
+	"k8s.io/client-go/1.5/kubernetes"
```
soamvasani: Would you mind splitting the k8s-client stuff into a separate file in this package?

fission/main.go
```diff
@@ -17,11 +17,105 @@ limitations under the License.
...
+		// TODO: find the port via name
+		ip, err3 := getNodeIP(k8sClientset)
+		if err3 != nil {
+			panic(err3)
+		}
+		return fmt.Sprintf("http://%v:%v",
+			ip, controller.Spec.Ports[0].NodePort)
+	case v1.ServiceTypeLoadBalancer:
+		return fmt.Sprintf("http://%v:%v",
+			controller.Status.LoadBalancer.Ingress[0].IP, controller.Spec.Ports[0].Port)
```
soamvasani: Sometimes the IP isn't set, and there's a hostname instead (seems to happen on AWS).

fission/main.go
```diff
@@ -124,6 +218,35 @@ func main() {
 				return nil
 			},
 		},
+		{
+			Name:  "get-env-variables",
```
soamvasani: Suggest removing this.

fission/common.go
yqf3139: We can get the server var only when the k8s config is present and valid. So keeping the environment variable is a good idea?

fission/main.go
yqf3139: Yes. Current implementation is not easy to use. But we should find a way to cache the fission url or we need to fetch the url every time cli runs.

fission/common.go
soamvasani: Yes we should allow the user to set the env var to override the default. But we shouldn't fail if it's unset.

fission/main.go
soamvasani: How slow is it to fetch the url every time? And btw, if we switch to using the apiserver proxy, we won't have to talk to the api to get the url.

yqf3139: Less than two seconds, depends on where your k8s cluster is.
For apiserver proxy, we still need to know the k8s master ip from the k8s local config. I haven't dig into that.

life1347: Two seconds is a little bit long to me. How about write related information into local config file (/home/.fission/config)?

yqf3139: @soamvasani 
Local config file is good to to me. What is your opinion?

soamvasani:
closing in favor of #431

PR #244, opened 2017-06-30
Update k8s client version to 3.0.0 beta

yqf3139:
Update k8s client from v1.5.0 to v3.0.0beta.
Fix the api changes between the two versions of client.


yqf3139:
@soamvasani 
I am working on service-catalog integration and the repo needs k8s-client version > 2.0.0.
If we want to use 2.0.0, I can rollback, since service-catalog PR requires several days to discuss and implement.

soamvasani:
We could wait until client-go release 3.0.0 is stable, and then merge this; if we do that there's no need to rollback.

PR #249, opened 2017-07-04
include path parameters in params hash for ruby environment

sbfaulkner:
this PR adds a helper method in the ruby context/request to make using path parameters easier

e.g.
given a route /path/:id
and a request for /path/123
a header would be present with the form `X-Fission-Params-Id: 123`
and with this update `request.params` would return the hash `{'id' => '123'}`

### DIFF
--- a/environments/ruby/README.md
+++ b/environments/ruby/README.md
@@ -2,9 +2,19 @@
 
 This is the Ruby environment for Fission.
 
-It's a Docker image containing a Ruby 2.4.1 runtime, along with a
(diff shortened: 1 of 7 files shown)


soamvasani:
Thanks @sbfaulkner ! Mind updating the README with this info?

sbfaulkner:
@soamvasani updated README(s) and improved structure of ruby env

sbfaulkner:
(logger addition will follow in another PR yet to be written)

soamvasani:
Awesome update! 😄 Merging now.

PR #251, opened 2017-07-06
Ruby logger

sbfaulkner:
This PR sets up the Rack logger, exposes it through the context, and updates the README and example

(also updates the error cases in the specializer and handler to log failures)


soamvasani:
Thanks @sbfaulkner !

PR #258, opened 2017-07-11
Fix nodejs env restarting issues (#254)

### LINKED ISSUE #254: Nodejs env cannot catch exceptions in callback functions and restarting causes failure
In node js functions, we often invoke some async functions and then exceptions in the callback functions cannot be catched by `server.js`. This often causes the process to exit.

Here is an example code causes this issue:
```
module.exports = function(context, callback) {
    setTimeout(function(){
        callback(200, "");
        throw new Error('');
    }, 1000);
}
```
The exception crashes the nodejs env, when kubernetes try to restart the nodejs env, another error occurs and then the pod continues to crash loop backoff.
```
Error: EEXIST: file already exists, symlink '/usr/src/app/node_modules' -> '/userfunc/node_modules'
```
The second issue is pretty easy to solve. But for the first, how can we handle this type of exceptions?

yqf3139:
- handle uncaught exceptions
- fix node modules symlinking failure when container restarts
- reload user code when container restarts

Closes #254

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/258)

### DIFF
--- a/environments/nodejs/server.js
+++ b/environments/nodejs/server.js
@@ -19,6 +19,12 @@ if (!argv.port) {
     argv.port = 8888;
 }
 
+// To catch unhandled exceptions thrown by user code async callbacks,
+// these exceptions cannot be catched by try-catch in user function invocation code below
+process.on('uncaughtException', (err) => {
+    console.error(`Caught exception: ${err}\n`);
+    console.log('Container resume running')
+});
 
 // Node resolves module paths according to a file's location. We load
 // the file from argv.codepath, but tell users to put dependencies in
@@ -27,7 +33,10 @@ if (!argv.port) {
 // function deps in the right place in argv.codepath; but for now we
 // just symlink the function's node_modules to the server's
 // node_modules.
-fs.symlinkSync('/usr/src/app/node_modules', `${path.dirname(argv.codepath)}/node_modules`);
+// try to symlink because the link exists if the container restarts
+if (!fs.existsSync(`${path.dirname(argv.codepath)}/node_modules`)) {
+    fs.symlinkSync('/usr/src/app/node_modules', `${path.dirname(argv.codepath)}/node_modules`);
+}
 
 // User function.  Starts out undefined.
 let userFunction;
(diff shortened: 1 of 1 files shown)


environments/nodejs/server.js
```diff
@@ -19,6 +19,12 @@ if (!argv.port) {
     argv.port = 8888;
 }
 
+// To catch unhandled exceptions thrown by user code async callbacks,
+// these exceptions cannot be catched by try-catch in user function invocation code below
+process.on('uncaughtException', (err) => {
+    console.error(`Caught exception: ${err}\n`);
+    console.log('Container resume running')
```
erwinvaneyk: You can leave this out. Seems to be a bit of a redundant message

environments/nodejs/server.js
```diff
@@ -72,8 +85,18 @@ app.post('/specialize', specialize);
 // Generic route -- all http requests go to the user function.
 app.all('/', function (req, res) {
     if (!userFunction) {
-        res.status(500).send("Generic container: no requests supported");
-        return;
+        if (fs.existsSync(argv.codepath)) {
```
erwinvaneyk: I would prefer inverting this condition, such that you can lose the `else`

```javascript 
if(!fs.exist(...)) {
   res.status(500).send(....);
   return;
}

try {
 //...
} catch(e) {
// ...
}
```

environments/nodejs/server.js
```diff
@@ -27,7 +33,10 @@ if (!argv.port) {
 // function deps in the right place in argv.codepath; but for now we
 // just symlink the function's node_modules to the server's
 // node_modules.
-fs.symlinkSync('/usr/src/app/node_modules', `${path.dirname(argv.codepath)}/node_modules`);
+// try to symlink because the link exists if the container restarts
```
erwinvaneyk: Very minor: `// Check for symlink, because....` would make more sense

andreialecu:
I'm just trying fission but I was hit by this problem right off the bat. It also seems to lock up the VM because of what I assume is an infinite loop while it keeps trying to symlink node_modules.

Is no one using nodejs with fission or am I doing something wrong?

vishal-biyani:
Hi @erwinvaneyk @yqf3139 Is this still relevant?

andreialecu:
Actually my problem was something else. The node handler function **needs** to have exactly 1 argument, even though I wasn't using the argument and didn't need it. So it was hanging because of that. 

So I had this error happen because of this check: https://github.com/fission/fission/blob/master/environments/nodejs/server.js#L136
Should be <= 1. so that `module.exports = function () { ... }` is also valid.

I was still seeing symlink `node_modules` related errors in the log, from subsequent runs of the same failing handler function. But after fixing the function it seems to have cleared up. 

Could still result in errors under some other circumstances, I guess, so a check could be useful.

erwinvaneyk:
@vishal-biyani I am not sure anymore; it has been too long ago since I reviewed it.

Issue #257, opened 2017-07-11
Add function invoke subcommand


erwinvaneyk:
There has already been some talk (can't find the exact issue at the moment) about adding a `fission fn invoke <name/id>` to bypass the http route setup, as it is generally irrelevant during the development/testing of the function.

~~To take it one step further, I propose that we expose functions by default. This could be enabled/disabled by toggling some development flag on the fission-bundle, or leave it to the users to block requests to the 'dev' HTTP namespace.~~

When you create a function "hello", given that the function name can be assumed to be unique. We immediately expose the function (for all HTTP methods) through something like this:
`$FISSION_ROUTER/_function/hello`. When the function versioning is added to the master branch, we could change the path to include the version: `$FISSION_ROUTER/_function/hello-<version>`. In that case the path without a version will simply point to the newest version.

When the function is deleted, the paths are also immediately deleted.

This will also allow a `fission function invoke <name/id>` to be trivial to implement.

yqf3139:
@erwinvaneyk  
Every function is default on `$FISSION_ROUTER/fission-function/$FUNCION_NAME` after the function is created. This feature is not well documented. We can add the cli `invoke` command and furthermore, toggling, blocking and versioning.

erwinvaneyk:
Ah thanks @yqf3139 , I missed that completely then 👍 

I will leave this issue open for the CLI improvements in that case.

soamvasani:
Thanks for bringing up this issue @erwinvaneyk and @yqf3139.

We need to separate those default `fission-function` routes to another router, so that the main router can be exposed outside the cluster without exposing all functions.

And once we do that, we will also need an API that will allow functions without triggers to be invoked. That API would be subject to authentication, of course.

soamvasani:
`fission fn test` does this

PR #259, opened 2017-07-11
Add 'function invoke' subcommand

erwinvaneyk:
This PR adds an initial "function invoke" subcommand to the CLI. 

Currently a downside is that it requires the FISSION_ROUTER to be present. This should be just temporary issue. As discussed #257 at some point the internal routes will be extracted and placed in an internal router, which will allow the CLI connect to the cluster API server and get proxied to the internal router. Until then this seems like a useful addition to the current CLI.

### DIFF
--- a/fission/common.go
+++ b/fission/common.go
@@ -29,6 +29,10 @@ func fatal(msg string) {
 	os.Exit(1)
 }
 
+func warn(msg string) {
+	os.Stderr.WriteString(msg + "\n")
+}
+
 func getClient(serverUrl string) *client.Client {
 
 	if len(serverUrl) == 0 {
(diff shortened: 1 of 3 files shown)


soamvasani:
I'd like to avoid the dependence on FISSION_ROUTER.  You can get at the router using the apiserver proxy instead:
https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls

erwinvaneyk:
Agreed, in that case, I am holding this PR until #242 is merged, in order to use the k8s client to find the router.

erwinvaneyk:
Closing in favor of #355

Issue #246, opened 2017-07-01
Idempotent commands


rochdev:
Right now most fission commands have a `create` and `update` subcommands but there is no equivalent idempotent subcommand.

`kubectl` has the `apply`command and I think it would make sense to add the same functionality as `fission <command> apply`.

The use case would be when you don't know if a function already exists on the cluster, for example when using automation and continuous delivery.

soamvasani:
While we don't have `fission apply` just yet, we do have `kubectl apply` since we switched to using kubernetes resources for our state.

life1347:
Close this since we have `fission spec apply` now

Issue #184, opened 2017-05-09
allow customization of fetcher image url


Slahser:
thx for your work first.

i am building a fission environment offline

because fetcher block it 

```
fission --server http://ip:31313 env create --name nodejs --image registry.me.com/node-env
```

is there any flag to solve it?

soamvasani:
This isn't configurable yet.  

If anyone would like to take a crack at fixing this before I get to it, this is the place where we have the fetcher URL:
https://github.com/fission/fission/blob/master/poolmgr/gp.go#L355

We should change that to an env var, and then the env var could be set through the deployment yaml.

Slahser:
sigh for my poor golang ability...

waiting for a good pr

soamvasani:
Available since 0.2.1 (specify `fetcherImage` and `fetcherImageTag` as helm variables).

Issue #129, opened 2017-02-18
Use Kubernetes ThirdPartyResources for storing state


kfox1111:
Kubernetes has a 3rd party resource type that allows extending the kubernetes api by 3rd parties.

It would be very interesting to add support for this in fission. The user could then upload an environment description as a 3rd party object, and the script file itself as a 3rd party object.

This even would allow the user to deploy those objects in helm charts along with other k8s objects.

soamvasani:
Closing, we've switched to TPRs a while ago (and will be switching to CRDs soon).

Issue #103, opened 2017-02-02
Environments v2 


soamvasani:
This is an issue that's tracking the work around redesigning the environment interface.  This covers #24 but also various other issues, such as dependency resolution, better handling of compile-time errors, etc.

Current status: writing up requirements and user stories.  The doc is in Documentation/wip/environments-v2.md.

soamvasani:
@ktrance @dgoujard you might be interested in this issue; I will update it as we progress. If you'd like to add anything to the requirements, from a C# and a PHP point of view, this is a good place to discuss it.

ktrance:
Will do. 
Just looked at the AWS deployment package concept and for dotnet something like this would probably be the easiest to quickly get up and running. The developer is responsible for for packaging every component he needs which we wouldn't have to deal with open ports to get nuget packages etc. 

We could just:
1. dynamically load the main binary either from some specification or we would need  to scan all of them 
2. find a predefined interface/contract which we in essence have today in the dotnet env
3. execute the function

Nice and easy. I would be interested in doing a prototype of this.

dgoujard:
@soamvasani For PHP env ideal workflow will be : 
1. Create a zip with multiples files (php and composer.json if needed) 
2. Function on /specialize phase will install composer.json package
3. Function will include "index.php" file on each request (and call handler if present)

This workflow did not allow to add php extension but it's more common to add package than extension (and i added many extension btw)

soamvasani:
@ktrance So in this design, the user is responsible for doing the builds and creating a zip file, right?  It's a good idea -- it would allow us to get started with compiled language support without solving the bigger problem of builder containers too.  We still need to solve storage for these packages, but I will try to spec something out for that.

soamvasani:
@dgoujard I'm sorry my knowledge of PHP isn't good enough to understand this... what is the difference between a package and an extension?

We're also trying to separate out the install steps from the runtime.  So basically we'd have a separate builder that would install dependencies, create some sort of package -- a zip file for example -- and then have the runtime load that when /specialize is called.

dgoujard:
@soamvasani an extension need a compilation and loading at php startup. An package is just a library like you load in python with package.txt.

In this case the builder pod will download packages (composer.json) and the function pod will just execute code as today

ktrance:
@soamvasani , yes that would be the general idea.

It would of course be a pain to go through if you only want to run a 5 line function but then you could still use the existing solution.

[...]

PR #266, opened 2017-07-26
V2 types and TPR

soamvasani:
This changes the core fission function, environment and trigger types. It also changes Fission's storage to use ThirdPartyResources.

1. Functions are now specified by packages. Functions can also have both source and deployment packages. A package can be specified by a literal, or by a URL.
2. Environments have a build and runtime component. 
3. Triggers reference functions by a FunctionReference. This is a layer of indirection between triggers and functions, and will allow things like incremental function upgrades in future releases.

See Documentation/wip/env-v2.md for design discussion about points 1 and 2. 

Operationally, moving our state to TPRs means that users no longer have to maintain an etcd instance just for fission, and they also don't have to use persistent volumes for the fission api server. It also means they can specify functions and triggers declaratively and version control these declarations.

This change breaks compatibility with the v1 Fission API. It removes versioning from the core API. Some points of rationale for this:
1. Moving our type system closer  to Kubernetes architectural patterns is important, gives us declarative infrastructure, allows better interoperability with other K8s projects, and the potential to use tools like Helm to install functions in the future. 
2. Versioning in the v1 Fission API was based on immutability of objects. It also had a very implicit notion of "latest". Both these concepts don't fit cleanly into K8s API patterns.
3. We also wanted to avoid doing too much in one layer; it is best to look at the core of fission as a platform that you deploy functions into, and implement some sort of "Versioned Function" object on top of this layer.
4. Adding a v1 compatibility layer was in theory possible, but messy because of the immutability in v1. At the current pre-beta stage of the project it wasn't worthwhile adding and maintaining this complexity.
5. Broadly, we take breaking compatibility very seriously, but we're early enough in the project to be able to do it.

Note that all v1 _environments_ are still compatible Fission even after the API change. This is largely because environments don't have fission-api-specific stuff in them.

The fission CLI retains all of its old commands, but loses the versioning semantics. To migrate from a v1 to v2 deployment, you have to use the v1 CLI to get functions, envs and triggers; and then use the v2 CLI to create them again. This is a rather manual process. We'll be happy to help anyone having trouble with this process on the fission slack channel.

This is not a complete Environment v2 implementation; the remaining changes for that are:
1. adding zip file support to fetcher
2. implementing a storage service for large packages
3. build management: a build manager to invoke builders, and API and CLI support for running builds, showing errors, etc.


soamvasani:
cc @yqf3139 @dgoujard This change breaks compatibility with the UI. I was initially planning on a compatiblity layer, but I think it would be better to update the UI for the new types. Let me know if I can help with that in any way.

yqf3139:
@soamvasani Thanks. I made some experiments with TPR in fission-benchmark. I will work on the new types.

soamvasani:
Issues #103, #129, #184, #246.

types.go
```diff
@@ -17,72 +17,151 @@ limitations under the License.
...
+	//
+	// Functions and packages
+	//
+
+	// Checksum of package contents when the contents are stored
+	// outside the Package struct. Type is the checksum algorithm;
+	// "sha256" is the only currently supported one. Sum is hex
+	// encoded.
+	Checksum struct {
+		Type string `json:"type"`
```
erwinvaneyk: You might want to alias this to ChecksumType, similar to PackageType

types.go
```diff
@@ -17,72 +17,151 @@ limitations under the License.
...
+	Package struct {
+		// Type defines how the package is specified: literal or URL.
+		Type PackageType `json:"type"`
+
+		// Literal contents of the package. Can be used for
+		// encoding packages below TODO (256KB?) size.
+		Literal []byte `json:"literal"`
+
+		// URL references a package.
+		URL string `json:"url"`
```
erwinvaneyk: I have not read through all the changes yet, but why do you need the distinction here between literal and URL? That might be something to extract from the actual `Package` struct to something like this:
```
PackageLocation (Url/Filesystem) ---fetch--> Package
```
That would make the package agnostic of what way it was retrieved.

types.go
soamvasani: done.

types.go
```diff
@@ -92,6 +175,52 @@ type (
 	}
 
 	errorCode int
+
+	//
+	// Fission-Environment interface. The following types are not
```
erwinvaneyk: Is this just about `FunctionLoadRequest`? For clarity it might be an idea to split these internal type definitions to a separate file/package.

types.go
```diff
@@ -17,72 +17,155 @@ limitations under the License.
...
-	// Environment identifies the language and OS specific
-	// resources that a function depends on.  For now this
-	// includes only the function run container image.  Later,
-	// this will also include build containers, as well as support
-	// tools like debuggers, profilers, etc.
-	Environment struct {
-		Metadata             `json:"metadata"`
-		RunContainerImageUrl string `json:"runContainerImageUrl"`
+	// FunctionSpec describes the contents of the function.
+	FunctionSpec struct {
```
erwinvaneyk: Is this the definition of a function in general or the definition of a (deployed) function instance.

soamvasani: It's the definition of a fission function. Not sure I get the question, what do you mean by "instance"?

types.go
soamvasani: Yes it is; I'll split it out.

types.go
erwinvaneyk: Ah, I thought there was also a plan to introduce a declarative function definition (something like the serverless.yml file in the serverless framework or npm's package.json). I thought this was it. :P So how do users have to specify package metadata (like the package entrypoint and version) now?

soamvasani: This is part of the plan for declarative definitions. You can have a yaml file that's the serialized version of these objects, just like any other kubernetes yaml files. (Client tooling to work with those files will be improved in future changes.)

PR #274, opened 2017-08-10
Different Minishift example

dmlond:
adds the cluster-admin policy to developer, then uses the developer role
exposes the router and service using the nip.io (or alternative) proxy that comes with minishift (on my mac, at least)

I use jq to parse the json output from oc export to get the FISSION_URL and FISSION_ROUTER. There may be a way to do it without jq.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/274)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -122,8 +122,8 @@ $ oc create -f https://github.com/fission/fission/releases/download/nightly20170
 Identically as with Kubernetes, you need to set the FISSION_URL and FISSION_ROUTER environment variables. If you're using minishift, use these commands:
 
 ```
-  $ export FISSION_URL=http://$(oc export route/controller -o json | jq -r '.spec.host')¬
-  $ export FISSION_ROUTER=$(oc export route/router -o json | jq -r '.spec.host')¬
+  $ export FISSION_URL=http://$(oc get route/controller -o=jsonpath='{.spec.host}')¬
+  $ export FISSION_ROUTER=$(oc get route/router -o=jsonpath='{.spec.host}')¬
 ```
 After these steps, you should be able to run fission client as with kubernetes.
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -94,20 +94,19 @@ permissions (like `system:admin`), as it needs to create a
 `ClusterRole` for deploying function containers from the `fission`
 namespace/project.
 
-Identically as with Kubernetes, you need to set the FISSION_URL and FISSION_ROUTER environment variables. If you're using minishift, use these commands:
-
-```
-  $ export FISSION_URL=http://$(minishift ip):31313¬
-  $ export FISSION_ROUTER=$(minishift ip):31314¬
-```
 #### Using Minishift or Local Cluster
```
soamvasani: This section is now missing the instructions to set FISSION_URL and FISSION_ROUTER?

dmlond: I moved this section to start on line 121. I also modified it to use the oc command instead of minishift ip.

soamvasani: But that section is for "other clouds", as an alternative to minishift or a local cluster.  Maybe it should just be a new section, "CLI setup".

dmlond: I see what you mean. One thing I do not have experience with is how relevant the 'using other clouds' section is, since I have only done this on minishift. If you notice, I modified both sections to use https://github.com/fission/fission/releases/download/nightly20170705/fission-cloud.yaml so they are no longer that much different.  It may be that you could use the same oc calls to expose the router and controller in minishift and other clouds.  In any case, it might make sense to separate the FISSION_URL stuff into a separate section since it should work the same in minishift and other clouds.

vishal-biyani:
Hi @dmlond We have moved the docs to another repo: https://github.com/fission/docs.fission.io
Do you mind please sending this PR there?

vishal-biyani:
Hi @dmlond Thanks again for your effort - request you to raise the PR in the docs repo as mentioned in earlier comment. I will close this PR for now here.

PR #280, opened 2017-08-11
Set fetcher image through poolmgr env

life1347:
This PR aims to update helm charts, so that a user can change fetcher image with poolmgr env (https://github.com/fission/fission/blob/master/poolmgr/gp.go#L90).

### DIFF
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -6,7 +6,7 @@
 ## For minikube, set this to NodePort, elsewhere use LoadBalancer.
 serviceType: LoadBalancer
 
-## Fission image repostiroy
+## Fission image repository
 image: fission/fission-bundle
 
 ## Image pull policy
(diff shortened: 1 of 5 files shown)


soamvasani:
Thanks @life1347 ! Can you please remove the etcd change? Better to have one thing at a time in a PR.

life1347:
I move the second commits to another PR. :)

soamvasani:
Can you rebase to master, please?  That should fix the travis ci breakage.

charts/fission-all/values.yaml
```diff
@@ -15,6 +15,12 @@ pullPolicy: ifNotPresent
 ## Fission image version
 imageTag: nightly20170705
 
+## Fission fetcher repostiroy
```
soamvasani: typo: repository

charts/fission-core/values.yaml
```diff
@@ -12,6 +12,12 @@ image: fission/fission-bundle
 ## Fission image version
 imageTag: alpha20170124
 
+## Fission fetcher repostiroy
```
soamvasani: typo: repository

PR #284, opened 2017-08-13
Parse metadata.Name before creating tpr resource

life1347:
This small PR aims to improve the usability of fission CLI when a user tries to create a resource with an invalid name. (https://github.com/fission/fission/issues/269)


soamvasani:
Thanks, @life1347 !

Issue #286, opened 2017-08-16
Fetcher failed to access TPR if RBAC is enabled


life1347:
```
2017/08/16 18:52:17 [hello1-1] calling fetcher to copy function
10.0.0.163 - - [16/Aug/2017:18:52:17 +0000] "POST /v2/getServiceForFunction HTTP/1.1" 500 119
2017/08/16 18:52:17 Error: 500: Failed to get function: the server does not allow access to the requested resource (get functions.fission.io hello1-1)
2017/08/16 18:52:17 Error in pod 'nodejs-0349cce7-82b4-11e7-a9d6-42010af000f2-lo3jqjpr-259088tp32', scheduling cleanup
```
With latest TPR commit, fetcher failed to access TPR due to RBAC.

soamvasani:
Did you use the latest helm chart? I thought I fixed this...

soamvasani:
This works now, and it's in the end-to-end tests.

PR #288, opened 2017-08-17
Fix fetcher failed to access TPR if RBAC is enabled

life1347:
This PR creates a service account for fetcher to access TPR(https://github.com/fission/fission/issues/286).

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -132,6 +132,27 @@ roleRef:
   name: cluster-admin
   apiGroup: rbac.authorization.k8s.io
 
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fission-fetcher
+  namespace: {{ .Values.functionNamespace }}
+
+---
+kind: ClusterRoleBinding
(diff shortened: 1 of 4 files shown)


charts/fission-all/templates/deployment.yaml
```diff
@@ -132,6 +132,27 @@ roleRef:
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: fission-fetcher-tpr
+subjects:
+- kind: ServiceAccount
+  name: fission-fetcher
+  namespace: {{ .Values.functionNamespace }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
life1347: I've tried to use "view" instead of cluster-admin here, however, fetcher still failed to access TPR. @soamvasani any ideas?

soamvasani: You're saying it works with cluster-admin but not with view?

life1347: yes!

soamvasani: Ok. That's fine. We'll narrow this down to a more specific role when we switch to CRD.

PR #292, opened 2017-08-22
Minor documentation fix for the Go example

georgebuckerfield:
Really minor documentation fix that might trip people up otherwise. Changes the `--env` referenced in the create command to match the env that is created earlier in the example.


soamvasani:
Thanks @georgebuckerfield !

PR #295, opened 2017-08-24
[wip] Split out the Package type into a first class Kubernetes resource

soamvasani:
Motivation:
 * Functions will need to share packages
 * A package storage system may be more generally useful than just functions (for example, for storing static assets)

### DIFF
--- a/types.go
+++ b/types.go
@@ -58,13 +58,21 @@ type (
 	}
 
 	PackageRef struct {
-		PackageName      string
-		PackageNamespace string
+		Name      string
+		Namespace string
+	}
+	FunctionPackageRef struct {
+		PackageRef PackageRef
 
-		// EntryPoint optionally specifies an entry point in
-		// the package. Each environment defines a default
-		// entry point, but that can be overridden here.
-		EntryPoint string `json:"entrypoint"`
+		// FunctionName specifies a specific function within the package. This allows
+		// functions to share packages, by having different functions within the same
+		// package.
+		//
+		// Fission itself does not interpret this path. It is passed verbatim to
+		// build and runtime environments.
+		//
+		// This is optional: if unspecified, the environment has a default name.
+		FunctionName string `json:"functionName"`
 	}
 
 	// FunctionSpec describes the contents of the function.
@@ -76,11 +84,11 @@ type (
 
 		// Source is an source package for this function; it's used for the build step if
 		// the environment defines a build container.
-		Source PackageRef `json:"source"`
+		Source FunctionPackageRef `json:"source"`
 
 		// Deployment is a deployable package for this function. This is the package that's
 		// loaded into the environment's runtime container.
-		Deployment PackageRef `json:"deployment"`
+		Deployment FunctionPackageRef `json:"deployment"`
 	}
 
 	FunctionReferenceType string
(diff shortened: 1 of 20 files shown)


soamvasani:
cc @life1347, this will affect your work with the build manager

types.go
```diff
@@ -55,6 +55,11 @@ type (
 		// Checksum ensures the integrity of packages
 		// refereced by URL. Ignored for literals.
 		Checksum Checksum `json:"checksum"`
+	}
+
+	PackageRef struct {
+		PackageName      string
+		PackageNamespace string
```
erwinvaneyk: What about versioning of the package? Btw, can't you just reuse the `api.Metadata` for this identification of the package?

erwinvaneyk: What is the idea behind the entrypoint being included in the PackageRef? Seems more like something that belongs to the PackageSpec.

soamvasani: Versioning belongs in a higher-level resource, IMO: we can create a `VersionedFunction` or a `VersionedPackage` and have a controller loop that watches git and creates `Functions` or `Packages`.

The entrypoint is what allows different functions to share a package. So it belongs either in FunctionSpec or PackageRef. Perhaps we should call it something else? "Handler" maybe?

erwinvaneyk: In that case shouldn't the actual PackageRef contain solely a UID for identification, to ensure it only reference one version/instance of the VersionedPackage? Potentially you could still add name and namespace for explainability/human-readability.

Ah, so the entrypoint points identifies a single function in a package with (potentially) multiple functions. Seems to me like this belongs more in the FunctionSpec or a separate PackageFunctionRef or something similar, as I would expect something called a `PackageRef` to be just a reference to a package.

Aside, wouldn't it be useful to have a list of all containing functions/files in the PackageSpec? That would allow Fission to easily check if a reference to a package/function is valid.

erwinvaneyk: Does the entrypoint always point to a function? If so, just call it `fn`, `path`, `resource`, or `target`, as it seems to be just for the identification/referencing of a function/file. Handler or entrypoint sound like a command or something special going on.

soamvasani: Namespace+Name uniquely identifies objects in the Kubernetes universe, equivalent to UID.  

It's worth noting that everything you see here is mutable. It's up to our implementation to cache correctly. However, immutability can be implemented as a matter of _policy_: VersionedFunction/VersionedPackage could avoid ever mutating a package.

I agree about the entrypoint.  I'll move it to FunctionSpec.

We want the Package mechanism to be more general than functions. Validation is a good point -- but I think we could do that at build time (or runtime, when there's no builder).

erwinvaneyk: How does that Namespace+Name identification work with multiple versions of the same Package? Does the name get suffixed with a version?

soamvasani: There are no versions in packages. If a package is mutated, the reference now points at a changed package, and the function has changed.

soamvasani: That hasn't changed in this PR, btw: the current FunctionSpec on `master` branch also allows updating a function.

erwinvaneyk: Ah okay, I was under the impression that versioning was also supposed to be supported with packages. :+1:

soamvasani: The idea is to separate packages and functions into two primitives: a low level "dumb" primitive that knows nothing about versions and a higher level thing that operates as a "factory" for packages or functions. Kinda like how Deployments act as factories for Pods.

PR #296, opened 2017-08-26
Add env builder & srcpkg through cli

life1347:
This PR is a first PR for Environment v2. In this PR, a user can add the env builder & source pkg through the cli.

env builder
```
$ fission env create --name nodejs --image fission/node-env --builder fission/node-env --buildcmd "./build.sh"
```

src pkg

```
$ fission fn create --name hello1 --env nodejs --srcpkg test.zip --method GET
```

### DIFF
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -41,6 +41,13 @@ func envCreate(c *cli.Context) error {
 		fatal("Need an image, use --image.")
 	}
 
+	envBuilderImg := c.String("builder")
+
+	envBuildCmd := c.String("buildcmd")
+	if len(envBuilderImg) > 0 && len(envBuildCmd) == 0 {
+		fatal("Need a build command for env builder to run with")
+	}
+
 	env := &tpr.Environment{
 		Metadata: api.ObjectMeta{
 			Name:      envName,
@@ -50,6 +57,10 @@ func envCreate(c *cli.Context) error {
 			Version: 1,
 			Runtime: fission.Runtime{
 				Image: envImg,
+			},
+			Builder: fission.Builder{
+				Image:   envBuilderImg,
+				Command: envBuildCmd,
 			},
 		},
 	}
@@ -91,10 +102,12 @@ func envUpdate(c *cli.Context) error {
 	if len(envName) == 0 {
 		fatal("Need a name, use --name.")
 	}
-
 	envImg := c.String("image")
-	if len(envImg) == 0 {
-		fatal("Need an image, use --image.")
+	envBuilderImg := c.String("builder")
+	envBuildCmd := c.String("buildcmd")
+
+	if len(envImg) == 0 && len(envBuilderImg) == 0 && len(envBuildCmd) == 0 {
+		fatal("Need --image to specify env image, or use --builder to specify env builder, or use --buildcmd to specify new build command.")
 	}
 
 	env, err := client.EnvironmentGet(&api.ObjectMeta{
@@ -103,7 +116,15 @@ func envUpdate(c *cli.Context) error {
 	})
 	checkErr(err, "find environment")
 
-	env.Spec.Runtime.Image = envImg
+	if len(envImg) > 0 {
+		env.Spec.Runtime.Image = envImg
+	}
+	if len(envBuilderImg) > 0 {
+		env.Spec.Builder.Image = envBuilderImg
+	}
+	if len(envBuildCmd) > 0 {
+		env.Spec.Builder.Command = envBuildCmd
+	}
 
 	_, err = client.EnvironmentUpdate(env)
 	checkErr(err, "update environment")
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"text/tabwriter"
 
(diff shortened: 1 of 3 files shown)


fission/environment.go
```diff
@@ -41,6 +41,13 @@ func envCreate(c *cli.Context) error {
 		fatal("Need an image, use --image.")
 	}
 
+	envBuilderImg := c.String("builder")
+
+	envBuildCmd := c.String("buildcmd")
```
soamvasani: Let's use a default value so users can avoid typing one more option on the CLI.
How about simply `"build"`.

fission/environment.go
```diff
@@ -91,10 +102,12 @@ func envUpdate(c *cli.Context) error {
...
 		fatal("Need a name, use --name.")
 	}
-
 	envImg := c.String("image")
-	if len(envImg) == 0 {
-		fatal("Need an image, use --image.")
+	envBuilderImg := c.String("builder")
+	envBuildCmd := c.String("buildcmd")
+
+	if len(envImg) == 0 && len(envBuilderImg) == 0 && len(envBuildCmd) == 0 {
```
soamvasani: We always need a runtime, so `--image` should not be optional.

fission/function.go
```diff
@@ -41,6 +42,19 @@ func fileSize(filePath string) int64 {
 	return info.Size()
 }
 
+func getPackageFileName(srcPkgName string, deployPkgName string) (string, error) {
+	if len(srcPkgName) == 0 && len(deployPkgName) == 0 {
+		return "", errors.New("Need --code or --package to specify deployment package, or use --srcpkg to specify source package.")
+	} else if len(srcPkgName) > 0 && len(deployPkgName) > 0 {
+		return "", errors.New("Cannot specify deployment package (--code, --package) and source package (--srcpkg) at the same time.")
```
soamvasani: Hmm, I think we should allow this. E.g. users might just want to save the source along with a build, to allow debugging for example.

fission/function.go
```diff
@@ -206,21 +231,47 @@ func fnUpdate(c *cli.Context) error {
...
-		fatal("Need --env or --code or --package argument.")
+	if len(envName) == 0 && len(deployPkgName) == 0 && len(srcPkgName) == 0 {
+		fatal("Need --env or --code or --package or --srcpkg argument.")
+	}
+
+	// We should not allow a user to update deployment pkg if the source
+	// pkg is not empty, since the deployment pkg may be overwritten with
+	// a new one, which is built from the old source pkg, once env builder
+	// finished it's build process.
+	if isSrcPkg && len(deployPkgName) > 0 && len(srcPkgName) == 0 {
```
soamvasani: How about allowing this, but choosing not to do a build. In other words, do a build only if a function has a source but no deployment package.

fission/environment.go
life1347: so the environment builder need to provide a default `build` command to build the package?

soamvasani: Yeah.

fission/environment.go
soamvasani: Nevermind, I see this is about the update.

life1347:
@soamvasani looks like it takes multiple changes if we move builds to package level. I will try to implement the build manager based on your suggestion (do a build only if a function has a source but no deployment package) to put env v2 online first, and change it after build status is introduced into the package spec.

soamvasani:
@life1347 Ok, that plan sounds good to me. I'm going to put #297 on hold for now, will get to it after the env v2 work is merged.

fission/function.go
```diff
@@ -41,6 +43,55 @@ func fileSize(filePath string) int64 {
 	return info.Size()
 }
 
+func getFnPkgRef(client *client.Client, fnName string, fileName string) fission.FunctionPackageRef {
```
soamvasani: 1. Please add a comment saying what this function does.
2. I think it should be renamed to something like "createPackageFromFile" to more accurately reflect what it's doing

fission/function.go
```diff
@@ -41,6 +43,55 @@ func fileSize(filePath string) int64 {
...
+			Literal: pkgContents,
+		},
+	}
+	_, err := client.PackageCreate(pkg)
+	checkErr(err, "upload package")
+
+	return fission.FunctionPackageRef{
+		PackageRef: fission.PackageRef{
+			Name:      pkgName,
+			Namespace: api.NamespaceDefault,
```
soamvasani: minor comment, but use `pkg.Metadata.Namespace`, instead of duplicating the api.NamespaceDefault constant

fission/function.go
```diff
@@ -41,6 +43,55 @@ func fileSize(filePath string) int64 {
...
+		},
+	}
+}
+
+func updatePkgContent(client *client.Client, pkgName string, fileName string) error {
```
soamvasani: Please rename to `updatePackageContents`

fission/function.go
```diff
@@ -206,34 +240,25 @@ func fnUpdate(c *cli.Context) error {
...
-		if fileSize(fileName) > fission.PackageLiteralSizeLimit {
-			// TODO fallback to uploading + setting a Package URL
-			fmt.Printf("File size >256k not supported yet")
-			os.Exit(1)
+	// Now builder manager only starts a build if a function has a source package
+	// but no deployment package. This behavior will be changed after we move builds
+	// to package level (https://github.com/fission/fission/pull/297).
+	for _, pkgName := range []string{srcPkgName, deployPkgName} {
+		if len(pkgName) > 0 {
+			err := updatePkgContent(client, function.Spec.Source.PackageRef.Name, srcPkgName)
```
soamvasani: is this correct? if `pkgName` is `deployPkgName`, you'll still only update the source package?

PR #297, opened 2017-08-26
Move builds to package level

soamvasani:
Rename the old "packages" to "archives".

"Packages" are now a pair of source and deployment archives.

Functions now have a single FunctionPackageRef into a single Package,
instead of two separate references.

The Package type will also have a PackageStatus, which will contain
build information.

### DIFF
--- a/controller/api_test.go
+++ b/controller/api_test.go
@@ -81,8 +81,10 @@ func TestFunctionApi(t *testing.T) {
 			Namespace: api.NamespaceDefault,
 		},
 		Spec: fission.FunctionSpec{
-			EnvironmentName: "nodejs",
-			Deployment: fission.FunctionPackageRef{
+			Environment: fission.EnvironmentReference{
+				Name: "nodejs",
+			},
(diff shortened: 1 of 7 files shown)


soamvasani:
cc @life1347 @erwinvaneyk

soamvasani:
The rationale for this change: since functions can share packages, it makes more logical sense to do the build at the package level than at the function level. (We could have duplicate package refs in multiple functions and de-duplicate the build process, but that's unnecessary complexity.)

soamvasani:
Going to keep this PR on hold until the builder is merged.

erwinvaneyk:
Nice, makes sense to me! 👍

soamvasani:
Hey @life1347 -- i think this PR may actually help #308 -- what do you think?

Also, it will be better for users if we merge this change before 0.2.1 on 9/9 -- that way we won't have to change the data types in the next upgrade (where as this 0.2.1 release already is a type change, since it contains the TPR change)

PR #298, opened 2017-08-27
Add Environment v2 Builder

life1347:
This PR introduces environment v2 builder into fission. 

Environment builder is consist of `builder` binary and language-specific container.

A env builder starts a build when receiving a build request from build manager(not existed yet) with the following example payload.

```
{
	"srcPkgFilename": "test.zip",
	"command": "pip3 install -r ${SRC_PKG}/requirements.txt -t ${SRC_PKG} && cp -r ${SRC_PKG} ${DEPLOY_PKG}"
}
```

Once the build is finished, builder replies builder manager with the following example payload
```
{
    "archiveFilename": "archive.zip",
    "buildLogs": "Collecting pyyaml (from -r /tmp/fission/srcPkg/requirements.txt (line 1))\n  Downloading PyYAML-3.12.tar.gz (253kB)\nInstalling collected packages: pyyaml\n  Running setup.py install for pyyaml: started\n    Running setup.py install for pyyaml: finished with status 'done'\nSuccessfully installed pyyaml-3.12\n"
}
```

Then builder manager will create a deployment package and update the `FunctionSpec.Deployment` reference.

For those who are interesting in this PR, here is the build tutorial 
```
$ cd ${GOPATH}/src/github.com/fission/fission/environments/python3/builder
$ ./build.sh
```
For further demonstration, please visit demo video [here](https://youtu.be/v-LEwq-oIQc).


environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	w.Header().Add("Content-Type", "application/json")
+	w.Write(rBody)
+	w.WriteHeader(http.StatusOK)
+}
+
+func (builder *Builder) build(command string, workDir string) (string, error) {
+	// use `/bin/sh -c` to run multiple commands at the same time
```
soamvasani: comment seems unnecessary, you're only running one command?

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
+package builder
```
soamvasani: please add the copyright header

soamvasani: Also, please move this out of the `environment/` dir. A top-level `builder` dir would be fine.

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+)
+
+func MakeBuilder(sharedVolumePath string) *Builder {
+	return &Builder{
+		sharedVolumePath: sharedVolumePath,
+	}
+}
+
+func (builder *Builder) Handler(w http.ResponseWriter, r *http.Request) {
```
soamvasani: So it seems this builder is designed to do only one build at a time, but doesn't actually enforce that. If multiple concurrent builds are going to break, a sanity check in this handler would be helpful (you can use a global `sync/atomic` variable to track if a build is in progress).

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		log.Printf("Error reading request body")
+		http.Error(w, err.Error(), 500)
+		return
+	}
+	var req fission.PackageBuildRequest
+	err = json.Unmarshal(body, &req)
+	if err != nil {
+		log.Printf("Error reading request body: %v", err)
+		http.Error(w, err.Error(), 500)
```
soamvasani: this should be a 400, it's the client's bug if the json is invalid. 500 is for things that are the server's fault.

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+	log.Printf("builder received request: %v", req)
+
+	srcPkgArchivePath := filepath.Join(builder.sharedVolumePath, req.SrcPkgFilename)
```
soamvasani: can you add some log messages here?  "Uncompressing package", "starting build", etc.

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// archive is a function that zips directory into a zip file
+func (builder *Builder) archive(src string, dst string) error {
+	var files []string
+	target, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+	if target.IsDir() {
```
soamvasani: why do you need this? Doesn't `archiver.Zip.Make` work on a folder?

environments/builder/client/client.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+
+	"github.com/fission/fission"
+)
+
+func DoBuildRequest(builderUrl string, br *fission.PackageBuildRequest) error {
```
soamvasani: I know it's just one function, but can you structure this like the other clients? (We may need to add other methods in the future, as we add stuff like progress reporting etc.)

i.e. a BuilderClient struct with a URL inside it, and a `build(br *fission.PackageBuildRequest)` method.

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+
+	w.Header().Add("Content-Type", "application/json")
+	w.Write(rBody)
+	w.WriteHeader(http.StatusOK)
+}
+
+func (builder *Builder) build(command string, workDir string) (string, error) {
+	// use `/bin/sh -c` to run multiple commands at the same time
+	cmdStrs := []string{"-c", command}
+	cmd := exec.Command("/bin/sh", cmdStrs...)
```
soamvasani: So... is there a way to stream output to stdout while this is running?

It would be nice to be able to look at build output "live", by using `kubectl logs`, while the build is in progress.

environments/python3/builder/defaultBuildCmd
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+pip3 install -r ${SRC_PKG}/requirements.txt -t ${SRC_PKG} && cp -r ${SRC_PKG} ${DEPLOY_PKG}
```
soamvasani: I think the builder should also load the user function, so that if it has syntax errors, we'll catch them in the build output. (I'm fine with doing that in a separate PR.)

types.go
```diff
@@ -215,6 +215,27 @@ type (
 	}
 )
 
+type (
+	PackageBuildRequest struct {
+		SrcPkgFilename string `json:"srcPkgFilename"`
+		// Command for builder to run with.
+		// A build command consists of commands, parameters and environment variables.
+		// Now, only two environment variables is support:
```
soamvasani: "For now, two environment variables are supported:"

types.go
```diff
@@ -215,6 +215,27 @@ type (
...
+		// 1. SRC_PKG: path to source package directory
+		// 2. DEPLOY_PKG: path to deployment package directory
+		// The environment variables will be replaced with correspond package directory path.
```
soamvasani: I don't think you need to explain this, it's just normal shell command variable substitution. I suggest removing the examples too, it makes the reader think there's something more complex going on.

environments/builder/builder.go
soamvasani: (Feel free to address this in a later change. You can probably do it with some complicated usage of the `exec` package.)

erwinvaneyk: Check out the binary environment for ways to work with the stdin and stdout of the forked process.

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+	cmd.Dir = workDir
+	// set env variables for build command
+	cmd.Env = append(os.Environ(),
+		envSrcPkg+"="+srcPkgPath,
+		envDeployPkg+"="+deployPkgPath,
```
erwinvaneyk: Minor: personally I find `fmt.Sprintf` to make this more readable

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+
+func MakeBuilder(sharedVolumePath string) *Builder {
+	return &Builder{
+		sharedVolumePath: sharedVolumePath,
+	}
+}
+
+func (builder *Builder) Handler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.Error(w, "", 404)
```
erwinvaneyk: Officially this would be a 405 (Method not allowed)

environments/builder/builder.go
```diff
@@ -0,0 +1,154 @@
...
+	defer func() {
+		elapsed := time.Now().Sub(startTime)
+		log.Printf("elapsed time in build request = %v", elapsed)
+	}()
+
+	// parse request
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("Error reading request body")
+		http.Error(w, err.Error(), 500)
```
erwinvaneyk: 404

soamvasani: @erwinvaneyk Hmm? I don't think 404 applies here

environments/python3/builder/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
erwinvaneyk: Add `set -ex` here

environments/builder/builder.go
erwinvaneyk: Oh, indeed, nevermind 👍

environments/builder/builder.go
erwinvaneyk: @soamvasani Thinking about it, this would be quite some effort to implement, as currently the builder is implemented in a synchronous way (if i understand @life1347 explanation correctly). It would be a nice to have, but that would require the client-builder communication to be asynchronous, in which the client would be polling the server for updates on a build.

soamvasani: I'm not talking about a change to the builder API. Just saying that it would be useful if the output were logged to stdout while the command is running.

soamvasani: Yeah. I'm trying to avoid having too much in one PR. Let's talk about a progress reporting API in a different issue/PR.

environments/builder/builder.go
life1347: It's more reasonable to accept multiple builds, I will try to adopt that. And there is one question, does the builder pod (fetcher + builder) need to keep alive till the environment is removed?  
In my original design, builder pod (fetcher + builder) is ephemeral. Every time a build request comes to builder manager, it creates a pod to do the build, and remove it when it's in the idle state. Or we can treat the builder pod like a special environment that only for the build. So that builder manager can fire an HTTP request to trigger the build request.

@soamvasani what do you think about it?

environments/builder/builder.go
life1347: User may use `&&` to concatenate multiple commands into one build command. In this case, we need to use `/bin/sh -c` to run them at once. For example:
```
pip3 install -r ${SRC_PKG}/requirements.txt -t ${SRC_PKG} && cp -r ${SRC_PKG} ${DEPLOY_PKG}
```
 But I can remove the comment if it's unnecessary.

environments/builder/builder.go
soamvasani: > accept multiple builds

sounds great!

> In my original design, builder pod (fetcher + builder) is ephemeral. Every time a build request comes to builder manager, it creates a pod to do the build,

Hmm. This question is more about the build manager, right? Does it affect this PR?

environments/builder/builder.go
soamvasani: well this comment won't be seen by users... also it would be easier to use a script inside the container instead of making a complex build command like this.

PR #301, opened 2017-08-30
Unarchive zip file after fetcher downloads the package

life1347:
In env v2, the fetcher has couple things to do:
1. archive deployment package and upload it after the build is finished
2. unarchive source package for builder to build with
2. unarchive deployment package for function container to run with

In this PR, fetcher will check the file type and unarchive it if it's a zip file. And since a deployment package may be a directory instead of a piece of code, a user needs to specify the entry point of function so that env runtime knows how to load the function.

### DIFF
--- a/environments/fetcher/fetcher.go
+++ b/environments/fetcher/fetcher.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -13,6 +14,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mholt/archiver"
 	"k8s.io/client-go/1.5/kubernetes"
 	"k8s.io/client-go/1.5/pkg/api"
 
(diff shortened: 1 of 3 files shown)


environments/fetcher/fetcher.go
```diff
@@ -196,16 +200,50 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
 
 	}
 
-	// move tmp file to requested filename
-	err = os.Rename(tmpPath, filepath.Join(fetcher.sharedVolumePath, req.Filename))
+	targetFilename := filepath.Join(fetcher.sharedVolumePath, req.Filename)
+	// check file type here, if the file is a zip file unarchive it.
+	if archiver.Zip.Match(tmpPath) {
+		if req.FetchType == FETCH_SOURCE {
+			// builder accepts multiple builds at the same time, use `pkg.Name`-`pkg.ResourceVersion`
```
soamvasani: it should be the caller's responsibility to ensure that `req.Filename` is unique for concurrent requests.

environments/fetcher/fetcher.go
```diff
@@ -196,16 +200,50 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
...
+			// to separate source pacakages for different functions.
+			targetFilename = filepath.Join(fetcher.sharedVolumePath, srcPkgFilename)
```
soamvasani: targetFilename should always be req.Filename -- that way the caller controls the path of the download

environments/fetcher/fetcher.go
```diff
@@ -196,16 +200,50 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
...
+		}
+		// unarchive tmp file to requested filename
+		err = fetcher.unarchive(tmpPath, targetFilename)
```
soamvasani: Instead of unarchiving in one step, you should unarchive into a temp directory, and then rename that to the target.  The reason is that "rename" is a single atomic operation, so with this two step unarchive+rename process, we will know that if the target directory exists then it means the unarchive completed. There will never be a target directory with an incomplete set of files.

environments/fetcher/fetcher.go
```diff
@@ -154,6 +157,7 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
 		var pkg *tpr.Package
 		if req.FetchType == FETCH_SOURCE {
 			pkg, err = fetcher.fissionClient.Packages(fn.Spec.Source.PackageRef.Namespace).Get(fn.Spec.Source.PackageRef.Name)
+			srcPkgFilename = fmt.Sprintf("%v-%v", pkg.GetObjectMeta().GetName(), pkg.GetObjectMeta().GetResourceVersion())
```
soamvasani: see comment below, i think we can avoid this

environments/fetcher/fetcher.go
life1347: that's a good point, I'll change it. thanks!

PR #308, opened 2017-09-03
Add builder manager support

life1347:
This PR aims to add builder manager support. 
Though this is a workable PR, it's still under development and not ready for code review.


life1347:
@soamvasani @erwinvaneyk 

Here is a question want to discuss

A deployment package may be a directory, in this situation the codepath of user function is no longer at `/userfunc/user`.  The env runtime is not able to load user function correctly due to the wrong codepath.

Following is the example directory structure after fetcher unarchives the deployment package.

```
/userfunc/
├── user
│   ├── PyYAML-3.12-py3.5.egg-info
│   │   ├── PKG-INFO
│   │   ├── SOURCES.txt
│   │   ├── dependency_links.txt
│   │   ├── installed-files.txt
│   │   └── top_level.txt
│   ├── requirements.txt
│   ├── user // real user codepath
│   └── yaml
│       ├── __init__.py
│       ├── <other module files>
└── user.tmp

```

In this case, the real codepath is `/userfunc/user/user`. Python runtime needs to load modules from `/userfunc/user` and load user function at `/userfunc/user/user`.

We need to find a way to solve this problem.

I think we can add a flag called`codepath` when creating the function. Then, poolmgr passes the value to env runtime when it tries to specialize the function pod

```
resp2, err := http.Post(specializeUrl, "application/json",
	bytes.NewReader([]byte(`{"codepath":"/userfunc/user/user"}`)))
```

Then, server.py uses the codepath in json body to load modules and user function.

```
@app.route('/specialize', methods=['POST'])
def load():
    global userfunc
    body = request.get_json()
    loadpath = body['codepath']
    if loadpath != '':
        sys.path.append(os.path.dirname(loadpath))
        codepath = loadpath
    userfunc = (imp.load_source('user', codepath)).main
    return ""
```
What do you think about it?

soamvasani:
@life1347 so are we talking about adding field for a file path, or for a function name? because there is already a field for the function name in `FunctionPackageRef`.

life1347:
@soamvasani oh! you're right. I missed that.

soamvasani:
Hey @life1347, this looks awesome and I'm still reviewing it. Can you add a test script under `/test/tests/test_builder.sh`? Those tests run automatically in master branch CI builds, and they are run with FISSION_URL and FISSION_ROUTER defined as env vars. You can also test your script manually.

life1347:
@soamvasani sure things, will add test script for this.

soamvasani:
(build failed because buildmanager/api.go is not gofmt'd)

life1347:
@soamvasani this PR is tested at branch `add-builder-mgr-ci-test` and ready for code review.

environments/fetcher/fetcher.go
```diff
@@ -229,6 +240,76 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
 
+func (fetcher *Fetcher) UploadHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.Error(w, "", 405)
```
soamvasani: please add a little error message there, like "only POST is supported on this endpoint".  Useful for debugging when things go wrong.

environments/fetcher/fetcher.go
```diff
@@ -229,6 +240,76 @@ func (fetcher *Fetcher) Handler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		e := fmt.Sprintf("Error archiving zip file: %v", err)
+		log.Println(e)
+		http.Error(w, e, 500)
+		return
+	}
+
+	ssClient := storageSvcClient.MakeClient(req.StorageSvcUrl)
+
+	fileID, err := ssClient.Upload(dstFilepath, nil)
```
soamvasani: add a log before starting the upload?

environments/python3/builder/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM alpine:3.5
+FROM alpine:3.4
```
soamvasani: why did this version go backwards, is this intentional?

environments/python3/server.py
```diff
@@ -18,6 +17,19 @@ def load():
     userfunc = (imp.load_source('user', codepath)).main
```
soamvasani: So what is `codepath` if the v1 endpoint is hit?  Won't this error out? Let's keep the v1 endpoints fully working for now-- we'll remove them in a few releases once v2 is stable.

examples/python/sourcepkg/build.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+pip3 install -r ${SRC_PKG}/requirements.txt -t ${SRC_PKG} && cp -r ${SRC_PKG} ${DEPLOY_PKG}
```
soamvasani: It's a *great* idea to let users provide their own build scripts! 

However, we should provide good defaults too. 

In other words, the python builder should contain a default build.sh that's invoked when user does not specify it. That default build script should basically be this file -- pip install + copy to destination.

examples/python/sourcepkg/user.py
```diff
@@ -0,0 +1,13 @@
+import sys
+sys.path.append('.')
```
soamvasani: Hmm... is there any way we can avoid making the user do this? Can the runtime env server.py do this?

poolmgr/gp.go
```diff
@@ -61,10 +62,13 @@ type (
 		poolInstanceId         string                // small random string to uniquify pod names
 		fetcherImage           string
 		fetcherImagePullPolicy v1.PullPolicy
+		runtimeImagePullPolicy v1.PullPolicy
```
soamvasani: could you please add a comment for this field

poolmgr/gp.go
```diff
@@ -61,10 +62,13 @@ type (
...
 		kubernetesClient       *kubernetes.Clientset
+		fissionClient          *tpr.FissionClient
 		instanceId             string // poolmgr instance id
 		labelsForPool          map[string]string
 		requestChannel         chan *choosePodRequest
+		sharedMountPath        string
```
soamvasani: Good idea to move this into the `GenericPool`. But please add a comment for this field.

(Also: this struct has gotten too big. We should break it up into some logical sub-structures. Later.)

poolmgr/gp.go
```diff
@@ -297,9 +316,23 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
 	// tell fetcher to get the function.
 	fetcherUrl := gp.getFetcherUrl(podIP)
 	log.Printf("[%v] calling fetcher to copy function", metadata.Name)
-	err := fetcherClient.DoFetchRequest(fetcherUrl, &fetcher.FetchRequest{
+
```
soamvasani: So, in general, we have to be very very careful adding anything to this function, because it slows down the cold-start. E.g. I think we should cache the package metadata for functions, that way we don't have to hit the k8s api in this function. However, that caching is a bit of a bigger change, and I'm ok with doing it later.

poolmgr/gp.go
```diff
@@ -297,9 +316,23 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
...
+	fn, err := gp.fissionClient.
+		Functions(metadata.Namespace).Get(metadata.Name)
+	if err != nil {
+		return err
+	}
+
+	pkg, err := gp.fissionClient.
```
soamvasani: Why do we need this? You just have to pass the package metadata, you don't need the full package object. In fact you just need the Name and Namespace in the metadata.  (You could even change the fetchRequest to take a PackageRef instead of api.ObjectMeta, if you want to.)

poolmgr/gp.go
```diff
@@ -311,12 +344,35 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
 
 	// get function run container to specialize
 	log.Printf("[%v] specializing pod", metadata.Name)
-	specializeUrl := gp.getSpecializeUrl(podIP)
 
 	// retry the specialize call a few times in case the env server hasn't come up yet
 	maxRetries := 20
+
+	fn, err = gp.fissionClient.Functions(metadata.Namespace).Get(metadata.Name)
```
soamvasani: This same function was retrieved from the API up on line 320; is it intentionally being fetched again?

poolmgr/gp.go
```diff
@@ -311,12 +344,35 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
...
+	if err != nil {
+		return err
+	}
+
+	loadReq := fission.FunctionLoadRequest{
+		FilePath:     filepath.Join(gp.sharedMountPath, "user"),
```
soamvasani: can you keep the string `"user"` in variable somewhere instead of hard-coding it twice?

poolmgr/gp.go
```diff
@@ -311,12 +344,35 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
...
+
+	body, err := json.Marshal(loadReq)
+	if err != nil {
+		return err
+	}
+
 	for i := 0; i < maxRetries; i++ {
-		resp2, err := http.Post(specializeUrl, "text/plain", bytes.NewReader([]byte{}))
+		var resp2 *http.Response
+		// if functionName is empty, assume its environment v1
```
soamvasani: Why? I'm not following the logic here...

soamvasani: The logic for v2 environments should be simpler: only use the v2 interface if the EnvironmentSpec has Version=2.

examples/python/sourcepkg/build.sh
life1347: A default build command will be packed into python env. 
https://github.com/life1347/fission/blob/cf30d9514ce850ee1b54fdaa6eb4bd029655e2a3/environments/python3/builder/Dockerfile#L8
The default build command will be specified when a package is created without specified a build command.
https://github.com/life1347/fission/blob/cf30d9514ce850ee1b54fdaa6eb4bd029655e2a3/fission/function.go#L149

buildermgr/envwatcher.go
```diff
@@ -0,0 +1,474 @@
...
+
+func (envw *environmentWatcher) getCacheKey(envName string, envResourceVersion string) string {
+	return fmt.Sprintf("%v-%v", envName, envResourceVersion)
+}
+
+func (envw *environmentWatcher) getLabels(envName string, envResourceVersion string) map[string]string {
+	sel := make(map[string]string)
+	sel[LABEL_ENV_NAME] = envName
+	sel[LABEL_ENV_RESOURCEVERSION] = envResourceVersion
+	return sel
```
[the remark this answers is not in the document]
life1347: According to this post (https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it), a map is a pointer to `runtime.hmap`. We can simply return map directly.

environments/fetcher/fetcher.go
```diff
@@ -239,7 +334,21 @@ func (fetcher *Fetcher) rename(src string, dst string) error {
...
 // archive is a function that zips directory into a zip file
 func (fetcher *Fetcher) archive(src string, dst string) error {
-	return archiver.Zip.Make(dst, []string{src})
+	var files []string
+	target, err := os.Stat(src)
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to zip file: %v", err))
+	}
+	if target.IsDir() {
+		// list all
```
soamvasani: You don't need this. The archiver recursively adds subdirs.

fission/environment.go
```diff
@@ -121,6 +130,11 @@ func envUpdate(c *cli.Context) error {
 	if len(envImg) > 0 {
 		env.Spec.Runtime.Image = envImg
 	}
+
+	if env.Spec.Version == 1 && (len(envBuilderImg) > 0 || len(envBuildCmd) > 0) {
+		fatal("Environment v1 API interface doesn't supported environment builder.")
```
soamvasani: How about: "Version 1 Environments do not support builders. Must specify --version=2."

fission/environment.go
```diff
@@ -42,12 +41,22 @@ func envCreate(c *cli.Context) error {
...
+		envVersion = 2
+		if len(envBuildCmd) == 0 {
+			envBuildCmd = "build"
+		}
+	}
+
+	// Environment API interface version is not specified and
+	// builder image is empty, set default interface version
+	if envVersion == 0 {
+		fmt.Println("Use default environment v1 API interface")
```
soamvasani: Suggest removing this, most users won't really understand it.

fission/function.go
```diff
@@ -66,6 +65,47 @@ func createArchive(client *client.Client, fileName string) *fission.Archive {
...
+			Namespace: api.NamespaceDefault,
+			Name:      envName,
+		},
+	}
+	var pkgStatus fission.BuildStatus = fission.BuildStatusSucceeded
+
+	if len(deployArchiveName) > 0 {
+		pkgSpec.Deployment = *createArchive(client, deployArchiveName)
+		if len(srcArchiveName) > 0 {
+			fmt.Println("Deployment may be overwritten by builder manager after source package compilation")
```
soamvasani: We should never overwrite a user's deployment package. We should only do a build if the deployment is absent.

fission/function.go
```diff
@@ -66,6 +65,47 @@ func createArchive(client *client.Client, fileName string) *fission.Archive {
...
+		pkgSpec.Source = *createArchive(client, srcArchiveName)
+		// set pending status to package
+		pkgStatus = fission.BuildStatusPending
+	}
+
+	if len(buildcmd) > 0 {
+		pkgSpec.BuildCommand = buildcmd
+	}
+
+	pkgName := strings.ToLower(uuid.NewV4().String())
```
soamvasani: UUIDs are hard to work with on the CLI. Why not accept a `name` parameter? And for implicitly created packages (from functions), use `<functionName>-<uniuri.NewLen(6)>`

fission/function.go
```diff
@@ -88,17 +132,24 @@ func fnCreate(c *cli.Context) error {
...
+	if len(srcArchiveName) == 0 && len(deployArchiveName) == 0 {
+		fatal("Need --code or --deploy to specify deployment archive, or use --src to specify source archive.")
 	}
 
-	if len(srcPkgName) == 0 && len(deployPkgName) == 0 {
-		fatal("Need --code or --package to specify deployment package, or use --srcpkg to specify source package.")
+	entrypoint := c.String("entrypoint")
+	buildcmd := c.String("buildcmd")
+	if len(buildcmd) == 0 {
+		buildcmd = "/builder"
```
soamvasani: Just leave it empty and let the builder handle default build commands. Also, from builder.go it looks like the default is "/build" not "/builder"?

poolmgr/gp.go
soamvasani: ^ Please address this comment

poolmgr/gp.go
```diff
@@ -297,10 +316,23 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
...
-	err := fetcherClient.DoFetchRequest(fetcherUrl, &fetcher.FetchRequest{
+
+	fn, err := gp.fissionClient.
+		Functions(metadata.Namespace).
+		Get(metadata.Name)
+	if err != nil {
+		return err
+	}
+
+	targetFilename := "user"
```
soamvasani: I think we could use `fn.Metadata.UID` here?

test/build_and_test.sh
```diff
@@ -12,12 +12,18 @@ source $(dirname $0)/test_utils.sh
 
 IMAGE=gcr.io/fission-ci/fission-bundle
 FETCHER_IMAGE=gcr.io/fission-ci/fetcher
+PYTHON_RUNTIME_IMAGE=gcr.io/fission-ci/python-env
```
soamvasani: We should generalize this a bit, accepting 'python' as a parameter instead of hard-coding it. That way we can extend these tests to other languages more easily.

test/test_utils.sh
```diff
@@ -50,6 +50,36 @@ build_and_push_fetcher() {
...
+    gcloud_login
+    
+    gcloud docker -- push $image_tag
+    popd
+}
+
+build_and_push_python_env_builder() {
+    image_tag=$1
+
+    pushd $ROOT/builder/cmd
```
soamvasani: Can you separate this into a different function? Builder is not python specific and shouldn't be built inside python build scripts

test/tests/test_buildermgr.sh
```diff
@@ -0,0 +1,73 @@
...
+fission fn create --name $fn --env python --src demo-src-pkg.zip --entrypoint "main" --buildcmd "./build.sh"
+trap "fission fn delete --name $fn" EXIT
+
+echo "Creating route"
+fission route create --function $fn --url /$fn --method GET
+
+echo "Waiting for router to catch up"
+sleep 3
+
+echo "Doing an HTTP POST on the builder manager's route to start a build"
```
soamvasani: You can remove this now, right?  The pkgwatcher will build the package anyway.

test/tests/test_buildermgr.sh
```diff
@@ -0,0 +1,73 @@
...
+pkg=$(kubectl --namespace default get functions $fn -o jsonpath='{.spec.package.packageref.name}')
+echo $pkg
+response=$(curl -X POST $FISSION_URL/proxy/buildermgr/v1/build \
+  -H 'content-type: application/json' \
+  -d "{\"package\": {\"namespace\": \"default\",\"name\": \"$pkg\"}}")
+
+echo "Waiting for builder manager to finish the build triggered by http request"
```
soamvasani: You should poll the buildstatus instead.

soamvasani:
@life1347 

The feedback above is important but this change has gotten too big to incrementally review. Please make these changes in your next PR, I'll merge this now.

environments/fetcher/client/client.go
```diff
@@ -33,3 +47,34 @@ func DoFetchRequest(fetcherUrl string, fr *fetcher.FetchRequest) error {
...
+	}
+
+	rBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Printf("Received upload response: %v", string(rBody))
+
+	uploadReq := fetcher.UploadResponse{}
```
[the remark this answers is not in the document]
life1347: oops

fission/function.go
soamvasani: I see what you're saying. How about we do this:
 1. Packages created implicitly during function create/update take the function name + unique id
 1. Users are allowed to rename packages with the upcoming package cli
 1. Users are allowed to name packages during creating if they use the package cli 
 1. Package metadata is annotated with a list of functions that use it
 1. Package metadata is labelled with a "createdForFunction=<function name>" label when it's created implictly by a function create

PR #315, opened 2017-09-08
Storage service helm chart integration + bugfixes

soamvasani:
Storage service in the helm charts, plus a few bugfixes in the proxy.

### DIFF
--- a/charts/fission-all/templates/pvc.yaml
+++ b/charts/fission-all/templates/pvc.yaml
@@ -19,4 +19,10 @@ spec:
   resources:
     requests:
       storage: {{ .Values.persistence.size | quote }}
+{{- if .Values.persistence.storageClass }}
+  storageClassName: {{ .Values.persistence.storageClass | quote }}
+{{- else }}
+  storageClassName: default
+{{- end }}
+
 {{- end }}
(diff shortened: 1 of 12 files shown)


charts/fission-all/templates/pvc.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: v1
+metadata:
+  name: fission-storage-pvc
+  labels:
+    app: fission-storage
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+  annotations:
+  {{- if .Values.persistence.storageClass }}
+    volume.beta.kubernetes.io/storage-class: {{ .Values.persistence.storageClass | quote }}
```
life1347: Should we use `storageClassName ` instead of annotation `volume.beta.kubernetes.io/storage-class`? 
According to the [doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1), the annotation will be deprecated in the future release.

soamvasani: Thanks, didn't know that!

soamvasani: Ah after reading this in more detail, it looks like leaving storageClassName unspecified is equivalent to choosing the default. As far as I can tell this is the only way to get a portable chart.

Issue #287, opened 2017-08-16
Update function not working with latest master branch


life1347:
With fission-bundle nightly20170705, function response is changed once user update the function source code.
```
$ ./fission fn create --name hello1-1 --env nodejs --code ~/Desktop/main.js --method GET
function 'hello1-1' created
$ ./fission route create --method GET --url /h1 --function hello1-1
trigger '9a48a8e6-2aa7-4c09-b19f-b38d46df7b0c' created
$ curl http://$FISSION_ROUTER/h1
hello world!
$ ./fission fn update --name hello1-1 --code ~/Desktop/main.js
function 'hello1-1' updated
$ curl http://$FISSION_ROUTER/h1
hello world 2!
$ curl http://$FISSION_ROUTER/h1
hello world 2!
```
But with the latest master branch, function response is not changed immediately due to functionService cache. 

```
$ curl http://$FISSION_ROUTER/h1
hello world!
$ ./fission fn update --name hello1-1 --code ~/Desktop/main.js
function 'hello1-1' updated
$ curl http://$FISSION_ROUTER/h1
hello world!
$ kubectl  delete pod router-2145821694-511rl
pod "router-2145821694-511rl" deleted
$ curl http://$FISSION_ROUTER/h1
hello world 2!
```

PR #316, opened 2017-09-08
Use latest function metadata to check cached function service.

life1347:
This PR aims to solve https://github.com/fission/fission/issues/287.

When an HTTP request comes to the router, it will send a tap request to pool manager with function metadata. The metadata within payload may be out-of-date, so pool manager always returns the IP address of old function pod to the router. In this situation, a user received wrong function response until the pod is killed by the pool manager.

With this PR, when a `getServiceForFunction` request comes to pool manager, pool manager will fetch the latest function metadata from TPR and use this metadata to find correspond service cache. If cache missed, then a new function pod will be created to serve the user request.

### DIFF
--- a/poolmgr/api.go
+++ b/poolmgr/api.go
@@ -179,21 +179,6 @@ func (poolMgr *Poolmgr) getFunctionEnv(m *api.ObjectMeta) (*tpr.Environment, err
 }
 
 func (poolMgr *Poolmgr) getServiceForFunction(m *api.ObjectMeta) (string, error) {
-	// Pool manager uses function metadata to generate a key
-	// for cache index. However the function metadata passed
-	// from router may be out-of-date (e.g. resourceVersion).
-	// We should always use the latest function metadata grab
-	// from TPR for the following operation.
-	fn, err := poolMgr.fissionClient.Functions(m.Namespace).Get(m.Name)
-	if err != nil {
-		e := fmt.Sprintf("Error getting TPR resource info for function %v: %v", m.Name, err)
-		log.Println(e)
-		return "", err
-	}
-
-	// update function metadata
-	m = &fn.Metadata
-
 	// Check function -> svc cache
 	log.Printf("[%v] Checking for cached function service", m.Name)
 	fsvc, err := poolMgr.fsCache.GetByFunction(m)
--- a/poolmgr/api.go
+++ b/poolmgr/api.go
@@ -143,6 +143,7 @@ func (poolMgr *Poolmgr) getServiceForFunctionApi(w http.ResponseWriter, r *http.
 		code, msg := fission.GetHTTPError(err)
 		log.Printf("Error: %v: %v", code, msg)
 		http.Error(w, msg, code)
+		return
 	}
 
 	w.Write([]byte(serviceName))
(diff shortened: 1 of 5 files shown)


poolmgr/api.go
```diff
@@ -178,6 +179,21 @@ func (poolMgr *Poolmgr) getFunctionEnv(m *api.ObjectMeta) (*tpr.Environment, err
 }
 
 func (poolMgr *Poolmgr) getServiceForFunction(m *api.ObjectMeta) (string, error) {
+	// Pool manager uses function metadata to generate a key
```
soamvasani: This changes the meaning of the poolmgr API. This bug needs to be solved in the router, I think: it should not be resolving trigger function references to outdated functions.

soamvasani:
Hmm, I think the bug is in the functionReferenceResolver's cache. That one is kept for one minute, and we don't invalidate it correctly.

soamvasani:
This is quite tricky to fix while still keeping the cache useful...how about something like this:

  1. Create a `functionReferenceResolver.Invalidate(newFuncs []tpr.Function)` method; 
  2. In this method, make a temporary `map[namespacedFunctionReference]tpr.Function` from `newFuncs`
  3. Get a copy of the cache, and delete each key that points to a function having a different resourceVersion than the one in the `newFuncs` map.

life1347:
After checking the source code, if I understand correctly the `ts.syncTriggers` method is called when HTTP triggers changed. How about adding a `sync` method in `functionServiceMap` to update cache periodically?

soamvasani:
It's not functionServiceMap that's the problem -- it's the resolver cache.  Function service map already has resourceVersion in the key.

soamvasani:
An easier fix is to just set resolver cache expiry to 100msec instead of 1 minute, then you don't need a complicated invalidation.
https://github.com/fission/fission/blob/master/router/functionReferenceResolver.go#L66

life1347:
After having a discussion with @soamvasani,
the router will watch functions changes and update `functionReferenceResolver.refCache` & `functionServiceMap.cache` if needed to solve this problem.

router/httpTriggers.go
```diff
@@ -162,6 +163,61 @@ func (ts *HTTPTriggerSet) watchTriggers() {
...
+					rr.functionMetadata.ResourceVersion != fn.Metadata.ResourceVersion {
+					err := ts.resolver.delete(key.namespace, &key.functionReference)
+					if err != nil {
+						log.Printf("Error deleting functionReferenceResolver cache: %v", err)
+					}
+					break
+				}
+			}
+
+			ts.syncTriggers()
```
life1347: Need call ts.syncTriggers() to update functionHandler or the functionHandler will pass out-of-date function metadata to poolmgr. (https://github.com/fission/fission/blob/master/router/functionHandler.go#L42)

router/functionServiceMap.go
```diff
@@ -75,3 +75,17 @@ func (fmap *functionServiceMap) assign(f *api.ObjectMeta, serviceUrl *url.URL) {
 		// ignore error
 	}
 }
+
```
soamvasani: Why do we need to invalidate the function service map? If the function's changed, it's resource version has changed. And the function service map key contains the resource version, which means a lookup of an existing function with a new resource version will actually be a cache miss.

router/httpTriggers.go
soamvasani: Ah, yup, makes sense.

test/tests/test_function_update.sh
```diff
@@ -0,0 +1,62 @@
...
+echo "Waiting for router to catch up"
+sleep 10
+
+echo "Doing an HTTP GET on the function's route"
+response=$(curl http://$FISSION_ROUTER/$fn)
+
+echo "Checking for valid response"
+echo $response | grep -i foo
+
+# Running a background process to keep access the
```
soamvasani: nice 👍

PR #327, opened 2017-09-14
Removed deprecated k8s templates

erwinvaneyk:
Since the Helm charts are now the authoritative way of deploying fission, removing the old, out-of-sync templates would avoid confusion (been there) and potential bugs due to users still using the old deployment approach. 

I quickly checked wether there are templates that do provide some functionality that is not provided by the helm chart. Only the `fission-cloud.yaml` might have a reason to still exist, but I am not entirely sure if it is still relevant (cc @soamvasani)

Ready for review :)


soamvasani:
Thanks @erwinvaneyk !

PR #329, opened 2017-09-17
Upgrade Node Environment to 8.x

MylesBorins:
7.x has not been maintained since April. 8.x is going into LTS
during October and will receive another 30 months of support.

This update should simply work out of the box. Let me know if
I can help with anything


soamvasani:
Thanks, @MylesBorins !

Issue #331, opened 2017-09-18
Function-service cache bug in poolmgr


soamvasani:
cc @erwinvaneyk 

This is exposed (only) by Fission Workflows, due to multiple specialization.  

The workflow environment specializes all workflows into the same pod (that pod is the workflow engine.)

In this bug GetFuncSvc assumed that a cache.Add failure was always due to the item existing in the cache.  However, with multiple specialization, the byFunction cache has unique keys but the byPod and byAddress caches do not. So fsc.Add gets past the function check but not past the pod and address check. 

The fix is for the function service cache to fully support multiple specialization. That means fsc.Add() should not error out if the by-pod and by-address cache fail with ErrorNameExists. It should only fail when there is a collision by function identifier.

PR #332, opened 2017-09-18
Function service cache partial support for multiple specialization

soamvasani:
This adds a little bit of support for multiple specialization to the
function service cache. It allows Add() to succeed if the pod and
address maps already contain entries.

But it doesn't deal with DeletePod and TouchByAddress. That's ok for
now, because the workflow engine never deletes the environment pod
(that pod is the workflow engine).

However, if/when we want full support for multiple specialization,
that will require updating both DeletePod and TouchByAddress.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cache
 
 import (
-	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


soamvasani:
closes #331

Issue #328, opened 2017-09-15
Missing environment images of the latest tag


y-taka-23:
I found that the following environment images lack the `latest` tag:

* `fission/binary-env`
* `fission/php-env`
* `fission/perl-env`
* `fission/dotnet-env`

Especially speaking of `binary-env`,  the current [README.md](https://github.com/fission/fission/blob/85983c9/examples/binary/README.md) refers to the `latest` image as:

```
fission env create --name binary-env --image fission/binary-env
```

So I'm afraid that newbies who follow the document will stall here. Do you have any reasons to omit the `latest` tags?

soamvasani:
Thanks @y-taka-23, this is indeed a bug in our release automation script.  

The fix would be to make hack/release.sh `build_and_push_env_image()` also tag and push the `latest` tag.

y-taka-23:
Successfully deployed into Docker Hub.

PR #333, opened 2017-09-19
Tag and push the latest environment images

y-taka-23:
Currently some of the `latest` images are not deployed into DockerHub, as described in the issue #328


soamvasani:
Thanks @y-taka-23 !

PR #336, opened 2017-09-20
Fission workflow env integration

erwinvaneyk:
This PR prepares the way for the integration with fission workflows. Specifically it changes the following:
- Add `AllowedFunctionsPerContainer` field to environment spec. Currently there are two possible values `single` (current functionality) and `infinite`. With the `infinite` value, the environment marks itself as being able handle multiple specializations. This will reduce the required pool size to 1, removes automatic gc'ing of the environment, and does not relabel the environment after being specialized.
- In the router params are added to the requests to allow environments to get the function metadata. This is needed to allow the environments with multiple specializations to differentiate between HTTP requests to the different functions it handles.
- Workflow apiserver proxy in the controller. Similar to the storagesvc-proxy, it proxies requests to the optional workflow-apiserver.

cc @soamvasani @life1347

### DIFF
--- a/controller/api.go
+++ b/controller/api.go
@@ -38,6 +38,7 @@ type (
 		fissionClient     *tpr.FissionClient
 		storageServiceUrl string
 		builderManagerUrl string
+		WorkflowApiUrl    string
 	}
 
 	logDBConfig struct {
@@ -62,6 +63,13 @@ func MakeAPI() (*API, error) {
 		api.builderManagerUrl = strings.TrimSuffix(u, "/")
 	} else {
 		api.builderManagerUrl = "http://buildermgr"
+	}
+
+	wfEnv := os.Getenv("WORKFLOW_API_URL")
+	if len(u) > 0 {
+		api.WorkflowApiUrl = strings.TrimSuffix(wfEnv, "/")
+	} else {
+		api.WorkflowApiUrl= "http://workflow-apiserver"
 	}
 
 	return api, err
@@ -170,6 +178,7 @@ func (api *API) Serve(port int) {
 	r.HandleFunc("/proxy/storage/v1/archive", api.StorageServiceProxy)
 	r.HandleFunc("/proxy/buildermgr/v1/build", api.BuilderManagerBuildProxy)
 	r.HandleFunc("/proxy/buildermgr/v1/builder", api.BuilderManagerEnvBuilderProxy)
+	r.HandleFunc("/proxy/workflow", api.WorkflowApiProxy)
 
 	address := fmt.Sprintf(":%v", port)
 
--- a/controller/api.go
+++ b/controller/api.go
@@ -39,6 +39,7 @@ type (
 		storageServiceUrl string
 		builderManagerUrl string
 		WorkflowApiUrl    string
+		workflowApiUrl    string
 	}
 
 	logDBConfig struct {
@@ -67,9 +68,9 @@ func MakeAPI() (*API, error) {
 
 	wfEnv := os.Getenv("WORKFLOW_API_URL")
 	if len(u) > 0 {
-		api.WorkflowApiUrl = strings.TrimSuffix(wfEnv, "/")
+		api.workflowApiUrl = strings.TrimSuffix(wfEnv, "/")
 	} else {
-		api.WorkflowApiUrl= "http://workflow-apiserver"
+		api.workflowApiUrl = "http://workflow-apiserver"
 	}
 
 	return api, err
--- a/controller/api.go
+++ b/controller/api.go
@@ -38,7 +38,6 @@ type (
 		fissionClient     *tpr.FissionClient
 		storageServiceUrl string
 		builderManagerUrl string
-		WorkflowApiUrl    string
 		workflowApiUrl    string
 	}
 
--- a/poolmgr/gp.go
+++ b/poolmgr/gp.go
@@ -313,7 +313,7 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
 	// for fetcher we don't need to create a service, just talk to the pod directly
 	podIP := pod.Status.PodIP
 	if len(podIP) == 0 {
-		return errors.New("pod has no IP")
+		return errors.New("Pod has no IP")
 	}
 
 	// tell fetcher to get the function.
--- a/poolmgr/gp.go
+++ b/poolmgr/gp.go
@@ -157,7 +157,10 @@ func MakeGenericPool(
 
 	go gp.choosePodService()
 
-	go gp.idlePodReaper()
+	// Unless specified otherwise, periodically cleanup inactive pods.
+	if env.Spec.AllowedFunctionsPerContainer != fission.AllowedFunctionsPerContainerInfinite {
+		go gp.idlePodReaper()
+	}
 
 	return gp, nil
 }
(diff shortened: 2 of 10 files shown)


poolmgr/gp.go
```diff
@@ -316,7 +319,13 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
...
 	maxRetries := 20
 	for i := 0; i < maxRetries; i++ {
-		resp2, err := http.Post(specializeUrl, "text/plain", bytes.NewReader([]byte{}))
+
+		md, err := json.Marshal(metadata)
+		if err != nil {
+			return err
+		}
+
+		resp2, err := http.Post(specializeUrl, "application/json", bytes.NewReader(md))
```
soamvasani: Why does the specialize endpoint need  function metadata? Does the workflow engine itself talk to the fission API to get the function? 

Can you load the workflow from a file instead? If so, #308 has everything you'll need.

poolmgr/gp.go
```diff
@@ -478,7 +487,7 @@ func (gp *GenericPool) GetFuncSvc(m *api.ObjectMeta) (*funcSvc, error) {
 	if gp.useSvc {
 		svcName := fmt.Sprintf("svc-%v", m.Name)
 		if len(m.UID) > 0 {
-			svcName += ("-" + string(m.UID))
+			svcName = fmt.Sprintf("%s-%v", svcName, m.UID)
```
soamvasani: curious about this change -- are you actually using it? I've been meaning to remove this code, it's dead code anyway.

poolmgr/gpm.go
```diff
@@ -86,9 +87,16 @@ func (gpm *GenericPoolManager) service() {
...
+				switch req.env.Spec.AllowedFunctionsPerContainer {
+				case fission.AllowedFunctionsPerContainerInfinite:
+					poolSize = 1
+				}
+				log.Println("poolsize: ", poolSize)
+				log.Println("setting: ", req.env.Spec.AllowedFunctionsPerContainer)
+
 				pool, err = MakeGenericPool(
-					gpm.kubernetesClient, req.env,
-					3, // TODO configurable/autoscalable
```
soamvasani: you dropped the TODO comment in the new code, mind adding it back? :)

poolmgr/gpm.go
```diff
@@ -86,9 +87,16 @@ func (gpm *GenericPoolManager) service() {
 			var err error
 			pool, ok := gpm.pools[tpr.CacheKey(&req.env.Metadata)]
 			if !ok {
+				var poolSize int32 = 3
+				switch req.env.Spec.AllowedFunctionsPerContainer {
+				case fission.AllowedFunctionsPerContainerInfinite:
+					poolSize = 1
+				}
+				log.Println("poolsize: ", poolSize)
+				log.Println("setting: ", req.env.Spec.AllowedFunctionsPerContainer)
```
soamvasani: Are these logs intended for commit? If so, they should also have the name of the env, should be combined into one log message, and should have "functions per container" instead of "setting".

router/util.go
```diff
@@ -0,0 +1,31 @@
...
+	"k8s.io/client-go/1.5/pkg/api"
+	"fmt"
+	"k8s.io/client-go/1.5/pkg/types"
+)
+
+const (
+	HEADERS_FISSION_FUNCTION_PREFIX = "Fission-Function"
+)
+
+func MetadataToHeaders(prefix string, meta *api.ObjectMeta, request *http.Request) error {
```
soamvasani: lowercase, not exported from router package

router/util.go
```diff
@@ -0,0 +1,31 @@
...
+	request.Header.Add(fmt.Sprintf("X-%s-Uid", prefix), string(meta.UID))
+	request.Header.Add(fmt.Sprintf("X-%s-Name", prefix), meta.Name)
+	request.Header.Add(fmt.Sprintf("X-%s-Namespace", prefix), meta.Namespace)
+	request.Header.Add(fmt.Sprintf("X-%s-ResourceVersion", prefix), meta.ResourceVersion)
+	return nil
+}
+
+func HeadersToMetadata(prefix string, headers http.Header) (*api.ObjectMeta, error) {
```
soamvasani: lowercase, not exported from router package

erwinvaneyk:
As discussed offline with @soamvasani, removed the addition of ObjectMetadata from the specialization request in favor of the request data structure in #308

soamvasani:
@erwinvaneyk sorry, you get to fix the merge conflicts this time 😁

erwinvaneyk:
Rebased, but did not test the changes in conjunction with the new environment spec yet.

router/util.go
```diff
@@ -0,0 +1,31 @@
...
+
+func metadataToHeaders(prefix string, meta *api.ObjectMeta, request *http.Request) error {
+	request.Header.Add(fmt.Sprintf("X-%s-Uid", prefix), string(meta.UID))
+	request.Header.Add(fmt.Sprintf("X-%s-Name", prefix), meta.Name)
+	request.Header.Add(fmt.Sprintf("X-%s-Namespace", prefix), meta.Namespace)
+	request.Header.Add(fmt.Sprintf("X-%s-ResourceVersion", prefix), meta.ResourceVersion)
+	return nil
+}
+
+func headersToMetadata(prefix string, headers http.Header) (*api.ObjectMeta, error) {
```
soamvasani: is this used anywhere?

erwinvaneyk: Yeah, in the workflow environment (which has Fission as a dependency). As it is the inverse of the `metadataToHeaders` it seems best placed here, even though fission itself doesn't use it itself yet

controller/api.go
```diff
@@ -38,6 +38,8 @@ type (
 		fissionClient     *tpr.FissionClient
 		storageServiceUrl string
 		builderManagerUrl string
+		WorkflowApiUrl    string
```
soamvasani: remove this?

poolmgr/gp.go
```diff
@@ -348,8 +351,9 @@ func (gp *GenericPool) specializePod(pod *v1.Pod, metadata *api.ObjectMeta) erro
 	maxRetries := 20
 
 	loadReq := fission.FunctionLoadRequest{
-		FilePath:     filepath.Join(gp.sharedMountPath, targetFilename),
-		FunctionName: fn.Spec.Package.FunctionName,
+		FilePath:         filepath.Join(gp.sharedMountPath, targetFilename),
+		FunctionName:     fn.Spec.Package.FunctionName,
+		FunctionMetadata: metadata,
```
soamvasani: I just noticed this right now, but `fn.Metadata` will be more complete (include up to date resource version etc).

router/util.go
```diff
@@ -0,0 +1,31 @@
...
+	"k8s.io/client-go/1.5/pkg/api"
+	"k8s.io/client-go/1.5/pkg/types"
+	"net/http"
+)
+
+const (
+	HEADERS_FISSION_FUNCTION_PREFIX = "Fission-Function"
+)
+
+func metadataToHeaders(prefix string, meta *api.ObjectMeta, request *http.Request) error {
```
soamvasani: this never errors out, remove the error return

router/util.go
```diff
@@ -0,0 +1,31 @@
...
+
+func headersToMetadata(prefix string, headers http.Header) (*api.ObjectMeta, error) {
+	meta := &api.ObjectMeta{
+		Name:            headers.Get(fmt.Sprintf("X-%s-Name", prefix)),
+		UID:             types.UID(headers.Get(fmt.Sprintf("X-%s-Uid", prefix))),
+		Namespace:       headers.Get(fmt.Sprintf("X-%s-Namespace", prefix)),
+		ResourceVersion: headers.Get(fmt.Sprintf("X-%s-ResourceVersion", prefix)),
+	}
+
+	return meta, nil
```
soamvasani: same comment as the other function, remove the error return

router/functionHandler.go
```diff
@@ -98,6 +98,14 @@ func (fh *functionHandler) handler(responseWriter http.ResponseWriter, request *
 		request.Header.Add(fmt.Sprintf("X-Fission-Params-%v", k), v)
 	}
 
+	// System Params
+	err := metadataToHeaders(HEADERS_FISSION_FUNCTION_PREFIX, fh.function, request)
+	if err != nil {
```
soamvasani: can remove this block once the function removes the error return

PR #337, opened 2017-09-20
Fission dotnet 2.0 env

joalmeid:
Adding a new dotnet environment, supporting dotnet core 2.0.
Project structure changed. Project.json was abandoned. Project..csproj is now used.

All tests, referenced in documentation were done. 

A new docker image should be created officially : fission/dotnet20-env

### DIFF
--- a/environments/dotnet20/README.md
+++ b/environments/dotnet20/README.md
@@ -1,0 +1,306 @@
+# Fission: dotnet 2.0 C# Environment
+
+This is a simple dotnet core 2.0 C# environment for Fission.
+
+It's a Docker image containing the dotnet 2.0.0 runtime. The image 
+uses Kestrel with Nancy to host the internal web server and uses 
+Roslyn to compile the uploaded code.
+
+The image supports compiling and running code with types defined in
+mscorlib and does not at present support other library references.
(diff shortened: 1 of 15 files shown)


soamvasani:
Hi @joalmeid, thanks for this contribution! Great documentation too! 
Going to merge this now.

You may be interested in the v2 Environments -- these environments separate out the compilation step so that the request path doesn't have the overheads of compilation. We're working on documenting it better and providing an example. (The design doc is [here](https://github.com/fission/fission/blob/master/Documentation/wip/environments-v2.md))

environments/dotnet20/README.md
```diff
@@ -0,0 +1,306 @@
...
+## Echo example
+
+### Setup fission environment
+First you need to setup the fission according to your cluster setup as 
+specified here: https://github.com/fission/fission
+
+
+### Create the class to run
+
+Secondly you need to create a file /tmp/func.cs containing the following code:
```
soamvasani: Actually one quick comment, would you mind putting these samples in separate files under `examples/dotnet20/`?

joalmeid: Done. Examples are similar to dotnet, however added an echo example.

Issue #167, opened 2017-03-30
"native-or-bluebird deprecated, use any-promise instead."


awitherow:
The package https://github.com/normalize/mz/blob/master/package.json has updated to quite a higher version number than which is currently implemented in the package.json for the fission-nodejs-runtime.

Is there anything holding us back from keeping these dependencies more up to date?

soamvasani:
You're right, the package.json needs to be updated. (Nothing holding us back, just something nobody's done yet.)

PR #339, opened 2017-09-21
Update a dependency in the package.json

watilde:
Since `mz` has already bumped for several times without
breaking change, it's better to follow the updates.

```console
$ npm outdated
Package  Current  Wanted  Latest  Location
mz         2.1.0   2.1.0   2.7.0  fission-nodejs-runtime
```

Fixes: https://github.com/fission/fission/issues/167


soamvasani:
Thanks, @watilde !

Issue #340, opened 2017-09-21
Multiple trigger definitions causes error in mqTrigger Manager


jsturtevant:
## Issue
When multiple trigger definitions are defined the following error occurs in `mqTrigger`:

```
level=warning msg="Failed to subscribe to message queue trigger trigger2: stan: duplicate durable registration" 
```

## Reproduce
Create a NodeJs env

```
fission env create --name nodejs --image fission/node-env:v0.2.1 
```

Create a function:

```
fission function create --name hello --env nodejs --code hello.js 
```

Create two triggers:

```
fission mqt create --name trigger1 --function hello --topic inputqueue --resptopic outputqueue 
fission mqt create --name trigger2 --function hello --topic inputqueue2 --resptopic outputqueue2
```

PR #341, opened 2017-09-21
Multiple Trigger Definitions Fix

jsturtevant:
Fix for #340 

The issue was that `range` allocates the `trigger` value once which causes the newTriggerMap to have reference to last instance trigger.

To fix the issue, create a new trigger by pulling it off the original map.


erwinvaneyk:
lgtm 👍

soamvasani:
Thanks @jsturtevant!

I am embarassed at how many times I've made this same mistake, argh. I wonder if one of the go tools would help find these automatically.

jsturtevant:
And thanks to @peterhuene for helping me debug it :-) Agree tooling should be able to help warn about this type of issue.

PR #343, opened 2017-09-22
Fix unsupported checksum type (issue 342)

life1347:


fission/function.go
```diff
@@ -62,6 +65,22 @@ func createArchive(client *client.Client, fileName string) *fission.Archive {
 
 		archive.Type = fission.ArchiveTypeUrl
 		archive.URL = archiveUrl
+
+		f, err := os.Open(fileName)
+		if err != nil {
+			fatal(fmt.Sprintf("Error finding file %v: %v", fileName, err))
```
soamvasani: please use checkErr like the rest of the cli -- `checkErr(err, fmt.Sprintf("find file %v", fileName))`

fission/function.go
```diff
@@ -62,6 +65,22 @@ func createArchive(client *client.Client, fileName string) *fission.Archive {
...
+		}
+		defer f.Close()
+
+		h := sha256.New()
+		if _, err := io.Copy(h, f); err != nil {
+			fatal(fmt.Sprintf("Error calculating checksum for file %v: %v", fileName, err))
```
soamvasani: use `checkErr`

donaherc:
I've verified that the branch does indeed solve my issue:

```
$ curl http://$FISSION_ROUTER/hello
Hello, World!
```

PR #346, opened 2017-09-23
Allow namespaces to be used in dotnet functions for better local dev exp

dicko2:
Issue: For a local development environment we create small projects with a series of functions in them, also with unit tests as well for testing them locally. Currently fission requires the class to be at the root name space, which means working locally i cannot have more than on FissionFunction in my project.

example working
```csharp
using System;
using Fission.DotNetCore.Api;

    public class FissionFunction
    {
        public string Execute(FissionContext context)
        {
            return "Hello World!";
        }
    }
```
example not working
```csharp
//func1.cs
using System;
using Fission.DotNetCore.Api;

namespace dotnet.HelloWorld
{
    public class FissionFunction
    {
        public string Execute(FissionContext context)
        {
            return "Hello World!";
        }
    }
}
```

```csharp
//func2.cs
using System;
using Fission.DotNetCore.Api;
namespace dotnet.HelloWorld2
{
    public class FissionFunction
    {
        public string Execute(FissionContext context)
        {
            return "Hello World Number 2!";
        }
    }
}
```

This change will allow the class to be placed inside a namespace. This means in cases like ours where we are building small project with a series of functions in them that we want to test and deploy together we can place them in separate namespaces and test and compile them locally together.

### DIFF
--- a/environments/dotnet/FissionCompiler.cs
+++ b/environments/dotnet/FissionCompiler.cs
@@ -64,7 +64,7 @@ namespace Fission.DotNetCore.Compiler
                     ms.Seek(0, SeekOrigin.Begin);
 
                     Assembly assembly = AssemblyLoadContext.Default.LoadFromStream(ms);
-                    var type = assembly.GetType("FissionFunction");
+                    var type = assembly.GetTypes().FirstOrDefault(x => x.Name.Equals("FissionFunction"));
                     var info = type.GetMember("Execute").First() as MethodInfo;
                     return new Function(assembly, type, info);
                 }
--- a/environments/dotnet20/FissionCompiler.cs
+++ b/environments/dotnet20/FissionCompiler.cs
@@ -64,7 +64,7 @@ namespace Fission.DotNetCore.Compiler
                     ms.Seek(0, SeekOrigin.Begin);
 
                     Assembly assembly = AssemblyLoadContext.Default.LoadFromStream(ms);
-                    var type = assembly.GetType("FissionFunction");
+                    var type = assembly.GetTypes().FirstOrDefault(x => x.Name.Equals("FissionFunction"));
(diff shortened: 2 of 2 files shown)


soamvasani:
Seems reasonable, thanks!

soamvasani:
@dicko2 Just to clarify -- this change doesn't break compatiblity, right? In other words, do functions that used to work with the old environment continue to work with the new one?

soamvasani:
@dicko2 ^ in case you missed the question about testing/compatibility. Thanks.

soamvasani:
Closing due to inactivity.  Please reopen if you've tested existing functions and made sure they don't break.

paraspatidar:
Hi , 
In case you are looking for this thing , here is how you can achieve it  , in environment file FissionCompile.cs  , modify & add this lines: 


```
                    
                    //support for Namespace , thus need fully qualified class name : 
                     // FuncNameSpace.FissionFunction
                     //var type = assembly.GetType("FissionFunction");
                    var type = assembly.GetType("FuncNameSpace.FissionFunction");

```
Now , create a function with many additional classes and namespace as FuncNameSpace 

```
namespace FuncNameSpace
{
public class FissionFunction 
{
    public string Execute(FissionContext context){
		    string respo = "orgval";
			var funcName="func1";
			var hostname=Environment.GetEnvironmentVariable("HOSTNAME");
            try
            {
				respo=$"{funcName}-{hostname}";
				context.Logger.WriteInfo($"{funcName}-{hostname}");
            }  
            catch(Exception ex)
            {
				context.Logger.WriteError(ex.Message);
				context.Logger.WriteInfo(ex.Message);
                respo = ex.Message;
            }
		return respo;
    }
}
public class another class
{
	
}
}

```

I am planning to raise PR for the same .

Issue #350, opened 2017-09-24
Failed to start storage service after restarting it


life1347:
Storage service failed to start with following error message because it doesn't check the existence of `/fission/fission-functions` before calling `CreateContainer`
```
panic: Error initializing storage: mkdir /fission/fission-functions: file exists
```

soamvasani:
Going to keep this open for fixing it while properly using the stow API (i.e. use WalkContainers instead of constructing the container id) @life1347

soamvasani:
fixed by #353

PR #352, opened 2017-09-24
Fix storage service failed to start after restarting it

life1347:
Fix https://github.com/fission/fission/issues/350


soamvasani:
Thanks for the quick fix! I'm going to merge this because the bug is so important, but it's a violation of the stow API -- can you instead use loc.Containers(), iterate over it (there will only be one result), and use that container?

PR #354, opened 2017-09-25
Make default node-env use alpine. List envs in documentation.

### LINKED ISSUE #319: Create alpine-node environment
The image fission/node-env is 679MB, which is quite big and causes pull issues on slower networks.
As discussed on slack, we should create a smaller node env image based on mhart/alpine-node, which is currently around 65MB. Users will then choose the appropriate env for their needs when creating the env, for example with syntax node-env:alpine or node-env:debian

rapitable:
Fixes #319


soamvasani:
Thanks for this work, @rapitable !

The current image uses the [`onbuild` image](https://github.com/nodejs/docker-node/blob/master/8.5/onbuild/Dockerfile), which copies package.json into the image and runs npm install. 

I think there's a couple ways to do this -- we could to these dockerfiles to copy package.json and run npm install, or we could do a multistage docker build, use the onbuild image as the first stage, and then copy out the node_modules into a second stage alpine image.

rapitable:
Hi @soamvasani , ah good point. I saw that the node community is looking to deprecate the use of onbuild images so I changed it but I hadn't appreciated that's what the base image is doing.
The official recommendation (https://hub.docker.com/_/node/) is to:

> adjust your Dockerfile to inherit from a non-onbuild variant and copy the commands from the onbuild variant Dockerfile (moving the ONBUILD lines to the end and removing the ONBUILD keywords) into your own file so that you have tighter control over them and more transparency for yourself and others looking at your Dockerfile as to what it does

So I'll do that and update the PR when I get chance

rapitable:
Hi @soamvasani, I think this is good to go. I've tested running with both alpine and debian images and all seems good.

soamvasani:
Excellent, thanks @rapitable !

Issue #236, opened 2017-06-21
CLI to call a function + output error logs


soamvasani:
Here's a really common workflow:

1. Create a function
2. Map a URL
3. Hit the URL
4. Get an internal server error because of a function exception or failure to load
5. Look for logs

We can combine steps 2-5 into one CLI command

So you'll be able to do

```
fission function test --name hello
```

This should create a route, and do a GET. If the function succeeds it should output the result of the request, and if there's an error it should output the function pod's logs.

soamvasani:
There's already #259, but we also need to retrieve function pod logs.

PR #355, opened 2017-09-25
Test functions 236

vishal-biyani:
- This is the first implementation of #236, tested with NodeJS so far. Ran into some issue with other env. will try again in a bit
- Changes from #259 have been cherry picked for this one
- Is still dependent on ```FISSION_ROUTER``` env as #242 is not merged yet

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/355)


fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+
+	url := fmt.Sprintf("http://%s/fission-function/%s", routerURL, fnName)
+
+	maxRetries := 3
+	retryDelay := 100 * time.Millisecond
+	var resp *http.Response
+	fnReachable := false
+
+	for i := 0; i < maxRetries; i++ {
+		time.Sleep(retryDelay)
```
soamvasani: No need to sleep before the first request?

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+		resp = httpRequest(c.String("method"), url, c.String("body"), c.StringSlice("header"))
+		if resp.StatusCode < 300 {
+			body, _ := ioutil.ReadAll(resp.Body)
```
soamvasani: Don't ignore errors

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+			fmt.Print(string(body))
+			defer resp.Body.Close()
+			fnReachable = true
+			break
+		}
+	}
+
+	if !fnReachable {
+		fmt.Print("Function not reachable \n")
+		body, _ := ioutil.ReadAll(resp.Body)
```
soamvasani: handle errors

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+		fmt.Print(string(body))
```
soamvasani: print status code too? Something like `Error calling function <function>: <code> <body>`

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+	defer resp.Body.Close()
+
+	return nil
+}
+
+func fnTest(c *cli.Context) error {
+	//client := getClient(c.GlobalString("server"))
+	fnName := c.String("name")
+
+	fnCreate(c)
```
soamvasani: suggest removing

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+
+	if !fnReachable {
+		fmt.Print("Function not reachable \n")
+		body, _ := ioutil.ReadAll(resp.Body)
+		fmt.Print(string(body))
+		defer resp.Body.Close()
+		fnLogs(c)
+	}
+
+	fnDelete(c)
```
soamvasani: suggest removing

fission/main.go
```diff
@@ -55,6 +58,8 @@ func main() {
 		{Name: "list", Usage: "List all functions", Flags: []cli.Flag{}, Action: fnList},
 		{Name: "logs", Usage: "Display function logs", Flags: []cli.Flag{fnNameFlag, fnPodFlag, fnFollowFlag, fnDetailFlag, fnLogDBTypeFlag}, Action: fnLogs},
 		{Name: "pods", Usage: "Display function pods", Flags: []cli.Flag{fnNameFlag, fnLogDBTypeFlag}, Action: fnPods},
+		{Name: "invoke", Usage: "Invoke a function", Flags: []cli.Flag{fnNameFlag, fnMethodFlag, fnBodyFlag, fnHeaderFlag}, Action: fnInvoke},
+		{Name: "test", Usage: "Test a function", Flags: []cli.Flag{fnNameFlag, fnEnvNameFlag, fnCodeFlag, fnPackageFlag, fnSrcPackageFlag, htMethodFlag}, Action: fnTest},
```
soamvasani: I don't think we need both. Just test (or just invoke) is fine, with all the functionality in one command.

fission/function.go
```diff
@@ -432,3 +433,69 @@ func fnPods(c *cli.Context) error {
...
+			break
+		}
+	}
+
+	if !fnReachable {
+		fmt.Print("Function not reachable \n")
+		body, _ := ioutil.ReadAll(resp.Body)
+		fmt.Print(string(body))
+		defer resp.Body.Close()
+		fnLogs(c)
```
soamvasani: don't ignore errors

fission/function.go
vishal-biyani: I had thought that we wanted an end to end flow: create-test-delete but rechecked the description, will remove it.

fission/function.go
vishal-biyani: Instead of a fixed sleep, I am retrying. Would a fixed sleep be good idea?

soamvasani: The retry is good, but the first request attempt can be sent without waiting (in other words the sleep should be after the request attempt)

fission/common.go
```diff
@@ -50,3 +55,36 @@ func checkErr(err error, msg string) {
 		fatal(fmt.Sprintf("Failed to %v: %v", msg, err))
 	}
 }
+
+func httpRequest(method, url, body string, headers []string) *http.Response {
+	if method != http.MethodGet &&
+		method != http.MethodDelete &&
+		method != http.MethodPost &&
+		method != http.MethodPut {
+		// Fatal causes some stale data for the calling function, needs improvement
```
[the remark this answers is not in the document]
vishal-biyani: Above is not valid now, my assumption of workflow was different. So removed relevant sections.

soamvasani:
@vishal-biyani Mind rebasing + addressing conflicts?

vishal-biyani:
@soamvasani Rebased + Solved merge conflicts.

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+	}
+
+	queryURL, err := url.Parse(c.GlobalString("server"))
+	checkErr(err, "Error parsing the base URL")
+	queryURL.Path = fmt.Sprintf("/proxy/logs/%s", fnName)
+
+	req, err := http.NewRequest("POST", queryURL.String(), nil)
+	checkErr(err, "Error creating logs request")
+
+	httpClient := http.Client{Timeout: 5 * time.Second}
```
soamvasani: How did you choose the timeout? Is it what kubectl does?

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	var data interface{}
+	json.Unmarshal(body, &data)
+	ldata, err := base64.StdEncoding.DecodeString(data.(string))
+
+	if ldata == nil {
+		return false
+	}
+	fmt.Println("Data=", string(ldata))
```
soamvasani: Remove "Data="

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		fmt.Println("Error in getting logs from POD directly, will check logstore, StatusCode:", resp.StatusCode)
+		return false
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	var data interface{}
+	json.Unmarshal(body, &data)
+	ldata, err := base64.StdEncoding.DecodeString(data.(string))
```
soamvasani: I'm confused about this. What does the body look like? Why use json.Unmarshal for just a string? Can you simply do `base64.StdEncoding.Decode(body)`?

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+
+	if ldata == nil {
```
soamvasani: when would this happen?

fission/function.go
```diff
@@ -497,3 +536,43 @@ func fnPods(c *cli.Context) error {
...
+func fnTest(c *cli.Context) error {
+	fnName := c.String("name")
+	routerURL := os.Getenv("FISSION_ROUTER")
+	if len(routerURL) == 0 {
+		fatal("Need FISSION_ROUTER set to your fission router.")
+	}
+
+	url := fmt.Sprintf("http://%s/fission-function/%s", routerURL, fnName)
+
+	maxRetries := 3
```
soamvasani: I don't think this command needs retries, let's keep it simple and the user can retry if they want.

fission/function.go
```diff
@@ -497,3 +536,43 @@ func fnPods(c *cli.Context) error {
...
+			defer resp.Body.Close()
+			fnReachable = true
+			break
+		}
+		time.Sleep(retryDelay)
+	}
+
+	if !fnReachable {
+		body, err := ioutil.ReadAll(resp.Body)
+		checkErr(err, "Function test failed")
```
soamvasani: See other uses of checkErr -- this doesn't fit the grammar of the error message. Also, this error is not about the function test failing, it's about failing to read the http response

fission/function.go
```diff
@@ -497,3 +536,43 @@ func fnPods(c *cli.Context) error {
...
+		fmt.Printf("Error calling function %v: %v %v", fnName, resp.StatusCode, string(body))
+		defer resp.Body.Close()
+		flag := getPodLogs(c)
```
soamvasani: Don't use `bool` for error checking in Go -- use an `error`.

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
 	return code
 }
 
+func getPodLogs(c *cli.Context) bool {
```
soamvasani: This function doesn't just get logs, it prints them too -- it should be called "show pod logs" or "print pod logs" or something like that

controller/functionApi.go
```diff
@@ -188,3 +193,62 @@ func (a *API) FunctionLogsApiPost(w http.ResponseWriter, r *http.Request) {
...
+
+	_, clientset, err := tpr.GetKubernetesClient()
+	if err != nil {
+		a.respondWithError(w, err)
+		return
+	}
+
+	// Get Unmanaged Pods first
+	selector := "unmanaged=true,functionName=" + fnName
+	podList, err := clientset.Core().Pods(ns).List(metav1.ListOptions{LabelSelector: selector})
```
soamvasani: Always need to handle errors!

controller/functionApi.go
```diff
@@ -188,3 +193,62 @@ func (a *API) FunctionLogsApiPost(w http.ResponseWriter, r *http.Request) {
...
+	// Get the logs for last Pod executed
+	pods := podList.Items
+	sort.Slice(pods, func(i, j int) bool {
+		itime := pods[i].ObjectMeta.CreationTimestamp.Time
+		jtime := pods[j].ObjectMeta.CreationTimestamp.Time
+		return itime.After(jtime)
+	})
+
+	podLogOpts := v1.PodLogOptions{Container: envName} // Only the env container, not fetcher
+	podLogsReq := clientset.Core().Pods(ns).GetLogs(pods[0].ObjectMeta.Name, &podLogOpts)
```
soamvasani: `pods[0]` won't exist if there are no pods, and you'll get a runtime range check panic. So you have to handle the case where there are zero pods.

fission/common.go
```diff
@@ -50,3 +55,37 @@ func checkErr(err error, msg string) {
...
+		fatal(fmt.Sprintf("Invalid HTTP method '%s'.", method))
+	}
+
+	req, err := http.NewRequest(method, url, strings.NewReader(body))
+	checkErr(err, "Error creating reuquest to function")
```
soamvasani: checkErr(err, "create HTTP request")

checkErr prints "Failed to <msg>" where msg is the second argument.

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+	fnName := c.String("name")
+	if len(fnName) == 0 {
+		fatal("Need --name argument.")
+	}
+
+	queryURL, err := url.Parse(c.GlobalString("server"))
+	checkErr(err, "Error parsing the base URL")
```
soamvasani: error message grammar, `checkErr(err, "parse server URL")`

fission/function.go
```diff
@@ -134,6 +138,41 @@ func getContents(filePath string) []byte {
...
+	queryURL.Path = fmt.Sprintf("/proxy/logs/%s", fnName)
+
+	req, err := http.NewRequest("POST", queryURL.String(), nil)
+	checkErr(err, "Error creating logs request")
```
soamvasani: error message grammar

vishal-biyani:
Hi @soamvasani I have addressed all comments except mechanism to find the "latest pod" for logs. For that one, I tried a few things, but the current logic seems to work based on sort in place. LMK and we can discuss it on slack if needed.

fission/function.go
```diff
@@ -497,3 +528,36 @@ func fnPods(c *cli.Context) error {
...
+func fnTest(c *cli.Context) error {
+	fnName := c.String("name")
+	routerURL := os.Getenv("FISSION_ROUTER")
+	if len(routerURL) == 0 {
+		fatal("Need FISSION_ROUTER set to your fission router.")
+	}
+
+	url := fmt.Sprintf("http://%s/fission-function/%s", routerURL, fnName)
+
+	fnReachable := false
```
soamvasani: this function could be written more simply without this flag, right?

fission/function.go
```diff
@@ -497,3 +528,36 @@ func fnPods(c *cli.Context) error {
...
+
+	resp := httpRequest(c.String("method"), url, c.String("body"), c.StringSlice("header"))
+	if resp.StatusCode < 300 {
```
soamvasani: 3xx are not really errors, right? I don't think we should call them errors. This should be 400...

fission/function.go
```diff
@@ -497,3 +528,36 @@ func fnPods(c *cli.Context) error {
...
+		body, err := ioutil.ReadAll(resp.Body)
+		checkErr(err, "Function test")
+		fmt.Print(string(body))
+		defer resp.Body.Close()
+		fnReachable = true
```
soamvasani: just return here instead of setting a flag?

fission/function.go
```diff
@@ -134,6 +137,34 @@ func getContents(filePath string) []byte {
...
+	req, err := http.NewRequest("POST", queryURL.String(), nil)
+	checkErr(err, "create logs request")
+
+	httpClient := http.Client{}
+	resp, err := httpClient.Do(req)
+	checkErr(err, "execute get logs request")
+
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("get logs from POD directly")
```
soamvasani: s/POD/pod/

controller/functionApi.go
```diff
@@ -188,3 +194,66 @@ func (a *API) FunctionLogsApiPost(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	f, err := a.fissionClient.Functions(api.NamespaceDefault).Get(fnName)
+	if err != nil {
+		a.respondWithError(w, err)
+		return
+	}
+	envName := f.Spec.Environment.Name
+
+	_, clientset, err := tpr.GetKubernetesClient()
```
soamvasani: It would be better to move this to the init of the api object -- so you don't have to login to kubernetes api on every logs request

fission/function.go
```diff
@@ -497,3 +528,36 @@ func fnPods(c *cli.Context) error {
...
+		defer resp.Body.Close()
+		fnReachable = true
+	}
+
+	if !fnReachable {
+		body, err := ioutil.ReadAll(resp.Body)
+		checkErr(err, "read log response from pod")
+		fmt.Printf("Error calling function %v: %v %v", fnName, resp.StatusCode, string(body))
+		defer resp.Body.Close()
+		err = printPodLogs(c)
```
soamvasani: This logic -- calling print pod logs and falling back to the DB -- would be useful in the `fission fn logs` command as well.  (But we can push that to a later change if you want to.)

vishal-biyani: @soamvasani I can integrate this and make it so that ``` fn logs``` calls pod logs and if not found then looks up DB logs. But there are two things which are inconsistencies from user perspective and we should address in future: 1) The logs from POD don't have a timestamp before every line while ```fn logs``` prints TS 2) The ```fn logs``` shows lot many lines as opposed to pod logs which shows only recent pod's logs. So the behavior/experience can vary based on if you test a function and call logs or call it without testing it. There might be some optimizations in fetching logs from DB side too. What do you think?

controller/functionApi.go
```diff
@@ -188,3 +194,66 @@ func (a *API) FunctionLogsApiPost(w http.ResponseWriter, r *http.Request) {
...
 }
+
+// FunctionPodLogs : Get logs for a function directly from pod
+func (a *API) FunctionPodLogs(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	fnName := vars["function"]
+	ns := vars["namespace"]
+
+	if len(ns) == 0 {
+		ns = "fission-function"
```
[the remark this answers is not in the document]
vishal-biyani: @soamvasani I will create an issue so we can track this.

vishal-biyani:
Thanks @soamvasani for extensive review on first PR :) I have one comment specific to using new logs feature in ```fn logs```, based on your thoughts I will create a issue or fix in existing PR itself.

vishal-biyani:
Review status: 0 of 6 files reviewed at latest revision, 39 unresolved discussions.

---

*[controller/functionApi.go, line 29 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-29:-Kxg5dDiegjmZEoBAOjd:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L29)):*

skip this blank link please, we're following a convention of three "paragraphs" of imports: [standard libs] [external deps] [internal deps]
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 33 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-33:-Kxg2Pf1DknkWXh6B_ej:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L33)):*

commented out, please remove
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 221 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-221:-Kxg38koO6nc00jXvOcV:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L221)):*

You can just use `functionName`, right? No need for matching on `unmanaged`, too? (In the future we will have function pods inside deployments, and so without the unmanaged label.)
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 222 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-222:-Kxg39QE6WwpkApJPEcG:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L222)):*

Always need to handle errors! 
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 224 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-224:-Kxg39ztFVcw9t1PjZZd:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L224)):*

Just nitpicking here, but this isn't the last pod executed, but the pod that was created last. Since we bind functions to pods late, the two can be different.  The actual code will work fine for now, I think, as a heuristic.
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 233 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-233:-Kxg3BEgktvN3HxcQYAZ:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L233)):*

`pods[0]` won't exist if there are no pods, and you'll get a runtime range check panic. So you have to handle the case where there are zero pods.
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 240 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r6-240:-Kxg3CckPOQMGCzVOdS5:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/controller/functionApi.go#L240)):*

This is ok as an initial implementation, but we'll end up reading logs into memory. We should stream them to the response (using `io.Copy` would work, I think). Please fix it if it's not too complex, but otherwise file an issue to come back to later.
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 205 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r7-205:-Kxg2nXcp34iqeULmP1p:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/controller/functionApi.go#L205)):*

@soamvasani I will create an issue so we can track this.
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 215 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r7-215:-Kxg35jJZhxwFIknpgy3:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/controller/functionApi.go#L215)):*

It would be better to move this to the init of the api object -- so you don't have to login to kubernetes api on every logs request
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 221 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r7-221:-Kxg37Mn9j22T709xXSD:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/controller/functionApi.go#L221)):*

nit: comment is outdated in relation to the code... should just say function pods now
</blockquote></details>

Done.

---

*[controller/functionApi.go, line 257 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9Y-r7-257:-Kxg5_A_Sz2ii8FaFkOq:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/controller/functionApi.go#L257)):*

I think this should move to be before the `io.Copy` call. Otherwise we won't close podLogs if there's an error writing to `w`.
</blockquote></details>

Done.

---

*[fission/common.go, line 64 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9U-r2-64:-Kxg3F9HYg10xJ7sZ7tr:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/common.go#L64)):*

Can you remove the comment too if it's not relevant?
</blockquote></details>

Done.

---

*[fission/common.go, line 73 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9U-r6-73:-Kxg52wqRJY6pNjIWHMX:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/common.go#L73)):*

checkErr(err, "create HTTP request")

checkErr prints "Failed to <msg>" where msg is the second argument.
</blockquote></details>

Done.

---

*[fission/common.go, line 78 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9U-r6-78:-Kxg54JMmgqEzBW51RvX:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/common.go#L78)):*

We should fail completely instead of just warning and skipping the header. The semantics of the request could be arbitrarily different when a header is skipped -- we don't know anything about the function.
</blockquote></details>

Done.

---

*[fission/common.go, line 87 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9U-r6-87:-Kxg55e8VrgPfDgaKky-:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/common.go#L87)):*

Log the error `err`
</blockquote></details>

Done.

---

*[fission/function.go, line 479 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-479:-Kxg5UBnMoztWtALtdVF:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L479)):*

The retry is good, but the first request attempt can be sent without waiting (in other words the sleep should be after the request attempt)
</blockquote></details>

Done.

---

*[fission/function.go, line 482 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-482:-Kxg5SEVWJzRSZabaHz7:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L482)):*

Don't ignore errors
</blockquote></details>

Done.

---

*[fission/function.go, line 492 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-492:-Kxg5R8ViNkha2KTwYxd:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L492)):*

handle errors
</blockquote></details>

Done.

---

*[fission/function.go, line 493 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-493:-Kxg5QgqSfq4TU4Fp4Qq:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L493)):*

print status code too? Something like `Error calling function <function>: <code> <body>`
</blockquote></details>

Done.

---

*[fission/function.go, line 495 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-495:-Kxg5JByRtUG0ynCZNIk:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L495)):*

don't ignore errors
</blockquote></details>

Done.

---

*[fission/function.go, line 498 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r2-498:-Kxg5IJdIJl_sx_h1lnL:b-896fix) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/function.go#L498)):*

suggest removing
</blockquote></details>

Done.

---

*[fission/function.go, line 141 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-141:-Kxg57IkPS6LYbsJUE4-:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L141)):*

This function doesn't just get logs, it prints them too -- it should be called "show pod logs" or "print pod logs" or something like that
</blockquote></details>

Done.

---

*[fission/function.go, line 148 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-148:-Kxg58jCadMnb649kdNd:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L148)):*

error message grammar, `checkErr(err, "parse server URL")`
</blockquote></details>

Done.

---

*[fission/function.go, line 152 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-152:-Kxg596RqPhQzGskMC1t:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L152)):*

error message grammar
</blockquote></details>

Done.

---

*[fission/function.go, line 154 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-154:-Kxg59nyiigRW9BQtF7A:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L154)):*

How did you choose the timeout? Is it what kubectl does? 
</blockquote></details>

Done.

---

*[fission/function.go, line 167 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-167:-Kxg5BQ-7hdHTn43Wv0g:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L167)):*

I'm confused about this. What does the body look like? Why use json.Unmarshal for just a string? Can you simply do `base64.StdEncoding.Decode(body)`?
</blockquote></details>

Done.

---

*[fission/function.go, line 169 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-169:-Kxg5BsZiARVhpDJlZfU:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L169)):*

when would this happen?
</blockquote></details>

Done.

---

*[fission/function.go, line 172 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-172:-Kxg5AxESGAZdL8kvPVb:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L172)):*

Remove "Data="
</blockquote></details>

Done.

---

*[fission/function.go, line 549 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-549:-Kxg5TZvofnqL2ErCE4L:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L549)):*

I don't think this command needs retries, let's keep it simple and the user can retry if they want.
</blockquote></details>

Done.

---

*[fission/function.go, line 569 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-569:-Kxg5PoKVLF6HO_OOJDI:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L569)):*

See other uses of checkErr -- this doesn't fit the grammar of the error message. Also, this error is not about the function test failing, it's about failing to read the http response
</blockquote></details>

Done.

---

*[fission/function.go, line 572 at r6](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r6-572:-Kxg5OoAy9qhKwuYoK9J:b-896fix) ([raw file](https://github.com/fission/fission/blob/3ef8396e8ec84d954e10628eaaa3da673ff5cc91/fission/function.go#L572)):*

Don't use `bool` for error checking in Go -- use an `error`.
</blockquote></details>

Done.

---

*[fission/function.go, line 159 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r7-159:-Kxg5ASfGS00mJx14ESm:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/fission/function.go#L159)):*

s/POD/pod/
</blockquote></details>

Done.

---

*[fission/function.go, line 541 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r7-541:-Kxg5T5UQ_1qhhTms5Sb:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/fission/function.go#L541)):*

this function could be written more simply without this flag, right?
</blockquote></details>

Done.

---

*[fission/function.go, line 544 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r7-544:-Kxg5Sg9GA9msZXnNutx:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/fission/function.go#L544)):*

3xx are not really errors, right? I don't think we should call them errors. This should be 400...
</blockquote></details>

Done.

---

*[fission/function.go, line 549 at r7](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9V-r7-549:-Kxg5Rruh-d6Nlhe0QAd:b-896fix) ([raw file](https://github.com/fission/fission/blob/9be1f29b27be795876067bd501d821712ddc5362/fission/function.go#L549)):*

just return here instead of setting a flag?
</blockquote></details>

Done.

---

*[fission/main.go, line 62 at r2](https://reviewable.io:443/reviews/fission/fission/355#-KxUTmgIRu2QafcV2t9W-r2-62:-Kxg5HlIUXakETNFMz8D:b2nhktp) ([raw file](https://github.com/fission/fission/blob/ca052f4a8fea13d20e9b2a0e29de28a6075e0d95/fission/main.go#L62)):*

I don't think we need both. Just test (or just invoke) is fine, with all the functionality in one command.
</blockquote></details>

OK.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/355)*

soamvasani:
:lgtm:

---

Reviewed 1 of 4 files at r6, 1 of 3 files at r7, 4 of 4 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/355#-:-Kxku7mKdC_VZ79nDXup:bnfp4nl)*

PR #363, opened 2017-09-29
Fix workflow apiserver proxy

erwinvaneyk:
This PR fixes the workflow apiserver proxy. Additionally it adds an experimental automated deploy script.


soamvasani:
Can you keep the deploy script in a separate PR please? Seems unrelated to this change.

PR #375, opened 2017-10-12
First phase of execution strategy layer for #193

vishal-biyani:
Implements a very thin layer on top of pool manager without any change in functionality. The router now interacts with executor and executor, in turn, interacts with pool manager.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/375)


soamvasani:
Thanks @vishal-biyani. Sorry if this was unclear in the design discussion, but I'm not sure we need an actual new server for this; why not just add a layer to the poolmgr (now executor) service itself? That executor can internally dispatch to the poolmgr as one backend. 

I don't see a lot of advantages in splitting this out as a server+deployment+service.

soamvasani:
Hey Vishal, looks like the tests broke because they still collect logs from a "poolmgr" pod.

executor/cleanup.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package poolmgr
+package executor
```
soamvasani: Hmm, this _is_ a poolmgr-specific file. How about we make a subdirectory under here, and move the poolmgr into it?

executor/gp.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package poolmgr
+package executor
```
soamvasani: Same comment as above.. we should keep a clear package boundary around the poolmgr, which we could do by moving it into its own package.

soamvasani:
Hey, I'm going to merge #381 before this one, since that one is blocking our compatibility with Kubernetes 1.8. It may be a bit of a big merge conflict, sorry about that.

vishal-biyani:
No issues @soamvasani, CRD is definitely good to have ASAP. I have merged with latest from the master.

vishal-biyani:
Closing this in favor of #384 and more smaller PRs.

Issue #366, opened 2017-09-30
Router takes long time to sync routes


life1347:
When there are multiple triggers or functions exist, router takes long time to sync routes.
```
$ kubectl logs -f router-1336853453-2rxcs
2017/09/30 08:16:47 Starting router at port 8888
2017/09/30 08:16:47 elapsed time in get router = 1.015145ms
2017/09/30 08:16:47 Syncing http triggers
2017/09/30 08:16:47 Syncing http triggers
2017/09/30 08:17:30 elapsed time in get router = 42.950303499s
2017/09/30 08:17:30 Syncing http triggers
2017/09/30 08:17:30 elapsed time in get router = 43.113514069s
2017/09/30 08:17:30 Syncing http triggers
2017/09/30 08:17:50 elapsed time in get router = 19.195917427s
2017/09/30 08:17:50 Syncing http triggers
2017/09/30 08:17:50 elapsed time in get router = 19.197475451s
2017/09/30 08:17:50 Syncing http triggers
2017/09/30 08:18:35 elapsed time in get router = 44.793002911s
2017/09/30 08:18:35 Syncing http triggers
2017/09/30 08:18:35 elapsed time in get router = 44.810635728s
2017/09/30 08:18:35 Syncing http triggers
```

PR #376, opened 2017-10-14
Reduce function resolving time

life1347:
Before this PR, resolver takes long time to resolve function because of `kube-api-qps` of api server.
In this PR, resolver uses informer cache to resolve function instead of making http call every time to avoid hitting the qps of api server. And now the resolve time is down to average 5µs (original ~400ms) each time. (solve issue https://github.com/fission/fission/issues/366)

### DIFF
--- a/router/httpTriggers.go
+++ b/router/httpTriggers.go
@@ -23,7 +23,11 @@ import (
 
 	"github.com/gorilla/mux"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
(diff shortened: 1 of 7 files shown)


router/httpTriggers.go
```diff
@@ -135,75 +142,77 @@ func (ts *HTTPTriggerSet) watchTriggers() {
...
+	listWatch := &k8sCache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			return watchlist.List(options)
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return watchlist.Watch(options)
+		},
+	}
+	resyncPeriod := 30 * time.Second
+	_, controller := k8sCache.NewInformer(listWatch, &tpr.Function{}, resyncPeriod,
```
soamvasani: How about keeping the store -- and keeping it in  a new `ts.store`? And then using it in `syncTriggers`?

soamvasani:
I'm going to merge this PR now, but please address the question above in a future change.

PR #378, opened 2017-10-16
changed helm install pullPolicy from Always to IfNotPresent when building local docker image

prithviramesh:
Containers will get stuck in ErrImagePull/ImagePullBackOff state otherwise


soamvasani:
Thanks @prithviramesh !

PR #388, opened 2017-11-01
Use shorter resync period to flush out-of-dated cache

life1347:
This small PR aims to fix the test failure during CI test due to long `resyncPeriod`.

Informer resync watch resources from k8s api server.  If `resyncPeriod` is too long, router may not be able to update with the latest resources cache and caused test failure.

It should fix the PRs (PR382, 381, 367) that failed at test case `test_node_hello_http.sh`. 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/388)

### DIFF
--- a/router/httpTriggers.go
+++ b/router/httpTriggers.go
@@ -151,7 +151,7 @@ func (ts *HTTPTriggerSet) watchTriggers() {
 			return watchlist.Watch(options)
 		},
 	}
-	resyncPeriod := 30 * time.Second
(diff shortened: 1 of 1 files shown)


soamvasani:
I don't understand why this change is needed. There's a watch going on, so the router cache should already be getting updated in < 1 sec. Why do we need a full resync so often?

life1347:
You're right. Close this PR.

PR #390, opened 2017-11-08
Fix functionReferenceResolver return out-of-date function metadata

life1347:
In router we launch two watcher to watch function changes. One for HTTPTriggerSet, another for functionReferenceResolver. 

In some situation, two watchers may not receive change events at the same time.  So functionReferenceResolver return out-of-date function metadata  back to httpTrigger. In this case, router proxies request to the old pod and caused test failure.

This PR aims to solve this problem by sharing http trigger resource store with functionReferenceResolver to prevent data consistency problem.

Following is the interaction diagram for this problem.

router-informer-store-bug

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/390)

### DIFF
--- a/router/functionReferenceResolver.go
+++ b/router/functionReferenceResolver.go
@@ -36,8 +36,6 @@ type (
 	// functionReferenceResolver provides a resolver to turn a function
 	// reference into a resolveResult
 	functionReferenceResolver struct {
-		fissionClient *crd.FissionClient
-
 		// FunctionReference -> function metadata
 		refCache *cache.Cache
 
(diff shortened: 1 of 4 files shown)


soamvasani:
💯 This is an *awesome* explanation of the problem. Thanks for the quick turnaround @life1347!

soamvasani:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[router/httpTriggers.go, line 148 at r1](https://reviewable.io:443/reviews/fission/fission/390#-KyS-93AHqFHEtjFd1Ng:-KyS-93BlEewzH2kHUo7:b-8sntff) ([raw file](https://github.com/fission/fission/blob/ead56a7180edde654e552421bd6f1620c1a15117/router/httpTriggers.go#L148)):*
> ```Go
> }
> 
> func (ts *HTTPTriggerSet) watchTriggers() (k8sCache.Store, k8sCache.Controller) {
> ```

The interface is a little bit confusing:

1. The function doesn't actually start watching the triggers.  It should be called something like initTriggerController().
2. It's strange to both set `ts.triggerStore` and also return store. Do one or the other. Avoid setting it and let `makeHTTPTriggerSet` set `ts.triggerStore`, OR set `ts.triggerStore` avoid returning it. Same for the controller.

---

*[router/httpTriggers.go, line 168 at r1](https://reviewable.io:443/reviews/fission/fission/390#-KyS0CWlmIKLCrFpZMDu:-KyS0CWlmIKLCrFpZMDv:b-nemmbo) ([raw file](https://github.com/fission/fission/blob/ead56a7180edde654e552421bd6f1620c1a15117/router/httpTriggers.go#L168)):*
> ```Go
> }
> 
> func (ts *HTTPTriggerSet) watchFunctions() (k8sCache.Store, k8sCache.Controller) {
> ```

Same comments on this interface as the `watchTriggers` function.

---

*[router/httpTriggers.go, line 204 at r1](https://reviewable.io:443/reviews/fission/fission/390#-KyS0T202oexK_PLyCbx:-KyS0T202oexK_PLyCby:b-3k67ha) ([raw file](https://github.com/fission/fission/blob/ead56a7180edde654e552421bd6f1620c1a15117/router/httpTriggers.go#L204)):*
> ```Go
> 		stop := make(chan struct{})
> 		defer func() {
> 			stop <- struct{}{}
> ```

I'm a bit confused about this. When does this line run? (Does it ever run?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/390)*

soamvasani:
:lgtm:

---

Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/390#-:-KyW2U7EUNESMdVXTsuG:bnfp4nl)*

PR #395, opened 2017-11-14
Added python example to demonstrate status codes.

c0dyhi11:
It took me a while to figure this out. So I figured I'd post an example to help others.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/395)


soamvasani:
Thanks @c0dyhi11 !

PR #397, opened 2017-11-15
Add fission/builder image

erwinvaneyk:
This PR tackles a couple of issues with builder(mgr)

It...:
- ...fixes the issue where the environment-scoped buildcmd was not used.
- ...allows files to be passed to the builder
- ...adds a `fission/builder` docker image avoiding the need to compile fission to get the binary.
- ...fixes the issue where the builder did not log the stderr

~~As a flyby, I updated the deploy script to correspond with the one in fission workflows~~

cc @life1347

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/397)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,13 @@ cache:
     - $HOME/google-cloud-sdk/
     - $HOME/k8scli
 
+services:
+  - docker
+
+before_install:
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+
 install:
   - go get github.com/Masterminds/glide
   - hack/travis-kube-setup.sh
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/dchest/uniuri"
 	"io"
+	"path"
 )
 
 const (
@@ -130,7 +131,17 @@ func (builder *Builder) Handler(w http.ResponseWriter, r *http.Request) {
 
 func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath string) (string, error) {
 	cmd := exec.Command(command)
-	cmd.Dir = srcPkgPath
+
+	fi, err := os.Stat(srcPkgPath)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("could not find srcPkgPath: '%s'", srcPkgPath))
+	}
+	if fi.IsDir() {
+		cmd.Dir = srcPkgPath
+	} else {
+		cmd.Dir = path.Dir(srcPkgPath)
+	}
+
 	// set env variables for build command
 	cmd.Env = append(os.Environ(),
 		fmt.Sprintf("%v=%v", envSrcPkg, srcPkgPath),
@@ -142,17 +153,9 @@ func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath s
 		return "", errors.New(fmt.Sprintf("Error creating stdout pipe for cmd: %v", err.Error()))
 	}
 
-	stderr,  err := cmd.StderrPipe()
+	stderr, err := cmd.StderrPipe()
 	if err != nil {
 		return "", errors.New(fmt.Sprintf("Error creating stderr pipe for cmd: %v", err.Error()))
-	}
-
-	out := io.MultiReader(stdout, stderr)
-	scanner := bufio.NewScanner(out)
-
-	err = cmd.Start()
-	if err != nil {
-		return "", errors.New(fmt.Sprintf("Error starting cmd: %v", err.Error()))
 	}
 
 	var buildLogs string
@@ -161,6 +164,14 @@ func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath s
 	// Init logs
 	fmt.Printf("command=%v\n", command)
 	fmt.Printf("env=%v\n", cmd.Env)
+
+	out := io.MultiReader(stdout, stderr)
+	scanner := bufio.NewScanner(out)
+
+	err = cmd.Start()
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Error starting cmd: %v", err.Error()))
+	}
 
 	// Runtime logs
 	for scanner.Scan() {
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/dchest/uniuri"
+	"io"
 )
 
 const (
@@ -136,12 +137,18 @@ func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath s
 		fmt.Sprintf("%v=%v", envDeployPkg, deployPkgPath),
 	)
 
-	cmdReader, err := cmd.StdoutPipe()
+	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		return "", errors.New(fmt.Sprintf("Error creating stdout pipe for cmd: %v", err.Error()))
 	}
 
-	scanner := bufio.NewScanner(cmdReader)
+	stderr,  err := cmd.StderrPipe()
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Error creating stderr pipe for cmd: %v", err.Error()))
+	}
+
+	out := io.MultiReader(stdout, stderr)
+	scanner := bufio.NewScanner(out)
 
 	err = cmd.Start()
 	if err != nil {
@@ -151,21 +158,28 @@ func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath s
 	var buildLogs string
 
 	fmt.Println("\n=== Build Logs ===")
+	// Init logs
+	fmt.Printf("command=%v\n", command)
+	fmt.Printf("env=%v\n", cmd.Env)
+
+	// Runtime logs
 	for scanner.Scan() {
 		output := scanner.Text()
 		fmt.Println(output)
 		buildLogs += fmt.Sprintf("%v\n", output)
 	}
-	fmt.Println("==================\n")
 
 	if err := scanner.Err(); err != nil {
+		fmt.Println(err)
 		return "", errors.New(fmt.Sprintf("Error reading cmd output: %v", err.Error()))
 	}
 
 	err = cmd.Wait()
 	if err != nil {
-		return "", errors.New(fmt.Sprintf("Error waiting for cmd: %v", err.Error()))
+		fmt.Println(err)
+		return "", errors.New(fmt.Sprintf("Error waiting for cmd '%v': %v", command, err.Error()))
 	}
+	fmt.Println("==================\n")
 
 	return buildLogs, nil
 }
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -181,14 +181,16 @@ func (builder *Builder) build(command string, srcPkgPath string, deployPkgPath s
 	}
 
 	if err := scanner.Err(); err != nil {
-		fmt.Println(err)
-		return "", errors.New(fmt.Sprintf("Error reading cmd output: %v", err.Error()))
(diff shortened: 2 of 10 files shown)


soamvasani:
@life1347 please take a look

erwinvaneyk:
@soamvasani test fails because of missing fission/builder image. If you agree with the approach taken here, we can push the image and rerun the tests to make it all green

soamvasani:
Hrm, what if a builder wants to do something like `FROM golang`?  Maybe we can add the builder binary to the image using a shared volume instead?

erwinvaneyk:
That would also be a solution. The old approach is still possible allowing for build environments to extend other images. Though the image is more of a convenience. It is just a pain to setup up cross-project build chains to get the binary.

soamvasani:
Yes I agree, the cross project thing is annoying, and the current approach (in the master branch, not this PR) is also kinda weird.

Okay, so let's go with this approach of having a builder image, given that it isn't stopping us from the alternatives.

I think there's a way to test it by removing the hard-coding of the `fission/builder` image in the Dockerfile. That way the CI actually tests the builder, e.g. if you have builder changes you want those tested, not the last release. You can parameterize this by using a docker "ARG" for the image name and tag: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact . So if you do that you also won't need the image to be pushed for the tests on this PR to pass.

erwinvaneyk:
@soamvasani thanks for the tip to use build args. I did have to add a docker update step to travis, because the default docker version of travis does not support build args yet.

soamvasani:
Reviewed 8 of 9 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[builder/builder.go, line 184 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzNZfk4vOZCl3_FwEno:-KzNZfk4vOZCl3_FwEnp:bc2tx5l) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/builder/builder.go#L184)):*
> ```Go
> 
> 	if err := scanner.Err(); err != nil {
> 		fmt.Println(err)
> ```

log some sort of message, not just the error; you could log the same message that's being returned.

---

*[builder/builder.go, line 190 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzNZu5bOc3X9EAE0bTb:-KzNZu5cUeSa2PtUhKxI:b-9ib12m) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/builder/builder.go#L190)):*
> ```Go
> 	err = cmd.Wait()
> 	if err != nil {
> 		fmt.Println(err)
> ```

same comment, log a message + the error

---

*[buildermgr/envwatcher.go, line 452 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_7X0B8Rasw7WsU3g:-KzN_7X0B8Rasw7WsU3h:bb0ejqq) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/envwatcher.go#L452)):*
> ```Go
> 							Name:                   "builder",
> 							Image:                  env.Spec.Builder.Image,
> 							ImagePullPolicy:        apiv1.PullIfNotPresent,
> ```

Doesn't this break the integration tests? I think you'll need an env var that can be set on the deployment, just like we do with fetcher in the tests.

---

*[buildermgr/pkgwatcher.go, line 85 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_RNN8JgZp-Zter1L:-KzN_RNN8JgZp-Zter1M:b-1gewrn) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/pkgwatcher.go#L85)):*
> ```Go
> 			rv = pkg.Metadata.ResourceVersion
> 
> 			// only build packages in pending state
> ```

please avoid unnecessary changes

---

*[hack/deploy.sh, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_smvOCMmQKxOiFSF:-KzN_smvOCMmQKxOiFSG:b-veyjxg) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/hack/deploy.sh#L1)):*
> ```Shell
> #!/usr/bin/env bash
> ```

Can you create a separate PR for this file?  I'm not sure how it relates to this PR?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

erwinvaneyk:
Review status: 9 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[buildermgr/envwatcher.go, line 452 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_7X0B8Rasw7WsU3g:-KzP7kHlld3MTxgkEnJV:b-mgoqx8) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/envwatcher.go#L452)):*

Doesn't this break the integration tests? I think you'll need an env var that can be set on the deployment, just like we do with fetcher in the tests.
</blockquote></details>

An env var or we could specify it in the env.Spec.Builder?

---

*[buildermgr/pkgwatcher.go, line 85 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_RNN8JgZp-Zter1L:-KzP8V3D-zVXpL0I5aiF:b-t1aywy) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/pkgwatcher.go#L85)):*

please avoid unnecessary changes
</blockquote></details>

Will try to avoid it in future prs :+1:

---

*[hack/deploy.sh, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_smvOCMmQKxOiFSF:-KzP8z0pq6jVW3-oAZt8:b-53zjqy) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/hack/deploy.sh#L1)):*

Can you create a separate PR for this file?  I'm not sure how it relates to this PR?
</blockquote></details>

It indeed is a flyby as specified in the main post. Will pull it out of this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

erwinvaneyk:
Review status: 9 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[builder/builder.go, line 184 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzNZfk4vOZCl3_FwEno:-KzPHJQK5PA8oepT3JQY:b-896fix) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/builder/builder.go#L184)):*

log some sort of message, not just the error; you could log the same message that's being returned.
</blockquote></details>

Done.

---

*[builder/builder.go, line 190 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzNZu5bOc3X9EAE0bTb:-KzPHJngO2RKjrkxTh4o:b-896fix) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/builder/builder.go#L190)):*

same comment, log a message + the error
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

erwinvaneyk:
Review status: 9 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[hack/deploy.sh, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_smvOCMmQKxOiFSF:-KzPLKeVe1jC5iXJtwh2:b-896fix) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/hack/deploy.sh#L1)):*

It indeed is a flyby as specified in the main post. Will pull it out of this PR.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

soamvasani:
Reviewed 1 of 9 files at r1, 2 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[builder/builder.go, line 184 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzNZfk4vOZCl3_FwEno:-KzPT0ZGAFBEHewv21xA:bv4q9fb) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/builder/builder.go#L184)):*

Done.
</blockquote></details>

Nitpick, but the var name `msg` suggests a string, while this is actually an error object.

---

*[builder/cmd/builder, line 0 at r5](https://reviewable.io:443/reviews/fission/fission/397#-KzQUgbTrtbEVY63id7X:-KzQUgbTrtbEVY63id7Y:b3wvshl) ([raw file](https://github.com/fission/fission/blob/a2a8a7e4ce443a509bc0e221ae1a50e7b2cc545d/builder/cmd/builder#L0)):*
unintentional checkin?

---

*[buildermgr/envwatcher.go, line 452 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_7X0B8Rasw7WsU3g:-KzPTIfKyfFML7ZDIYYU:b-76cr32) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/envwatcher.go#L452)):*

An env var or we could specify it in the env.Spec.Builder?
</blockquote></details>

Either will work but I'd like to keep things consistent with other places in the project, such as fetcher. I suggest using an env var right now and adding it to the spec in a separate change.

---

*[buildermgr/pkgwatcher.go, line 85 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_RNN8JgZp-Zter1L:-KzPTfaBFi20SMH5arI8:b-it55tb) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/pkgwatcher.go#L85)):*

Will try to avoid it in future prs :+1:
</blockquote></details>

Thanks. But please remove this one too.

---

*[charts/fission-all-0.4.0rc.tgz, line 0 at r4](https://reviewable.io:443/reviews/fission/fission/397#-KzPTjY6fqcyLYjrb0D8:-KzPTjY7JZX75qYQKiay:ba8u56v) ([raw file](https://github.com/fission/fission/blob/188df1e6e7068fe01d752f5402c43c7ab63e3a4b/charts/fission-all-0.4.0rc.tgz#L0)):*
Unintentionally checked in?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

erwinvaneyk:
Review status: 8 of 9 files reviewed at latest revision, 6 unresolved discussions.

---

*[buildermgr/envwatcher.go, line 452 at r2](https://reviewable.io:443/reviews/fission/fission/397#-KzN_7X0B8Rasw7WsU3g:-KzQ_ZB-8dPRIoUrSUCb:b-n7vo20) ([raw file](https://github.com/fission/fission/blob/84b3bb93813c4d51b56b194cace814d9e9553141/buildermgr/envwatcher.go#L452)):*

Either will work but I'd like to keep things consistent with other places in the project, such as fetcher. I suggest using an env var right now and adding it to the spec in a separate change.
</blockquote></details>

For the sake of getting this PR merged soonish, I want to revert it back to pullAlways, and deferring the pull policy change to a later PR. Agreed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397)*

soamvasani:
LGTM

---

Reviewed 5 of 5 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/397#-:-KzSYtSujG4wd0oJ0zJy:bb6ywmx)*

PR #399, opened 2017-11-15
Functions have access to secrets/configmaps specified by the user

prithviramesh:
This PR focuses on allowing functions to access certain secrets/configmaps specified by the user. The user can either specify these through CLI arguments while creating the function or edit and append the secrets/configmaps to the function yaml after function creation.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/399)


vishal-biyani:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[environments/fetcher/fetcher.go, line 249 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZeeo8fYJFDkuAq-Y9:-KzZeeo8fYJFDkuAq-YA:b-w0z6rw) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L249)):*
> ```Go
> 	log.Printf("Checking secrets/cfgmaps")
> 	if len(req.SecretList) > 0 {
> 		log.Println("writing secrets to file")
> ```

Sorry but this won't give us much useful information since there will be multiple secrets and it will just repeat.  Suggest removing/adding more useful info to it, same with previous log line

---

*[environments/fetcher/fetcher.go, line 253 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZf2nStocrG0h5E85d:-KzZf2nTRzoZFVHIoBdK:b2ngtl1) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L253)):*
> ```Go
> 		for _, secret := range req.SecretList {
> 
> 			if len(secret.Name) == 0 {
> ```

Should this be considered an error?

---

*[environments/fetcher/fetcher.go, line 267 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZdLXo7hGYqdKiRlCg:-KzZdLXo7hGYqdKiRlCh:b-ukr78z) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L267)):*
> ```Go
> 			secretPath := secret.Namespace + "/" + secret.Name
> 			secretDir := filepath.Join(fetcher.sharedSecretPath, secretPath)
> 			err = os.MkdirAll(secretDir, 0777)
> ```

I would give the dir more restricted mod like 0644

---

*[environments/fetcher/fetcher.go, line 305 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZgpdxqDSbeYXM01Zn:-KzZgpdxqDSbeYXM01Zo:b-ps8ee) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L305)):*
> ```Go
> 				return
> 			}
> ```

Probably this whole block below this line can be put in a separate method which takes config and data.Data as input since it is common for both secret and configmap. Separates writing to file in a separate reusable block

---

*[environments/fetcher/cmd/main.go, line 23 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZhhFy2a4aCaDGvRmE:-KzZhhFy2a4aCaDGvRmF:bp1hl90) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L23)):*
> ```Go
> 	}
> 
> 	secret_dir := os.Args[2]
> ```

This will work for now, but since newdeploy backend is using flags, it might be good idea to switch to flag.Arg(2) and so on

---

*[environments/fetcher/cmd/main.go, line 32 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZih0cZ4F2oI18tEs1:-KzZih0cZ4F2oI18tEs2:b-wdzfvw) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L32)):*
> ```Go
> 		}
> 	}
> ```

To me it feels like a good idea to move creation of dir, secret_dir and config_dir inside MakeFetcher - that ways it initializes all directories and clients needed as part of creating Fetcher. This is too much boilerplate outside right now.

---

*[environments/fetcher/cmd/main.go, line 43 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZj0Skv176KE9lB7Ju:-KzZj0Skv176KE9lB7Jv:bvcffay) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L43)):*
> ```Go
> 	}
> 
> 	fetcher := fetcher.MakeFetcher(dir, secret_dir, config_dir)
> ```

nit pick - dir => func_dir (Also Go follows a camel case so funcDir may be)

---

*[fission/function.go, line 249 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZjZDFxpNHqdpAoTdW:-KzZjZDFxpNHqdpAoTdX:b68dfwb) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/fission/function.go#L249)):*
> ```Go
> 			},
> 
> 			SecretList: []fission.SecretReference{},
> ```

Just for my understanding - are we using a [] so that in future we could support multiple secrets?

---

*[fission/function.go, line 393 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZjkpssigzJvYnA0sG:-KzZjkpssigzJvYnA0sH:bk7dmso) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/fission/function.go#L393)):*
> ```Go
> 	}
> 
> 	/*secretName := c.String("secret")
> ```

This looks like commented to me - are we keeping for future enhancement?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

vishal-biyani:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[environments/fetcher/fetcher.go, line 267 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZdLXo7hGYqdKiRlCg:-Kz_Ss7xX287FeTZEPEr:bio9zpp) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L267)):*
> 0644
I meant 0700 like the previous one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

prithviramesh:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[environments/fetcher/fetcher.go, line 249 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZeeo8fYJFDkuAq-Y9:-KzqXirRLFvddwvP9vYF:bo8s0di) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L249)):*

Sorry but this won't give us much useful information since there will be multiple secrets and it will just repeat.  Suggest removing/adding more useful info to it, same with previous log line
</blockquote></details>

Removed both

---

*[environments/fetcher/fetcher.go, line 253 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZf2nStocrG0h5E85d:-KzqXnlQ7vCOwKiYncnH:bquw0iw) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L253)):*

Should this be considered an error?
</blockquote></details>

Removed, it was left from an older commit. No need anymore

---

*[environments/fetcher/fetcher.go, line 267 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZdLXo7hGYqdKiRlCg:-KzqXy7i1eFf8CNVvSU0:b-c3opkl) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L267)):*

> 0644
I meant 0700 like the previous one.
</blockquote></details>

New MakeDir method in Fetcher sets it to 0700

---

*[environments/fetcher/fetcher.go, line 305 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZgpdxqDSbeYXM01Zn:-KzqY2BUnpQva2-OmY8e:b-tf8sbd) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/fetcher.go#L305)):*

Probably this whole block below this line can be put in a separate method which takes config and data.Data as input since it is common for both secret and configmap. Separates writing to file in a separate reusable block
</blockquote></details>

Finished this, method is called WriteSecretOrConfigMap 

---

*[environments/fetcher/cmd/main.go, line 32 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZih0cZ4F2oI18tEs1:-KzqYALGGFjDvd5hNeIJ:bb6un1j) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L32)):*

To me it feels like a good idea to move creation of dir, secret_dir and config_dir inside MakeFetcher - that ways it initializes all directories and clients needed as part of creating Fetcher. This is too much boilerplate outside right now.
</blockquote></details>

Done

---

*[environments/fetcher/cmd/main.go, line 43 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZj0Skv176KE9lB7Ju:-KzqYBMCSlSOPwUg_Jn2:bb6un1j) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L43)):*

nit pick - dir => func_dir (Also Go follows a camel case so funcDir may be)
</blockquote></details>

Done

---

*[fission/function.go, line 249 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZjZDFxpNHqdpAoTdW:-KzqYFmq_Ug6s8My1slc:b-tuik1l) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/fission/function.go#L249)):*

Just for my understanding - are we using a [] so that in future we could support multiple secrets?
</blockquote></details>

Exactly

---

*[fission/function.go, line 393 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZjkpssigzJvYnA0sG:-KzqYH0omBcbU3RXylEV:ba3dfob) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/fission/function.go#L393)):*

This looks like commented to me - are we keeping for future enhancement?
</blockquote></details>

Removed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

prithviramesh:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[environments/fetcher/cmd/main.go, line 23 at r2](https://reviewable.io:443/reviews/fission/fission/399#-KzZhhFy2a4aCaDGvRmE:-KztwqQLqF4Di_XXgcOo:bb6un1j) ([raw file](https://github.com/fission/fission/blob/dd571598d6c18e119155cbf73fcd82efcf42b76f/environments/fetcher/cmd/main.go#L23)):*

This will work for now, but since newdeploy backend is using flags, it might be good idea to switch to flag.Arg(2) and so on
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

soamvasani:
Reviewed 1 of 7 files at r3.
Review status: 1 of 11 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[types.go, line 125 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-49TTESD2_xI0zJ4jx:-L-49TTESD2_xI0zJ4jy:b2khr70) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/types.go#L125)):*
> ```Go
> 		Package FunctionPackageRef `json:"package"`
> 
> 		SecretList    []SecretReference    `json:"secretList"`
> ```

Add a comment please. Also how about just `secrets` for the json name.

---

*[types.go, line 126 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-49ab6AqqkiBSmoGat:-L-49ab6AqqkiBSmoGau:bjmvupo) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/types.go#L126)):*
> ```Go
> 
> 		SecretList    []SecretReference    `json:"secretList"`
> 		ConfigMapList []ConfigMapReference `json:"configMapList"`
> ```

and maybe just `configmaps` for the json name.

---

*[environments/fetcher/fetcher.go, line 158 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-4Ea966ji_wEGYAun8:-L-4Ea966ji_wEGYAun9:bfpshfc) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/environments/fetcher/fetcher.go#L158)):*
> ```Go
> }
> 
> func writeSecretOrConfigMap(dataMap interface{}, dirPath string, w http.ResponseWriter) {
> ```

as we discussed, it'll be a lot easier to read if this just takes a map[string][]byte

---

*[environments/fetcher/fetcher.go, line 291 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-4EOKmH6hx0Rv_P2e4:-L-4EOKmH6hx0Rv_P2e5:b-r8wte3) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/environments/fetcher/fetcher.go#L291)):*
> ```Go
> 	}
> 
> 	log.Printf("Checking secrets/cfgmaps")
> ```

add functionname etc to the log

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

prithviramesh:
Review status: 0 of 11 files reviewed at latest revision, 13 unresolved discussions.

---

*[types.go, line 125 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-49TTESD2_xI0zJ4jx:-L-4fwo0yKX1lpmeCm8z:bb6un1j) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/types.go#L125)):*

Add a comment please. Also how about just `secrets` for the json name.
</blockquote></details>

Done

---

*[types.go, line 126 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-49ab6AqqkiBSmoGat:-L-4fxQ0VOd0foepcLB6:bb6un1j) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/types.go#L126)):*

and maybe just `configmaps` for the json name.
</blockquote></details>

Done

---

*[environments/fetcher/fetcher.go, line 158 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-4Ea966ji_wEGYAun8:-L-4fymPjzcJpHGakjEB:b-i2oy1h) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/environments/fetcher/fetcher.go#L158)):*

as we discussed, it'll be a lot easier to read if this just takes a map[string][]byte
</blockquote></details>

Done (begrudgingly :sob: )

---

*[environments/fetcher/fetcher.go, line 291 at r4](https://reviewable.io:443/reviews/fission/fission/399#-L-4EOKmH6hx0Rv_P2e4:-L-4g6Y7Qr1ExgSdz9DC:b3lma1i) ([raw file](https://github.com/fission/fission/blob/d2c4c69e14eb057a0eb0220fbffd97bbca56ab43/environments/fetcher/fetcher.go#L291)):*

add functionname etc to the log
</blockquote></details>

As discussed, will create a new PR to add functionname to the fetch request, kept as is for now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399)*

soamvasani:
This needs a rebase and a final review of the API change.

life1347:
I'll take over this PR.

soamvasani:
Thanks for taking this up!!  Looks good to me except for some minor comments. After they are addressed please feel free to merge without waiting for me.

---

Reviewed 16 of 16 files at r7.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[types.go, line 132 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZS3Oz_-Mfn96Smdzw:-L4ZS3P-l6pgEduqakjs:b3ged0w) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/types.go#L132)):*
> ```Go
> 		Package FunctionPackageRef `json:"package"`
> 
> 		SecretList    []SecretReference    `json:"secrets"`
> ```

Field names should also be changed, to `Secrets` and `ConfigMaps`.

---

*[environments/fetcher/fetcher.go, line 37 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZSmGgiusmC8VnY0YX:-L4ZSmGgiusmC8VnY0YY:btc0hi7) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/environments/fetcher/fetcher.go#L37)):*
> ```Go
> 		StorageSvcUrl string                       `json:"storagesvcurl"`
> 		Filename      string                       `json:"filename"`
> 		SecretList    []fission.SecretReference    `json:"secretList"`
> ```

Please call this Secrets, ConfigMaps

---

*[environments/fetcher/fetcher.go, line 70 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZTBB4EfKHEnh-z0Kq:-L4ZSy1PnylmoaGHwVNo:b-guomx5) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/environments/fetcher/fetcher.go#L70)):*
> ```Go
> )
> 
> func makeDir(dirPath string, perm os.FileMode) error {
> ```

This function can be removed, because it is equivalent to simply `os.MkdirAll`, because MkdirAll already returns nil (no error) if the directory exists (https://golang.org/pkg/os/#MkdirAll)

---

*[environments/fetcher/fetcher.go, line 322 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZUOVD9qxWTkddaFcV:-L4ZUOVEXcH-JhgR7yvg:bqfczyl) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/environments/fetcher/fetcher.go#L322)):*
> ```Go
> 			}
> 
> 			secretPath := secret.Namespace + "/" + secret.Name
> ```

Please use filepath.Join

---

*[environments/fetcher/fetcher.go, line 353 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZUqDQ0_N2m51LSgdS:-L4ZUqDQ0_N2m51LSgdT:bqfczyl) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/environments/fetcher/fetcher.go#L353)):*
> ```Go
> 			}
> 
> 			configPath := config.Namespace + "/" + config.Name
> ```

Please use filepath.Join

---

*[fission/function.go, line 132 at r7](https://reviewable.io:443/reviews/fission/fission/399#-L4ZWLVREAh8SIBN1c-z:-L4ZWLVSkGu3pJbm8mZU:b-z8ecjf) ([raw file](https://github.com/fission/fission/blob/77df5076906ae3981699a20d145d3a9eefa8aeeb/fission/function.go#L132)):*
> ```Go
> 	cfgMapName := c.String("configmap")
> 
> 	secretNameSpace := c.String("secretns")
> ```

Sorry for the tiniest nitpick ever, but it should be "secretNamespace", not "secretNameSpace" ("Namespace" is one word so not camel cased :) )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/399#-:-L4ZXFkcHmfKF96aLEMN:budd0ih)*

fission/main.go
```diff
@@ -61,12 +61,16 @@ func main() {
 	fnHeaderFlag := cli.StringSliceFlag{Name: "header, H", Usage: "request headers"}
 	fnEntryPointFlag := cli.StringFlag{Name: "entrypoint", Usage: "entry point for environment v2 to load with"}
 	fnBuildCmdFlag := cli.StringFlag{Name: "buildcmd", Usage: "build command for builder to run with"}
+	fnSecretFlag := cli.StringFlag{Name: "secret", Usage: "function access to secret"}
+	fnSecretnsFlag := cli.StringFlag{Name: "secretns", Usage: "namespace of secret"}
```
smruthi2187: I was thinking that we may not need to provide the user an option for secret namespace and configMap namespace. Mainly because we dont provide the user an option to specify namespace for their function.
If the reason that we want to give this as an option is that the secret names and configmap names shouldnt collide, then we can probably prefix or suffix them with function names since they are unique. This is just a thought.

fission/function.go
```diff
@@ -204,11 +218,29 @@ func fnCreate(c *cli.Context) error {
 					ResourceVersion: pkgMetadata.ResourceVersion,
 				},
 			},
+			SecretList:     []fission.SecretReference{},
```
smruthi2187: I was wondering in what scenario will a function have more than one secrets or more than one configMaps.
It seems like we are not allowing CRUD on secrets and configMaps, instead, we are only allowing Create operation on them both. 
So, we may not need to have a list of secrets and list of configMaps. right?

environments/fetcher/cmd/main.go
```diff
@@ -20,9 +20,12 @@ import (
 // Usage: fetcher <shared volume path>
 func main() {
 	flag.Usage = fetcherUsage
+	specializeOnStart := flag.Bool("specialize-on-startup", false, "Flag to activate specialize process at pod starup")
 	fetchPayload := flag.String("fetch-request", "", "JSON Payload for fetch request")
 	loadPayload := flag.String("load-request", "", "JSON payload for Load request")
-	specializeOnStart := flag.Bool("specialize-on-startup", false, "Flag to activate specialize process at pod starup")
+	secretDir := flag.String("secret-dir", "", "Path to shared secrets directory")
```
smruthi2187: One question here. Choosing the path for mounting secrets and configMaps in the fetcher means that all functions are forced to use the same path for their secrets and configs. right?
There may be a few users who would like to choose a specific path for their secrets and configMaps during `fn create --secret <name> --secretPath </foo/bar>`. 
Would it be a good idea to not make these volumes at the startup of fetcher, instead, create them during `specializePod` function? 
I do understand that this approach may slow down the start of function a bit, but, i think it might be worth a tradeoff of speed for flexibility here. 
This is just a though. I'm willing to hear what you all have to say.

fission/function.go
life1347: It is intentional. Though CLI only accept one secret/configmap. For advance user, they can use YAML to declare multiple secrets/configmaps (PR422). I will write down the constraint in the tutorial.

PR #408, opened 2017-11-26
Prepare Fission for IPv6 uses

valentin2105:
Hi,

Here is a small contribution to allow Fission to talk to IPv6 function pods. 

I'm able to make all work except to communications to IPv6 function pod : 

poolmgr logs : 
```
 Error: 500: Post http://1404:f600:f::0:e:16:d0bb:8000/: invalid URL port "404:f600:f::0:e:16:d0bb"
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/408)

### DIFF
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -309,15 +309,15 @@ func (gp *GenericPool) getSpecializeUrl(podIP string, version int) string {
 	}
 	t := IsIPv6(podIP)
 	if t == false {
-		if version == 1 {
-			return fmt.Sprintf("http://%v:8888/specialize", podIP)
-		}
-		return fmt.Sprintf("http://%v:8888/v%v/specialize", podIP, version)
-	} else if t == true {
-		if version == 1 {
-			return fmt.Sprintf("http://[%v]:8888/specialize", podIP)
-		}
-		return fmt.Sprintf("http://[%v]:8888/v%v/specialize", podIP, version)
+		r := fmt.Sprintf("http://%v:8888", podIP)
+	} else if t == true { //Use bracket if IP == IPv6
+		r := fmt.Sprintf("http://[%v]:8888", podIP)
+	}
+
+	if version == 1 {
+	 	return fmt.Sprintf("%v/specialize", r)
+	} else {
+	 	return fmt.Sprintf("%v/v%v/specialize", r, version)
 	}
 }
 
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -285,6 +285,12 @@ func (gp *GenericPool) scheduleDeletePod(name string) {
 	}()
 }
 
+func IsIPv6(podIP string) bool {
+	ip := net.ParseIP(podIP)
+	return ip != nil && strings.Contains(podIP, ":")
+}
+
+
 func (gp *GenericPool) getFetcherUrl(podIP string) string {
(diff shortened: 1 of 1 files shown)


executor/poolmgr/gp.go
```diff
@@ -297,15 +297,28 @@ func (gp *GenericPool) getFetcherUrl(podIP string) string {
...
+}
+
 func (gp *GenericPool) getSpecializeUrl(podIP string, version int) string {
 	u := os.Getenv("TEST_SPECIALIZE_URL")
 	if len(u) != 0 {
 		return u
 	}
-	if version == 1 {
-		return fmt.Sprintf("http://%v:8888/specialize", podIP)
+	t := IsIPv6(podIP)
```
soamvasani: Thanks, @valentin2105 !

May I suggest reorganizing this if-else to set the host part of the URL separately from the version check? That way you won't need the nested if-s.

valentin2105:
@soamvasani Thanks for the review. 

It seem better for you ?

executor/poolmgr/gp.go
```diff
@@ -297,15 +297,28 @@ func (gp *GenericPool) getFetcherUrl(podIP string) string {
...
+}
+
 func (gp *GenericPool) getSpecializeUrl(podIP string, version int) string {
 	u := os.Getenv("TEST_SPECIALIZE_URL")
 	if len(u) != 0 {
 		return u
 	}
+	t := IsIPv6(podIP)
+	if t == false {
+		r := fmt.Sprintf("http://%v:8888", podIP)
```
soamvasani: This defines `r` only inside this block, so line 318 below causes a build failure. I think you'll have to define `r` as a var outside the if statement. 

(Also, what does `r` stand for? Maybe use just `baseUrl` or something? And could you also rename `t` to something like `v6`?)

valentin2105:
@soamvasani It should be better ;)

executor/poolmgr/gp.go
```diff
@@ -297,15 +297,29 @@ func (gp *GenericPool) getFetcherUrl(podIP string) string {
...
+
 func (gp *GenericPool) getSpecializeUrl(podIP string, version int) string {
 	u := os.Getenv("TEST_SPECIALIZE_URL")
+	isv6 := IsIPv6(podIP)
+	var baseUrl string
 	if len(u) != 0 {
 		return u
 	}
+	if isv6 == false {
+		baseUrl := fmt.Sprintf("http://%v:8888", podIP)
```
soamvasani: subtle thing, but this should be `baseUrl = ...` not `baseUrl := ...`. With the `:=` operator you're creating a new variable inside this scope and hiding the baseUrl variable defined on line 308.

valentin2105:
I have made the correction, but the tests fail again.

valentin2105:
@soamvasani  It seem ok for you ?

executor/poolmgr/gp.go
```diff
@@ -297,15 +297,29 @@ func (gp *GenericPool) getFetcherUrl(podIP string) string {
 	return fmt.Sprintf("http://%v:8000/", podIP)
```
soamvasani: This needs the same change?

valentin2105:
@soamvasani It seem good 😄  I always forget to FMT 🤣 .

soamvasani:
👍 thanks @valentin2105 !

Issue #410, opened 2017-11-27
Glide Update breaks fission-bundle compilation 


prithviramesh:
Updating the nats-streaming-server package to the latest version causes the compilation to fail; the method "IsSubjectValid" in "github.com/nats-io/nats-streaming-server/util" is deprecated and is now called "IsChannelNameValid"

[change made in this commit to nats-streaming-server](https://github.com/nats-io/nats-streaming-server/commit/240a2c024e1ad2abb8387b157693df7fa0a10f00#diff-99f38c7c07bfcc03a528180716bafab5)

https://godoc.org/github.com/nats-io/nats-streaming-server/util

life1347:
thanks @prithviramesh! I pushed a PR to solve this problem.

prithviramesh:
Thanks @life1347 for your prompt solution!

erwinvaneyk:
Fixed in #411

PR #411, opened 2017-11-28
Update nats dependencies

life1347:
This small PR aims to solve compilation problem after nats related dependencies being updated. (issue https://github.com/fission/fission/issues/410)

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/411)


soamvasani:
Thanks @life1347. 

@erwinvaneyk looks like this is just a function name change in the nats client.

PR #413, opened 2017-11-28
Add container spec config options  to (build) environments

erwinvaneyk:
~This PR adds the ability to add environment variables to (build) environments.~

~To keep the spec api consistent with the Kubernetes api, I choose to follow the EnvVar approach taken there using a `env` field in the object.~

~I created a separate EnvVar implementation, because I did not want to introduce a dependency on kubernetes types in the fission types, but I am open to change that. (your thoughts on this @soamvasani ?)~

This PR adds the ability for users to provide a template/container to influence the deployment of the containers. The current motivation behind this is to allow the setting of environment variables in the runtime environment containers.

I have been testing quite a bit (due to the incompatible client issue), and in the process added a script to run a single e2e test on your current cluster. Additionally, this PR updates the compiling.md document to address the need for nested vendor exclusion.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/413)

### DIFF
--- a/common.go
+++ b/common.go
@@ -69,6 +69,10 @@ func LoggingMiddleware(next http.Handler) http.Handler {
 	})
 }
 
+// MergeContainerSpecs merges container specs using a predefined order.
+//
+// The order of the arguments indicates which spec has precedence (lower index takes precedence over higher indexes).
+// Slices and maps are merged; other fields are set only if they are a zero value.
 func MergeContainerSpecs(specs ...*apiv1.Container) apiv1.Container {
 	result := &apiv1.Container{}
 	for _, spec := range specs {
--- a/common.go
+++ b/common.go
@@ -27,6 +27,8 @@ import (
 	"syscall"
 
 	"github.com/gorilla/handlers"
+	"github.com/imdario/mergo"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
 )
 
 func UrlForFunction(name string) string {
@@ -65,4 +67,19 @@ func LoggingMiddleware(next http.Handler) http.Handler {
 			handlers.LoggingHandler(os.Stdout, next).ServeHTTP(w, r)
 		}
 	})
+}
+
+func MergeContainerSpecs(specs ...*apiv1.Container) apiv1.Container {
+	result := &apiv1.Container{}
+	for _, spec := range specs {
+		if spec == nil {
+			continue
+		}
+
+		err := mergo.Merge(result, spec)
+		if err != nil {
+			panic(err)
+		}
+	}
+	return *result
 }
--- a/controller/api_test.go
+++ b/controller/api_test.go
@@ -23,6 +23,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"reflect"
 	"testing"
 	"time"
 
@@ -32,7 +33,6 @@ import (
 	"github.com/fission/fission"
 	"github.com/fission/fission/controller/client"
 	"github.com/fission/fission/crd"
-	"reflect"
 )
 
 var g struct {
@@ -263,9 +263,9 @@ func TestWatchApi(t *testing.T) {
 
 	w, err := g.client.WatchGet(m)
 	panicIf(err)
-	assert((testWatch.Spec.Namespace == w.Spec.Namespace &&
+	assert(testWatch.Spec.Namespace == w.Spec.Namespace &&
 		testWatch.Spec.Type == w.Spec.Type &&
-		testWatch.Spec.FunctionReference == w.Spec.FunctionReference), "watch should match after reading")
+		testWatch.Spec.FunctionReference == w.Spec.FunctionReference, "watch should match after reading")
 
 	testWatch.Metadata.Name = "yyy"
 	m2, err := g.client.WatchCreate(testWatch)
--- a/controller/api_test.go
+++ b/controller/api_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/fission/fission"
 	"github.com/fission/fission/controller/client"
 	"github.com/fission/fission/crd"
+	"reflect"
 )
 
 var g struct {
@@ -214,11 +215,7 @@ func TestEnvironmentApi(t *testing.T) {
 
 	e, err := g.client.EnvironmentGet(m)
 	panicIf(err)
-	assert(testEnv.Spec.AllowedFunctionsPerContainer == e.Spec.AllowedFunctionsPerContainer, "env AllowedFunctionsPerContainer should match after reading")
-	assert(testEnv.Spec.Poolsize == e.Spec.Poolsize, "env Poolsize should match after reading")
-	assert(testEnv.Spec.Builder == e.Spec.Builder, "env Builder should match after reading")
-	assert(testEnv.Spec.Runtime == e.Spec.Runtime, "env Runtime should match after reading")
-	assert(testEnv.Spec.Version == e.Spec.Version, "env Version should match after reading")
+	assert(reflect.DeepEqual(testEnv.Spec, e.Spec), "env should match after reading")
 
 	testEnv.Metadata.ResourceVersion = m.ResourceVersion
 	testEnv.Spec.Runtime.Image = "another-img"
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -524,7 +524,7 @@ func (gp *GenericPool) createPool() error {
 						},
 					},
 					Containers: []apiv1.Container{
-						{
+						fission.MergeContainerSpecs(gp.env.Spec.Runtime.Container, &apiv1.Container{
 							Name:                   gp.env.Metadata.Name,
 							Image:                  gp.env.Spec.Runtime.Image,
 							ImagePullPolicy:        gp.runtimeImagePullPolicy,
@@ -560,8 +560,7 @@ func (gp *GenericPool) createPool() error {
 									},
 								},
 							},
-							Env: gp.env.Spec.Runtime.Env,
-						},
+						}),
 						{
 							Name:                   "fetcher",
 							Image:                  gp.fetcherImage,
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -476,8 +476,6 @@ func (gp *GenericPool) createPool() error {
 	if err != nil {
 		return err
 	}
-	env := fission.K8sEnvVars(gp.env.Spec.Runtime.Env)
-	log.Printf("Env: %v", env)
 
 	// Use long terminationGracePeriodSeconds for connection draining in case that
 	// pod still runs user functions.
@@ -562,7 +560,7 @@ func (gp *GenericPool) createPool() error {
 									},
 								},
 							},
-							Env: env,
+							Env: gp.env.Spec.Runtime.Env,
 						},
 						{
 							Name:                   "fetcher",
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -524,7 +524,7 @@ func (gp *GenericPool) createPool() error {
 						},
 					},
 					Containers: []apiv1.Container{
-						fission.MergeContainerSpecs(gp.env.Spec.Runtime.Container, &apiv1.Container{
+						fission.MergeContainerSpecs(&apiv1.Container{
 							Name:                   gp.env.Metadata.Name,
 							Image:                  gp.env.Spec.Runtime.Image,
 							ImagePullPolicy:        gp.runtimeImagePullPolicy,
@@ -560,7 +560,7 @@ func (gp *GenericPool) createPool() error {
 									},
 								},
 							},
-						}),
+						}, gp.env.Spec.Runtime.Container),
 						{
 							Name:                   "fetcher",
 							Image:                  gp.fetcherImage,
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -476,6 +476,8 @@ func (gp *GenericPool) createPool() error {
 	if err != nil {
 		return err
 	}
+	env := fission.K8sEnvVars(gp.env.Spec.Runtime.Env)
+	log.Printf("Env: %v", env)
 
 	// Use long terminationGracePeriodSeconds for connection draining in case that
 	// pod still runs user functions.
@@ -560,6 +562,7 @@ func (gp *GenericPool) createPool() error {
 									},
 								},
 							},
+							Env: env,
 						},
 						{
 							Name:                   "fetcher",
--- a/types.go
+++ b/types.go
@@ -208,6 +208,10 @@ type (
 		// server listens for function requests. Optional;
 		// default 8888.
 		FunctionEndpointPort int32 `json:"functionendpointport"`
+
+		// Optional
+		// Environment variables to be set in environment
+		Env []EnvVar `json:"env"`
 	}
 	Builder struct {
 		// Image for containing the language runtime.
@@ -215,6 +219,10 @@ type (
 
 		// (Optional) Default build command to run for this build environment.
 		Command string `json:"command,omitempty"`
+
+		// Optional
+		// Environment variables to be set in environment
+		Env []EnvVar `json:"env"`
 	}
 	EnvironmentSpec struct {
 		// Environment API version
@@ -241,6 +249,10 @@ type (
 
 		// The initial pool size for environment
 		Poolsize int `json:"poolsize,omitempty"`
+	}
+	EnvVar struct {
+		Name  string `json:"name"`
+		Value string `json:"value"`
 	}
 
 	AllowedFunctionsPerContainer string
--- a/types.go
+++ b/types.go
@@ -209,9 +209,15 @@ type (
 		// default 8888.
 		FunctionEndpointPort int32 `json:"functionendpointport"`
 
-		// Env is a list of environment variables to be set
-		// in the runtime environment. (optional)
-		Env []apiv1.EnvVar `json:"env"`
+		// Container allows the modification of the deployed runtime
+		// container using the Kubernetes Container spec. Fission overrides
+		// the following fields:
+		// - Name
+		// - Image; set to the Runtime.Image
+		// - TerminationMessagePath
+		// - ImagePullPolicy
+		// (optional)
+		Container *apiv1.Container `json:"container,omitempty"`
 	}
 	Builder struct {
 		// Image for containing the language runtime.
@@ -220,9 +226,17 @@ type (
 		// (Optional) Default build command to run for this build environment.
 		Command string `json:"command,omitempty"`
 
-		// Env is a list of environment variables to be set
-		// in the build environment. (optional)
-		Env []apiv1.EnvVar `json:"env"`
+		// Container allows the modification of the deployed builder
+		// container using the Kubernetes Container spec. Fission overrides
+		// the following fields:
+		// - Name
+		// - Image; set to the Builder.Image
+		// - Command; set to the Builder.Command
+		// - TerminationMessagePath
+		// - ImagePullPolicy
+		// - ReadinessProbe
+		// (optional)
+		Container *apiv1.Container `json:"container,omitempty"`
 	}
 	EnvironmentSpec struct {
 		// Environment API version
--- a/types.go
+++ b/types.go
@@ -18,7 +18,7 @@ package fission
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/pkg/api/v1"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
 )
 
 type (
@@ -133,7 +133,7 @@ type (
 		ConfigMaps []ConfigMapReference `json:"configmaps"`
 
 		// cpu and memory resources as per K8S standards
-		Resources v1.ResourceRequirements `json:"resources"`
+		Resources apiv1.ResourceRequirements `json:"resources"`
 
 		// InvokeStrategy is a set of controls which affect how function executes
 		InvokeStrategy InvokeStrategy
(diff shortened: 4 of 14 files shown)


soamvasani:
Can you describe the use case you're intending to address with this?

erwinvaneyk:
Hey @soamvasani - it currently is specific to my use case. Here: https://github.com/fission/fission-workflows/blob/master/charts/fission-workflows/templates/deployment.yaml#L64 I need to be able to pass in some env variables related to NATS and Fission. Before these were hardcoded into the docker image of the Fission Workflows environment, which is not.. great. You would have to hardcode the key of NATS into docker container for example and you couldn't run your fission installation in another place other than the one specified in the docker image (like 
here https://github.com/fission/fission-workflows/blob/master/build/env/Dockerfile).

Beyond workflows, I can image this will be useful in any situation in which you want to add some configurability to your environment.

types.go
```diff
@@ -175,6 +183,10 @@ type (
 		// Defaults to 'Single'
 		AllowedFunctionsPerContainer AllowedFunctionsPerContainer `json:"allowedFunctionsPerContainer"`
 	}
+	EnvVar struct {
```
soamvasani: Hmmm, let's not do this.  I get why you're trying to avoid pulling in the whole k8s API package in here, but I think that's better than duplicating little pieces of it in incomplete ways. For example this is missing `EnvVarSource`, which is in fact the right way to solve the problem you're trying to solve (because it handles secrets correctly).  So we should just use an actual K8s EnvVar slice and copy it into pod spec's container spec.

erwinvaneyk: Yeah I am also okay with that. 👍

soamvasani: 👍 thanks!

types.go
```diff
@@ -149,13 +150,21 @@ type (
...
 	}
 	Builder struct {
 		// Image for containing the language runtime.
 		Image string `json:"image"`
 
 		// (Optional) Default build command to run for this build environment.
 		Command string `json:"command"`
+
+		// Optional
+		// Environment variables to be set in environment
```
soamvasani: same as above

erwinvaneyk:
@soamvasani - addressed your comments

soamvasani:
Change LGTM, @erwinvaneyk please look into the test failures; @life1347 please merge when ready.

soamvasani:
We had a few conversations about this elsewhere, and I'm wary of adding the whole contents of the Container into this spec piece by piece.

I think we can instead allow the user to supply a full Kubernetes `Container` struct in the builder and runtime env specs; then fission would place that in the PodSpecs, with the shared volumes added.

erwinvaneyk:
Discussed with @soamvasani 

> [...] main concern is users not being able to figure out how their provided container spec changes. If it can be neatly documented that might be good enough.

For now this means the following: 
- [ ] Add (detailed) documentation to types.go

erwinvaneyk:
@soamvasani please review; naming wise I went for the most obvious name: `container`. We could also go for `template`, to make it more consistent with the Kubernetes naming (although it might lead to more confusion?).

common.go
```diff
@@ -18,6 +18,8 @@ package fission
 
 import (
 	"fmt"
+	"github.com/imdario/mergo"
+	apiv1 "k8s.io/client-go/pkg/api/v1"
```
vishal-biyani: Nitpick, formatting the imports with the convention used elsewhere.

vishal-biyani: I am sorry, this looks odd here, but is correctly formatted when I cross-checked in IDE locally. Please ignore.

controller/api_test.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/fission/fission"
 	"github.com/fission/fission/controller/client"
 	"github.com/fission/fission/crd"
+	"reflect"
```
vishal-biyani: nit: Would be good to move this with other system-level imports instead with Fission imports

controller/api_test.go
```diff
@@ -214,11 +215,7 @@ func TestEnvironmentApi(t *testing.T) {
 
 	e, err := g.client.EnvironmentGet(m)
 	panicIf(err)
-	assert(testEnv.Spec.AllowedFunctionsPerContainer == e.Spec.AllowedFunctionsPerContainer, "env AllowedFunctionsPerContainer should match after reading")
-	assert(testEnv.Spec.Poolsize == e.Spec.Poolsize, "env Poolsize should match after reading")
-	assert(testEnv.Spec.Builder == e.Spec.Builder, "env Builder should match after reading")
-	assert(testEnv.Spec.Runtime == e.Spec.Runtime, "env Runtime should match after reading")
-	assert(testEnv.Spec.Version == e.Spec.Version, "env Version should match after reading")
+	assert(reflect.DeepEqual(testEnv.Spec, e.Spec), "env should match after reading")
```
vishal-biyani: Suggest using google/go-cmp as a safer alternative to `reflect.DeepEqual`, may be as a separate issue if you would like

executor/poolmgr/gp.go
```diff
@@ -471,7 +471,7 @@ func (gp *GenericPool) createPool() error {
 						},
 					},
 					Containers: []apiv1.Container{
-						{
+						fission.MergeContainerSpecs(gp.env.Spec.Runtime.Container, &apiv1.Container{
```
vishal-biyani: Should a similar merge not be applied to `newdeploy` environment in the specs, or am I missing something?

controller/api_test.go
erwinvaneyk: I didn't know about google/go-cmp. How is it different/safer from the reflect approach? Anyway, for the sake of avoiding too much lingering, let's defer this to a later pr.

executor/poolmgr/gp.go
erwinvaneyk: Done 👍

erwinvaneyk:
Review status: 0 of 13 files reviewed at latest revision, 7 unresolved discussions.

---

*[common.go, line 22 at r7](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0I-r7-22:-L6MXAKZvNQKKqGOOPs9:b-896fix) ([raw file](https://github.com/fission/fission/blob/bf70a03095eeb889f8f65d9af410fedc7d79969e/common.go#L22)):*

I am sorry, this looks odd here, but is correctly formatted when I cross-checked in IDE locally. Please ignore.
</blockquote></details>

Done.

---

*[types.go, line 155 at r3](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0L-r3-155:-L6MXB7Vu6Yh6MTbVxV0:b-896fix) ([raw file](https://github.com/fission/fission/blob/63c83c04986c8db604b2cbd233c593d2127084ca/types.go#L155)):*

Godoc convention, first word of comment should be "Env" (same as name of var).  e.g. "Env is the list of ...".  Also it's a bit confusing.  How about "Env is the list of environment variables set in the runtime fission environment process. Optional.".
</blockquote></details>

Done.

---

*[types.go, line 166 at r3](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0L-r3-166:-L6MXCbsAAGwOmBlYbRG:b-896fix) ([raw file](https://github.com/fission/fission/blob/63c83c04986c8db604b2cbd233c593d2127084ca/types.go#L166)):*

same as above
</blockquote></details>

Done.

---

*[controller/api_test.go, line 35 at r7](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0J-r7-35:-L6MXDRaKGLIe23b72eD:b-896fix) ([raw file](https://github.com/fission/fission/blob/bf70a03095eeb889f8f65d9af410fedc7d79969e/controller/api_test.go#L35)):*

nit: Would be good to move this with other system-level imports instead with Fission imports
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413)*

soamvasani:
Thanks @erwinvaneyk !  I have some thoughts on the overriding behaviour.

---

Reviewed 3 of 7 files at r6, 4 of 6 files at r7, 1 of 1 files at r8, 5 of 5 files at r9.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[common.go, line 47 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6NwWk87Vc0GIBhcFmT:-L6NwWk87Vc0GIBhcFmU:bo7obup) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/common.go#L47)):*
> ```Go
> }
> 
> func MergeContainerSpecs(specs ...*apiv1.Container) apiv1.Container {
> ```

Function comment please.  Specifically:

1. Which spec "wins"? Last one, I assume?
2. What happens to slices, like VolumeMounts? Do they get "merged" or just plain overridden?

---

*[types.go, line 155 at r3](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0L-r3-155:-L6Nwvzz0Bd4zt7TfY-x:b-jiffpi) ([raw file](https://github.com/fission/fission/blob/63c83c04986c8db604b2cbd233c593d2127084ca/types.go#L155)):*

Done.
</blockquote></details>

Thanks for the helpful comment. Now that it's in words, I actually understand the diff better.  And I think the overriding rules are not great...

If the user sets both Container.Image and Runtime.Image, maybe we should just error out. In other words we should not override Image.  And if we do override, then the one in the container seems like it should "win".

I also think user should be allowed to control ImagePullPolicy, so there's no good reason to override this.

Why do we want to override TerminationMessagePath?

---

*[buildermgr/envwatcher.go, line 452 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6O0Prkb6FXVHNmECvZ:-L6O0Prkb6FXVHNmECv_:btml9e8) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/buildermgr/envwatcher.go#L452)):*
> ```Go
> 					},
> 					Containers: []apiv1.Container{
> 						fission.MergeContainerSpecs(env.Spec.Builder.Container, &apiv1.Container{
> ```

So I think the right thing is to merge 3 specs:

Container{
    image
    image pull policy
    command
    ReadinessProbe
}

env.Spec.Builder.Container

Container{
   Name: "builder"
   VolumeMounts
}


That way the user can override everything except what fission needs to actually work.

---

*[executor/newdeploy/newdeploy.go, line 121 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6O95hnlSKt23xGvhXl:-L6O95honq4Q9BZKiwN9:bwmesx4) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/executor/newdeploy/newdeploy.go#L121)):*
> ```Go
> 						},
> 						Containers: []apiv1.Container{
> 							fission.MergeContainerSpecs(env.Spec.Runtime.Container, &apiv1.Container{
> ```

Same comments as above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413#-:-L6OAG1adt0Wq8sjsCqp:b-3c4p8j)*

erwinvaneyk:
Review status: 6 of 13 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[common.go, line 47 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6NwWk87Vc0GIBhcFmT:-L6SAlZKJeEC9fM0ikTD:b2c8v19) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/common.go#L47)):*

Function comment please.  Specifically:

1. Which spec "wins"? Last one, I assume?
2. What happens to slices, like VolumeMounts? Do they get "merged" or just plain overridden?
</blockquote></details>

This indeed needs some documentation to explain the behavior for the situations you describe. Currently, the 'first' wins. If a field is set to a non-zero value, it is not overriden by other specs. Slices and maps are a special case: the contents of specs are concatenated together. 

We probably need to add a couple of edge cases (e.g. you probably want to have the command or args field overriden rather than merged. Looking at the source of the Container spec, the fields are annotated how to be merged (using the JSON patch specification). Unfortunately, there is no implementation that allows merging structs using these annotations as far as I can find. Kubernetes has an util library with a forked json library called 'strategicpatch' that implements the JSON patch spec, but only for JSON documents. So we are getting deeper and deeper into the rabbit hole if we go that way with the approach to merging specs.

---

*[types.go, line 155 at r3](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0L-r3-155:-L6SCZoHtAZKBehT2YqZ:bz00bbi) ([raw file](https://github.com/fission/fission/blob/63c83c04986c8db604b2cbd233c593d2127084ca/types.go#L155)):*

Thanks for the helpful comment. Now that it's in words, I actually understand the diff better.  And I think the overriding rules are not great...

If the user sets both Container.Image and Runtime.Image, maybe we should just error out. In other words we should not override Image.  And if we do override, then the one in the container seems like it should "win".

I also think user should be allowed to control ImagePullPolicy, so there's no good reason to override this.

Why do we want to override TerminationMessagePath?

</blockquote></details>

Okay, looking at the cases you have listed, we cannot avoid a largish function to handle the merging, dealing with all the cases. In that case we should take time to go over all fields in the container spec and how we want to handle them.

I can see the usefulness of this early/clear feedback to the user in the form of an error when the image is set. I am doubting whether it (erroring on the image)  should happen in the actual merging/deployment of the environment, or in a separate validation step in the controller.

As for why override the TerminationMessagePath: I simply listed all fields that are currently overriden. 

---

*[buildermgr/envwatcher.go, line 452 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6O0Prkb6FXVHNmECvZ:-L6SFTbmvxXF7uC9kuAU:bel8c39) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/buildermgr/envwatcher.go#L452)):*

So I think the right thing is to merge 3 specs:

Container{
    image
    image pull policy
    command
    ReadinessProbe
}

env.Spec.Builder.Container

Container{
   Name: "builder"
   VolumeMounts
}


That way the user can override everything except what fission needs to actually work.

</blockquote></details>

Hmm, nice solution! So basically the environment is a merge of `system-defaults | user-defined | system-defined`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413)*

soamvasani:
Review status: 6 of 13 files reviewed at latest revision, 8 unresolved discussions.

---

*[common.go, line 47 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6NwWk87Vc0GIBhcFmT:-L6SRmXzayNTfZfHV4bv:b-ft1w9l) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/common.go#L47)):*

This indeed needs some documentation to explain the behavior for the situations you describe. Currently, the 'first' wins. If a field is set to a non-zero value, it is not overriden by other specs. Slices and maps are a special case: the contents of specs are concatenated together. 

We probably need to add a couple of edge cases (e.g. you probably want to have the command or args field overriden rather than merged. Looking at the source of the Container spec, the fields are annotated how to be merged (using the JSON patch specification). Unfortunately, there is no implementation that allows merging structs using these annotations as far as I can find. Kubernetes has an util library with a forked json library called 'strategicpatch' that implements the JSON patch spec, but only for JSON documents. So we are getting deeper and deeper into the rabbit hole if we go that way with the approach to merging specs.
</blockquote></details>

Ok, please add your first paragraph as the function comment.

Let's not make this more complex than it is. We do not have to implement a generic container spec merge. We just have to figure out which of the user's fields to override ourselves.  I don't think the Kubernetes generic merge thing is relevant in this particular case, simply because we don't really have two containers to merge.

---

*[types.go, line 155 at r3](https://reviewable.io:443/reviews/fission/fission/413#-L-2Mej46zCIICRG7j0L-r3-155:-L6SSFEmrL6x5GCDpXGu:b-cq98c5) ([raw file](https://github.com/fission/fission/blob/63c83c04986c8db604b2cbd233c593d2127084ca/types.go#L155)):*

Okay, looking at the cases you have listed, we cannot avoid a largish function to handle the merging, dealing with all the cases. In that case we should take time to go over all fields in the container spec and how we want to handle them.

I can see the usefulness of this early/clear feedback to the user in the form of an error when the image is set. I am doubting whether it (erroring on the image)  should happen in the actual merging/deployment of the environment, or in a separate validation step in the controller.

As for why override the TerminationMessagePath: I simply listed all fields that are currently overriden. 
</blockquote></details>

1. We don't have to analyze and discuss every field in the Container spec because most of them simply get passed into the pod.  Again, let's not make this more complex than it is; there aren't really two containers here, just a container and a few fields that we want to override.

2. Ok, yes, we can tackle that under the CRD validation work

3. Ok

---

*[buildermgr/envwatcher.go, line 452 at r9](https://reviewable.io:443/reviews/fission/fission/413#-L6O0Prkb6FXVHNmECvZ:-L6SSsmvDx5fv6dA3F12:b-2lmov6) ([raw file](https://github.com/fission/fission/blob/45e9198b850d8f40566ea9b33f3649fb9462a1da/buildermgr/envwatcher.go#L452)):*

Hmm, nice solution! So basically the environment is a merge of `system-defaults | user-defined | system-defined`?
</blockquote></details>

Yes. I got the order wrong since I thought the last one wins. Since the first one wins it's probably 
`mergo(overriden fields, user-defined, defaults)`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413)*

soamvasani:
Gonna move this to 0.6.1, @erwinvaneyk -- we can release that asap when this merges.

soamvasani:
I have a few very minor comments, looks good other than that! Thanks for patiently working through all the stages feedback.

---

Reviewed 3 of 7 files at r10, 6 of 6 files at r11.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[common_test.go, line 1 at r11](https://reviewable.io:443/reviews/fission/fission/413#-L7lIfAl_1cRFn5nt03r:-L7lIfAmCgstgX2-eAkh:b-nuoo45) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/common_test.go#L1)):*
> ```Go
> package fission
> ```

(a) please rename file to merge_test or something
(b) copyright header

---

*[common_test.go, line 6 at r11](https://reviewable.io:443/reviews/fission/fission/413#-L7lImjqGoYRx1DH0LBX:-L7lImjqGoYRx1DH0LBY:bx2ts7a) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/common_test.go#L6)):*
> ```Go
> 	"github.com/stretchr/testify/assert"
> 	apiv1 "k8s.io/client-go/pkg/api/v1"
> 	"testing"
> ```

import order, stdlibs in first "paragraph"

---

*[test/tests/test_env_vars.sh, line 19 at r11](https://reviewable.io:443/reviews/fission/fission/413#-L7me3GfB4tFaOkDL--U:-L7me3GgAwBb_30rf_9w:b-4fwskg) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/test/tests/test_env_vars.sh#L19)):*
> ```Shell
> FN_FILE=${TEST_DIR}/${FN}.yaml
> 
> explain() {
> ```

explain doesn't imply that the command is executed; call it `log_run` or something

---

*[test/tests/test_env_vars.sh, line 46 at r11](https://reviewable.io:443/reviews/fission/fission/413#-L7meOHm3i9ylwrkvnz9:-L7meOHm3i9ylwrkvnzA:b-qbaixv) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/test/tests/test_env_vars.sh#L46)):*
> ```Shell
>     NS=$1
>     POD=$2
>     kubectl -n ${NS} get po -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' \
> ```

I'm ok with this but just fyi you can also do `-o name`(And then `|cut -f2 -d'/'`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413#-:-L7mf4-Ljy8SthTfDYdE:b8gw2gq)*

erwinvaneyk:
Review status: 12 of 14 files reviewed at latest revision, 11 unresolved discussions.

---

*[common_test.go, line 1 at r11](https://beta.reviewable.io/reviews/fission/fission/413#-L7lIfAl_1cRFn5nt03r:-L7xr_QYEJ_kGR8ofwJQ:b-5sizxz) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/common_test.go#L1)):*

(a) please rename file to merge_test or something
(b) copyright header

</blockquote></details>

I think common_test is more idiomatic go, as the merging is part of the common.go file. But lets discuss this later; I'll change it.

Would be great to have a check for the copyright header in the travis tests btw :D 

---

*[common_test.go, line 6 at r11](https://beta.reviewable.io/reviews/fission/fission/413#-L7lImjqGoYRx1DH0LBX:-L7xs_RS1HyVfuo2zKCh:byuo23i) ([raw file](https://github.com/fission/fission/blob/9af5409a857dfdd7541798af8d167add9d06dd7a/common_test.go#L6)):*

import order, stdlibs in first "paragraph"
</blockquote></details>

Okay, I didn't know we consider this part of the mandatory style. We could consider using `goimport` instead of `go fmt` for the `verify-format` check to check for this type of style violation.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/413)*

soamvasani:
LGTM

---

Reviewed 3 of 3 files at r12.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/413#-:-L8BU4Ci5h71Mm91uBcw:bb6ywmx)*

soamvasani:
🎉

PR #418, opened 2017-11-30
Fix pool manager crash problem if failed at http call

life1347:
Pool manager crashed due to nil pointer problem when it failed at calling fetcher to fetch code.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/418)


soamvasani:
Argh. This is such a common mistake. We really should start using `go vet` or something.

PR #419, opened 2017-11-30
Overwrite request host with internal host to prevent request rejection

life1347:
In `http.Request` contains a `Host` field that holds the hostname the HTTP request targeting. When a request comes to the router, the router will proxy the request to one of backend function pods. 

Normally, it will be fine not changing the Host field of the request. But after some experiment for istio integration, the request with mismatched Host and Url was rejected by istio-proxy automatically. And the solution is to replace Host with right hostname the request targeting.

Since it's not really related to istio, so I separated it into an independent PR.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/419)

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -141,6 +141,11 @@ func (fh *functionHandler) handler(responseWriter http.ResponseWriter, request *
 		// function metadata here.
 		req.URL.Path = "/"
 
+		// Overwrite request host with internal host,
+		// or request will be blocked in some situations
+		// (e.g. istio-proxy)
+		req.Host = serviceUrl.Host
+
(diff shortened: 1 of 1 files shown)


soamvasani:
Makes sense.  Can you add an X-Forwarded-For header too?

life1347:
The ServeHTTP already do that for us. https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L183-L191

life1347:
The ServeHTTP already do that for us. 
https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L183-L191

PR #422, opened 2017-12-04
Declarative application specifications for Fission

soamvasani:
This change enables users to have declarative specifications for Fission resources. Users can specify their "app" in a set of spec files, and use a new fission CLI command to "apply" these specs to a running cluster.

```
NAME:
   fission spec - Manage a declarative app specification

USAGE:
   fission spec command [command options] [arguments...]

COMMANDS:
     init      Create an initial declarative app specification
     validate  Validate Fission app specification
     apply     Create or update Fission resources from app specification
     destroy   Delete all Fission resources in the app specification
     save      Save an existing Fission resource to the app specification
     helm      Create a helm chart from the app specification
```

```
NAME:
   fission spec apply - Create, update or delete Fission resources from app specification

USAGE:
   fission spec apply [command options] [arguments...]

OPTIONS:
   --specdir value  Directory to store specs, defaults to ./specs
   --delete         Allow apply to delete resources that no longer exist in the specification
   --watch          Watch local files for change, and re-apply specs as necessary
```

This capability is not strictly new -- since Fission resources are Kubernetes CustomResources (using CRDs), you could already do a lot of the stuff in this change with `kubectl apply`. 

We have a few goals with this feature:
 1. Users should be able to check in a set of Fission resources (functions, triggers, etc.) to source control and re-create the app
 1. Function source code should live in function files, in some editing-friendly way (not serialized into YAMLs)
 1. Users get the full power of being able to customize various parameters without having to write a big YAML from scratch.  
 1. Idempotent setup: users should be able to apply specs to a running cluster, and the apply should create, update, delete resources if and when needed.
 1. Make fission apps more distribution friendly

(This change solves or will solve most of the above goals, but there's a bunch of work to do for the last one.)

Since Fission resources are CRDs, there's not a whole lot to do. We just have to deserialize and create/update/delete each resource.  (More about delete below).  

However, one type of Fission resource is _not_ a CRD: a fission _archive_.  Archives are zip files/tgz files that contain a set of files. The archive may be a source archive (for the builder) or a deployment archive (for the runtime).  Archives are necessarily stored somewhere other than Kubernetes resources (because of size limits).  The fission storage service fills this need.

We need a way for users to specify a set of files, and reference that set from package specs.  This change creates a spec-only resource called an `ArchiveUploadSpec`, and a new type of URL `archive://<archive-name>` to point at the archive created by an `ArchiveUploadSpec`.  The tool then resolves these `archive://` URLs after uploading the files. (Note that there is no ArchiveUploadSpec CRD; this has to be handled on the client itself.)

The command `fission spec apply --watch` will continuously keep their cluster updated while files change in the repo. This automates the full pipeline from upload to build to deploy. It's a very useful tool to use while in the code-test-fix-repeat cycle. The implementation in this PR is very naïve, and there's room for optimizations in the future (like modifying archives instead of rebuilding them from scratch, and even doing rsync or something like it, instead of shuttling around full archives).

A _deployment config_ in the spec directory specifies a name and a UID. `fission spec init` creates the deployment config and the spec directory.  The name and UID are added as annotations to every resource created or updated by `fission spec apply`.  The name is just for user-friendliness and can be changed any time. The UID should not be changed.

Updates and deletes are only applied to resources that have been created by `apply`. For updates this ensures that accidental name collisions don't end up modifying someone else's resources. For deletes, this allows us to find resources that were part of the spec but aren't any more. (The other way to do this is to keep deleted objects in the spec but mark them as deleted in some way, either as an annotation or by putting them in a special "deleted objects" directory. That seems more complex for the user. Easier to simply delete resources that you want removed.)

This is a purely client-side change.  This means we can't be transactional -- the user can end up with a half-installed state if they hit control-C or lose their connection to the cluster.  But they can repair their state by running the apply command again, since it's idempotent.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/422)


erwinvaneyk:
Reviewed 8 of 10 files at r1.
Review status: 8 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[examples/spec-example/specs/hello.yaml, line 1 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sTpgn1aVqyLbiHwMO:-L-sTpgn1aVqyLbiHwMP:b-3drbjr) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/examples/spec-example/specs/hello.yaml#L1)):*
> ```YAML
> apiVersion: fission.io/v1
> ```

What about adding a comment in the header of these generated yaml specs to indicate that they have been generated. That should avoid any impression of users thinking that they have to write these themselves. Plus it adds some useful debug information. Something like: 

```yaml
# Generated using `fission spec <cmd>` with fission cli <version>
apiVersion: fission.io/v1
...
```

This suggestion holds for all generated specs - not going to repeat this comment.

---

*[fission/common.go, line 278 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sX237B_fZq8pFxIq-:-L-sX238CBBZ8d66L-fE:b8z8s1r) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L278)):*
> ```Go
> // make a kubernetes compliant name out of an arbitrary string
> func kubifyName(old string) string {
> 	maxLen := 63
> ```

extract to package constant

---

*[fission/common.go, line 294 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sVcbcxtX_UIx1US9-:-L-sVcbcxtX_UIx1US90:bp6urza) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L294)):*
> ```Go
> 	// trim trailing
> 	trailing, err := regexp.Compile("[^a-z0-9]+$")
> 	checkErr(err, "compile regexp")
> ```

Seems a bit redundant to recompile these static expressions. What about moving all the `regexp.compile` outside of this function to the package level?

---

*[fission/common.go, line 298 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sWl_sJCKi4QgpAQy4:-L-sWl_sJCKi4QgpAQy5:bbs48aw) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L298)):*
> ```Go
> 
> 	// truncate to length
> 	if len(newName) > maxLen-4 {
> ```

Needs a comment or a constant to explain why 4. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/422)*

soamvasani:
Review status: 1 of 24 files reviewed at latest revision, 4 unresolved discussions.

---

*[examples/spec-example/specs/hello.yaml, line 1 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sTpgn1aVqyLbiHwMO:-L1q5ZqE0qXd40NIEMdu:b-s0vjdj) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/examples/spec-example/specs/hello.yaml#L1)):*

What about adding a comment in the header of these generated yaml specs to indicate that they have been generated. That should avoid any impression of users thinking that they have to write these themselves. Plus it adds some useful debug information. Something like: 

```yaml
# Generated using `fission spec <cmd>` with fission cli <version>
apiVersion: fission.io/v1
...
```

This suggestion holds for all generated specs - not going to repeat this comment.
</blockquote></details>

I see what you're saying, but since they can be edited by users, they're not purely generated files.  The `create --spec` stuff is intended as a starting point for specs, not a complete solution.  The idea is that YAML is quite annoying to write initially but not that bad to edit later.

---

*[fission/common.go, line 278 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sX237B_fZq8pFxIq-:-L1q8bR_2nHQm4olqRRa:b-kmlccg) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L278)):*

extract to package constant
</blockquote></details>

It's only used in this function, but we'll lift it to the package if it's ever used in other places.

---

*[fission/common.go, line 294 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sVcbcxtX_UIx1US9-:-L1q7wUQ8wxpRmdixPvS:b-7vd10r) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L294)):*

Seems a bit redundant to recompile these static expressions. What about moving all the `regexp.compile` outside of this function to the package level?
</blockquote></details>

Hmm.  Gonna favor readability/simplicity over performance in this case... 

---

*[fission/common.go, line 298 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sWl_sJCKi4QgpAQy4:-L1q8HUCui8xZXVwsL4l:b-a4m798) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/fission/common.go#L298)):*

Needs a comment or a constant to explain why 4. 
</blockquote></details>

Good catch! I removed it (names can become non-unique but we'll just allow that to fail. The user can manually change the names to make them unique)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/422)*

erwinvaneyk:
Reviewed 1 of 10 files at r1, 23 of 23 files at r2.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[types.go, line 168 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20EI1wyscTZJkokn3J:-L20EI1wyscTZJkokn3K:b-rbdtxd) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/types.go#L168)):*
> ```Go
> 
> 		// Optional
> 		Builder Builder `json:"builder"`
> ```

What is the guideline you use to decide where to add `omitempty`? Seems like it can be added here among other places as well.

---

*[examples/go/specs/env.yaml, line 12 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20IjHA0axll9ghrD-Q:-L20IjHA0axll9ghrD-R:b-y472yn) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/env.yaml#L12)):*
> ```YAML
>     image: fission/go-build-env:20171206-2
>   runtime:
>      image: fission/go-env:20171206
> ```

formatting is off here

---

*[examples/go/specs/fission-config.yaml, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20Ez8Mh1XzcqDwaQi4:-L20Ez8Mh1XzcqDwaQi5:bsn8919) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/fission-config.yaml#L1)):*
> ```YAML
> kind: DeploymentConfig
> ```

In addition to the comment regarding the `generated with...` comment. It would be useful to add more comments to these generated files to explain what the purpose is. Without looking into the source code it is hard to grasp for me what exactly is the purpose of some of the fields. For example, is it intended that this spec does not contain an `apiVersion`, what exactly is the purpose of the uid here, etc...? Ideally these generated specs would be self-explanatory, like how the generated vagrant config looks like.

---

*[examples/go/specs/fission-config.yaml, line 3 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20GSN_JiYWyYR8eTme:-L20GSN_JiYWyYR8eTmf:bquy6es) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/fission-config.yaml#L3)):*
> ```YAML
> kind: DeploymentConfig
> name: hello-go
> uid: a8cdb63c-9be8-4a59-9427-89051afecd7b
> ```

Name of this file `fission-config.yaml` is not consistent with the other files in this directory, and does not really reflect/explain the purpose of this spec. What about renaming it to `config.yaml` or `deployment.yaml` or `fissionfile.yaml`...

---

*[examples/go/specs/hello.yaml, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20Iqgew2XMF7BVtJVN:-L20Iqgew2XMF7BVtJVO:bf1y880) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/hello.yaml#L1)):*
> ```YAML
> include:
> ```

Not sure if this is manually named or generated. In case of the latter you might want to prepend `hello.yaml` with a function indicator to avoid conflicts with other files in the spec directory. `fn-hello.yaml`, `function-hello.yaml`?

---

*[examples/go/specs/hello.yaml, line 2 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20GzlUZkIiUOv_DXE3:-L20GzlUZkIiUOv_DXE4:b3l2tq1) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/hello.yaml#L2)):*
> ```YAML
> include:
> - hello.go
> ```

Minor: if  it is easy to fix, 1) move the include below the kind/name. 2) similar to the deployment

---

*[examples/go/specs/hello.yaml, line 15 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20HQf_d1t-4SsxeR9C:-L20HQf_d1t-4SsxeR9D:b-54c754) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/hello.yaml#L15)):*
> ```YAML
> spec:
>   deployment:
>     checksum: {}
> ```

Left as a future todo? if not, why is it empty?

---

*[examples/go/specs/hello.yaml, line 24 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20HjcWIn7fsj9THhBv:-L20HjcWIn7fsj9THhBw:buvnhgt) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/go/specs/hello.yaml#L24)):*
> ```YAML
>     url: archive://hello-go
> status:
>   buildstatus: pending
> ```

Minor: can this be left out and have the default be `pending`. This seems redundant as well as confusing.

---

*[examples/spec-example/README, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20JV_l90Uy22ShPdYx:-L20JV_l90Uy22ShPdYy:bis7uqb) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/spec-example/README#L1)):*
> ```
> This is the root directory of a sample fission "app".  The app contains source code for one function (hello world) in the hello/hello.py file.
> ```

Not sure how Github handles these types of files. Anyhow, for the sake of consistency rename the file to `README.md`.

---

*[examples/spec-example/hello/hello.py, line 4 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20KQ-2cfqKYu-5OnSU:-L20KQ-2cfqKYu-5OnSV:b-t5nl1x) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/examples/spec-example/hello/hello.py#L4)):*
> ```Python
> 
> def main():
>     return "hello, world!!\n"
> ```

Minor: would be great to actually use yaml - for the sake of completeness

---

*[examples/spec-example/specs/hello.yaml, line 1 at r1](https://reviewable.io:443/reviews/fission/fission/422#-L-sTpgn1aVqyLbiHwMO:-L20CO6T7yHy8iNe97zd:bglkcm0) ([raw file](https://github.com/fission/fission/blob/44e803c8742910ee154216c7a09922881d526839/examples/spec-example/specs/hello.yaml#L1)):*

I see what you're saying, but since they can be edited by users, they're not purely generated files.  The `create --spec` stuff is intended as a starting point for specs, not a complete solution.  The idea is that YAML is quite annoying to write initially but not that bad to edit later.
</blockquote></details>

~~I understand. Still I think it would benefit the user experience to indicate that they shouldn't be writing from scratch. Maybe we can come up with another comment. Otherwise we should make it very clear in the documentation. ~~

I see that you added. Could you update the specs of the examples to reflect the changes?

---

*[fission/buildwatch.go, line 32 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20LA2r2YL0lNWRNKs0:-L20LA2r2YL0lNWRNKs1:b-3hmok4) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/fission/buildwatch.go#L32)):*
> ```Go
> 
> type (
> 	packageBuildWatcher struct {
> ```

Needs a comment what the purpose is of this `packageBuildWatcher`.

---

*[fission/buildwatch.go, line 80 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20M6sj8XbT3GzvSVIo:-L20M6sko5kqA2eZ8-uc:bx4odqo) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/fission/buildwatch.go#L80)):*
> ```Go
> 				continue
> 			}
> 			if pkg.Status.BuildStatus == fission.BuildStatusNone {
> ```

Minor: a switch would be more readable here

---

*[fission/buildwatch.go, line 118 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20MvVNc8JNcdKpbCbB:-L20MvVNc8JNcdKpbCbC:bavaj5b) ([raw file](https://github.com/fission/fission/blob/d66c946e547fcbef63cb5650b1c52f3c879ad73e/fission/buildwatch.go#L118)):*
> ```Go
> func pkgKey(pkg *crd.Package) string {
> 	// packages are mutable so we want to keep track of them by resource version
> 	return fmt.Sprintf("%v:%v:%v", pkg.Metadata.Name, pkg.Metadata.Namespace, pkg.Metadata.ResourceVersion)
> ```

Sounds like something that we will run into more in the future. What about adding this as a method to `crd.package`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/422)*

soamvasani:
Review status: 16 of 26 files reviewed at latest revision, 14 unresolved discussions.

---

*[types.go, line 168 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20EI1wyscTZJkokn3J:-L2wV4C35gIJ224bnSiW:bkr1bhi) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/types.go#L168)):*

What is the guideline you use to decide where to add `omitempty`? Seems like it can be added here among other places as well.
</blockquote></details>

You can't omitempty a struct type, AFAIK (I mean you can add the annotation of course, but it gets ignored): https://stackoverflow.com/questions/18088294/how-to-not-marshal-an-empty-struct-into-json-with-go

---

*[fission/buildwatch.go, line 118 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20MvVNc8JNcdKpbCbB:-L3AxTPdNAP_LoyjF_MJ:b-3nlobr) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/fission/buildwatch.go#L118)):*

Sounds like something that we will run into more in the future. What about adding this as a method to `crd.package`?
</blockquote></details>

Not sure if the implementation is general enough yet, but I'll keep an eye out for that.

---

*[examples/go/specs/fission-config.yaml, line 1 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20Ez8Mh1XzcqDwaQi4:-L2wVp4yXTg314bitWLj:blmwhl9) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/examples/go/specs/fission-config.yaml#L1)):*

In addition to the comment regarding the `generated with...` comment. It would be useful to add more comments to these generated files to explain what the purpose is. Without looking into the source code it is hard to grasp for me what exactly is the purpose of some of the fields. For example, is it intended that this spec does not contain an `apiVersion`, what exactly is the purpose of the uid here, etc...? Ideally these generated specs would be self-explanatory, like how the generated vagrant config looks like.
</blockquote></details>

It should have an API version.  I'll update the examples.

---

*[examples/go/specs/fission-config.yaml, line 3 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20GSN_JiYWyYR8eTme:-L2wVZlsBquPdO1E6UPe:b-7bn6zd) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/examples/go/specs/fission-config.yaml#L3)):*

Name of this file `fission-config.yaml` is not consistent with the other files in this directory, and does not really reflect/explain the purpose of this spec. What about renaming it to `config.yaml` or `deployment.yaml` or `fissionfile.yaml`...
</blockquote></details>

Changed to fission-deployment-config.yaml.

---

*[examples/go/specs/hello.yaml, line 15 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20HQf_d1t-4SsxeR9C:-L3AmoaiBQPPYDCQNfrj:blkjry3) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/examples/go/specs/hello.yaml#L15)):*

Left as a future todo? if not, why is it empty?
</blockquote></details>

It's generated; Go doesn't let us use omitempty on structs. But I've manually edited it to make the sample more readable.

---

*[examples/go/specs/hello.yaml, line 24 at r2](https://reviewable.io:443/reviews/fission/fission/422#-L20HjcWIn7fsj9THhBv:-L3An6ZQIsOClal4cloT:bore9op) ([raw file](https://github.com/fission/fission/blob/1a850bf14f4f6bc21b347a4617497c8ee4c10cb8/examples/go/specs/hello.yaml#L24)):*

Minor: can this be left out and have the default be `pending`. This seems redundant as well as confusing.
</blockquote></details>

It's not redundant, it's used to tell the builder to start the build. It's the tricky thing about invoking a task in a declarative context. I'll add some explanatory comments for users. There is some history behind the decision to use this flag to trigger builds.  You can find it in past discussions on builder pull requests.  (Also feel free to bring it up on the slack; I'm open to changing how this is done, if anyone has better ideas.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/422)*

PR #424, opened 2017-12-04
Removed openshift specifics as they are no longer necessary

karmab:
tested fission on openshift 3.7.0 and installed fine, without having to use openshift variables
those variables were used prior to switching to custom resource definitions, and as such are obsolete

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/424)


soamvasani:
Thank you for this @karmab !

Issue #423, opened 2017-12-04
Build logs not saved on build error


soamvasani:
Hey @life1347, it looks like the buildermgr only updates pkg.Status.BuildLogs when the build succeeds -- it should also save them when the build fails (i.e. the build script exits with non-zero exit code)

erwinvaneyk:
Fixed by #426 - @life1347 you can use "fixed <issue>" to automatically close issues on PR merges https://help.github.com/articles/closing-issues-using-keywords/

PR #426, opened 2017-12-04
[Issue 423] build logs not saved on build error

life1347:
This PR tends to fix the bug ~introduced in PR #397~ that builder manager updates package with empty build logs when build script exited with non-zero code. Also fix a potential nil pointer issue. (issue #423)

(Update: 2017/12/05 )
Looks like this bug existed at the beginning

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/426)


soamvasani:
The change LGTM. Did this really break in #397?

life1347:
Sorry, I check the PR again and found this bug was introduced at the beginning. Comment is updated

Issue #428, opened 2017-12-07
Funtion can be published to env which do not exist 


AmitKrVarman:
would it be worth putting in validation to check if env is defined and created before we publish functions?

fission function create --name funtion-name --env i-do-not-exist --deploy function.so

Sometime type while publishing functions go unnoticed.

soamvasani:
Thanks Amit, yes, we should add a warning in this case.

PR #436, opened 2017-12-20
Show warning when user tries to create a function with a non-existed environment

life1347:
This PR aims to fix issue https://github.com/fission/fission/issues/428 when a user tries to create a function with a non-existed environment.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/436)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -113,7 +113,7 @@ func fnCreate(c *cli.Context) error {
 		})
 		if err != nil {
 			if e, ok := err.(fission.Error); ok && e.Code == fission.ErrorNotFound {
-				fmt.Printf("Environment \"%v\" does not exist. Without the environment the function cannot be executed!\n", envName)
+				fmt.Printf("Environment \"%v\" does not exist. Please create the environment before executing the function. \nFor example: `fission env create --name %v --image <image>`\n", envName, envName)
 			} else {
 				checkErr(err, "execute retrieving environment information")
 			}
--- a/fission/function.go
+++ b/fission/function.go
@@ -106,6 +106,19 @@ func fnCreate(c *cli.Context) error {
 			fatal("Need --env argument.")
 		}
 
+		// examine existence of given environment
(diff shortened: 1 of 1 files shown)


fission/function.go
```diff
@@ -106,6 +106,19 @@ func fnCreate(c *cli.Context) error {
...
 		}
 
+		// examine existence of given environment
+		_, err := client.EnvironmentGet(&metav1.ObjectMeta{
+			Namespace: metav1.NamespaceDefault,
+			Name:      envName,
+		})
+		if err != nil {
+			if e, ok := err.(fission.Error); ok && e.Code == fission.ErrorNotFound {
+				fmt.Printf("Environment \"%v\" is not existed, please create it before accessing the function.\n", envName)
```
erwinvaneyk: Error message is not correct. `... is not existed, please create...` => `... does not exist. Please create it before creating the function.`

life1347: `Please create it before creating the function` means that we need to stop the function creation if an environment not existed? I think we can simply show the warning here instead of stopping the creation.

erwinvaneyk: Oh right, I thought it was an error. What about making it a bit more explicit `Without the environment the function cannot be executed!`?

erwinvaneyk: It is just the word 'accessing' that is bothering me btw. It is too vague. But I am also not sure what word would be best to replace that. ;)

life1347: How about `Function with non-existing environment cannot be executed. Please create environment "%v" after the function creation.` ?

erwinvaneyk: I liked the first sentence as it is very short and direct about what is wrong. The second sentence provides some more explanation. What did you think about the `Without the environment the function cannot be executed!` one?

life1347: After discussion with @erwinvaneyk on slack, we think it's better to guide a user what should do next when he/she saw the warning message. The decision is 
```
"Environment \"%v\" does not exist. Please create the environment before executing the function. \nFor example: `fission env create --name %v --image <image>`\n"
```

Issue #400, opened 2017-11-16
Builds fail if builder isn't ready


erwinvaneyk:
When a builder environment is not yet ready, build requests fail. However, once the builder environment becomes ready, the failed build requests are not retried.

To reproduce:
- create a build environment 
- create a function using that build environment before the build environment is actually available.

soamvasani:
Mind taking this one, @life1347 ?

life1347:
Yes, I can handle this. We can let builder manager check and retry builds for packages in the failed status. To prevent infinite build failure, I will also add a `RetryTimes` in PackageStatus. If `RetryTimes` reaches the threshold then stop retrying for that package.

```
PackageStatus struct {
    BuildStatus BuildStatus `json:"buildstatus"`
    BuildLog    string      `json:"buildlog"` // output of the build (errors etc)
    RetryTimes  int         `json:"retrytimes"`
}
```

soamvasani:
Instead of retrying, would it make sense to wait for the builder to be ready? You can even add a readiness probe to the builder pod spec to make sure everything is up.

soamvasani:
fixed by #437

PR #437, opened 2017-12-20
Block build requests before environment builder is ready

life1347:
This PR adds readiness probe to fetcher & builder so that we can make sure that the builder is ready to accept build requests. https://github.com/fission/fission/issues/400

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/437)


soamvasani:
Thanks @life1347, the change LGTM.  Please investigate test failures and merge when ready.

soamvasani:
You'll have to modify the unit test somehow to make it not start the full storage service.  Maybe a simple flag to disable the pruner would be good enough for the unit test.  The integration test can be used to test the pruner.

life1347:
Just want to use this [gif](http://www.shaoerw.com/dongtaitu/img45876adfeafg/434.gif) to express how I feel when green check marks appear.

PR #438, opened 2017-12-21
WIP - Design Proposal for functionVersioning

smruthi2187:
This is just a proposal to support function versioning in fission. Please suggest any modifications or feel free to propose a different approach towards function versioning.
Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/438)

### DIFF
--- a/Documentation/wip/FunctionVersioningProposal.md
+++ b/Documentation/wip/FunctionVersioningProposal.md
@@ -4,7 +4,7 @@
 1. Create a new version of a function with code changes ( perhaps adding some new features to a function )
 2. Have two versions of a function deployed and perform canary testing by creating an HTTP trigger for both versions of a function.
 3. Create an HTTP Trigger for a function with a specific version, if not fall back to the `“latest”` version of a function.
-4. Create other kings of triggers to a function with a specific version.
+4. Create other kinds of triggers to a function with a specific version.
 5. Create a set of functions and group them as one version. ( perhaps run a group of functions as a cronjob with one timeTrigger, haven’t thought this through )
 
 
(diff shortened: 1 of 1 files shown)


Documentation/wip/FunctionVersioningProposal.md
```diff
@@ -0,0 +1,128 @@
...
+	FunctionRef FunctionRef
+} 
+```
+* Introducing a new CRD of type `"FunctionMaster"`. This will have a reference to all aliases of a function. Also will have a list of references to all versions created for this function.
+```
+Type FunctionMaster struct {
+	Name string
+	Alias []Alias
+	Versions []FunctionRef
+}
```
vishal-biyani: How is an Alias mapped to a certain version of FunctionRef, just thinking if a Map might better suit here.

Documentation/wip/FunctionVersioningProposal.md
```diff
@@ -0,0 +1,128 @@
...
+Type Tag struct {
+	Name string
+	FunctionRef []FunctionRef
+}
+```
+
+#### General Flow 
+1. When a user wants to create a function with different versions, he can do so with the following commands
+    ```bash
+    $ fission function version create --name hello --env nodejs --code hello1.js --version v1.0.0
```
vishal-biyani: This is my personal opinion but using a SemVar like versioning (i.e. x.y.z) makes it a bit user un-friendly and might be an overkill. I like simpler naming scheme with a single number. Also in practice how many versions of a function might user create? Even if it is in range of 1000, the single number versioning is far simpler. Open to more discussion on this.

Documentation/wip/FunctionVersioningProposal.md
smruthi2187: For each version of the function, this design intends to create a new function object, with name "functionName_versionValue". The Alias type holds a reference to a specific functionName. Actually, the "General Flow" explains the flow of commands that the user would use to create different versions of the function.

Documentation/wip/FunctionVersioningProposal.md
smruthi2187: Yes, I agree too. But it's upto the user to create versions according to any syntax. I've also listed this point under "Note" section in the end. I'm not very keen on fission maintaining the version values.

smruthi2187:
Not valid anymore.

PR #442, opened 2018-01-07
Added instructions for deploying Fission on  docker for mac/windows

erwinvaneyk:
Playing around with the Kubernetes support in Docker for Mac today, one of the first things to check is of course how this affects Fission. So I documented my experiences

So far I haven't run into any issues with Fission. Docker for Mac automatically ensures that the single node cluster is exposed on localhost. Seems like a really promising alternative to Minikube.

There is a small gotcha, at least when you have been using Minikube on the same machine. To get it working you need to switch the config context to `docker-for-mac`.

It would be great if someone could try this out (both on Windows and Mac) to verify that there aren't any other issues.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/442)

### DIFF
--- a/Documentation/docs-site/content/install.md
+++ b/Documentation/docs-site/content/install.md
@@ -72,7 +72,7 @@ $ helm init
 
 ### Install Fission
 
-#### Minikube
+#### Minikube, Docker for Mac/Windows
 
 ```
 $ helm install --namespace fission --set serviceType=NodePort https://github.com/fission/fission/releases/download/0.4.0/fission-all-0.4.0.tgz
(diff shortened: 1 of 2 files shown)


vishal-biyani:
Hey @erwinvaneyk Do you mind retrofitting this to docs repo in the following page and closing this PR please : https://github.com/fission/docs.fission.io/blob/master/docs/content/installation/installation.en.md

vishal-biyani:
Closing here will need to be moved to go docs repo.

erwinvaneyk:
@vishal-biyani it should also be checked with our new use of the proxies to connect to the Fission API

Issue #446, opened 2018-01-16
Function logs breakage


soamvasani:
`fission fn logs` doesn't print the actual log message from the function (it prints the timestamp correctly). The bug is either in the query in the CLI or in the fluent config.

PR #448, opened 2018-01-16
Fix fission function logs

prithviramesh:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/448)

### DIFF
--- a/test/tests/test_logging/test_function_logs.sh
+++ b/test/tests/test_logging/test_function_logs.sh
@@ -30,7 +30,7 @@ fission route create --function $fn --url /$fn --method GET
 trap cleanup EXIT
 
 echo "Waiting for router to catch up"
-sleep 3
+sleep 15
 
 echo "Doing 4 HTTP GETs on the function's route"
 for i in 1 2 3 4
@@ -40,7 +40,7 @@ done
 
 echo "Grabbing logs, should have 4 calls in logs"
 
-sleep 25
+sleep 15
 
 fission function logs --name $fn --detail > logfile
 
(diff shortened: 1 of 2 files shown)


test/tests/test_logging/test_function_logs.sh
```diff
@@ -43,11 +42,18 @@ echo "Grabbing logs, should have 4 calls in logs"
 
 sleep 15
 
-logs=$(fission function logs --name $fn --detail)
+fission function logs --name $fn --detail > logfile
```
soamvasani: Can you move this 'logfile' to /tmp? Also, can you add code to delete it before the test exits?

test/tests/test_logging/test_function_logs.sh
```diff
@@ -43,11 +42,18 @@ echo "Grabbing logs, should have 4 calls in logs"
...
+
+size=$(wc -c <logfile)
+if [ $size = 0 ]
+then
+    fission function logs --name $fn --detail > logfile
```
soamvasani: This is the same command as line 45?  What's the goal of doing this again?

prithviramesh: sometimes the 'fission function logs' command doesn't work the first time around, and it works with a retry of the command.. this is probably due to the time it takes for the logs to be shipped into Influx. The retry is to ensure the test does not fail because of this delay; if the retry fails and the command does not return anything the second time around then we can assume that it does not work properly

test/tests/test_logging/test_function_logs.sh
prithviramesh: sure thing

prithviramesh:
#446

PR #452, opened 2018-01-21
Convert build.sh to a multi-stage Dockerfile.

justinbarrick:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/452)

### DIFF
--- a/environments/binary/Dockerfile
+++ b/environments/binary/Dockerfile
@@ -1,10 +1,18 @@
+FROM golang:onbuild
+
+WORKDIR /go
+COPY *.go /go/
+
+RUN GOOS=linux GOARCH=386 go build -o server .
+
 FROM alpine:3.5
+
+WORKDIR /app
 
 RUN apk update
 RUN apk add coreutils binutils findutils grep
 
-COPY . /app
-WORKDIR /app
+COPY --from=0 /go/server /app/server
 
 EXPOSE 8888
 ENTRYPOINT ["./server"]
--- a/environments/binary/README.md
+++ b/environments/binary/README.md
@@ -36,11 +36,6 @@ All output that is provided to the server over the STDOUT will be transformed in
 
 
 ## Compiling
-In order to build the Dockerfile, the server needs to be compiled to the right architecture.
-
-```bash
-sh ./build.sh
-```
 
 Build the Dockerfile:
 ```bash
--- a/environments/binary/build.sh
+++ b/environments/binary/build.sh
@@ -1,4 +1,0 @@
-#!/bin/sh
-
-GOOS=linux GOARCH=386 go build -o server .
-
--- a/environments/binary/builder/Dockerfile
+++ b/environments/binary/builder/Dockerfile
@@ -1,0 +1,4 @@
+ARG BUILDER_IMAGE=fission/builder:latest
+FROM ${BUILDER_IMAGE}
+
+ADD build.sh /usr/local/bin/build
--- a/environments/binary/builder/build.sh
+++ b/environments/binary/builder/build.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+apk update
+
+CWD=$(pwd)
+
+if [ -f ${SRC_PKG}/setup.sh ]; then
+    cd ${SRC_PKG}
+    ./setup.sh
+    cd ${CWD}
+fi
+
+cp -rf ${SRC_PKG} ${DEPLOY_PKG}
--- a/environments/binary/builder/build.sh
+++ b/environments/binary/builder/build.sh
@@ -4,9 +4,9 @@ apk update
 
 CWD=$(pwd)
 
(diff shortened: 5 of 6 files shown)


environments/binary/Dockerfile
```diff
@@ -1,10 +1,18 @@
...
+
+RUN GOOS=linux GOARCH=386 go build -o server .
+
 FROM alpine:3.5
 
+WORKDIR /app
+
 RUN apk update
-RUN apk add coreutils binutils findutils grep
+RUN apk add coreutils binutils findutils grep curl jq
```
erwinvaneyk: I don't have a strong opinion on which packages to include or not, and these two seem really useful for many cases. However, this setup (of having the packages hardcoded in the image) is not scalable or user-friendly for that matter.

We could use a build environment here similar to the [golang](https://github.com/fission/fission/tree/master/environments/go/builder) or [python](https://github.com/fission/fission/tree/master/environments/python/builder) environments. The `binary-build` environment could execute a file in the package called `setup.sh` (if present), which could install the necessary package management. 

@justinbarrick if you want to dive deeper into Fission, that would actually be a perfect follow-up PR to this one. 😁 

cc @soamvasani - your opinion on this? If you agree, let's file this as an issue to keep track of it.

justinbarrick: Okay, I got this mostly setup. The builder will run `setup.sh`, if it exists and I modified `binary/server.go` to support specialize v2 requests and directories.

However, since setup.sh runs in the builder, any packages installed won't persist.

I'm not sure what the correct way to handle that is, since I don't want to cause the packages to be installed.

erwinvaneyk: Ey @justinbarrick - awesome work so far :+1: Sorry for the confusion, I was actually expecting this v2 conversion of the binary env to happen in another/next PR - this PR was already good to go. In either case, I would be happy to accept these changes as well.  

The fact that we don't capture changes to filesystem in the builder is indeed an issue for this environment. Not quite sure what the best way is to solve that issue. I see three options here:

1. We probably need a way to wrap the added packages (or changes to the filesystem in general) and add them to the deploy package. The binary environment can then unwrap the changes and update the runtime environment with the changes during specialization.

2. Another option, which might slower but simpler, is to add support for the execution of another script during the specialization phase. That would allow you to directly manipulate the runtime environment - but at a performance cost as the specialization will become more expensive.

3. The simplest but worst option UX-wise is to leave it at this builder and expect the users to work around it. The user could move the downloaded apk packages (like curl) to the deploy package and add those packages manually to the PATH in the runtime. This sounds like a *okay* solution until we have another solution figured out. For a longer term solution I don't really like it; it is not user-friendly, error prone, and requires users to have a relatively deep understanding of how the build process of Fission works.

^ Either of these solutions is completely out of scope for this PR. So let's keep the PR as it is now. Of course, if you want to help out with it, join us on the Slack so we can discuss this thoroughly before committing to a solution. :)

Minor thing: could you rename `setup.sh` to `build.sh`? It is more consist with the naming of build files in the rest of the project. My bad for suggesting it.

justinbarrick: I think that makes sense, if we go with option 3 for now, we'll come up with a clean workaround eventually.

As for a long term solution, we could potentially achive option 1 by running the builder script in a chroot, copying over the whole directory to the function container, and then running the function in a chroot.

erwinvaneyk: I like the chroot suggestion - that is definitely an option worth exploring.

erwinvaneyk:
LGTM 👍

life1347:
*[environments/binary/server.go, line 56 at r4](https://reviewable.io:443/reviews/fission/fission/452#-L3adEBtpEOw1WBz9Szd:-L3adEBtpEOw1WBz9Sze:b-szokmj) ([raw file](https://github.com/fission/fission/blob/b1dedc3d7ceb32c0062dc86fff5f510fe9b975cf/environments/binary/server.go#L56)):*
> ```Go
> 		switch mode := fileStat.Mode(); {
> 		case mode.IsDir():
> 			codePath = request.FilePath + "/" + request.FunctionName
> ```

Minor thing: it will better to use filepath.Join here to concatenate the codepath

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/452)*

life1347:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 2 of 3 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/452)*

soamvasani:
Hi @justinbarrick , looks like this is good to go, would you mind addressing this comment from @life1347's comment above:

> Minor thing: it will be better to use filepath.Join here to concatenate the codepath

justinbarrick:
Fixed!

PR #449, opened 2018-01-17
use time.Since instead of time.Now().Sub

wgliang:
use time.Since instead of time.Now().Sub

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/449)

PR #456, opened 2018-01-24
Fix fission bundle build failure

life1347:
This is a quick fix to solve the build failure problem introduced in https://github.com/fission/fission/pull/449

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/456)


life1347:
Sorry, my bad I should be more carefully next time.

PR #458, opened 2018-01-27
Prevent leaking of packages during fn update

smruthi2187:
1. I'd recommend going through "groupingFunctionsWithLabels.md" first for a summary of changes.
2. I moved some archive related code from fission/common.go to fission/archive.go. I can move it back if not desirable.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/458)


life1347:
*[fission/package.go, line 377 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vPklYn1voFYGOgszf:-L3vPklYn1voFYGOgszg:b-p96d1c) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/fission/package.go#L377)):*
> ```Go
> 		return nil
> 	}
> ```

it's a little bit confusing if users type a command like `pkg delete --name foo --orphan`.
So we can add a check and show prompt to notify them that `--orphan` will be ignored.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[fission/function.go, line 143 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vTMuNopY9cJ-_VYaX:-L3vTAnOPoVObi_qlCfa:b-4ktj3x) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/fission/function.go#L143)):*
> ```Go
> 	labels["package"] = pkgMetadata.Name
> 	labels["environment"] = envName
> ```

Minor thing: It's ok to me, but I am more preferred like following 

labels := map[string]string{
	"package":     pkgMetadata.Name,
	"environment": envName,
}

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[controller/functionApi.go, line 44 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vVJkdSS_w1akkGXSA:-L3vVJkdSS_w1akkGXSB:bete1bz) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/controller/functionApi.go#L44)):*
> URL.Query()
How about using `r.URL.Query().Get("labelSelector")` here?
doc & example: https://golang.org/pkg/net/url/#Values

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[fission/function.go, line 352 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vaXFNZ1jJX84FuNL0:-L3vaXFNZ1jJX84FuNL1:bqpa9ni) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/fission/function.go#L352)):*
> ```Go
> 		// we don't want to leak the pkg referenced earlier by this function, if not used by any other function
> 		if len(fnList) == 1 {
> 			err = deletePackage(client, function.Spec.Package.PackageRef.Name)
> ```

It's not a good idea to delete package directly without any prompt since we don't do this before.
Can we provide a flag like `--removeorpahn` and delete package only when the flag is set to true?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[fission/function.go, line 254 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3viF5uOy2hDC6ZuV47:-L3viF5uOy2hDC6ZuV48:bvxnz2n) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/fission/function.go#L254)):*
> ```Go
> 
> 	// This is for updating existing functions with labels
> 	label := c.Bool("label")
> ```

Hey, can we move this part to the controller and trigger it when the controller starts? Since it's more like a schema upgrade from the existing.
In this way, we don't need to tell users to do the upgrade by themselves. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[fission/package.go, line 342 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vj3PFRUws5XNVHAwT:-L3vj3PGp3J4oytyJ5Gt:b-36cc8e) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/fission/package.go#L342)):*
> ```Go
> 	if listOrphans {
> 		// Take this opportunity to label all functions. There may be a few functions created in older fission releases not having label.
> 		err = labelFunctions(client)
> ```

Instead of doing this every time, it's better to let controller do the upgrade when it starts 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
*[controller/functionApi.go, line 47 at r1](https://reviewable.io:443/reviews/fission/fission/458#-L3vm5r9Hr1ZmKCQ1wPf:-L3vm5r9Hr1ZmKCQ1wPg:b-fxbv88) ([raw file](https://github.com/fission/fission/blob/8b5eb0f02dea25928cdacbed766f65e5130edff8/controller/functionApi.go#L47)):*
> ```Go
> 	}
> 
> 	funcs, err := a.fissionClient.Functions(metav1.NamespaceAll).List(options)
> ```

From the usage of `isOrphanPackage I am afraid we encounter the same problem related to `kube-api-qps` of api server (Please refer: https://github.com/fission/fission/pull/376). Cloud you fire an issue to track this problem? And let's try to solve this in another PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/458)*

life1347:
Hey @smruthi2187 , looks good so far. And please address the comments above, Thanks :)

PR #468, opened 2018-01-31
Delete and list orphan pkgs

smruthi2187:
1. Changes to delete the package with `removeOrphan` flag during `fn update`
2. Two CLIs - to list and to delete orphan packages with `fission package list --orphan` and `fission package delete --orphan`


life1347:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[fission/function.go, line 350 at r1](https://reviewable.io:443/reviews/fission/fission/468#-L49QazSQtQynr0H22e2:-L49QazTte-BVgw3YGLf:bhtltv9) ([raw file](https://github.com/fission/fission/blob/6ffdab514aee78ea35a2721dbce2a3ab40e5f70b/fission/function.go#L350)):*
> ```Go
> func fnDelete(c *cli.Context) error {
> 	client := getClient(c.GlobalString("server"))
> ```

We also need to provide flag `removeOrphan` when `fn delete`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/468)*

life1347:
Hey @smruthi2187 , thanks for the PR. Please address the comment above and I think it's ok to merge after the changes.

life1347:
*[fission/function.go, line 324 at r1](https://reviewable.io:443/reviews/fission/fission/468#-L4KzUybkO7dqMXcSxBW:-L4KzUybkO7dqMXcSxBX:b-cl2naz) ([raw file](https://github.com/fission/fission/blob/4390f2c7efc493a932dd04afac0ce6aa8a059be4/fission/function.go#L324)):*
> ```Go
> 			}
> 		}
> 	} else if len(pkgName) != 0 && removeOrphan {
> ```

I found a potential problem for this if condition

In a case that a package is used by only one function and users enter the following command
```
fission fn update --name test --entrypoint "xxxx" --removeorphan
```
the package will be deleted and caused some problems.
We better add a check to check whether pkgName is equal to function.Spec.Package.PackageRef.Name or not. 
If matched, then ignore the flag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/468)*

life1347:
LGTM

PR #483, opened 2018-02-06
Fix possible context leak problem

life1347:
Go vet logs
```
$ ./hack/verify-govet.sh
./fission/spec.go:431: the pkgWatchCancel function is not used on all paths (possible context leak)
./fission/spec.go:473: this return statement may be reached without using the pkgWatchCancel var defined on line 431
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/483)

### DIFF
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -33,8 +34,6 @@ import (
 	"github.com/satori/go.uuid"
 	"github.com/urfave/cli"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"io/ioutil"
 
 	"github.com/fission/fission"
 	"github.com/fission/fission/controller/client"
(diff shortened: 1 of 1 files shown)


erwinvaneyk:
Why do we create a context on each iteration by the way? You can just move the context outside of the for-loop; that avoids the context leaking as well. (https://github.com/fission/fission/blob/8e5d63b3743b22504abfb839fe376cc620204ab9/fission/spec.go#L431)

life1347:
No, we cannot do that or PackageBuildWatcher will create tons of gorountines that never stop.
https://github.com/fission/fission/blob/8e5d63b3743b22504abfb839fe376cc620204ab9/fission/spec.go#L435-L438

erwinvaneyk:
You are right @life1347 :) I misread the function - it is more complex than I thought.

soamvasani:
Thanks for fixing this @life1347 , and sorry for merging my change without checking go vet.

fission/spec.go
```diff
@@ -441,6 +441,7 @@ func specApply(c *cli.Context) error {
 		}
 
 		if !watchResources {
+			pkgWatchCancel()
```
soamvasani: Won't this crash in the case that (watchResources==false && waitForBuild==false)?

life1347: You're right, didn't notice this problem. Fix in #483

life1347: I moved L431 out of if condition, it's more clean and easy to understand.
https://github.com/fission/fission/blob/fe9453a92eec70ff63dd38166da84e7d85c7381e/fission/spec.go#L429-L433

PR #485, opened 2018-02-06
Fix potential nil pointer problem

life1347:
Fix potential nil pointer problem when both watchResources and waitForBuild are equal to false.
I moved L431 out of if condition, it's more clean and easy to understand.

https://github.com/fission/fission/blob/fe9453a92eec70ff63dd38166da84e7d85c7381e/fission/spec.go#L427-L433

https://github.com/fission/fission/blob/fe9453a92eec70ff63dd38166da84e7d85c7381e/fission/spec.go#L443-L446

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/485)

### DIFF
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -424,13 +424,12 @@ func specApply(c *cli.Context) error {
 		checkErr(err, "apply specs")
 		printApplyStatus(as)
 
-		var ctx context.Context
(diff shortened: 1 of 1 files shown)


soamvasani:
LGTM, thanks, this is much better than my version.

PR #487, opened 2018-02-07
Fix binary environment build failure due to package not found

life1347:
Subpackage k8s.io/client-go/pkg/api/v1 is no longer exist in HEAD.
(https://github.com/kubernetes-incubator/external-storage/issues/324)
For now, just duplicate the FunctionLoadRequest struct like what go environment did.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/487)

### DIFF
--- a/environments/binary/server.go
+++ b/environments/binary/server.go
@@ -11,8 +11,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-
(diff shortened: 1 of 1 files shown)


soamvasani:
I think sooner or later the environments (binary and go) will have to use glide or equivalent.

PR #489, opened 2018-02-07
Version -> 0.5.0

life1347:
Will merge it after final check

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/489)

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -8,10 +8,11 @@ BUILDDIR=$(realpath $DIR)/build
 
 # Ensure we're on the master branch
 check_branch() {
+    version=$1
     curr_branch=$(git rev-parse --abbrev-ref HEAD)
-    if [ $curr_branch != "master" ]
+    if [ $curr_branch != "v${version}" ]
     then
-	echo "Not on master branch."
+	echo "Not on v${version} branch."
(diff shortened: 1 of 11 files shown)


hack/release.sh
```diff
@@ -8,10 +8,11 @@ BUILDDIR=$(realpath $DIR)/build
 
 # Ensure we're on the master branch
 check_branch() {
+    version=$1
     curr_branch=$(git rev-parse --abbrev-ref HEAD)
-    if [ $curr_branch != "master" ]
+    if [ $curr_branch != "v${version}" ]
```
soamvasani: Can you elaborate on this a bit?  Why is this better than releasing from master?

life1347: If we release from master branch, we need to merge changes to docs, charts and things related to fission version into master branch first. And then add changelog.md in a separate commit.
If we release from a branch other than the master (i.e v0.5.0), we can commit changes in that branch first. And after some final check (or review) we can merge these changes into master with only one commits rather than multiple ones.

soamvasani: Why can't we add the changelog in the same commit? Does the changelog generator depend on there being a version branch?

Issue #472, opened 2018-02-01
Test Stabilisation and optimisation


vishal-biyani:
The tests have been a little flaky and we need to improve certain things. This issue is a placeholder for all issues devs have faced and  work can be split into multiple PRs potentially.

1) Port binding issues

There are instances where during installation of fission chart, the ports cannot be assigned which causes the build failures. It is not clear if this is the problem or side effect of some other problem but this has been seen multiple times across multiple PR builds.

```
 $ curl http://$FISSION_ROUTER/hello
  Hello, world!
E0113 14:11:02.968228   25185 portforward.go:212] Unable to create listener: Error listen tcp6 [::1]:38644: bind: cannot assign requested address
```
Sample failed build: https://travis-ci.org/fission/fission/builds/328236257 

2) Intermittant failures

Some tests fail intermittently - mostly because the operation expected at some time does not finish. Some samples:

https://travis-ci.org/fission/fission/builds/336053006
https://travis-ci.org/fission/fission/builds/335775155

3) Debugging info not available

In some cases, if a helm chart installation takes too long (More than 10 minutes) - the build fails. But this information is not sufficient to solve the probelm. The sample build below for example had a wrong image name and the helm chart installation kept waiting forever for the image to be pulled. But this information is only visible in tiller pod log. If there is a mechanism to tail some of the tiller pod logs into Travis build log, it will be helpful.

https://travis-ci.org/fission/fission/builds/335474547

4) Tiller logs not verbose about issues

In case of (2) above, even tiller logs were not verbose enough. I had to actually go to describe pod and then the problem was found. There could be potentially two ways we could enhance this. One is to increase the verbosity of the tiller in the CI cluster. Another option could be to describe all objects created as part of helm chart installation if there is a failure/timeout.

vishal-biyani:
@life1347 @soamvasani @erwinvaneyk Please add if you see any other cases not covered in issue.

soamvasani:
Thanks Vishal, this is a *great* start.

> E0113 14:11:02.968228   25185 portforward.go:212] Unable to create listener: Error listen tcp6 [::1]:38644: bind: cannot assign requested address

I think this error is usually harmless, e.g. from a successful `helm delete`:
```
E0201 13:39:34.135626   32029 portforward.go:212] Unable to create listener: Error listen tcp6 [::1]:45791: bind: cannot assign requested address
release "96ed08" deleted
namespace "f-96ed08" deleted
```

[...]

PR #491, opened 2018-02-08
CI modifications

smruthi2187:
This PR is for https://github.com/fission/fission/issues/472

1. Readiness and liveness probes are added to most fission containers, along with corresponding `healthz` http handlers.
2. A signal Handler for SIGTERM (used by liveness probe to restart containers) is added to most fission containers so we have enough information to debug issues. 
(I found that the above 2 helped troubleshoot/get rid of most of the `getsockopt` errors in my CI runs.)
4. For helm install failures,  `kubernetes events` and `tiller logs`( of previous 30 minutes ) are added to the CI logs.
5. Helm install timeout has been changed from 10 minutes to 9 minutes to prevent travis from killing the build script before dumping all logs and objects.( this because travis has a build timeout of 10 minutes too)
3. Added `--quiet` option to some commands in the test to gain some space in CI log since its limited in size.

With these changes, 4 continuous CI runs passed on this branch. ( https://travis-ci.org/fission/fission/branches for proof :))
These changes in no way have solved any existing bugs. But only has some extra logs and debugging information.
Oh and the `Unable to create listener: Error listen tcp6 [::1]:38644: bind: cannot assign requested address` warnings are gone!!

Also, changing the docker repo for ubuntu_slim docker image is coming in a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/491)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ services:
 before_install:
   - sudo apt-get update
   - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+  - sudo sysctl net.ipv6.conf.all.disable_ipv6=0
 
 install:
   - go get github.com/Masterminds/glide
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: required
 
 dist: trusty
 
+branches:
+  only:
+  - master
+
 language: go
 
 go:
@@ -14,10 +18,6 @@ cache:
 
 services:
   - docker
-
-#branches:
-#  only:
-#  - master
 
 before_install:
   - sudo apt-get update
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ cache:
 services:
   - docker
 
+#branches:
+#  only:
+#  - master
+
 before_install:
   - sudo apt-get update
   - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -169,7 +169,7 @@ spec:
             port: "8888"
           initialDelaySeconds: 5
           periodSeconds: 2
-         livenessProbe:
+        livenessProbe:
           httpGet:
             path: "/router-healthz"
             port: "8888"
@@ -226,7 +226,7 @@ spec:
             port: "8888"
           initialDelaySeconds: 5
           periodSeconds: 2
-         livenessProbe:
+        livenessProbe:
           httpGet:
             path: "/healthz"
             port: "8888"
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -129,6 +129,18 @@ spec:
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
         args: ["--controllerPort", "8888"]
+        readinessProbe:
+          httpGet:
+            path: "/healthz"
+            port: "8888"
+          initialDelaySeconds: 5
+          periodSeconds: 2
+        livenessProbe:
+          httpGet:
+            path: "/healthz"
+            port: "8888"
+          initialDelaySeconds: 16
+          periodSeconds: 5
       serviceAccount: fission-svc
 
 ---
@@ -151,6 +163,18 @@ spec:
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
         args: ["--routerPort", "8888", "--executorUrl", "http://executor.{{ .Release.Namespace }}"]
+        readinessProbe:
+          httpGet:
+            path: "/router-healthz"
+            port: "8888"
+          initialDelaySeconds: 5
+          periodSeconds: 2
+         livenessProbe:
+          httpGet:
+            path: "/router-healthz"
+            port: "8888"
+          initialDelaySeconds: 16
+          periodSeconds: 5
       serviceAccount: fission-svc
 
 ---
@@ -196,6 +220,18 @@ spec:
           value: "{{ .Values.pullPolicy }}"
         - name: RUNTIME_IMAGE_PULL_POLICY
           value: "{{ .Values.pullPolicy }}"
+        readinessProbe:
+          httpGet:
+            path: "/healthz"
+            port: "8888"
+          initialDelaySeconds: 5
+          periodSeconds: 2
+         livenessProbe:
+          httpGet:
+            path: "/healthz"
+            port: "8888"
+          initialDelaySeconds: 16
+          periodSeconds: 5
       serviceAccount: fission-svc
 
 ---
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -132,13 +132,13 @@ spec:
         readinessProbe:
           httpGet:
             path: "/healthz"
-            port: "8888"
+            port: 8888
           initialDelaySeconds: 5
           periodSeconds: 2
         livenessProbe:
           httpGet:
             path: "/healthz"
-            port: "8888"
+            port: 8888
           initialDelaySeconds: 16
           periodSeconds: 5
       serviceAccount: fission-svc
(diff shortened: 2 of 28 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ sudo: required
 
 dist: trusty
 
+branches:
+  only:
+  - master
```
soamvasani: Wait doesn't this end up stopping pull request builds?

charts/fission-all/templates/deployment.yaml
```diff
@@ -129,6 +129,19 @@ spec:
...
+            path: "/healthz"
+            port: 8888
+          initialDelaySeconds: 1
+          periodSeconds: 1
+          failureThreshold: 30
+        livenessProbe:
+          httpGet:
+            path: "/healthz"
+            port: 8888
+          initialDelaySeconds: 35
```
soamvasani: 👍 I initially thought that liveness probes start after the pod is ready, but they don't; so this value is correct.

charts/fission-all/templates/deployment.yaml
```diff
@@ -151,6 +164,19 @@ spec:
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
         args: ["--routerPort", "8888", "--executorUrl", "http://executor.{{ .Release.Namespace }}"]
+        readinessProbe:
+          httpGet:
+            path: "/router-healthz"
```
soamvasani: why not just /healthz?

executor/executor.go
```diff
@@ -181,9 +185,23 @@ func (executor *Executor) getFunctionEnv(m *metav1.ObjectMeta) (*crd.Environment
 	return env, nil
 }
 
+func dumpStackTrace() {
```
soamvasani: this and the signal/notify goroutine are repeated a few times.  Mind moving them to somewhere common and having just a call like `fissionDebug.setupStackTraceHandler()` in StartWhatever() functions?

executor/poolmgr/gp.go
```diff
@@ -519,6 +519,33 @@ func (gp *GenericPool) createPool() error {
 								"-secret-dir", gp.sharedSecretPath,
 								"-cfgmap-dir", gp.sharedCfgMapPath,
 								gp.sharedMountPath},
+							ReadinessProbe: &apiv1.Probe{
```
soamvasani: I thought you told me that this can't be added, because existing envs would break?

router/httpTriggers.go
```diff
@@ -150,6 +154,9 @@ func (ts *HTTPTriggerSet) getRouter() *mux.Router {
 		muxRouter.HandleFunc(fission.UrlForFunction(function.Metadata.Name), fh.handler)
 	}
 
+	// Healthz endpoint for the router.
+	muxRouter.HandleFunc("/router-healthz", routerHealthHandler).Methods("GET")
```
soamvasani: Just /healthz would work

storagesvc/stowClient.go
```diff
@@ -60,6 +60,7 @@ var (
 )
 
 func MakeStowClient(storageType StorageType, storagePath string, containerName string) (*StowClient, error) {
+	log.Infof("start : MakeStowClient")
```
soamvasani: not really needed, or at least make it a debug log

storagesvc/stowClient.go
```diff
@@ -104,6 +105,7 @@ func MakeStowClient(storageType StorageType, storagePath string, containerName s
 	}
 	stowClient.container = con
 
+	log.Infof("end : MakeStowClient")
```
soamvasani: remove, or set to debug

test/test_utils.sh
```diff
@@ -174,14 +174,16 @@ helm_install_fission() {
     echo "Deleting old releases"
     helm list -q|xargs -I@ bash -c "helm_uninstall_fission @"
 
+    # deleting ns does take a while after command is issued.
+    sleep 45
```
soamvasani: this is a pretty long wait to be doing unconditionally, something like this might be better?
```
while kubectl get ns | grep $ns # or $fns i'm not sure
do 
    sleep 5
done
```

test/test_utils.sh
```diff
@@ -174,14 +174,16 @@ helm_install_fission() {
...
+
     echo "Installing fission"
     helm install		\
 	 --wait			\
-	 --timeout 600	        \
+	 --timeout 540	        \
 	 --name $id		\
 	 --set $helmVars	\
 	 --namespace $ns        \
-	 --debug                \
```
soamvasani: isn't --debug useful?

test/test_utils.sh
```diff
@@ -190,29 +192,64 @@ helm_install_fission() {
...
+        retry=$((retry+1))
+        if ((retry == max_retries)); then
+            echo "Waiting for $svc to be routable exceeded max retries. Quitting.."
+            exit 1
+        fi
+        ip=$(kubectl -n $ns get svc $svc -o jsonpath='{...ip}')
+        if [ -z $ip ]; then
+            continue
+        fi
+        http_status=`curl -sw "%{http_code}" "http://$ip/$health_endpoint"`
```
soamvasani: 👍 this is a good thing to do but I'm curious if it ever fails. (Since helm --wait should be waiting for pods to get ready, and the pods are only ready after http is up because of the readiness probes.)

test/test_utils.sh
```diff
@@ -402,7 +464,9 @@ install_and_test() {
     trap "helm_uninstall_fission $id" EXIT
     if ! helm_install_fission $id $image $imageTag $fetcherImage $fetcherImageTag $controllerPort $routerPort $fluentdImage $fluentdImageTag $pruneInterval
     then
-	dump_logs $id
+	describe_all_pods $id
+	dump_kubernetes_events $id
+    dump_tiller_logs
```
soamvasani: indentation

test/test_utils.sh
```diff
@@ -417,6 +481,8 @@ install_and_test() {
 
     if [ $FAILURES -ne 0 ]
     then
+    # describe each pod in fission ns and function namespace
```
soamvasani: indentation

test/test_utils.sh
```diff
@@ -343,11 +399,17 @@ dump_logs() {
     dump_fission_logs $ns $fns router
     dump_fission_logs $ns $fns buildermgr
     dump_fission_logs $ns $fns executor
+    dump_fission_logs $ns $fns storagesvc
     dump_function_pod_logs $ns $fns
     dump_builder_pod_logs $bns
     dump_fission_crds
 }
 
+echo_log() {
```
soamvasani: why not just call it `log`? `echo_log` feels kinda odd to write.

.travis.yml
smruthi2187: Its funny I suspected the same thing initially but then I checked that this is a separate repo setting called `Build pushed Pull Request` here : https://travis-ci.org/fission/fission/settings
https://docs.travis-ci.com/user/pull-requests/ for reference and this PR did run a CI test.

charts/fission-all/templates/deployment.yaml
smruthi2187: Oh because there might be a function that might want to create a route to `/healthz`. so I didnt want to hog up healthz :) I can change it to `healthz`.

executor/executor.go
smruthi2187: sure. I'll do that.

executor/poolmgr/gp.go
smruthi2187: I added it only to fetcher container. We must soon make it mandatory for all environments too. right?

test/test_utils.sh
smruthi2187: I just felt the info it dumped was repeated later as part of the `dump objects` and `dump logs` functions. Also, it used up 801 lines in the log. I can add it back if its needed.

test/test_utils.sh
smruthi2187: yes that's true, but readiness probe relies on `PodIP` and this function ensures the service `externalIPs` are indeed routing to the pods correctly.

test/test_utils.sh
smruthi2187: oops, I'll fix it.

test/test_utils.sh
smruthi2187: oops, I'll fix it.

test/test_utils.sh
smruthi2187: yeah, now that you mentioned it. sure, I'll change it to log.

test/test_utils.sh
soamvasani: Makse sense, no need to add it back then.

soamvasani:
LGTM, please merge when ready

PR #500, opened 2018-02-19
Adds default resources for fetcher pod

### LINKED ISSUE #478: Default ResourceRequirements for Fetcher container
In NewDeploy executor type, currently, Fetcher resources are same as function container. Instead, Fetcher should use smaller and default resources. (Without ResourceRequirements on both containers, scaling won't work)

In Poolmgr executor type, the resources are unassigned to fetcer container. Defaults should be assigned to fetcher container.

vishal-biyani:
This PR adds default resources for Fetcher pod for both executors - pool manager and new deployment (And fixes #478 ). Currently, the resource numbers are based on experimentation. In future maybe we could look at the numbers from monitoring (Prometheus?)

Also, the getFetcherResources method is duplicated in both executors. On one hand, I wanted different fetcher resources for both but also wanted to avoid code duplication. I can not move the method up at executor level as it will cause cyclical dependency of packages, maybe create a util package within executor and then import in both executors?

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/500)

### DIFF
--- a/executor/newdeploy/newdeploy.go
+++ b/executor/newdeploy/newdeploy.go
@@ -25,8 +25,10 @@ import (
 	"time"
 
 	k8s_err "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/pkg/api/v1"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
 	asv1 "k8s.io/client-go/pkg/apis/autoscaling/v1"
 	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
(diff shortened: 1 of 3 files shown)


executor/newdeploy/newdeploy.go
```diff
@@ -45,6 +47,30 @@ const (
 	envVersion = "ENV_VERSION"
 )
 
+func (deploy *NewDeploy) getFetcherResources() (v1.ResourceRequirements, error) {
+	//TBD Hardcoded as of now, should be configurable?
+	mincpu, err := resource.ParseQuantity("10m")
+	minmem, err := resource.ParseQuantity("16Mi")
+	maxcpu, err := resource.ParseQuantity("40m")
+	maxmem, err := resource.ParseQuantity("128Mi")
```
erwinvaneyk: you can move this outside of the function (either ignore the errors as they are hardcoded anyway, or check in the `init()`, or wrap them in a helper function (like `MustParseQuantity`))

executor/newdeploy/newdeploy.go
```diff
@@ -87,6 +113,12 @@ func (deploy *NewDeploy) createOrGetDeployment(fn *crd.Function, env *crd.Enviro
 			return nil, err
 		}
 
+		fetcherResources, err := deploy.getFetcherResources()
+		if err != nil {
+			log.Printf("Error while parsing fetcher resources: %v", err)
```
erwinvaneyk: Why print here, rather than wrapping the `err`?

executor/poolmgr/gp.go
```diff
@@ -715,3 +723,27 @@ func (gp *GenericPool) destroy() error {
 
 	return nil
 }
+
+func (gp *GenericPool) getFetcherResources() (v1.ResourceRequirements, error) {
```
erwinvaneyk: Extract this to a common/util file + it doesn't seem to be needed to attach this function the GenericPool.

executor/newdeploy/newdeploy.go
```diff
@@ -214,6 +246,10 @@ func (deploy *NewDeploy) createOrGetHpa(hpaName string, execStrategy *fission.Ex
 		return existingHpa, err
 	}
 
+	if depl == nil {
+		return nil, errors.New("Failed to create HPA, found empty deployment")
```
erwinvaneyk: Minor: try to stick with idiomatic go: error messages should not be capitalized.

vishal-biyani:
Thanks, @erwinvaneyk !

executor/poolmgr/gp.go
```diff
@@ -715,3 +723,27 @@ func (gp *GenericPool) destroy() error {
...
+	//TBD Hardcoded as of now, should be configurable?
```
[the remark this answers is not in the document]
soamvasani: Ok to do in a later change if you like.

Issue #506, opened 2018-02-22
poolsize environment attribute breaks upgrades


soamvasani:
The addition of the poolsize attribute to environment spec broke upgrades.  Here's why:

 1. New environment spec has a poolsize attribute
 1. But old environments don't have poolsize defined
 1. Reading old environments into the new env spec yields poolsize equal to zero
 1. Poolsize==0 causes functions using poolmgr to fail 

There's a lot of options on how to fix this:

 * Option 1: Use env versioning -- create a new environment version (3) and only read poolsize when version>=3. 
 * Option 2: Ignore poolsize==0 when functions exist that use poolmgr
 * Option 3: Dispatch functions to newdeploy executor backend when poolsize==0.

Option 1 seems cleanest to think about.

vishal-biyani:
Closed by PR.

soamvasani:
Wouldn't changing int to int* also be a compatiblity-breaking change?

life1347:
yes, so we may need to add another field for this purpose.

PR #508, opened 2018-02-23
Fixes the backward compatibility with older environment versions

vishal-biyani:
This fixes the compatibility part of the issue #506. There are two things that had to be done to make this work in a correct way:

- Add a check that if `env.Spec.Version < 3` then default `poolsize` to 3
- Since we changed the name from `poolmgrInstanceId` to `executorInstanceId` - we had to add a check for previously created deployments etc. to be cleaned up. Technically this is one time activity after upgrade and can be done manually - so can remove if that is what we conclude to do. Or we can leave for 2 releases and then remove it.

The testing that is done to prove that change works is described below:

### Test with 0.4.0

**Install Helm**
```helm install --name fission --namespace fission https://github.com/fission/fission/releases/download/0.4.1/fission-all-0.4.1.tgz```

**Create environment & verify**

```
$ fission env create --name node --image fission/node-env:latest
environment 'node' created

$ k -n default get environment node -o yaml
apiVersion: fission.io/v1
kind: Environment
metadata:
  clusterName: ""
  creationTimestamp: 2018-02-23T08:22:01Z
  deletionGracePeriodSeconds: null
  deletionTimestamp: null
  initializers: null
  name: node
  namespace: default
  resourceVersion: "5625578"
  selfLink: /apis/fission.io/v1/namespaces/default/environments/node
  uid: 9f5703df-1872-11e8-9bd0-42010aa00010
spec:
  allowedFunctionsPerContainer: ""
  builder:
    command: ""
    image: ""
  documentationurl: ""
  runtime:
    functionendpointport: 0
    image: fission/node-env:latest
    loadendpointpath: ""
    loadendpointport: 0
  version: 1


$ fpod
NAME                                                              READY     STATUS    RESTARTS   AGE
node-9f5703df-1872-11e8-9bd0-42010aa00010-xouku8bf-7cf6549f6vv8   2/2       Running   0          1m
node-9f5703df-1872-11e8-9bd0-42010aa00010-xouku8bf-7cf6549m57cq   2/2       Running   0          1m
node-9f5703df-1872-11e8-9bd0-42010aa00010-xouku8bf-7cf6549npc75   2/2       Running   0          1m
```
**Test with a function**

```
$ fission fn create --name hello --env node --code ../../fission_work/hello.js
function 'hello' created

$ curl http://$FISSION_ROUTER/fission-function/hello
Hello, Fission!
```

### Reproducing the issue

If we upgrade to the current master on top of the above-installed release, you will run into:

```
$ fission env list
NAME UID                                  IMAGE                   POOLSIZE MINCPU MAXCPU MINMEMORY MAXMEMORY
node 9f5703df-1872-11e8-9bd0-42010aa00010 fission/node-env:latest 0        0      0      0         0

$ curl http://$FISSION_ROUTER/fission-function/hello
Internal server error (fission)
```

### After fix (This PR)

```
$ fission env list
NAME UID                                  IMAGE                   POOLSIZE MINCPU MAXCPU MINMEMORY MAXMEMORY
node 9f5703df-1872-11e8-9bd0-42010aa00010 fission/node-env:latest 0        0      0      0         0

$ curl http://$FISSION_ROUTER/fission-function/hello
Hello, Fission!
```
You can also notice the pool getting created despite the poolsize being zero, because the env version is 1:
```
$ fpod
NAME                                                              READY     STATUS    RESTARTS   AGE
node-9f5703df-1872-11e8-9bd0-42010aa00010-r5i5ovyf-f5987dc5pm57   2/2       Running   0          29m
node-9f5703df-1872-11e8-9bd0-42010aa00010-r5i5ovyf-f5987dcctqxn   2/2       Running   0          26m
node-9f5703df-1872-11e8-9bd0-42010aa00010-r5i5ovyf-f5987dcln2z2   2/2       Running   0          29m
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/508)

### DIFF
--- a/executor/poolmgr/gpm.go
+++ b/executor/poolmgr/gpm.go
@@ -89,14 +89,14 @@ func (gpm *GenericPoolManager) service() {
 			var err error
 			pool, ok := gpm.pools[crd.CacheKey(&req.env.Metadata)]
 			if !ok {
-				var poolSize = int32(req.env.Spec.Poolsize)
+				poolsize := gpm.getEnvPoolsize(req.env)
 				switch req.env.Spec.AllowedFunctionsPerContainer {
 				case fission.AllowedFunctionsPerContainerInfinite:
-					poolSize = 1
+					poolsize = 1
 				}
 
 				pool, err = MakeGenericPool(
-					gpm.fissionClient, gpm.kubernetesClient, req.env, poolSize,
+					gpm.fissionClient, gpm.kubernetesClient, req.env, poolsize,
 					gpm.namespace, gpm.fsCache, gpm.instanceId)
 				if err != nil {
 					req.responseChannel <- &response{error: err}
@@ -110,7 +110,7 @@ func (gpm *GenericPoolManager) service() {
 			latestEnvPoolsize := make(map[string]int)
 			for _, env := range req.envList {
 				latestEnvSet[crd.CacheKey(&env.Metadata)] = true
-				latestEnvPoolsize[crd.CacheKey(&env.Metadata)] = env.Spec.Poolsize
+				latestEnvPoolsize[crd.CacheKey(&env.Metadata)] = int(gpm.getEnvPoolsize(&env))
 			}
 			for key, pool := range gpm.pools {
 				_, ok := latestEnvSet[key]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: executor/poolmgr/gpm.go [hunk 1]
```diff
@@ -178,3 +178,13 @@ func (gpm *GenericPoolManager) eagerPoolCreator() {
 		gpm.CleanupPools(envs.Items)
 	}
 }
+
+func (gpm *GenericPoolManager) getEnvPoolsize(env *crd.Environment) int32 {
```
soamvasani: Just curious, why does this return int32 and not int?

vishal-biyani: This is because the `Replicas` in `DeploymentSpec` needs a int32

vishal-biyani: @soamvasani I will merge this if looks good to you?

PR #512, opened 2018-02-26
Fix CLI not update function's secret/configmap correctly

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/512)


vishal-biyani:
Are we going to make this to 0.6.0?

life1347:
@vishal-biyani Thank you for point this out! And yes, it's for 0.6.0.

PR #519, opened 2018-03-01
Detect fission namespace in cli

soamvasani:
This PR removes the hard coded default "fission" namespace that I added in the last PR, replacing it with some logic to figure out where fission is installed and connecting to it.  If there is more than one fission install, error out with a useful message.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/519)


soamvasani:
That test failure is unrelated, filed issue #521

Issue #521, opened 2018-03-01
panic: segfault in executor while creating/updating a newdeploy function


soamvasani:
https://api.travis-ci.org/v3/job/347778131/log.txt
```
------- Running /test_secret_cfgmap/test_secret_cfgmap.sh -------
~/gopath/src/github.com/fission/fission/test/tests/test_secret_cfgmap ~/gopath/src/github.com/fission/fission
2018/03/01:14:17:19 Pre-test cleanup
Forwarding from 127.0.0.1:8888 -> 8888
Failed to delete environment: Delete failed: NotFound

2018/03/01:14:17:21 Creating python env
environment 'python' created
2018/03/01:14:17:22 Creating secret
secret "testsecret-1519913839" created
2018/03/01:14:17:22 Creating function with secret
function 'testsecret-1519913839' created
2018/03/01:14:17:22 Creating route
trigger 'f617ceb6-193f-4cbd-9894-ea7d400ab680' created
2018/03/01:14:17:23 Waiting for router to catch up
2018/03/01:14:17:28 Doing an HTTP GET on the function's route
  % Total    % Received % Xferd Handling connection for 8888
 Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:06 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:07 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:08 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:09 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:10 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:11 --:--:--     0
[...]

PR #523, opened 2018-03-02
Check component pods status and wait until ready

life1347:
This PR aims to solve https://github.com/fission/fission/issues/521

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/523)

### DIFF
--- a/test/test_utils.sh
+++ b/test/test_utils.sh
@@ -193,6 +193,26 @@ helm_install_fission() {
     helm list
 }
 
+wait_for_components() {
+    id=$1
+    ns=f-$id
+    while true; do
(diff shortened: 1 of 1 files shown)


soamvasani:
Hmm, with the readiness probes and the install being done with `helm install --wait ...`, I'm not sure this would be any different.
```
helm install --help
...
      --wait                   if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment are in a ready state before marking the release as successful. It will wait for as long as --timeout
```

life1347:
I found the root cause for #521 , it's related to function update with newdeploy executorType. Now trying to solve the problem.

PR #524, opened 2018-03-02
Fix executor tries to create a new deployment when a function is updated

life1347:
This PR aims to solve the segfault problem (#521 ) when a function with newdeploy executorType is updated.

Root cause:
The initial idea for newdeploy manager to update deployment and hpa when a function is updated. So the root causes of the problem are:
1. The fscache in executor and newdeploy manager saved fsvc with cache key which contains function resource version.
2. Newdeploy manager doesn't check the existence of kubeobjs of function

How to fix:
Add a small cache `fnRecords` to save fsvc & function info with cache key which is consist of `<function name>-<executor id>`. Then when newdeploy manager checks the `fnRecords` before trying to create kubeobjs.

Though the problem can be fixed with the current solution, what `fnRecords` does is quite similar to `fscache`. It will be a big change if we try to modify function service cache. So, maybe we can have some discussion on how to modify function service cache.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/524)

### DIFF
--- a/executor/cleanup.go
+++ b/executor/cleanup.go
@@ -117,6 +117,7 @@ func idleObjectReaper(kubeClient *kubernetes.Clientset,
 			for _, fsvc := range funcSvcs {
 
 				fn, err := fissionClient.Functions(fsvc.Function.Namespace).Get(fsvc.Function.Name)
+
 				if err != nil {
 					log.Printf("Error getting function: %v", fsvc.Function.Name)
 					continue
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -136,8 +136,6 @@ func (executor *Executor) createServiceForFunction(meta *metav1.ObjectMeta) (*fs
 		return nil, err
 	}
 
-	log.Printf("Trying to get function service: %v", fn.Metadata.Name)
-
 	switch fn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType {
 	case fission.ExecutorTypeNewdeploy:
 		fs, err := executor.ndm.GetFuncSvc(meta)
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -136,6 +136,8 @@ func (executor *Executor) createServiceForFunction(meta *metav1.ObjectMeta) (*fs
 		return nil, err
 	}
 
+	log.Printf("Trying to get function service: %v", fn.Metadata.Name)
+
 	switch fn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType {
 	case fission.ExecutorTypeNewdeploy:
 		fs, err := executor.ndm.GetFuncSvc(meta)
--- a/executor/fscache/functionServiceCache.go
+++ b/executor/fscache/functionServiceCache.go
@@ -205,8 +205,16 @@ func (fsc *FunctionServiceCache) Add(fsvc FuncSvc) (*FuncSvc, error) {
 		return nil, err
 	}
 
+	// Add to byFunctionUID cache. Ignore NameExists errors
+	// because of multiple-specialization. See issue #331.
 	err, _ = fsc.byFunctionUID.Set(fsvc.Function.UID, *fsvc.Function)
 	if err != nil {
+		if IsNameExistError(err) {
+			err = nil
+		}
+		if err != nil {
+			log.Printf("error caching fsvc by function uid: %v", err)
+		}
 		return nil, err
 	}
 
--- a/executor/fscache/functionServiceCache.go
+++ b/executor/fscache/functionServiceCache.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/pkg/api"
 
 	"github.com/fission/fission"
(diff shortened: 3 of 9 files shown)


soamvasani:
LGTM except for some questions. The test still looks too forgiving of errors? Nice work catching this bug!

---

Reviewed 6 of 6 files at r1, 1 of 2 files at r2, 1 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[executor/newdeploy/newdeploy.go, line 55 at r3](https://reviewable.io:443/reviews/fission/fission/524#-L6x2LCF9dV0_AJ6zBI-:-L6x2LCG0XTMqcpkicMa:b41mra9) ([raw file](https://github.com/fission/fission/blob/1f6e0d77243841d5248dfd798ed55985e40121c4/executor/newdeploy/newdeploy.go#L55)):*
> ```Go
> 	deployName string, deployLabels map[string]string) (*v1beta1.Deployment, error) {
> 
> 	waitForDeploy := func(depl *v1beta1.Deployment, replicas int32) (*v1beta1.Deployment, error) {
> ```

This can be a normal function defined outside this function, right?

---

*[executor/newdeploy/newdeploymgr.go, line 179 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6wTVa9JbjnO3gG5Z1q:-L6wTVaAMaIWImGZtB13:b-7n8i05) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/executor/newdeploy/newdeploymgr.go#L179)):*
> ```Go
> 		case FnDelete:
> 			_, err := deploy.fnDelete(req.fn)
> 			//delete(deploy.fnRecords, objName)
> ```

commented line

---

*[executor/newdeploy/newdeploymgr.go, line 370 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6x1ogTw7N_u4HskAZA:-L6x1ogUWpssQy2W0acE:bq4o98x) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/executor/newdeploy/newdeploymgr.go#L370)):*
> ```Go
> 	}
> 
> 	deployChanged := false
> ```

can you move this nearer to where it's used? maybe just before hpaChanged=false

---

*[test/tests/test_backend_nd_update.sh, line 62 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6wRzS7FnQ8b0yzXokT:-L6wRzS7FnQ8b0yzXokU:b-tyjszd) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/test/tests/test_backend_nd_update.sh#L62)):*
> ```Shell
>     echo "Checking for valid response"
> 
>     while true; do
> ```

Hmm, why retry? Don't we want to fail on any error?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/524#-:-L6x4bo2iNI6F6-7WEkC:b-2mv1xz)*

life1347:
It will retry until the expected value is returned or the test will fail after 60s (timeout).

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[executor/newdeploy/newdeploy.go, line 55 at r3](https://reviewable.io:443/reviews/fission/fission/524#-L6x2LCF9dV0_AJ6zBI-:-L7-exPg7abqtnlFRhDj:b-896fix) ([raw file](https://github.com/fission/fission/blob/1f6e0d77243841d5248dfd798ed55985e40121c4/executor/newdeploy/newdeploy.go#L55)):*

This can be a normal function defined outside this function, right?
</blockquote></details>

Done.

---

*[executor/newdeploy/newdeploymgr.go, line 179 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6wTVa9JbjnO3gG5Z1q:-L7-f6fGGig8I8oX3VE8:b-896fix) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/executor/newdeploy/newdeploymgr.go#L179)):*

commented line
</blockquote></details>

Done.

---

*[executor/newdeploy/newdeploymgr.go, line 370 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6x1ogTw7N_u4HskAZA:-L7-fpSTf5aFBTdUgnrk:b-k57in5) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/executor/newdeploy/newdeploymgr.go#L370)):*

can you move this nearer to where it's used? maybe just before hpaChanged=false
</blockquote></details>

No, this variable is used in the if statement.
https://github.com/fission/fission/pull/524/files#diff-052ae1a0c8f7c2d81207f00a5948bb06R405

---

*[test/tests/test_backend_nd_update.sh, line 62 at r1](https://reviewable.io:443/reviews/fission/fission/524#-L6wRzS7FnQ8b0yzXokT:-L7-ge2ShmxjwkOZZxT-:b9p50no) ([raw file](https://github.com/fission/fission/blob/0715fac3d69db94742d5cf68a9c18ea9e43bf42c/test/tests/test_backend_nd_update.sh#L62)):*

Hmm, why retry? Don't we want to fail on any error?
</blockquote></details>

The idea here is that it takes time for newdeploy manager to update the deployment of function and deployment needs time to perform the rolling update. So the response with old value may be returned when the first request to the function after the update. That's I add a retry here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/524#-:-L7-hzI19piN5yIqjQQc:b-tsckae)*

soamvasani:
LGTM, thanks!

---

Reviewed 3 of 3 files at r4.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/524#-:-L70lZoW7iNv1ODIlnu4:b2c0zcb)*

PR #527, opened 2018-03-05
Proposal for versioning.

smruthi2187:
This is a proposal for versioning functions and packages.

### DIFF
--- a/Documentation/wip/VersioningProposal.md
+++ b/Documentation/wip/VersioningProposal.md
@@ -1,0 +1,1221 @@
+Table of Contents :
+
+* [Use cases of creating versions of functions and packages](#use-cases-of-creating-versions-of-functions-and-packages)
+* [High level proposal](#high-level-proposal)
+* [Implementation details](#implementation-details)
+  * [Implementation details for fission spec](#implementation-details-for-fission-spec)
+     * [TL;DR of intended behavior of fission code through different scenarios](#tldr-of-intended-behavior-of-fission-code-through-different-scenarios)
+        * [Version creation](#version-creation)
+        * [Version deletion](#version-deletion)
+     * [Pseudocode for spec](#pseudocode-for-spec)
+  * [Implementation details for fission cli commands](#implementation-details-for-fission-cli-commands)
+     * [TL;DR walking you through various scenarios](#tldr-walking-you-through-various-scenarios)
+        * [Version create for packages](#version-create-for-packages)
+        * [Version create for function](#version-create-for-function)
+        * [Version update for packages](#version-update-for-packages)
+        * [Version update for functions](#version-update-for-functions)
+        * [Version delete for packages](#version-delete-for-packages)
+        * [Version delete for functions](#version-delete-for-functions)
+        * [Version list for packages](#version-list-for-packages)
+        * [Version list for functions](#version-list-for-functions)
+     * [Pseudocode](#pseudocode)
+* [Note](#note)
+* [Conclusion](#conclusion)
+   
+# Use cases of creating versions of functions and packages
+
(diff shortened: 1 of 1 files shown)


life1347:
Just some comments so far.

1. Maybe we can use replace deepCmp with hashSum like what Kubernetes does in pod template hash (https://github.com/mruediger/kubernetes/blob/master/pkg/util/pod/pod.go#L34-L38). It will be easier to implement and we don't need to change the code if we add more fields to the spec in future.

2. Will function versioning supports rollback? didn't see any place mentioned the function rollback or I missed that when meeting. A new object of type `GroupVersionTracker` will be created after each `fission spec version apply`, that means there are multiple revisions of `GroupVersionTracker`. Then the user can choose any one of it to perform the rolling update/rollback.

smruthi2187:
Thanks a lot for the comments @life1347 . Really appreciate the time you invested in it. 

I really like the idea you have under point 1. will definitely try and leverage it. will sync up with you too when i actually implement it.

For point 2, we can support rollback with fission cli commands, but, not sure how we can do that with spec. spec behaves more like git versioning. so in git, if you like to rollback(revert) you basically end up making a new commit. but i'll think about it some more and keep you posted.

For point 3, i see your perspective. but you know, i was thinking the analogy is a bit different in this case. here's what i think : pod can still act as a separate identity, even without configMap. So, configMap seems more like an attribute of pod and pod can function perfectly without this.
But, a fission function really has no identity without it's package. more like function's definition comes from only it's package. 
but then again, i am totally happy doing it the way you suggested. we can chat more about it too, what do you think?

vishal-biyani:
Hey, this is an excellent document @smruthi2187 , it took a couple of readings because there are a lot of details, but of course, it is really helpful to have this level of detail early on. I have a few smallish comments/questions:

1.  You mentioned Canary testing of functions in use cases, but later on you mentioned trigger pointing to the latest version of the function. And I think this is more from the future perspective - but how we will let's say send 10% traffic to a newer version and rest to the current version?

2. How will a newer version of Function CRD will play is something I thought over a bit (On a general note I think https://github.com/kubernetes/kubernetes/pull/60113 and the Google doc linked is a good read on CRD versioning - you will need to join SIG API machinery ( https://goo.gl/q0UT3A) to get access to doc though)

soamvasani:
@vishal-biyani not sure I understand your second point? The linked issue is about CRD schema versioning.

vishal-biyani:
@soamvasani I think I am trying to find out if a change in schema version will have any correlation/impact on the version of functions themselves. I guess not and not relevant for now at least.

smruthi2187:
Thanks for taking the time to review this @vishal-biyani. really appreciate it!

For canary testing, that's a good point you brought up. I have plans to modify the spec of the http trigger object itself. For now, there's just one field for version. My intention is to capture the percentage values along with the versions.
While it's easy to do this with fission cli commands, im still deciding how to make it work with `spec version`. mainly because with `spec version` the versions are created in the same iteration as the trigger object itself.
so there's no way a user can mention version string, instead, he can probably use some aliases like "latest" or "latest-1" or "n", "n-1". so our code should be able to interpret these aliases and map them to the right version uids and thereby resolve them to right function references.

i'll need to think about it more about design of canary deployments in fission. i'll update this document or make a new one for design of canary deployments soon. will sync up with you too then and get your feedback.

soamvasani:
@smruthi2187 I suggest merging this for future reference, even though we're not implementing this right now.

soamvasani:
Can we merge the proposal into the tree somewhere?  We learned useful things from the proposal and the discussions around it, and I feel like this will come up again at some point, so it'll be good not to start from zero again.

Issue #533, opened 2018-03-06
Executor not clean function's kubeObjs if function was deleted before the periodic clean process.


life1347:
Currently, the cleanup process only tries to clean related cache & kubeObjs if a function still exists.
But a function may be deleted before the periodic clean process. We should check the error type in https://github.com/fission/fission/blob/master/executor/cleanup.go#L119 .  And clean related cache & kubeObjs if an error is notExists error.

https://api.travis-ci.org/v3/job/349330314/log.txt

test script log
```
2018/03/06:17:07:59 Cleanup everything   <-  functions were deleted at 17:07:59
secret "testsecret-1520356015" deleted
configmap "testcfgmap-1520356015" deleted
function 'testsecret-1520356015' deleted
function 'testsecret-1520356015-1' deleted
function 'testcfgmap-1520356015' deleted
function 'testcfgmap-1520356015-1' deleted
function 'testnormal-1520356015' deleted
```

executor log
```
2018/03/06 17:10:49 Error getting function: testnormal-1520356015  <- periodic cleanup process 
2018/03/06 17:10:49 Error getting function: testsecret-1520356015
2018/03/06 17:10:49 Error getting function: testcfgmap-1520356015
2018/03/06 17:12:24 Creating pool for environment {nodejs  default /apis/fission.io/v1/namespaces/default/environments/nodejs 89cf46ed-2161-11e8-aec9-42010a80021b 27448748 0 2018-03-06 17:12:24 +0000 UTC <nil> <nil> map[] map[] [] nil [] }
```

vishal-biyani:
@life1347 When a function is deleted the event will trigger:
https://github.com/fission/fission/blob/master/executor/newdeploy/newdeploymgr.go#L152 

And the cleanup will be done as part of: https://github.com/fission/fission/blob/master/executor/newdeploy/newdeploymgr.go#L445

Or am I missing something?

life1347:
Oh, you're are right! I missed the L152! I will add a check here to prevent print error log for the deleted function.

life1347:
Reopen this issue. Newdeploy manager handler the delete event when a function is deleted. But this issue still exists for poolmanager.

sanketsudake:
This has been inactive for 1653 days. We're scrubbing the backlog and closing stale items to keep it actionable. If this is still relevant, please reopen or comment and we'll pick it back up.

PR #534, opened 2018-03-07
Fix executor failed to clean cache & kubeobjs after function deleted (#533)

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/534)


vishal-biyani:
LGTM

Issue #362, opened 2017-09-28
Show fission deployment version when using fission-v


erwinvaneyk:
Currently `fission -v` just states the version of the cli (even though it does not indicate that the version only reflects the cli)

Similar to `kubectl version` or `helm version` the command should provide the version of both the cli and the actual fission deployment. This should also allow us to diagnose these type of cli/deployment mismatches causing issues.

PR #538, opened 2018-03-08
Show fission deployment version with cli

life1347:
Set fission version through go `ldflags`. Following is the command output. (https://github.com/fission/fission/issues/362)
```
$ fission --version
Client Version: {"GitCommit":"84bf6d32ed9eccba48f0016fecc3bec94a47f043","BuildDate":"2018-03-08T15:54:56Z","Version":"test"}
Server Version: {"GitCommit":"81db94f3db335efbc209994b5bdaaafde4f9cec5","BuildDate":"2018-03-08T16:23:48Z","Version":"dummy-version"}
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/538)

### DIFF
--- a/builder/cmd/build.sh
+++ b/builder/cmd/build.sh
@@ -1,6 +1,18 @@
 #!/bin/bash
 version=$1
+if [ -z $version ]; then
+    version=$(git rev-parse HEAD)
+fi
+
 date=$2
+if [ -z $date ]; then
+    date=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+fi
+
 gitcommit=$3
+if [ -z $gitcommit ]; then
+    gitcommit=$(git rev-parse HEAD)
+fi
+
 CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -ldflags "-X github.com/fission/fission.GitCommit=$gitcommit -X github.com/fission/fission.BuildDate=$date -X github.com/fission/fission.Version=$version" -o builder .
 
--- a/builder/cmd/build.sh
+++ b/builder/cmd/build.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -o builder .
+version=$1
+date=$2
+gitcommit=$3
+CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -ldflags "-X github.com/fission/fission.GitCommit=$gitcommit -X github.com/fission/fission.BuildDate=$date -X github.com/fission/fission.Version=$version" -o builder .
 
--- a/environments/fetcher/cmd/build.sh
+++ b/environments/fetcher/cmd/build.sh
@@ -1,2 +1,5 @@
 #!/bin/sh
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -o fetcher .
+version=$1
+date=$2
+gitcommit=$3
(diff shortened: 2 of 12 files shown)


version.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+)
+
+func GetVersionInfo() string {
+	info := Info{
+		GitCommit: GitCommit,
+		BuildDate: BuildDate,
+		Version:   Version,
+	}
+	v, _ := json.Marshal(info)
```
erwinvaneyk: Don't marshal it here already; probably better to do the marshalling only where you actually need it (in request handlers). Otherwise it becomes confusing.

version.go
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+package fission
+
+import (
+	"encoding/json"
+)
+
+var (
+	GitCommit string // $(git rev-parse HEAD)
```
erwinvaneyk: Add a comment that the values of these variables are generated.

fission/main.go
```diff
@@ -41,11 +46,25 @@ func getKubeConfigPath() string {
...
+func getFissionAPIVersion(apiUrl string) string {
+	resp, err := http.Get(apiUrl)
+	if err != nil {
+		fatal(fmt.Sprintf("Error connection Fission API server: %v", err))
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		fatal(fmt.Sprintf("Error parsing Fission API version: %v", err))
```
erwinvaneyk: I don't think completely crashing is the best option here; the user might still want to know their client version. Currently it will just crash, not showing the client version either.

Just make this a multivalue return with an error, and handle error in the VersionPrinter.

environments/fetcher/cmd/build.sh
```diff
@@ -1,2 +1,5 @@
 #!/bin/sh
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -o fetcher .
+version=$1
+date=$2
+gitcommit=$3
```
erwinvaneyk: I am not sure about this: do you always need to specify these values? If so, maybe add some of default values here. Otherwise, it becomes pretty annoying to compile locally. Maybe something like: (date: now(), gitcommit: git rev-parse HEAD, version: $gitcommit)

fission-bundle/build.sh
```diff
@@ -1,2 +1,5 @@
 #!/bin/sh
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH
+version=$1
+date=$2
+gitcommit=$3
```
erwinvaneyk: Same here as in the fetcher/cmd/build.sh

builder/cmd/build.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
-CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags=-trimpath=$GOPATH -asmflags=-trimpath=$GOPATH -o builder .
+version=$1
+date=$2
+gitcommit=$3
```
erwinvaneyk: Same here as in the fetcher build.sh

fission/main.go
```diff
@@ -241,6 +266,7 @@ func main() {
 		{Name: "spec", Aliases: []string{"specs"}, Usage: "Manage a declarative app specification", Subcommands: specSubCommands},
 		{Name: "upgrade", Aliases: []string{}, Usage: "Upgrade tool from fission v0.1", Subcommands: upgradeSubCommands},
 		{Name: "tpr2crd", Aliases: []string{}, Usage: "Migrate tool for TPR to CRD", Subcommands: migrateSubCommands},
+		//{Name: "version", Aliases: []string{"v"}, Usage: "Print the client and server version information", Action: getVersionInfo},
```
erwinvaneyk: What's up with this? :D

version.go
```diff
@@ -0,0 +1,45 @@
...
+
+type (
+	Info struct {
+		GitCommit string `json:"GitCommit,omitempty"`
+		BuildDate string `json:"BuildDate,omitempty"`
+		Version   string `json:"Version,omitempty"`
+	}
+)
+
+func GetVersionInfo() string {
```
erwinvaneyk: Minor: I would prefer just `VersionInfo` to make it more go-canonical

fission/main.go
life1347: oops

soamvasani:
Nice work, I didn't know about those ldflags. LGTM.

Issue #238, opened 2017-06-21
Creating route with function that doesn't exist should give a warning


CamelCaseNotation:
```
$ fission function list
NAME     UID                                  ENV
hello    6cd9b8af-f3eb-4e3e-a3ce-965c37d305f8 nodejs
tst_hllo 48dab9a9-39b4-4876-bf1d-4f2f63001186 nodejs
```
Running the below should give a warning since no function by that name exists. It creates the route but then fails when hitting the endpoint with an internal fission error.

```
$ fission route create --method GET --url /test --function blah
trigger 'a7311368-4b87-464c-975b-a3b4be267ea4' created

$ curl http://$FISSION_ROUTER/test
Internal server error (fission)
```

PR #539, opened 2018-03-08
Show warning when trying to create a route with non-existent function (#238)

life1347:
https://github.com/fission/fission/issues/238

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/539)

### DIFF
--- a/fission/httptrigger.go
+++ b/fission/httptrigger.go
@@ -65,7 +65,7 @@ func checkFunctionExistence(fissionClient *client.Client, fnName string) {
 	}
 	_, err := fissionClient.FunctionGet(meta)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


fission/httptrigger.go
```diff
@@ -57,6 +58,17 @@ func getMethod(method string) string {
...
 }
 
+func checkFunctionExistence(fissionClient *client.Client, fnName string) {
+	meta := &metav1.ObjectMeta{
+		Name:      fnName,
+		Namespace: metav1.NamespaceDefault,
+	}
+	_, err := fissionClient.FunctionGet(meta)
+	if err != nil {
+		fmt.Printf("function '%v' not exists, use 'fission function create --name %v ...' to create function later\n", fnName, fnName)
```
soamvasani: not exists -> does not exist

to create function later -> to create the function

soamvasani:
LGTM other than that ^

Issue #515, opened 2018-02-27
Use a separate controller loop to watch functions change and create a service


life1347:
This is the follow-up issue for #421 . In PR421, the creation of service is triggered when a function creation request goes to function creation API, which is not good for users to create/modify resources with helm/kubectl. We should use a separate controller loop to watch functions changes and create/delete service when needed.

PR #544, opened 2018-03-12
Use a separate controller loop to watch functions change and create a service

life1347:
This aims to solve issue #515

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/544)

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -139,8 +139,6 @@ spec:
         env:
           - name: FISSION_FUNCTION_NAMESPACE
             value: "{{ .Values.functionNamespace }}"
-          - name: ENABLE_ISTIO
-            value: "{{ .Values.enableIstio }}"
         readinessProbe:
           httpGet:
             path: "/healthz"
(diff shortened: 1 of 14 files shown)


smruthi2187:
LGTM. Just one question, is there a reason why the ListWatch is not part of poolmgr?
The reason I ask is if controller process is removed from fission in the future, we may have to migrate the ListWatch to poolmgr.

life1347:
Good question! I didn't think of that before. I will push a change for moving ListWatch to poolmgr.

soamvasani:
LGTM.

soamvasani:
Looks like the tests failed with an executor crash:

```
2018/03/31 09:52:26 Creating NewDeploy ExecutorType
2018/03/31 09:52:26 starting executor at port 8888
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13610fa]
goroutine 24 [running]:
github.com/fission/fission/executor/poolmgr.(*GenericPoolManager).Run(0xc420381f20, 0x20dbba0, 0xc420584d80)
	src/github.com/fission/fission/executor/poolmgr/gpm.go:104 +0x3a
github.com/fission/fission/executor.(*Executor).Serve(0xc420584c00, 0x22b8)
	src/github.com/fission/fission/executor/api.go:114 +0x52a
created by github.com/fission/fission/executor.StartExecutor
	src/github.com/fission/fission/executor/executor.go:218 +0x2e6

```

Issue #486, opened 2018-02-07
HTTP request send to wrong Kubernetes pod


oif:
Hi,
I found an issue about HTTP Trigger function.  `fission version 0.4.1`

Steps
1. Create an env with only single runtime pod(like workflow env)
2. Deploy some functions with HTTP trigger on this env
3. Delete the env pod or this pod killed for some reasons (auto recover due to deployment controller)
4. curl this HTTP trigger
5. Check router logs find out request is forwarded to old pod IP

I'm not quite sure this issue is caused by Executor or other components

erwinvaneyk:
Hi @oif thanks for this issue. Yesterday I noticed similar behavior with the workflow environment as well. It doesn't seem related to the HTTP trigger, just triggering the function through `$FISISON_ROUTER/fission-function/<fn>` causes the same issue. I am not sure if this

I am not sure if this is specifically related to the workflow environment or whether other environments also exhibit this behavior.

cc @soamvasani

ananichev:
@oif @erwinvaneyk I faced the same issue. Seems the problem is in the function's cache. When a pod is created its address is put to the function's cache. And when you delete the env pod or this pod killed seems cache is not cleared. And the rotuer tries to proxy request to a died address.

soamvasani:
Thanks for the report @oif.

@erwinvaneyk this is unrelated to workflows.

@ananichev Yes, that's exactly it. The caches in both executor and router(s) keep the pod info for a minute or so, and for that duration requests are proxied to the dead IP.

To fix this there are a couple of changes we can make:
 1. watch for pod status changes and invalidate the caches in executor and router
 1. retry on connection errors 

The first change is definitely something we should do, but it doesn't solve all errors (it's still racy, since the invalidation might occur after a request is proxied to a dead IP).  

The second change doesn't suffer from that racy-ness, but may cause functions to be executed twice; this may be undesirable and probably shouldn't be the default.  This retrying behaviour would need to be controlled by some sort of new flag in the function and/or trigger spec.

smruthi2187:
The CI runs are failing. needs more testing.

smruthi2187:
Finally, CI runs did pass with this change and a subsequent change to prevent releasing idle transport connections.

PR #546, opened 2018-03-13
Invalidate stale router cache entry with podIP's for deleted pods.

smruthi2187:
This is to solve the issue : https://github.com/fission/fission/issues/486

The solution works with my tests. But, I 've made this WIP because I had a few questions around how I've done it and few other general questions.
These questions are added as TODOs in the respective places in code.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/546)


soamvasani:
Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[executor/api.go, line 78 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L81BLDOvErSsd6Li4p2:-L81BLDPUH-2kw2RSBv7:b-f7sxtz) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/api.go#L78)):*
> ```Go
> 			return fsvc.Address, nil
> 		} else {
> 			log.Printf("Deleting cache entry for invalid address : %s", fsvc.Address)
> ```

can you log the function name too?

---

*[executor/executor.go, line 122 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L81BcvB3AyHIUAf2DMt:-L81BcvCX5r9jjmu7HL5:bwus5al) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/executor.go#L122)):*
> ```Go
> }
> 
> func (executor *Executor) getFunctionExecutorType(meta *metav1.ObjectMeta) (string, error) {
> ```

Return `fission.ExecutorType` instead of a string.  That way we get type checks.

---

*[executor/client/client.go, line 89 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L81CkPqSIZLnYYDh9t6:-L81CkPqSIZLnYYDh9t7:bvbxz1v) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/client/client.go#L89)):*
> ```Go
> 
> func (c *Client) serveGetServiceRequests() {
> 	for {
> ```

This function seems to serialize cold start requests on the client side, and we really should not be doing that, since it can impact performance quite a bit.

Can you remind me why these changes are needed? There was something about needing to run this request in a different goroutine than the caller?

---

*[router/functionHandler.go, line 62 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83ae2tcC3uDju3gJCU:-L83YK9WlZfTYjZTH831:bpvap5p) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L62)):*
> ```Go
> }
> 
> // RoundTrip is a custom transport for http requests
> ```

Can you add a summary for this comment?

---

*[router/functionHandler.go, line 79 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83Xj2_SK2EUYJHSMk-:-L83Xj2aUxQMWIjR3yjj:bl4yp65) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L79)):*
> ```Go
> // While this RoundTripper handles the case where a previously cached address of the function pod isn't valid anymore
> // (probably because the pod got deleted somehow), by making a request to executor to get a new service for this function,
> // it doesn't handle a case where a newly specialized pod gets deleted just after the GetServiceForFunction succeeds.
> ```

The router should not try to be a reliable messaging service, we just need a reasonable chance at forwarding the request, e.g. today's behaviour of forwarding to a minutes-old cached IP needs to be fixed.  But we don't need to try to handle various race conditions of failure.  After all the pod can fail while actually serving the request, and in that case we'd not do anything.

Non-HTTP triggers like NATS should be used for things like at-least-once messaging.  

---

*[router/functionHandler.go, line 83 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83ZUWe3HNvNr3lqHDi:-L83ZUWe3HNvNr3lqHDj:b-91siku) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L83)):*
> ```Go
> // However, the subsequent http call for this function will ensure the cache is invalidated.
> //
> // Such a case can be handled by retrying requests for newly returned address for 5 times perhaps and then setting
> ```

Is this something this function does, or is this a suggestion for something we could do later?

---

*[router/functionHandler.go, line 98 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83YQdlAafpuK_Tud9D:-L83YQdlAafpuK_Tud9E:b-shghdy) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L98)):*
> ```Go
> 
> 	timeout := fh.roundTripper.initalTimeout
> 	isNewService := false
> ```

What is `isNewService`?  Does `false` mean the service doesn't exist yet? Or that it exists but isn't new?

---

*[router/functionHandler.go, line 104 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83ZhN1RcMJOAC6-mHK:-L83ZhN1RcMJOAC6-mHL:b-3oyjf2) ([raw file](https://github.com/fission/fission/blob/048ab6149ac3b14c426ae94e2581a9e2e4efb5c4/router/functionHandler.go#L104)):*
> ```Go
> 	MetadataToHeaders(HEADERS_FISSION_FUNCTION_PREFIX, fh.function, request)
> 
> 	// cache lookup
> ```

Do we need to remove this? If we keep it would it simplify the RoundTripper?

---

*[router/functionHandler.go, line 116 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83axS_gUCgNEJpbNFM:-L83axS_gUCgNEJpbNFN:b-im8j72) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L116)):*
> ```Go
> 			delay := time.Since(reqStartTime)
> 			if delay > 100*time.Millisecond {
> 				log.Printf("Request delay for %v: %v", fh.function.Name, delay)
> ```

This is only the time taken for the latest executor call. Can you move `reqStartTime` outside the loop?

---

*[router/functionHandler.go, line 119 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83_4lXJLPK-sVxo0Pz:-L83_4lXJLPK-sVxo0Q-:b9l2nja) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L119)):*
> ```Go
> 			}
> 
> 			// modify the request to reflect the new service address
> ```

Why is this here, isn't it already done in the director?

---

*[router/functionHandler.go, line 135 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83XT4pwJpQGFNNAKjt:-L83XT4pwJpQGFNNAKju:b-jwf7to) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L135)):*
> ```Go
> 		// over-riding default settings.
> 		transport.DialContext = (&net.Dialer{
> 			Timeout:   fh.roundTripper.initalTimeout,
> ```

shouldn't this be `timeout`, not `initialTimeout`?  (Also, would you mind fixing the typo in the "initalTimeout" name?  I guess I typo'd it the first time and have been autocompleting it ever since.)

---

*[router/functionHandler.go, line 139 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83c8w5phbeuOrg83gE:-L83c8w6spHlLKO1LQru:b-nqi03o) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L139)):*
> ```Go
> 		}).DialContext
> 
> 		// make the actual call with the request
> ```

suggest something like "forward the request to the function service". ("actual call" is not specific enough.)

---

*[router/functionHandler.go, line 141 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83cPA__dVX9akDL4pT:-L83cPA__dVX9akDL4pU:b18uy22) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L141)):*
> ```Go
> 		// make the actual call with the request
> 		resp, roundTripErr = transport.RoundTrip(req)
> 		if roundTripErr != nil {
> ```

It is a bit hard to follow these nested if/elses.  Some early exits will help, e.g. here you can check for roundtrip err == nil first, then you don't need a nested else

---

*[router/functionHandler.go, line 157 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83e165Sl_KeQfaTpAV:-L83TCqi4-W6ONC0qahJ:b9fgj3o) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L157)):*
> ```Go
> 					req.URL.Host, fh.function.Name)
> 				fh.fmap.remove(fh.function)
> 			} else if !isNetDialError(roundTripErr) {
> ```

If you check for this before, again you can early-exit and avoid more nested `else`s.

---

*[router/functionHandler.go, line 164 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83cwJWMOXKmi-_f3zS:-L83cwJWMOXKmi-_f3zT:bcoknsp) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L164)):*
> ```Go
> 			// if transport.RoundTrip succeeds and it was a cached entry, then tapService
> 			if !isNewService {
> 				serviceUrl, _ := fh.fmap.lookup(fh.function)
> ```

If we know it's cached then why do we need to lookup the cache?  i.e. don't we already have serviceUrl? 

Also, don't ignore errors!

---

*[router/functionHandler.go, line 165 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L83bjQ0mVMvhFSb_Al9:-L83bjQ1b9DFJsKs9-4P:b-7b8pn8) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L165)):*
> ```Go
> 			if !isNewService {
> 				serviceUrl, _ := fh.fmap.lookup(fh.function)
> 				// if we're using our cache, asynchronously tell
> ```

this old comment is the same as the new comment above it, you could remove this one

---

*[router/functionHandler.go, line 221 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L81PachdymR5PPkjjZL:-L81PachdymR5PPkjjZM:bahdf0g) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L221)):*
> ```Go
> 			// adding a custom header to request as a way to let RoundTripper know that the address
> 			// wasn't cached
> 			req.Header.Add("X-Custom-CacheMiss", "miss")
> ```

Can we use some other way to switch this behavior? The header is something an external request could set too, so it's awkward to use it for an internal flag like this.

---

*[router/functionHandler.go, line 234 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L838vUAx_N__ZbrUWiT:-L838vUBrOirIOdSxYXe:b-6y72i2) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L234)):*
> ```Go
> 	proxy := &httputil.ReverseProxy{
> 		Director:  director,
> 		Transport: *fh, // fh implements RoundTrip method of http.RoundTripper interface
> ```

You could make and call something like a new `fh.getRoundTripper(isCacheMiss)` here instead of reusing fh.  Then you won't need a new header.

---

*[test/tests/test_router_cache_invalidation.sh, line 2 at r1](https://reviewable.io:443/reviews/fission/fission/546#-L839By6Y34O8IGMhB52:-L839By6Y34O8IGMhB53:bb065e6) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/test/tests/test_router_cache_invalidation.sh#L2)):*
> ```Shell
> #!/bin/bash
> ```

Add a short comment describing the test?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/fission/fission/546)*

smruthi2187:
Review status: 4 of 9 files reviewed at latest revision, 19 unresolved discussions.

---

*[executor/api.go, line 78 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L81BLDOvErSsd6Li4p2:-L86K6cQ51UNxsHb6zvC:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/api.go#L78)):*

can you log the function name too?
</blockquote></details>

Done.

---

*[executor/executor.go, line 122 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L81BcvB3AyHIUAf2DMt:-L86K8AoAQloM1_TmU0y:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/executor.go#L122)):*

Return `fission.ExecutorType` instead of a string.  That way we get type checks.
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 62 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83ae2tcC3uDju3gJCU:-L86KAS62B0PJ-lN6gF7:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L62)):*

Can you add a summary for this comment?
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 83 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83ZUWe3HNvNr3lqHDi:-L86KDaB9hSqVRuEM4Cs:b-9iswmh) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L83)):*

Is this something this function does, or is this a suggestion for something we could do later?
</blockquote></details>

Just a suggestion. and an open question for code review feedback.

---

*[router/functionHandler.go, line 98 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83YQdlAafpuK_Tud9D:-L86KQlX0VkcQPdxqrKS:b851fcs) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L98)):*

What is `isNewService`?  Does `false` mean the service doesn't exist yet? Or that it exists but isn't new?
</blockquote></details>

Sorry about the confusing name, removed it and made it simpler (hopefully) like discussed.

---

*[router/functionHandler.go, line 104 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83ZhN1RcMJOAC6-mHK:-L86LZow2Ez8SebMfZDN:b-bs6vy9) ([raw file](https://github.com/fission/fission/blob/048ab6149ac3b14c426ae94e2581a9e2e4efb5c4/router/functionHandler.go#L104)):*

Do we need to remove this? If we keep it would it simplify the RoundTripper?
</blockquote></details>

I think we need it. 

---

*[router/functionHandler.go, line 116 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83axS_gUCgNEJpbNFM:-L86KXrtDW6-jvk92_E0:b895mbc) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L116)):*

This is only the time taken for the latest executor call. Can you move `reqStartTime` outside the loop?
</blockquote></details>

Like discussed, removed this. Will make an issue as a follow up PR to measure cold start from router perspective.

---

*[router/functionHandler.go, line 119 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83_4lXJLPK-sVxo0Pz:-L86KgVK-kZf7VTC3KnJ:b1wy84e) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L119)):*

Why is this here, isn't it already done in the director?
</blockquote></details>

I unified both occurrences to avoid confusion. but we need the request to be over-written to the newly obtained service address or original cached address. 

---

*[router/functionHandler.go, line 135 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83XT4pwJpQGFNNAKjt:-L86L5geCKfP0esMgVBe:b9clnu) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L135)):*

shouldn't this be `timeout`, not `initialTimeout`?  (Also, would you mind fixing the typo in the "initalTimeout" name?  I guess I typo'd it the first time and have been autocompleting it ever since.)
</blockquote></details>

Initially I thought, since there is a time.Sleep with an exponential backoff, we didnt need a huge timeout. but, after discussion, i realized that it's needed. so changed it accordingly.

---

*[router/functionHandler.go, line 139 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83c8w5phbeuOrg83gE:-L86LOAuEk-XgfczB7XW:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L139)):*

suggest something like "forward the request to the function service". ("actual call" is not specific enough.)
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 141 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83cPA__dVX9akDL4pT:-L86LPKW5QiTPJKLnUNF:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L141)):*

It is a bit hard to follow these nested if/elses.  Some early exits will help, e.g. here you can check for roundtrip err == nil first, then you don't need a nested else
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 157 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83e165Sl_KeQfaTpAV:-L86LPyM5PnryGhNX92U:brqnzk7) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L157)):*

If you check for this before, again you can early-exit and avoid more nested `else`s.
</blockquote></details>

Sorry about the nested elses. fixed them. hopefully.

---

*[router/functionHandler.go, line 164 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83cwJWMOXKmi-_f3zS:-L86LVD7DHvPtCwwJ9_k:bdr73wn) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L164)):*

If we know it's cached then why do we need to lookup the cache?  i.e. don't we already have serviceUrl? 

Also, don't ignore errors!
</blockquote></details>

this is removed.

---

*[router/functionHandler.go, line 165 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83bjQ0mVMvhFSb_Al9:-L86LX_WCEQrOYXxsv_A:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L165)):*

this old comment is the same as the new comment above it, you could remove this one
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 221 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L81PachdymR5PPkjjZL:-L86Lb9k2wXYy8p5Hm_c:b-be2aca) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L221)):*

Can we use some other way to switch this behavior? The header is something an external request could set too, so it's awkward to use it for an internal flag like this.
</blockquote></details>

Done. I originally thought instantiating a new object for every request can be avoided by re-using the fh object. but changed it as suggested.

---

*[router/functionHandler.go, line 234 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L838vUAx_N__ZbrUWiT:-L86LriL-C_DlS5di7Ls:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L234)):*

You could make and call something like a new `fh.getRoundTripper(isCacheMiss)` here instead of reusing fh.  Then you won't need a new header.
</blockquote></details>

Done.

---

*[test/tests/test_router_cache_invalidation.sh, line 2 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L839By6Y34O8IGMhB52:-L86LtNN9MuMhXxyWfQ2:b-896fix) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/test/tests/test_router_cache_invalidation.sh#L2)):*

Add a short comment describing the test?
</blockquote></details>

Done.

---

*[executor/client/client.go, line 89 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L81CkPqSIZLnYYDh9t6:-L86M-SK6nfwAe6fsBUE:bz8i9ws) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/executor/client/client.go#L89)):*

This function seems to serialize cold start requests on the client side, and we really should not be doing that, since it can impact performance quite a bit.

Can you remind me why these changes are needed? There was something about needing to run this request in a different goroutine than the caller?
</blockquote></details>

like discussed, it was my bad. in some old test run, my http requests were hanging when tried from within the tranport.RoundTrip function. So thought we need to make the request from a separate go-routine with a new http client. changed it back to just invoking the http post method without any serialization.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546)*

life1347:
Review status: 4 of 9 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[executor/poolmgr/gpm.go, line 205 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcMwLAPeLq-iuixiE:-L8CcMwLAPeLq-iuixiF:b5tsp60) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/executor/poolmgr/gpm.go#L205)):*
> ```Go
> 		if obj.Kind == "pod" {
> 			pod, err := gpm.kubernetesClient.CoreV1().Pods(obj.Namespace).Get(obj.Name, metav1.GetOptions{})
> 			if err == nil && strings.Contains(podAddress, pod.Status.PodIP) {
> ```

We also need to check the container status like what builder manager does in
https://github.com/fission/fission/blob/master/buildermgr/pkgwatcher.go#L127-L133

---

*[router/functionHandler.go, line 47 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcQ4l7OIg1n5xf0lI:-L8CcQ4l7OIg1n5xf0lJ:b4w8akh) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L47)):*
> ```Go
> }
> 
> func isNetDialError(err error) bool {
> ```

Since it's a useful function we can put it into github.com/fission/fission/common.go

---

*[router/functionHandler.go, line 134 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcWClEG81zVAyWLWU:-L8CcWClEG81zVAyWLWV:blj0429) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L134)):*
> ```Go
> 		req.URL.Scheme = serviceUrl.Scheme
> 		req.URL.Host = serviceUrl.Host
> 		// To keep the function run container simple, it
> ```

nitpick: please add a blank line between code and comments

---

*[router/functionHandler.go, line 169 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcZGyCuBD5moW8_dG:-L8CcZGyCuBD5moW8_dH:bndkqlq) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L169)):*
> ```Go
> 		// means its a newly created service and it returned a network dial error.
> 		// just retry after backing off for timeout period.
> 		if isNetDialError(err) && serviceUrlFromExecutor {
> ```

Maybe if serviceUrlFromExecutor is enough? since we checked and return non dial error at Line 163.

---

*[router/functionHandler_test.go, line 27 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CclbW7ajyKmNyANQG:-L8CclbW7ajyKmNyANQH:bcrkjyd) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler_test.go#L27)):*
> ```Go
> 
> 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
> 	"time"
> ```

please sort the import

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546)*

smruthi2187:
Review status: 3 of 10 files reviewed at latest revision, 24 unresolved discussions.

---

*[executor/poolmgr/gpm.go, line 205 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcMwLAPeLq-iuixiE:-L8JAiWk6_zwgLcMgYoN:b28j4k9) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/executor/poolmgr/gpm.go#L205)):*

We also need to check the container status like what builder manager does in
https://github.com/fission/fission/blob/master/buildermgr/pkgwatcher.go#L127-L133
</blockquote></details>

so here's what i was thinking ..
1. if the containers are not ready yet, because it's taking some time for them to come up, we should not return them as InvalidPod. because, that's the exact reason we have retries for. lets say we treat them as invalid pod just because the containers are not ready yet, then that'll result in one more pod becoming specialized and when we check if that pod is valid pod, again it returns false and it results in one more pod becoming specialized and it can go on and on.
2. instead, we can only check if the podStatus is !Failed and podIP is still the same. that way, if the containers are taking some time to become ready and respond to readiness probes, we wont end up specializing more and more pods.
sound good?

---

*[router/functionHandler.go, line 47 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcQ4l7OIg1n5xf0lI:-L8JAnlY4pWE1F8KprsR:b-896fix) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L47)):*

Since it's a useful function we can put it into github.com/fission/fission/common.go
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 134 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcWClEG81zVAyWLWU:-L8JApq40NMjVp0GiRWV:b-896fix) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L134)):*

nitpick: please add a blank line between code and comments
</blockquote></details>

Done.

---

*[router/functionHandler.go, line 169 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CcZGyCuBD5moW8_dG:-L8JArLB02Uduw3QlBie:b-896fix) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler.go#L169)):*

Maybe if serviceUrlFromExecutor is enough? since we checked and return non dial error at Line 163.
</blockquote></details>

Done.

---

*[router/functionHandler_test.go, line 27 at r2](https://beta.reviewable.io/reviews/fission/fission/546#-L8CclbW7ajyKmNyANQG:-L8JAtu--bSbmnSy-OEM:b-896fix) ([raw file](https://github.com/fission/fission/blob/ade12868f17a7ccfb5ea724e2c61ed6895a4a51e/router/functionHandler_test.go#L27)):*

please sort the import
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546)*

smruthi2187:
Review status: 3 of 10 files reviewed at latest revision, 24 unresolved discussions.

---

*[router/functionHandler.go, line 83 at r1](https://beta.reviewable.io/reviews/fission/fission/546#-L83ZUWe3HNvNr3lqHDi:-L8JBFNr0ViqC8FB9xM8:b-2rb90n) ([raw file](https://github.com/fission/fission/blob/863bb7261a4cd72e623265785727029e99e91910/router/functionHandler.go#L83)):*
<details><summary><i>Previously, smruthi2187 wrote…</i></summary><blockquote>

Just a suggestion. and an open question for code review feedback.
</blockquote></details>

oops, sorry, yes this is something this function does. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546)*

soamvasani:
Thanks, this looks great!  LGTM, just a very minor comment below.

---

Reviewed 4 of 7 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[executor/poolmgr/gpm.go, line 204 at r4](https://beta.reviewable.io/reviews/fission/fission/546#-L8oNAhB8-9MyAynHzXK:-L8oNAhB8-9MyAynHzXL:bua8ps0) ([raw file](https://github.com/fission/fission/blob/e2c1ec897cc0be6487d5dc6f4a50065419c156e4/executor/poolmgr/gpm.go#L204)):*
> ```Go
> 	for _, obj := range kubeObjects {
> 		if obj.Kind == "pod" {
> 			// local function to check if all containers in a pod are ready
> ```

why not keep this function outside, at the top level?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546#-:-L8pq8Wf3Pjh7dDpTEGO:b-s06ra0)*

life1347:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[executor/poolmgr/gpm.go, line 204 at r4](https://beta.reviewable.io/reviews/fission/fission/546#-L8oNAhB8-9MyAynHzXK:-L8q1z75E4sjh4rrVE-E:biy3ns5) ([raw file](https://github.com/fission/fission/blob/e2c1ec897cc0be6487d5dc6f4a50065419c156e4/executor/poolmgr/gpm.go#L204)):*

why not keep this function outside, at the top level?
</blockquote></details>

Maybe move to common.go? It's a quite useful function.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/fission/fission/546)*

Issue #502, opened 2018-02-20
fission-cli : Prefix the route relativeUrl with a slash if user omits it during route create


smruthi2187:
users feel that adding a "/" during a `fission route create --url /hello <bla>` is a little non-intuitive. 
probably best to check if prefix `/` exists otherwise prepend it at the time of creating httptrigger object.

PR #547, opened 2018-03-13
Prepending a slash to user input url if missing.

smruthi2187:
This is to solve issue : https://github.com/fission/fission/issues/502

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/547)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -274,6 +274,10 @@ func fnCreate(c *cli.Context) error {
 	if len(triggerUrl) == 0 {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


life1347:
We should add a check to the router too if users create CRD resources by spec not CLI. 
https://github.com/fission/fission/blob/be43e186d09bb0d6ab47f4746802cb2313c58fe8/router/httpTriggers.go#L221-L240

smruthi2187:
I added the check for spec in spec.go. Feels like it would be slightly expensive to add it in router. Do let me know if this works.

life1347:
LGTM

Issue #510, opened 2018-02-23
Tests for Function update


vishal-biyani:
Current tests for function update for the new deployment executor only test updates to a package. We need a few more tests to cover various scenarios and execution paths such as:

- When a function changes the type from pool manager to new deployment executor
- When the min & max scale changes
- When environment changes for a function
- When secrets and config maps get added to function (Existing secret/configmap test is relevant and helpful)
- When resource requirements for a function are changed

PR #550, opened 2018-03-14
Tests for function update

vishal-biyani:
This aims to solve #510

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/550)


test/tests/test_fn_update/test_resource_change.sh
```diff
@@ -0,0 +1,77 @@
...
+log "Creating route"
+fission route create --function ${fn} --url /${fn} --method GET
+
+log "Waiting for updates to take effect"
+sleep 5
+
+#If variable not used, shell assumes 'function' to be a real function
+func=function
+
+maxcpu_actual=$(kubectl describe $func $fn -n default|grep Cpu|head -1|cut -f 2|tr -dc '0-9')
```
smruthi2187: would it be better to use `kubectl` with `-o jsonpath` to extract the value of a specific field in the object? ex: to find controller_port from the controller pod spec, something like this could be done controller_port=`kubectl get svc controller -n $FISSION_NAMESPACE -ojsonpath='{.spec.ports[0].targetPort}'`.
that way we can avoid the cut with picking specific fields.

test/tests/test_fn_update/test_scale_change.sh
```diff
@@ -0,0 +1,59 @@
...
+timeout 60 bash -c "test_fn $fn 'world'"
+
+log "Updating function scale and target CPU percent for $fn"
+fission fn update --name $fn --env $env --code $ROOT/examples/python/hello.py --minscale $targetMinScale --maxscale $targetMaxScale --targetcpu $targetCpuPercent --executortype newdeploy --mincpu 20 --maxcpu 100 --minmemory 128 --maxmemory 256
+
+log "Waiting for update to catch up"
+sleep 5
+
+#If variable not used, shell assumes 'function' to be a real function
+actualMinScale=$(fission fn list|grep $fn|awk '{ print $5; }')
```
smruthi2187: kubectl with -ojsonpath could be used here too.

smruthi2187:
LGTM.

vishal-biyani:
Thanks, @smruthi2187 fixed the review comments. The tests in this PR are passing through the logging test is failing - but we have #582 for that, so I will merge this one.

PR #560, opened 2018-03-20
Spec validator, better errors, apply waits for previous build 

soamvasani:
This change implement a first cut of `fission spec validate`.  It implements basic checks for duplicate resources, dangling references, and unused resources.  These checks can be extended in the future to add warnings around best practices and so on.

It also collects all errors before exiting, so the the user can fix more errors before re-running `fission spec`.  

The validator is also called during `fission spec apply`.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/560)


smruthi2187:
This is awesome and super convenient for users!

PR #563, opened 2018-03-22
This change fixes an error in a yaml file in the fission-core chart.

yiqunding:
The Deployment definition for executor, controller, router contains readines/liveness probes. The port definition for those probes should be numbers, instead of string, i.e. port: "8888" should be port 8888.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/563)


soamvasani:
Thanks, @yiqunding !

Issue #559, opened 2018-03-20
fission env create CLI overrides user-specified version


soamvasani:
When the user specifies a builder, we automatically set the environment version = 2.  Which is fine if the user didn't set it, but if the user set it to 3 (or above) we'd end up lowering the version, and losing v3 features, like poolsize.

soamvasani:
@vishal-biyani can you take this one?

vishal-biyani:
@soamvasani Sure will take a look at it.

vishal-biyani:
@soamvasani , I have done changes but two questions:

1) If the builder is set and environment version is not set - we are setting env version to 2. Isn't it better to set it to latest env version (i.e.3)?

2) If builder and env version - both are not set then we are defaulting to env version 1. I think even in this case default to latest version might be better IMO. Thoughts?

PR #569, opened 2018-03-22
Fixes the issue #559 with env versions

vishal-biyani:
Env version if not set by the user while using builder should be equal or greater than 2.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/569)

### DIFF
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -57,19 +57,15 @@ func envCreate(c *cli.Context) error {
 	envExternalNetwork := c.Bool("externalnetwork")
 
 	if len(envBuilderImg) > 0 {
-		envVersion = 2
+		if !c.IsSet("version") {
+			envVersion = 2
+		}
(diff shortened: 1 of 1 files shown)


fission/environment.go
```diff
@@ -57,20 +57,16 @@ func envCreate(c *cli.Context) error {
...
+			envVersion = 2
+		}
 		if len(envBuildCmd) == 0 {
 			envBuildCmd = "build"
 		}
 	}
 
 	resourceReq := getResourceReq(c)
 
-	// Environment API interface version is not specified and
```
soamvasani: Wait why remove this? The new code now just leaves envVersion at zero when neither builder nor version is specified?

vishal-biyani: Oh ok, I will revert this back

soamvasani:
LGTM, thanks.

PR #581, opened 2018-03-27
Initial draft of testing proposal

vishal-biyani:


Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+- Script will use the new fission CLI
+- Script will use tests in fission_old to run the tests
+
+3. The script for (1) can be used from existing codebase while the part (2) will need developing a part of script specifically for upgrade.
+
+4. After the run - if both set of tests succeed, the upgrade can be considered fairly successful (Although not entirely - as there can be potential gap)
+
+### Alternative approach considered
+
+It is possible to get the binary from previous release instead of source code - but it has two drawbacks:
```
soamvasani: The runtime is much smaller, though.  Our tests are already ~20 minutes. You'll get to avoid building fission-bundle, fetcher, builder, cli, etc.

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. The effort for (2) in previous section still remains the same
+2. With binary install, we loose the tests from older version of Fission.
```
soamvasani: I don't think I agree with this? The tests are fully separate from installation.

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+
+## A small Diversion
```
soamvasani: Can we discuss this one separately? It does seem like we could add upgrade tests without overhauling the whole test system.

vishal-biyani: Sure, I agree this can be separated.

Documentation/wip/testing-proposal.md
vishal-biyani: What I meant here is if we don't check out the older version of Fission - and only use binary, then the tests from the previous version will have to be retrieved separately, as binary does not container tests. And we need tests from the previous version for "before upgrade" tests.

soamvasani: Oh, interesting.  How about using the tests from the current version, with just a header comment?  So just like we have "disabled:true" available right now, we could have a "pre-upgrade: true", and tests with that comment would be run.

Also, note that you'd need tests that don't clean up, so that we can test that functions etc. created before upgrade still work after the upgrade.

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+- Most of our tests are sequential - but in real life there are multiple things going on a server and we should mimic that - i.e. run tests in parallel. This is also benefetial to reduce overall test time. (Of course there are exceptions)
+
+Ginkgo allows tests to run in parallel and fires them in separate threads. It also allows randomizing the tests for every execution
+
+
+- Run selective test suites with certain purpose, for ex. test all performance tests is very rudimentry  with shell scripts. Also it is difficult to specify some some actions which should be run "before executing test suite 1" and after finishing execution and so on.
+
+Ginkgo provides constructs such as ```BeforeSuite``` and ```AfterSuite``` and the same for each test with ```BeforeEach`` and ```AfterEach```. This allows clear separation of fixtures from actual tests and also common setup and teardown
+
+- Some of our flaky tests succeed sometime and fail sometime. A lot of time is spent in debugging and fixinng in PRs on those tests. While the test itself should be fixed in longer term, it should be possible to retry flaky test as part of execution. Ginkgo provides a mechanism to retry certain tests.
```
clee: > Some of our flaky tests succeed sometime and fail sometime. A lot of time is spent in debugging and fixinng in PRs on those tests. While the test itself should be fixed in longer term, it should be possible to retry flaky test as part of execution.

A few small grammatical errors. Maybe instead:

*Sometimes, some of our tests fail unexpectedly and pass on the next attempt. A lot of time is spent in debugging and fixing in PRs on those tests. While each test should be fixed eventually, it should be possible to retry a flaky test as part of execution.*

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Any test which needs to absolutely disrupt cluster should not be run in parallel by using some tag.
```
clee: > absolutely disrupt cluster

absolutely disrupt *the* cluster

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Various runtime behaviours such as fail on first failure or keep testing till failure or randomize order of tests etc. is difficult to model with shell scripts.
```
clee: > is difficult to model

*are* difficult to model

Documentation/wip/testing-proposal.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Ability to measure time for all tests without doing time manipulation in shell scripts. Ginkgo provides built in constructs to measure test suite time and each individual execution time.
+
+### Next steps
+
+The best way to know usefulness of the test framework would be to model some of existing tests using the framework and demonstrate the value. This PR will add mode details with few of tests remodelled to use the framework.
```
clee: > mode details

*more* details, maybe? "mode details" doesn't make sense to me.

Documentation/wip/testing-proposal.md
vishal-biyani: You make a great point about keeping objects through the upgrade process and then run tests. I think these can be special tests intended for upgrade only and for these tests we can use the tag like "pre-upgrade: true"

The other aspect - which tests not to run before the upgrade, for ex. tests added in 0.7 should not be run at 0.6 level. I don't like the idea of using tags for this purpose. Are we on same page with this or else I can explain a little more.

vishal-biyani:
@soamvasani Ok to merge this design proposal and track the testing framework in a separate issue?

soamvasani:
@vishal-biyani Yup, sounds good.

Issue #554, opened 2018-03-16
Fetcher should not error out on EEXISTS 


soamvasani:
When an environment (like workflows) allows multiple functions to be loaded into the same pod, fetcher may be called to load a file that it already has. This case should be treated as a fast successful no-op, not a failure.

This is affecting workflows stability, so we should get to it in 0.7.

PR #584, opened 2018-03-27
Check if the requested file already exists in fetcher and skip fetch

smruthi2187:
This PR is for https://github.com/fission/fission/issues/554.
I haven't written a unit test for this trivial change. But tested it manually and it works.
I'll add a separate issue to write unit test case for fetcher and cover this case as part of that.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/584)


soamvasani:
LGTM

PR #587, opened 2018-03-29
Uses proper way to get server URL

### LINKED ISSUE #582: Function logs flaky behaviour
The fission function logs command not reliable in tests specially and has significant delay in general. 
For example below code snippet hits the function 4 times and waits for 60 seconds before calling ```fission fn logs``` but still the log is empty (Snippet from test_function_log.sh)

```
log "Doing 4 HTTP GETs on the function's route"
for i in 1 2 3 4
do
    curl -s http://$FISSION_ROUTER/$fn
done

sleep 60

fission function logs --name $fn --detail > /tmp/logfile
log "---function logs---"
cat /tmp/logfile
log "------"
```
The log output of the same:
```
2018/03/26:19:59:41 Waiting for router to catch up
2018/03/26:19:59:56 Doing 4 HTTP GETs on the function's route
Handling connection for 8888
log test
Handling connection for 8888
log test
Handling connection for 8888
log test
Handling connection for 8888
log test
2018/03/26:19:59:57 Grabbing logs, should have 4 calls in logs
2018/03/26:20:00:57 ---function logs---
2018/03/26:20:00:57 ------
2018/03/26:20:00:57 Cleanup route
trigger 'fc5bf700-022e-4d4d-9942-cbe6a7dcbae0' deleted
2018/03/26:20:00:58 delete logfile
[FAILED]: /test_logging/test_function_logs.sh
```
As can be seen from log, the function responds to curl calls, but the log is not reflected in ```fission function logs``` even after 60 seconds

vishal-biyani:
There were two problems with `fission fn logs` implementation on client side:

1. In case of error, it was returning null so user feedback was missing that there is an error in getting logs

2. After port forward implementation, the proper way to get server URL was missing in `GetLogDB`

That was the reason that test_function_log.sh was failing.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/587)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -599,7 +599,7 @@ func fnLogs(c *cli.Context) error {
 	checkErr(err, "get function")
 
 	// request the controller to establish a proxy server to the database.
-	logDB, err := logdb.GetLogDB(dbType, c.GlobalString("server"))
+	logDB, err := logdb.GetLogDB(dbType, getServerUrl())
 	if err != nil {
 		fatal("failed to connect log database")
 	}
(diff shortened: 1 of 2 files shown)


vishal-biyani:
Confirmation that `fission function logs` worked:
```[SUCCESS]: /test_logging/test_function_logs.sh```

life1347:
LGTM

PR #591, opened 2018-03-30
Adding routerUrl parameter for kubewatch, timer, message queue trigge…

smruthi2187:
Issue : The default value for routerUrl (https://github.com/fission/fission/blob/master/fission-bundle/main.go#L154) that is being passed to `KubeWatcher`, `Timer` and `MessageQueueMgr` processes (https://github.com/fission/fission/blob/master/fission-bundle/main.go#L172-L181) is defaulting to `http://router.fission`.

Fix : Adding the routerUrl parameter to their deployment objects in charts.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/591)


soamvasani:
LGTM, thanks!

Issue #590, opened 2018-03-29
Remove `fission fn pods` command


soamvasani:
It's not particularly useful.

PR #592, opened 2018-03-30
Add deprecated message to subcommand pods

life1347:
Add the deprecated message to subcommand `pod` and remove it in next the realease. So that users have time to migrate existing scripts. #590

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/592)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -118,7 +118,7 @@ func fnCreate(c *cli.Context) error {
 	client := getClient(c.GlobalString("server"))
 
 	if len(c.String("package")) > 0 {
-		fatal("--package is deprecated, please use --deploy instead.")
+		fatal("--package is deprecated and will be remove in the next release, please use --deploy instead.")
 	}
 
 	fnName := c.String("name")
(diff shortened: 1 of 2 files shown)


vishal-biyani:
I actually have the code ready for removing it, just so we don't duplicate the effort.

fission/function.go
```diff
@@ -118,7 +118,7 @@ func fnCreate(c *cli.Context) error {
 	client := getClient(c.GlobalString("server"))
 
 	if len(c.String("package")) > 0 {
-		fatal("--package is deprecated, please use --deploy instead.")
+		fatal("--package is deprecated and will be remove in the next release, please use --deploy instead.")
```
soamvasani: It's been a fatal error and deprecated for a few releases already, right?  We can just remove this one.

fission/function.go
```diff
@@ -654,6 +654,8 @@ func fnLogs(c *cli.Context) error {
 func fnPods(c *cli.Context) error {
 	client := getClient(c.GlobalString("server"))
 
+	fmt.Println("This subcommand is deprecated and will be remove in the next release. Please use `kubectl -n <namespace> logs -f -c <container> <pod>` instead.")
```
soamvasani: will be removed*

The command actually prints out pods names, so the replacement would be something like `kubectl get pods -l functionName=xyz --all-namespaces`

soamvasani:
LGTM, thanks!

PR #593, opened 2018-03-30
Add steps to set FISSION_ROUTER env variable & update docs

life1347:
Though we can test function by `fission test`, we still need steps for setting up FISSION_ROUTER. It's quite important for a newcomer to use fission.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/593)

### DIFF
--- a/Documentation/docs-site/config.toml
+++ b/Documentation/docs-site/config.toml
@@ -1,4 +1,4 @@
-baseURL = "http://fission.io/docs/0.5.0"
+baseURL = "http://fission.io/docs/0.6.0"
 languageCode = "en-US"
 defaultContentLanguage = "en"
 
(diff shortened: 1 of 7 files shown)


smruthi2187:
LGTM!

soamvasani:
Sorry, no, we shouldn't do this. The reason we removed the env variables is that a lot of people had trouble setting up on clusters without ingress and without loadbalancer support. For the test use case port-forward works great. For other use cases we have to document this stuff, but it should not be the default.

vishal-biyani:
Hi @life1347 I just noticed that some of these documentation changes never made it on docs site mostly because the docs moved in between. The changes should ideally be visible in https://github.com/fission/docs.fission.io/blob/master/docs/content/usage/functions.en.md 

Can you please make the same on docs repo and raise a PR?

life1347:
Oh, yes, I will open a PR to doc-site!

PR #595, opened 2018-04-02
bug fix: spec dir flag

xiekeyang:
Here the flag name should be `specdir`, not `specs`, that should match to:

```go
specDirFlag := cli.StringFlag{Name: "specdir", Usage: "Directory to store specs, defaults to ./specs"}
```

in fission/main.go

Currently it cause the specific specs dir not to work.

```
$ fission spec apply --specdir <my spec dir>
Everything up to date.
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/595)

### DIFF
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -115,7 +115,7 @@ type (
 )
 
 func getSpecDir(c *cli.Context) string {
-	specDir := c.String("specs")
+	specDir := c.String("specdir")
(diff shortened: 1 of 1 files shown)


soamvasani:
Oof. Thanks for fixing this, @xiekeyang !

(Maybe we should use a bunch of centrally defined constants instead of strings all over the cli for these options...)

Issue #585, opened 2018-03-28
Router starts before CRDs exist and causes test failures


vishal-biyani:
This is something which occurs relatively less frequently but I noticed in one of CI builds and needs to be fixed.

The router fails to start properly:
```
k -n f-54ef2f log router-89c87b54-fk9f9
2018/03/28 07:02:09 Starting router at port 8888 E0328 07:02:09.433805     1 reflector.go:201] 
src/github.com/fission/fission/router/httpTriggers.go:217:
Failed to list *crd.HTTPTrigger: the server could not find the requested resource 
(get httptriggers.fission.io)

E0328 07:02:09.445643       1 reflector.go:201] 
src/github.com/fission/fission/router/httpTriggers.go:217: 
Failed to list *crd.Function: the server could not find the requested 
resource (get functions.fission.io)
```
Also confirmed by pod status:
```
Readiness probe failed: Get http://10.52.2.163:8888/router-healthz: net/http: request canceled (Client.Timeout exceeded while awaiting headers)```
```
Strangely enough the pod is still assumed ready:
```
Conditions:
  Type		Status
  Initialized 	True 
  Ready 	True 
  PodScheduled 	True
```
This causes all tests to fail as the router is not reachable. I think this happens when the router starts even before the CRD creation is completed. We should improve liveness probe so that it restarts the router in case of above error.

soamvasani:
Can you please add a link the failed build?

soamvasani:
The reason I'm asking for the failed build, is that this isn't necessarily the cause of any failures.  The controller.Run() runs in a loop (see httpTriggers.go:217), so this error, in theory, should be harmless.  Unless this error continuously appears forever; then that would be a problem. But if it's just the initial few seconds, it's probably not, by itself, the cause of any failures.

vishal-biyani:
I think I restarted the build for that commit and can't locate in Travis anymore. The test log for all tests complained about not able to reach port 8888 though I don't recall entire line. Maybe close this for now and re-open when it appears next?

PR #602, opened 2018-04-03
Fix components crash before crds creation

life1347:
Some of the components use listwatch to watch CRDs changes. The component pod will enter crash state if it starts before CRDs exist.
```
2018/04/03 09:17:28 Creating NewDeploy ExecutorType
2018/04/03 09:17:28 starting executor at port 8888
E0403 09:17:28.388019       1 reflector.go:201] src/github.com/fission/fission/executor/newdeploy/newdeploymgr.go:142: Failed to list *crd.Function: the server could not find the requested resource (get functions.fission.io)
2018/04/03 09:17:28 Failed to get environment list: the server could not find the requested resource (get environments.fission.io)
```
This PR tends to solve this problem by adding `WaitForCRDs` to components startup process. Also, fix `ensureCRD` return nil while the error may not be empty. 

https://github.com/fission/fission/issues/585

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/602)

### DIFF
--- a/crd/client.go
+++ b/crd/client.go
@@ -169,7 +169,7 @@ func waitForCRDs(crdClient *rest.RESTClient) error {
 			return nil
 		}
 
-		if time.Since(start) > 60*time.Second {
+		if time.Since(start) > 30*time.Second {
 			return errors.New("timeout waiting for CRDs")
 		}
 	}
--- a/crd/client.go
+++ b/crd/client.go
@@ -169,7 +169,7 @@ func waitForCRDs(crdClient *rest.RESTClient) error {
 			return nil
 		}
 
-		if time.Since(start) > 30*time.Second {
+		if time.Since(start) > 60*time.Second {
 			return errors.New("timeout waiting for CRDs")
 		}
 	}
@@ -212,8 +212,8 @@ func (fc *FissionClient) Packages(ns string) PackageInterface {
 	return MakePackageInterface(fc.crdClient, ns)
 }
 
-func (fc *FissionClient) WaitForCRDs() {
-	waitForCRDs(fc.crdClient)
+func (fc *FissionClient) WaitForCRDs() error {
+	return waitForCRDs(fc.crdClient)
 }
 func (fc *FissionClient) GetCrdClient() *rest.RESTClient {
 	return fc.crdClient
--- a/buildermgr/buildermgr.go
+++ b/buildermgr/buildermgr.go
@@ -31,6 +31,11 @@ func Start(storageSvcUrl string, envBuilderNamespace string) error {
 		return err
 	}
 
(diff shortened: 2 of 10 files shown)


vishal-biyani:
I am not sure if this is an issue, as pointed by @soamvasani in related #585 the controller is a loop and will wait till resources are created. For example in [router log of the build here](https://api.travis-ci.org/v3/job/361548695/log.txt), the router finally starts working after looking for and getting CRDs 30 seconds later:

```
E0403 10:16:55.068569       1 reflector.go:201] src/github.com/fission/fission/router/httpTriggers.go:217: Failed to list *crd.HTTPTrigger: the server could not find the requested resource (get httptriggers.fission.io)
2018/04/03 10:17:04 Not cached, getting new service for &ObjectMeta{Name:hellopy-828343488,GenerateName:,Namespace:default,SelfLink:/apis/fission.io/v1/namespaces/default/functions/hellopy-828343488,UID:24c5576d-3728-11e8-8bc4-42010a8000de,ResourceVersion:31051690,Generation:0,CreationTimestamp:2018-04-03 10:16:58 +0000 UTC,DeletionTimestamp:<nil>,DeletionGracePeriodSeconds:nil,Labels:map[string]string{},Annotations:map[string]string{},OwnerReferences:[],Finalizers:[],ClusterName:,Initializers:nil,}

```

Relevant code block from `httpTriggers.go`:

```
func (ts *HTTPTriggerSet) runWatcher(ctx context.Context, controller k8sCache.Controller) {
	go func() {
		controller.Run(ctx.Done())
	}()
}
```

So similar code snippet from `newdeploymgr.go` should yield the same result, or am I missing something obvious?

```
func (deploy *NewDeploy) Run(ctx context.Context) {
	go deploy.funcController.Run(ctx.Done())
}
```

life1347:
Yes, the controller will create CRDs when it starts. But the problem is that component pods other than controller may start up first. Once the component tries to watch resources change before the controller is up, it crashes. Though K8S can restart pods automatically, it takes time to do that and cause tests fail.

crd/client.go
```diff
@@ -169,7 +169,7 @@ func waitForCRDs(crdClient *rest.RESTClient) error {
 			return nil
 		}
 
-		if time.Since(start) > 30*time.Second {
+		if time.Since(start) > 60*time.Second {
```
soamvasani: Why increase this?  It feels like 30 sec should be enough?

life1347: ok, I will rollback to 30 sec. And increase it in future if it's not enough.

soamvasani:
Thanks for the update, LGTM.

life1347:
Make CI happy. Merged! 🎉

Issue #530, opened 2018-03-06
Upgrade tests in CI


soamvasani:
Current CI tests only test `helm install`; they need to test upgrades as well.

vishal-biyani:
Hi @soamvasani Thanks for filing this, some questions:
- Do we want to upgrade only from the previous release or how much backward compatibility should be tested?
- Should this happen on every commit (Which seems a bit too much to me), or once let's say every day?

soamvasani:
* Previous release is a good start.
* We can use travis cron jobs, and conditionally run the upgrade tests when env var TRAVIS_EVENT_TYPE=="cron"

PR #605, opened 2018-04-04
Fission upgrade tests

vishal-biyani:
This PR aims to create a simple upgrade framework in which we can add more tests later.
Related to #530


vishal-biyani:
This PR tests upgrade of Fission from last known stable version to current version/master with a simple test.

`fission_upgrade_test.sh` is the main file which is called for upgrade test workflow, it should be called only if type of travis job is `cron` (Though that part is commented for now for testing). Also if upgrade test is called, normal build & test workflow is not called.

The flow of `fission_upgrade_test.sh` is roughly:

- Install Fission from the charts in Github repo based on last known version. Install CLI and setup port forwarding etc.
- `setup_fission_objects` from `fixture_tests.sh` - will create a function and associated route
- `upgrade_tests ` from `fixture_tests.sh` - will conduct tests, right now just a single test. The same set of tests will be called after upgrade too. Although the tests are tightly coupled with setup part in the previous point - as objects required for the test are created there
- Now build from source code, push images to the registry and then run the `helm upgrade`. Again set up port forwarding on a different port as the earlier port would have an issue. 
- Run `upgrade_tests ` from `fixture_tests.sh` again. 
- Finally `cleanup_fission_objects` is called as part of the trap, cleans up helm release and namespace etc.

test/build_and_test.sh
```diff
@@ -2,6 +2,11 @@
 
 set -euo pipefail
 
+if [[ "$TRAVIS_EVENT_TYPE" -e "cron" ]]
```
life1347: Just wondering where we can set Travis event type?

test/upgrade/fission_upgrade_test.sh
```diff
@@ -0,0 +1,131 @@
...
+--name $id \
+--wait \
+--timeout 540 \
+--set $helmVars \
+--namespace $ns \
+https://github.com/fission/fission/releases/download/${CURRENT_VERSION}/fission-all-${CURRENT_VERSION}.tgz
+
+mkdir temp && cd temp && curl -Lo fission https://github.com/fission/fission/releases/download/${CURRENT_VERSION}/fission-cli-linux && chmod +x fission && sudo mv fission /usr/local/bin/ && cd .. && rm -rf temp
+
+port_forward_services $id $routerNodeport
```
life1347: Will we use Loadbalancer or keep using port-forward for the test?

test/upgrade/fission_upgrade_test.sh
```diff
@@ -0,0 +1,131 @@
...
+
+build_and_push_fetcher $FETCHER_IMAGE:$TAG
+
+build_and_push_builder $BUILDER_IMAGE:$TAG
+
+ENV='python'
+
+build_and_push_env_runtime $ENV $REPO/$ENV-env:$TAG
+
+build_and_push_env_builder $ENV $REPO/$ENV-env-builder:$TAG $BUILDER_IMAGE:$TAG
```
life1347: L90-L96, from the changes, looks like we don't use these images for upgrade test. Is it intentional? If not, we can remove them to save some test time.

vishal-biyani: @life1347 We are using these at line 106 where we define the `helmVars` again

test/upgrade/fission_upgrade_test.sh
vishal-biyani: Yes I noticed change from Smruthi, will switch to load balancers too

test/build_and_test.sh
vishal-biyani: You can create a cron build in Travis like this, in which case the env. variable TRAVIS_EVENT_TYPE is set to value "cron", link to docs: https://docs.travis-ci.com/user/cron-jobs/ 
image

test/upgrade/fission_upgrade_test.sh
life1347: But the `helmVars` doesn't contain  `python-env`, `python-env-builder`, `builder`, also not used in any other place.
```
# L106
image=$IMAGE,
imageTag=$TAG,
fetcherImage=$FETCHER_IMAGE,
fetcherImageTag=$TAG,
logger.fluentdImage=$FLUENTD_IMAGE,
logger.fluentdImageTag=$TAG,
functionNamespace=$fns,
controllerPort=$controllerNodeport,
routerPort=$routerNodeport,
pullPolicy=Always,
analytics=false,
pruneInterval=$pruneInterval,
routerServiceType=$routerServiceType
```

test/build_and_test.sh
life1347: Oh! Cool!

life1347:
LGTM, I saw some comments `Remove before merging the PR`, will you remove them and merge the PR? Or, I can help to do that.

vishal-biyani:
Holding on to merging this till #611 is merged as there is an interdependency.

PR #611, opened 2018-04-06
Remove port forward in tests for router, controller and nats pods 

smruthi2187:
This prevents commands timing out due to the additional lag introduced by port-forward.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/611)

### DIFF
--- a/test/test_utils.sh
+++ b/test/test_utils.sh
@@ -148,6 +148,18 @@ clean_tpr_crd_resources() {
     kubectl --namespace default get crd| grep -v NAME| grep "fission.io"| awk '{print $1}'|xargs -I@ bash -c "kubectl --namespace default delete crd @"  || true
 }
 
+set_environment() {
+    id=$1
+    ns=f-$id
+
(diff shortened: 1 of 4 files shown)


test/test_utils.sh
```diff
@@ -148,6 +148,18 @@ clean_tpr_crd_resources() {
     kubectl --namespace default get crd| grep -v NAME| grep "fission.io"| awk '{print $1}'|xargs -I@ bash -c "kubectl --namespace default delete crd @"  || true
 }
 
+set_environment() {
+    id=$1
+    ns=f-$id
+
+    export FISSION_URL=http://$(kubectl -n $ns get svc controller -o jsonpath='{...ip}')
```
vishal-biyani: I think we would need to add a check to see if LB is provisioned and then assign. In some cases, LB provisioning & IP assignment may not be fast enough.

life1347: Hey @smruthi2187, I think we can simply add following code snippet back and merged this PR once its done.
https://github.com/fission/fission/blob/c031e046c19702d777b9f98daeb6a02cb582255a/test/test_utils.sh#L190-L215

Issue #572, opened 2018-03-23
SEGFAULT when builder manager failed to update package


life1347:
Builder manager encounter segmentation fault after failing to update package.
```
2018/03/23 07:01:13 Error setting package pending state: Operation cannot be fulfilled on packages.fission.io "demo-src-pkg-zip-ryb2": StorageError: invalid object, Code: 4, Key: /registry/fission.io/packages/default/demo-src-pkg-zip-ryb2, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: cf0a4157-2e67-11e8-aec9-42010a80021b, UID in object meta: 

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1f0 pc=0x1329f5f]
goroutine 298 [running]:
github.com/fission/fission/buildermgr.updatePackage(0xc420440d68, 0x0, 0x174246a, 0x6, 0xc42071e300, 0x168, 0x0, 0x0, 0x20c58a0, 0xc420587a80)
	src/github.com/fission/fission/buildermgr/common.go:123 +0x3f
github.com/fission/fission/buildermgr.(*packageWatcher).build(0xc42043c8c0, 0xc4202bfd40, 0x0)
	src/github.com/fission/fission/buildermgr/pkgwatcher.go:91 +0x63c
created by github.com/fission/fission/buildermgr.(*packageWatcher).watchPackages.func1
	src/github.com/fission/fission/buildermgr/pkgwatcher.go:203 +0x62
```

PR #626, opened 2018-04-16
Fix buildmgr SEGFAULT when it failed to update package

life1347:
### Description
This PR aims to fix https://github.com/fission/fission/issues/572 .

### Root cause
Build manager use the same variable to store function (`updatePackage`) return value, which is nil pkg pointer if the function failed to update package. And builder manager tries to print pkg information with that nil pointer.

### Solution
Rename function parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/626)


soamvasani:
LGTM, thanks!

PR #628, opened 2018-04-16
Update readme to point to the proper link

jgall:
It looks like at some point these readme links got out of sync with the website url schema. This PR fixes some of those links.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/628)


soamvasani:
Oops, thanks @jgall !

PR #630, opened 2018-04-17
Delete deployment with proper delete propagation policy

life1347:
1. Migrate DeleteOptions `OrphanDependents` to `PropagationPolicy`. 
2. Remove legacy code (workaround) for old k8s version.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/630)

### DIFF
--- a/executor/cleanup.go
+++ b/executor/cleanup.go
@@ -76,7 +76,11 @@ func cleanup(client *kubernetes.Clientset, namespace string, instanceId string) 
 	// still route to an old instance, i.e. router cache expiry
 	// time.
 	time.Sleep(6 * time.Minute)
+
 	err = cleanupPods(client, namespace, instanceId)
+	if err != nil {
+		return err
+	}
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


executor/cleanup.go
```diff
@@ -65,26 +69,14 @@ func cleanup(client *kubernetes.Clientset, namespace string, instanceId string)
 	if err != nil {
 		return err
 	}
-	// See K8s #33845 and related bugs: deleting a deployment
-	// through the API doesn't cause the associated ReplicaSet to
-	// be deleted.  (Fixed recently, but we may be running a
-	// version before the fix.)
-	err = cleanupReplicaSets(client, namespace, instanceId)
-	if err != nil {
```
vishal-biyani: I am sorry but why did we remove the error check here?

vishal-biyani:
Looks good except for 2 minor comments @life1347 
I don't think we have a good way to test this automatically though ATM. Can we have a placeholder bug for future?

executor/cleanup.go
life1347: The bug described in the comment was fixed in k8s 1.6. Since fission required 1.7+ , remove the legacy code here.

life1347:
https://github.com/fission/fission/issues/658

executor/cleanup.go
life1347: I guess you mean L84, fixed.

PR #633, opened 2018-04-18
Fix executor does not reap specialized function pod when env no longer exists

### LINKED ISSUE #629: IdleObjectReaper not reaps idle specialized pod if environment was deleted
### What happened
Notice the problem that executor does not delete specialized pod during the benchmark test. A specialized pod never get deleted once the environment, the pod belongs to, was deleted. 

### Root cause
IdleObjectReaper only iterates over the existing environment (L102). We also need to iterate over pods as well to check whether to delete.

https://github.com/fission/fission/blob/0194f53e124442986de20ee3bd3032bea61cca68/executor/cleanup.go#L96-L110

life1347:
### Description

Fix #629 

### What happened
Notice the problem that executor does not delete specialized pod during the benchmark test. A specialized pod never get deleted once the environment, the pod belongs to, was deleted. 

### Root cause
IdleObjectReaper only iterates over the existing environment (L102). We need to check all old funcSvcs even the environment no longer exists.

https://github.com/fission/fission/blob/0194f53e124442986de20ee3bd3032bea61cca68/executor/cleanup.go#L96-L110

### Solution
1. Modify `functionServiceCache` to list all old funcSvcs without requiring the environment metadata
2. Iterate over old funcSvcs and remove them if necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/633)

### DIFF
--- a/executor/cleanup.go
+++ b/executor/cleanup.go
@@ -22,9 +22,9 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api"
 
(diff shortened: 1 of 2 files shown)


soamvasani:
Makes sense! LGTM.

soamvasani:
You have to be careful because of the AllowedFunctionsPerContainer special case which is important for workflows, but it looks like this change handles it correctly.

PR #635, opened 2018-04-19
Fix SEGFAULT issue when buildmgr failed to update package

life1347:
This is the follow up PR to PR #626 . Looks like the issue is not really fix last time...
Test this PR locally and confirm the problem solved.

### DIFF
--- a/buildermgr/common.go
+++ b/buildermgr/common.go
@@ -131,12 +131,12 @@ func updatePackage(fissionClient *crd.FissionClient,
 	}
 
 	// update package spec
-	pkg, err := fissionClient.Packages(metav1.NamespaceDefault).Update(pkg)
+	newPkg, err := fissionClient.Packages(metav1.NamespaceDefault).Update(pkg)
 	if err != nil {
 		log.Printf("Error updating package: %v", err)
 		return nil, err
 	}
 
 	// return resource version for function to update function package ref
-	return pkg, nil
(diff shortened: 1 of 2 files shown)


buildermgr/common.go
```diff
@@ -131,12 +131,12 @@ func updatePackage(fissionClient *crd.FissionClient,
 	}
 
 	// update package spec
-	pkg, err := fissionClient.Packages(metav1.NamespaceDefault).Update(pkg)
```
soamvasani: Why is this necessary?  I must be missing something but it looks like it doesn't really change anything...

buildermgr/pkgwatcher.go
```diff
@@ -88,7 +88,7 @@ func (pkgw *packageWatcher) build(buildCache *cache.Cache, srcpkg *crd.Package)
 	if err != nil {
 		e := fmt.Sprintf("Error setting package pending state: %v", err)
 		log.Println(e)
-		updatePackage(pkgw.fissionClient, pkg, fission.BuildStatusFailed, e, nil)
```
soamvasani: This makes sense.

buildermgr/common.go
life1347: No, this is not necessary. I can rollback to previous version.

soamvasani:
LGTM, thanks.

PR #647, opened 2018-04-26
Fix test cases occasional failure

life1347:
This PR tends to address the test cases failure problem. (Log: https://api.travis-ci.org/v3/job/371377211/log.txt)
1. test_node_hello_http.sh
2. test_env_vars.sh

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/647)


life1347:
Both target test cases are passed. Faile at `test_function_logs` which is should be fixed in PR #646 . Merged.

PR #656, opened 2018-04-30
Java env alpha

vishal-biyani:
This is an alpha implementation of Java environment.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/656)


soamvasani:
Can you add something to the integration tests?

examples/jvm/java/src/main/java/io/fission/HelloWorld.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.fission.FissionFunction;
+import io.fission.FissionContext;
+
+public class HelloWorld implements FissionFunction<RequestEntity, FissionContext> {
```
erwinvaneyk: A bit of a design question, can users get around not having to import the entire springframework? If not, you might want to consider one of two options:
- rename the environment to `java` or `java-spring` to indicate that it is specific to spring.
- also provide an option for users to use `RequestEntity` and `ResponseEntity`-like objects that are in the `io.fission` package. (Such that users can choose to only import the `io.fission` package to create their function)

examples/jvm/java/src/main/java/io/fission/HelloWorld.java
```diff
@@ -0,0 +1,31 @@
+package io.fission;
+
+import java.util.LinkedHashMap;
+
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.fission.FissionFunction;
```
erwinvaneyk: Same as in Golang I would prefer `io.fission.Function` and `io.fission.Context` over these verbose names. However I am not sure if you would run into conflicts with Function from the stdlib. Might be something you could explore.

examples/jvm/java/pom.xml
```diff
@@ -0,0 +1,67 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
erwinvaneyk: This file needs to be formatted

examples/jvm/java/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+		<dependency>
+		    <groupId>com.fasterxml.jackson.core</groupId>
+		    <artifactId>jackson-core</artifactId>
+		</dependency>
+		<dependency>
+		    <groupId>io.fission</groupId>
+		    <artifactId>fission-java-core</artifactId>
+		    <version>0.0.1</version>
+		    <scope>provided</scope>
+		</dependency>
```
erwinvaneyk: This maven file is pretty big for a single, simple function. I am not sure if we can, but we might try to reduce it a bit. Maybe we could avoid the need for some configuration by creating a generic parent pom, similar to how springframework.

examples/jvm/java/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+		    <groupId>com.fasterxml.jackson.core</groupId>
+		    <artifactId>jackson-databind</artifactId>
+		</dependency>
+		<dependency>
+		    <groupId>com.fasterxml.jackson.core</groupId>
+		    <artifactId>jackson-core</artifactId>
+		</dependency>
+		<dependency>
+		    <groupId>io.fission</groupId>
+		    <artifactId>fission-java-core</artifactId>
```
erwinvaneyk: Rename this artifact to `io.fission.function-core`.  It seems inherently java to me.

environments/jvm/pom.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>io.fission</groupId>
+	<artifactId>jvmenv</artifactId>
```
erwinvaneyk: To keep the naming consistent, use `io.fission.env-jvm` or `io.fission.env-spring` or `io.fission.env-java` (see comment about environment naming

environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -0,0 +1,136 @@
...
+			try {
+				//cl.close();
+				jarFile.close();
+			} catch (IOException e) {
+				e.printStackTrace();
+				return ResponseEntity.badRequest().body("Error closing the file while loading the class");
+			}
+		}
+    	long elapsedTime = System.nanoTime() - startTime;
+    	System.out.println("Specialize call done in: " + elapsedTime/1000000 +" ms");
```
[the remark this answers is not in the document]
vishal-biyani: @erwinvaneyk Checking the specialization time in the executor is not ideal, even more so for new deployment executor as the specialization call happens in a different flow.

examples/jvm/java/src/main/java/io/fission/HelloWorld.java
vishal-biyani: @erwinvaneyk If we have to provide a Request/Response object as part of Fission library - that would be a lot of work. Using the spring framework is like using Flask framework in Python environment - we choose a fairly commonly used web framework. I am ok with renaming environment to java-spring as that clearly indicates the decision.

examples/jvm/java/pom.xml
vishal-biyani: Can look at a generic parent POM in future version

environments/jvm/src/main/java/io/fission/Server.java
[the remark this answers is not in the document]
vishal-biyani: For logging - using a library would mean environment JVM has already loaded the library. This can cause a lot of issues when users import their own logging libraries etc. So I am trying to avoid using a logging library. I will see if Java native logger is more performant and use that

examples/jvm/java/src/main/java/io/fission/HelloWorld.java
erwinvaneyk: @vishal-biyani I understand :) Renaming the environment to `java-spring` is more self-documenting to and sets the expectations for users. It also gives us the option to later add support for a different (whichever is most popular by then :P) framework or plain java without having to maintain and explain the caveat that `java` == `spring`. 👍

examples/jvm/java/pom.xml
erwinvaneyk: 👍

environments/jvm/src/main/java/io/fission/Server.java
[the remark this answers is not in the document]
erwinvaneyk: Specialization issue: 👍 

Logging: you can use `java.util.Logging` which is included in the standard library. Or, you could use SLF4J which decouples the logging from the underlying logging library, allowing users to choose their own logger.

environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -0,0 +1,122 @@
...
+    		}
+    		
+    		if (cl == null) {
+    			return ResponseEntity.status(500).body("Failed to initialize the classloader");
+    		}
+    		
+    		// Load all dependent classes from libraries etc. 
+    		while (e.hasMoreElements()) {
+    		    JarEntry je = e.nextElement();
+    		    if(je.isDirectory() || !je.getName().endsWith(".class")){
```
erwinvaneyk: 1) So I am not 100% sure how packaged Java applications are structured internally, but are you sure that this won't lead to problems with more complex applications (that might have nested structures)?
2) Resources in the package are not added to the classpath. So if the function interacts with (for example) some config or data file in its own classpath, it will break here.

I think both issues are not relevant to simple use cases/functions. So I would be in favor of adding these potential issues to GH issues, instead of stalling this PR.

environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -0,0 +1,122 @@
...
+    		        continue;
+    		    }
+    		    String className = je.getName().substring(0,je.getName().length()-CLASS_LENGTH);
+    		    className = className.replace('/', '.');
+    		    cl.loadClass(className);
+    		}
+    		
+    		// Instantiate the function class
+			fn = (Function) cl.loadClass(entryPoint).newInstance();
```
erwinvaneyk: indentation is off

environments/jvm/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+	<dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+		<dependency>
+		    <groupId>io.fission</groupId>
+		    <artifactId>fission-java-core</artifactId>
+		    <version>0.0.2-SNAPSHOT</version>
+		</dependency>
```
erwinvaneyk: indentation is off

examples/jvm/java/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            <artifactId>spring-boot-starter-web</artifactId>
+            <version>2.0.1.RELEASE</version>
+            <scope>provided</scope>
+        </dependency>
+		<dependency>
+		    <groupId>io.fission</groupId>
+		    <artifactId>fission-java-core</artifactId>
+		    <version>0.0.2-SNAPSHOT</version>
+		    <scope>provided</scope>
+		</dependency>
```
erwinvaneyk: indentation is off

environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -0,0 +1,122 @@
...
+				//cl.close();
+				jarFile.close();
+			} catch (IOException e) {
+				e.printStackTrace();
+				return ResponseEntity.badRequest().body("Error closing the file while loading the class");
+			}
+		}
+    	long elapsedTime = System.nanoTime() - startTime;
+    	logger.log(Level.INFO, "Specialize call done in: " + elapsedTime/1000000 +" ms");
+		return ResponseEntity.ok("Done");
```
erwinvaneyk: indentation

examples/jvm/java/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM maven:3.5-jdk-8 as BUILD
+COPY src /usr/src/myapp/src
+COPY pom.xml /usr/src/myapp
+ENTRYPOINT mvn -f /usr/src/myapp/pom.xml clean package
```
erwinvaneyk: You might want to add a simple `build.sh` for building the package to help the users understand what to do. Especially, since you need to do add some non-trivial/easy-to-forget arguments to it (like, volume mounting).

Even better: remove this dockerfile, and simply use the maven image directly in a `build.sh`, like explained here: https://docs.docker.com/samples/library/maven/
Probably something like:

_examples/jvm/java/build.sh_
```bash
docker run -it --rm -v "$(pwd)":/usr/src/myapp -w /usr/src/myapp maven:3.3-jdk-8 mvn clean package
```

examples/jvm/java/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+	</properties>
+	
+	<dependencies>
+		<dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+            <version>2.0.1.RELEASE</version>
+            <scope>provided</scope>
```
erwinvaneyk: Minor (no need to fix now): with this config you can't run the function locally (because the spring and fission are not PROVIDED). Better would be to exclude these dependencies in the build step, see: https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/including-and-excluding-artifacts.html

vishal-biyani: I think that `exclude` will not work as you state. In future we can have profiles - one for dev which has dependencies included - and other for deploy which will not ship dependencies. I will add this to list of items to be done in next iteration.

examples/jvm/java/Dockerfile
vishal-biyani: Oh cool, this is superb. I wanted to do something like this but did not know about this utility! Awesome!!

environments/jvm/src/main/java/io/fission/Server.java
vishal-biyani: On (1) - I am not entirely sure but I have seen most people using JarEntry to load classes, which I will check is safe in terms of nested & dependent classes.

on (2) - the recent convention is not to use the properties file but put such properties in Java classes. The properties which change often - supply from outside such as a environment variable/configmap or similar such service.

environments/jvm/src/main/java/io/fission/Server.java
vishal-biyani: For some reason, this is perfect in my IDE but looks off here.

examples/jvm/java/Dockerfile
vishal-biyani: And this is only for test and not intended for users.

environments/jvm/src/main/java/io/fission/Server.java
erwinvaneyk: For (2), okay forget I said configs :) What about data files or any other resources that are packaged into the JAR?

examples/jvm/java/Dockerfile
erwinvaneyk: IMO it is really useful to users to not have to think how to build the example.. it is an example after all. Just add it as a `build.sh` even if you don't use it in anywhere else.

environments/jvm/src/main/java/io/fission/Server.java
vishal-biyani: I agree that there can be resources such as SQL file in case of Liquibase and such, but I would take them for specific use cases.

examples/jvm/java/Dockerfile
vishal-biyani: Done, thanks

Issue #659, opened 2018-04-30
Seeing this issue: (Invalid argument - Failed to download url : Get : unsupported protocol scheme "") in executor


smruthi2187:
When executor tries to specialize a Pod, i notice this print in the executor logs 
(Invalid argument - Failed to download url : Get : unsupported protocol scheme "") 

Looking at the test scenario , my guess is that buildermgr hasnt finished the build yet and the deployment url is empty in the package spec, while fetcher is already trying to download the deploy package with empty url.

I'm yet to try to repro this issue and confirm if this is what is the issue. but one CI run of mine has this issue = https://api.travis-ci.org/v3/job/373173522/log.txt
Look for ""2018/04/30 19:15:53 [func3] calling fetcher to copy function with fetcher url: http://10.52.0.90:8000/"

vishal-biyani:
Hi @smruthi2187 Can this be closed as indicated in PR?

vishal-biyani:
Closing this as PR has been merged.

PR #661, opened 2018-05-01
meaningful error message when fetch request is received for a package when build is not successful.

smruthi2187:
This PR is for issue https://github.com/fission/fission/issues/659

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/661)

### DIFF
--- a/environments/fetcher/fetcher.go
+++ b/environments/fetcher/fetcher.go
@@ -254,6 +254,15 @@ func (fetcher *Fetcher) Fetch(req FetchRequest) (int, error) {
 		if req.FetchType == FETCH_SOURCE {
 			archive = &pkg.Spec.Source
 		} else if req.FetchType == FETCH_DEPLOYMENT {
+			// sometimes, the user may issue curl request to the function even before the source code is built into a deploy pkg.
+			// this results in executor sending a fetch request of type FETCH_DEPLOYMENT and since pkg.Spec.Deployment.Url will be empty,
+			// we hit this "Get : unsupported protocol scheme "" error.
+			// it may be useful to the user if we can send a more meaningful error in such a scenario.
+			if pkg.Status.BuildStatus != fission.BuildStatusSucceeded {
+				e := fmt.Sprintf("Build status for the function's pkg is : %s, can't fetch deployment", pkg.Status.BuildStatus)
+				log.Printf(e)
+				return 500, errors.New(e)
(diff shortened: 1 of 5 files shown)


life1347:
LGTM, please rebase the master and will merge it when the CI pass.

environments/fetcher/fetcher.go
```diff
@@ -254,6 +254,15 @@ func (fetcher *Fetcher) Fetch(req FetchRequest) (int, error) {
 		if req.FetchType == FETCH_SOURCE {
 			archive = &pkg.Spec.Source
 		} else if req.FetchType == FETCH_DEPLOYMENT {
+			// sometimes, the user may issue curl request to the function even before the source code is built into a deploy pkg.
```
soamvasani: small nit: "may issue curl request" -> "may invoke the function" (since there are other ways to invoke functions besides http)

environments/fetcher/fetcher.go
```diff
@@ -254,6 +254,15 @@ func (fetcher *Fetcher) Fetch(req FetchRequest) (int, error) {
...
+			// this results in executor sending a fetch request of type FETCH_DEPLOYMENT and since pkg.Spec.Deployment.Url will be empty,
+			// we hit this "Get : unsupported protocol scheme "" error.
+			// it may be useful to the user if we can send a more meaningful error in such a scenario.
+			if pkg.Status.BuildStatus != fission.BuildStatusSucceeded {
+				e := fmt.Sprintf("Build status for the function's pkg is : %s, can't fetch deployment", pkg.Status.BuildStatus)
+				log.Printf(e)
+				return 500, errors.New(e)
```
soamvasani: I think we should give some specific code here, such as 502 or 503.  BTW, does this code make it all the way back through the executor, router, and back to the original request?

smruthi2187: yeah makes it to router. but you know how router just sends 502 irrespective of the error.
for sample print, you can peek into https://api.travis-ci.org/v3/job/373309088/log.txt, look for below : 
2018/05/01 02:15:00 http: proxy error: Internal error - Build status for the function's pkg is : failed, can't fetch deployment

not sure if 502 or 503 makes sense in this case. i thought about it a lot and decided on 500.

soamvasani:
LGTM, thanks!

PR #669, opened 2018-05-04
Fetcher java changes

### LINKED ISSUE #665: Fetcher should support other archive formats
Currently, fetcher fetches the deployable artifact and extracts from a zip archive. In some languages, for example, Java, it is better to have the JAR file instead of an extracted version. Fetcher should support for Java jar file format as a default.

vishal-biyani:
Currently, I don't know of a good way to check if it is a Jar file (The way Go checks if it is zip file based on first 4 bytes of the file).

This implementation relies on `/META-INF/MANIFEST.MF` file in Jar file to determine that is a Jar/Java archive. Most of Java projects have this file. This is not a perfect solution, so need to check if we can find something better.

If the file is found to be JAR based on the presence of MANIFEST.MF, then instead of passing the exploded directory, the original file is passed to the environment.

This fixes #665 and is needed by #656

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/669)

### DIFF
--- a/environments/fetcher/fetcher.go
+++ b/environments/fetcher/fetcher.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/mholt/archiver"
@@ -304,16 +303,11 @@ func (fetcher *Fetcher) Fetch(req FetchRequest) (int, error) {
 		}
 		manifestPath := tmpUnarchivePath + "/META-INF/MANIFEST.MF"
 		_, err = os.Stat(manifestPath)
-		fmt.Println("tmpPath=", tmpPath)
-		fmt.Println("tmpUnarchivePath=", tmpUnarchivePath)
 		//TODO - Find a better way to detect a Java Jar
 		//If /META-INF/MANIFEST.MF exists - it is a Java environment
 		if err != nil {
 			tmpPath = tmpUnarchivePath
-		} else {
-			tmpPath = strings.Replace(tmpPath, ".tmp", ".jar", -1)
 		}
-		fmt.Println("Final tmpPath=", tmpPath)
 
 	}
 
--- a/environments/fetcher/fetcher.go
+++ b/environments/fetcher/fetcher.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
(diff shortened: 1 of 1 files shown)


soamvasani:
Hmm, this is pretty Java-specific.  How about some sort of flag in the env spec?

vishal-biyani:
Fair point @soamvasani 

So I looked at a way to add some sort of flag to indicate whether the fetcher should pass on the extracted version or the unextracted version of the archive. Something like:

```
Archive struct {
		// Type defines how the package is specified: literal or URL.
		Type ArchiveType `json:"type,omitempty"`
.... other truncated stuff
		// ExtractArchive is used by fetcher to determine if the extrcted archive
		// or unarchived file should be placed, which is then used by specialize handler
		ExtractArchive bool
	}
```

But most of the flags around source/binary (Basically related to `Package` & `Archive`) are passed at two places: `package` subcommands and `function` subcommand for example `buildcmd`. 

So this new flag - would have to be added also at both places - package and function. Are you ok with this?

soamvasani:
I was thinking it could be in the environment spec -- after all we're thinking of it as an env-level property, it wouldn't be different for 2 different Java functions for example.  Then the user doesn't have to keep setting this flag for each package/function.

vishal-biyani:
@soamvasani I agree. I initially had put the attribute in `EnvironmentSpec` itself, but when I looked at fetcher request structure, it felt like a good place to put that attribute.

Also, the Fetcher does not know anything about environment per se directly, if you look at FetchRequest & nowhere in the fetcher code environment is retrieved. So it is kind of loosely coupled with the environment. 

```
FetchRequest struct {
		FetchType     FetchRequestType             `json:"fetchType"`
		Package       metav1.ObjectMeta            `json:"package"`
		Url           string                       `json:"url"`
		StorageSvcUrl string                       `json:"storagesvcurl"`
		Filename      string                       `json:"filename"`
		Secrets       []fission.SecretReference    `json:"secretList"`
		ConfigMaps    []fission.ConfigMapReference `json:"configMapList"`
	}
```

So I think we should keep the attribute at the `environmentspec`/environment level but pass it as an attribute in `FetcherRequest` - so the fetcher code does not talk to the environment as such (Nothing against it, just current loose coupling seems like right way to do)

Comments?

vishal-biyani:
Closing in favour of #675

PR #672, opened 2018-05-08
Replace Werkzeug  with Bjoern as underlying WSGI server

life1347:
Werkzeug as underlying WSGI server of Flask is not suitable for serving high load traffic. After some research and end up using Bjoern as the alternative. The performance graph will be attached later.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/672)


soamvasani:
Nice, LGTM.

PR #675, opened 2018-05-10
Added a flag to control the extraction of archive based on user input

vishal-biyani:
This PR adds a flag at environment level which allows the user to decide if they want to extract the binary archive inside the function. This is useful for example in case of Java environments where you want the Jar file to be used instead of extracted contents of Jar file.

This PR also replaces earlier planned implementation #669 which was a bit too specific to Java.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/675)


types.go
```diff
@@ -268,6 +268,10 @@ type (
 		// The grace time for pod to perform connection draining before termination. The unit is in seconds.
 		// Optional, defaults to 360 seconds
 		TerminationGracePeriod int64
+
+		// ExtractArchive is used by fetcher to determine if the extrcted archive
+		// or unarchived file should be placed, which is then used by specialize handler
+		ExtractArchive bool
```
soamvasani: Specify json field name

executor/poolmgr/gp.go
```diff
@@ -363,6 +363,13 @@ func (gp *GenericPool) specializePod(pod *apiv1.Pod, metadata *metav1.ObjectMeta
 		return err
 	}
 
+	env, err := gp.fissionClient.
```
soamvasani: Don't add this to specialize; it slows down cold starts.  Move it to the gp struct construction

environments/fetcher/fetcher.go
```diff
@@ -293,14 +294,16 @@ func (fetcher *Fetcher) Fetch(req FetchRequest) (int, error) {
 	}
 
 	// check file type here, if the file is a zip file unarchive it.
-	if archiver.Zip.Match(tmpPath) {
+	// If extract true then extract else use archive as it is
```
soamvasani: Please improve comment, since the code doesnt just check if the extract flag is true, it also checks if it's an archive.  Though the code seems readable and I would be ok with just removing the comment

buildermgr/common.go
```diff
@@ -56,9 +56,10 @@ func buildPackage(fissionClient *crd.FissionClient, builderNamespace string,
 	builderC := builderClient.MakeClient(fmt.Sprintf("http://%v:8001", svcName))
 
 	fetchReq := &fetcher.FetchRequest{
-		FetchType: fetcher.FETCH_SOURCE,
-		Package:   pkg.Metadata,
-		Filename:  srcPkgFilename,
+		FetchType:      fetcher.FETCH_SOURCE,
+		Package:        pkg.Metadata,
+		Filename:       srcPkgFilename,
+		ExtractArchive: true,
```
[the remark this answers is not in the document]
vishal-biyani: @soamvasani No, this is not environment specific and actually needs to be `true` for all environments. The builder fetches source code as an archive, extracts it and then fires the `buildcmd`. If the `extractarchive` is false, it never finds the `buildcmd` script! 

Ideally in our case `extractarchive` should be always true but bool defaults to false if not specified. I am not sure if we should use something to enable default to true or for now leave it a coding decision.

vishal-biyani:
Related #665

vishal-biyani:
@soamvasani I will merge this and we can handle extraction part in it's own PR?

fission/environment.go
```diff
@@ -90,6 +90,11 @@ func envCreate(c *cli.Context) error {
 		}
 	}
 
+	extractArchive := true
```
life1347: Usage in fission/main.go says "defaults to false", but here is true.
And I will suggest setting `Value` field in cli.BoolFlag instead of hard-code here.

PR #677, opened 2018-05-11
Fission metrics integration

soamvasani:
Add prometheus metrics collection endpoints to router and executor.  Add prometheus annotations to router and executor pods.

No grafana integration just yet.  Let's do that in a separate change.

Fission will not install Prometheus; users can just use the regular prometheus installation methods (I tested this change with the prometheus helm chart).

Docs tracked in https://github.com/fission/docs.fission.io/issues/22

[This change builds on a bunch of work from @yqf3139 and @prithviramesh, several months ago, that didn't get merged at the time.]

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/677)


soamvasani:
I'm going to merge this; I'll address review comments in a follow up PR.

PR #688, opened 2018-05-21
Working version of Ingress integration

vishal-biyani:
Working Ingress integration for routes.
For ease of use need to add a flag driven deployment of ingress controller.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/688)


router/ingress.go
```diff
@@ -0,0 +1,161 @@
...
+func createIngress(trigger *crd.HTTPTrigger, kubeClient *kubernetes.Clientset) {
+
+	if !trigger.Spec.CreateIngress {
+		log.Printf("Skipping creation of ingress for trigger: %v", trigger.Metadata.Name)
+	}
+
+	ing := &v1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:       getDeployLabels(trigger),
+			GenerateName: "ingress-" + trigger.Spec.FunctionReference.Name + "-",
```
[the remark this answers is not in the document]
smruthi2187: on the same, is it ok to create an ingress object with the same name as the http trigger since there's 1-1 mapping?

router/ingress.go
```diff
@@ -0,0 +1,161 @@
...
+
+	_, err := kubeClient.ExtensionsV1beta1().Ingresses(podNamespace).Create(ing)
+	if err != nil {
+		log.Printf("Failed to create ingress: %v", err)
+	}
+}
+
+func getDeployLabels(trigger *crd.HTTPTrigger) map[string]string {
+	return map[string]string{
+		"triggerName":  trigger.Metadata.Name,
```
smruthi2187: since the trigger names are unique, i think it's enough to just have that. we may not need function names.

test/tests/test_ingress.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -euo pipefail
+
+ROOT=$(dirname $0)/../..
```
smruthi2187: while this test is good and tests that the ingress objects get updated when the relative Url and host parameters of the corresponding httpTrigger object gets updated.
but, i do feel like it'll be good to have another test that verifies that the ingress does works and the request gets routed to the correct relativeUrl,

router/ingress.go
```diff
@@ -0,0 +1,161 @@
...
+			GenerateName: "ingress-" + trigger.Spec.FunctionReference.Name + "-",
+			// The Ingress NS MUST be same as Router NS, check long discussion:
+			// https://github.com/kubernetes/kubernetes/issues/17088
+			// We need to revisit this in future, once Kubernetes supports cross namespace ingress
+			Namespace: podNamespace,
+		},
+		Spec: v1beta1.IngressSpec{
+			Rules: []v1beta1.IngressRule{
+				{
+					Host: trigger.Spec.Host,
```
smruthi2187: pardon my ignorance here but i dont understand the actual use of "host" parameter in the context of a normal http trigger. but now with the ingress using it, does it mean the router can be reachable using the host FQDN?

test/tests/test_ingress.sh
```diff
@@ -0,0 +1,45 @@
...
+log "Creating route for URL $relativeUrl"
+route_name=$(fission route create --url $relativeUrl --function $functionName --createingress| grep trigger| cut -d" " -f 2|cut -d"'" -f 2)
+trap "cleanup $route_name" EXIT
+
+log "Route $route_name created"
+
+sleep 5
+
+log "Verifying to route in ingress"
+
```
smruthi2187: may be add a log message here that says "dumping all ingress objects"

router/ingress.go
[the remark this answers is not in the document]
smruthi2187: also had another question, how will the users provide the annotations for ingress resources?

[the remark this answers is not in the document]
vishal-biyani: I thought about having the same name for ingress and trigger and now with functions etc., it makes sense. 

As for annotations - as of now we are not supporting it. In a future version, we might.

router/ingress.go
vishal-biyani: Since these are just labels, I will for now keep the label of both route and function name. Might be useful in some ways later. Maybe also add a trigger namespace reference might be useful.

test/tests/test_ingress.sh
vishal-biyani: Yeah, I would have liked it too, but that means deploying an ingress controller. Will take in another change.

router/ingress.go
vishal-biyani: I don't fully understand it, but if you map your LB to a DNS name, then the `host` becomes useful.

test/tests/test_ingress.sh
smruthi2187: ok, 👍

router/ingress.go
smruthi2187: ok, 👍

router/ingress.go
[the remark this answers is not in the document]
smruthi2187: ok 👍

vishal-biyani:
Hi @smruthi2187 So once the user starts using ingress controller, they will create a public-facing ELB with associated ingress controller such as with Nginx (ELB, in turn, will be mapped to a DNS name, but that's irrelevant to our discussion as of now).

So technically it is possible to have a single ingress resource and point all incoming requests to the router. In which case, all functions will be publically exposed. The forwarding will be something like (Although I must say that there are caveats and implementations are specific to ingress controller):

```
Ingress LB/fnName -> Ingress Controller/fnName -> Rewrite and forward to router/fnName or router/fission-function/fnName

```
But this means **all** functions are exposed to outside world. If we choose to do this on route level, then only some selected function are exposed. Well this is not entirely true as of today because the Fission routrer itself is public, so any route which is created and not exposed via ingress controller is still available publically through the Fission router.

But I believe in future - we might move to a implementation where:

- The router is only used for APIs exposed to Fission client and not for exposing function routes.
- Functions which need to be exposed internal to cluster can be exposed via a separate internal router (A router without a public facing IP)
- All functions which need to be accessed externally will be exposed via ingress controller. Of course some of this could be configurable.

That's my understanding so far. @soamvasani might have some thoughts too.

smruthi2187:
hmmm, so what i meant was, have one ingress object and map only those relativeUrls for which user opts for "createIngress" in that one object. instead of creating multiple ingress objects.
that will not expose all functions, but only those functions that want their relativeUrls exposed.

vishal-biyani:
@smruthi2187 Yes, what you are saying is possible. But then annotations won't be possible at function/route level and hence I feel 1:1 mapping between route & ingress is better. Does that sound reasonable? I am happy to discuss this more if needed in a quick call.

smruthi2187:
yes 👍 , makes sense! thanks !

smruthi2187:
LGTM.

vishal-biyani:
@smruthi2187 Mind taking a look, I did some more changes after your LGTM.

smruthi2187:
LGTM. i like that the names of triggers are now the same as ingress objects.

PR #692, opened 2018-05-22
Python Env: Support JSON Object return

xiekeyang:
If one Python function return a json data (Python `dict`), error will happen:

```
TypeError: 'dict' object is not callable
```

So, I reformat the return part to string type, then the objects can be parsed successfully.

The example:

```
def main():
    return {0: "Hello", 1: "World"}
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/692)

### DIFF
--- a/environments/python/server.py
+++ b/environments/python/server.py
@@ -78,7 +78,7 @@ def f():
     #   g.myKey = myValue
     # And the user func can then access that (after doing a "from flask import g").
     #
-    return userfunc()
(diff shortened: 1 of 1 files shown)


soamvasani:
Sorry, we can't do this as it breaks compatibility with existing functions.  Here's an example that would break: https://github.com/fission/fission/blob/master/examples/python/guestbook/add.py

Issue #664, opened 2018-05-02
Package creation for bigger archives needs env variables


vishal-biyani:
It looks like if we have an deploy archive, which is bigger than 256KB, we need to set Kubeconfig as well as FISSION_NAMESPACE environment variables:

```
$ fission package create  --env java --deploy hello-world-1.0-SNAPSHOT-jar-with-dependencies.jar 
unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined

$ export KUBECONFIG=~/.kube/config
$ fission package create  --env java --deploy hello-world-1.0-SNAPSHOT-jar-with-dependencies.jar 
Error getting storage service object from kubernetes :an empty namespace may not be set when a resource name is provided

$ export FISSION_NAMESPACE=fission
$ 
$ fission package create  --env java --deploy hello-world-1.0-SNAPSHOT-jar-with-dependencies.jar 
Package 'hello-world-1-0-snapshot-jar-with-dependencies-jar-ehhg' created
```
And relevant piece of code from https://github.com/fission/fission/blob/master/fission/common.go#L172:

```
if fileSize(fileName) < fission.ArchiveLiteralSizeLimit {
        contents := getContents(fileName)
        archive.Type = fission.ArchiveTypeLiteral
        archive.Literal = contents
    } else {
        // make a kubernetes client
        _, kubeClient, _, err := crd.GetKubernetesClient()
        if err != nil {
            fatal(err.Error())
        }
```

soamvasani:
This should be solved the same way as controller and router portforwards -- label the storage service and search for it by label, so user doesn't need to specify namespace (unless they have multiple installations)

PR #697, opened 2018-05-22
Archives bigger than 256K size need env variable for uploading

vishal-biyani:
This PR fixes https://github.com/fission/fission/issues/664

**Issue**

If you uploaded archives bigger than 256KB, then you had to set the KUBECONFIG and FISSION_NAMESPACE environment variables explicitly. This was because Kubeclient was used to get a handle of storage service and upload artefacts.

**Solution**

Instead of using Kubeclient to directly to find and talk to storage service, the following was done:

- A new API was created to expose the name of the storage service. This would give the name of storage service with the namespace. The input is the label of the storage service.
- Existing port forwarding mechanism was used to talk to storage service using name retrieved above. 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/697)

### DIFF
--- a/fission/common.go
+++ b/fission/common.go
@@ -156,20 +156,17 @@ func createArchive(client *client.Client, fileName string, specFile string) *fis
 		id, err := ssClient.Upload(fileName, nil)
 		checkErr(err, fmt.Sprintf("upload file %v", fileName))
 
-		url := fmt.Sprintf("%s/proxy/svcname?application=fission-storage", client.Url)
-		resp := httpRequest("GET", url, "", nil)
-		storageSvc := "http://storagesvc.fission"
-		if resp.StatusCode < 400 {
-			body, err := ioutil.ReadAll(resp.Body)
-			checkErr(err, "Package create")
-			storageSvc = "http://" + string(body)
-			defer resp.Body.Close()
-		}
+		storageSvc, err := client.GetSvcURL("application=fission-storage")
+		storageSvcURL := "http://" + storageSvc
+		checkErr(err, "get fission storage service name")
 
-		archiveUrl := ssClient.GetURL(id, storageSvc)
+		// We make a new client with actual URL of Storage service so that the URL is not
+		// pointing to 127.0.0.1 i.e. proxy. DON'T reuse previous ssClient
+		pkgClient := storageSvcClient.MakeClient(storageSvcURL)
+		archiveURL := pkgClient.GetUrl(id)
 
 		archive.Type = fission.ArchiveTypeUrl
-		archive.URL = archiveUrl
+		archive.URL = archiveURL
 
 		csum, err := fileChecksum(fileName)
 		checkErr(err, fmt.Sprintf("calculate checksum for file %v", fileName))
--- a/fission/common.go
+++ b/fission/common.go
@@ -156,16 +156,17 @@ func createArchive(client *client.Client, fileName string, specFile string) *fis
 		id, err := ssClient.Upload(fileName, nil)
 		checkErr(err, fmt.Sprintf("upload file %v", fileName))
 
-		archiveUrl := ssClient.GetUrl(id)
-
-		/*url := fmt.Sprintf("http://%s/proxy/svcname?application=fission-storage", client.Url)
-		resp := httpRequest("GET", url, nil, nil)
+		url := fmt.Sprintf("%s/proxy/svcname?application=fission-storage", client.Url)
+		resp := httpRequest("GET", url, "", nil)
+		storageSvc := "http://storagesvc.fission"
 		if resp.StatusCode < 400 {
 			body, err := ioutil.ReadAll(resp.Body)
 			checkErr(err, "Package create")
-			fmt.Print(string(body))
+			storageSvc = "http://" + string(body)
 			defer resp.Body.Close()
-		}*/
+		}
+
+		archiveUrl := ssClient.GetURL(id, storageSvc)
 
 		archive.Type = fission.ArchiveTypeUrl
 		archive.URL = archiveUrl
--- a/fission/common.go
+++ b/fission/common.go
@@ -149,20 +149,6 @@ func createArchive(client *client.Client, fileName string, specFile string) *fis
 		archive.Type = fission.ArchiveTypeLiteral
 		archive.Literal = contents
 	} else {
-		// make a kubernetes client
-		_, kubeClient, _, err := crd.GetKubernetesClient()
-		if err != nil {
-			log.Fatal(err.Error())
-		}
-
-		fissionNamespace := os.Getenv("FISSION_NAMESPACE")
-
-		// get svc end point for storagesvc
-		service, err := kubeClient.CoreV1().Services(fissionNamespace).Get("storagesvc", metav1.GetOptions{})
-		if err != nil {
-			log.Fatal(fmt.Sprintf("Error getting storage service object from kubernetes :%v", err.Error()))
-		}
-
 		u := strings.TrimSuffix(client.Url, "/") + "/proxy/storage"
 		ssClient := storageSvcClient.MakeClient(u)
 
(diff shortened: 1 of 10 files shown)


fission/common.go
```diff
@@ -168,31 +168,14 @@ func createArchive(client *client.Client, fileName string, specFile string) *fis
...
 
 		// TODO add a progress bar
 		id, err := ssClient.Upload(fileName, nil)
 		checkErr(err, fmt.Sprintf("upload file %v", fileName))
 
-		// this needs to be storagesvc.fission
-		storageSvcEndpoint := fmt.Sprintf("http://%s.%s/", service.Name, service.Namespace)
-		storageServiceClient := storageSvcClient.MakeClient(storageSvcEndpoint)
-		archiveUrl := storageServiceClient.GetUrl(id)
+		archiveUrl := ssClient.GetUrl(id)
```
life1347: This will work if FISSION_URL or server flag is set. But if both of them are not set, the CLI will set up local port forwarding to the controller. In this case, the curl will be `127.0.0.1:<localpost>`, and make the archive URL point to a local address, not the storage service URL. I think that's the main reason why CLI trying to get storagesvc service object from k8s API server to make an internal archive URL.

vishal-biyani: Oh right, good catch, I checked the package I created and indeed it has a URL which points to `127.0.0.1` - which won't work. I will re-work and update the PR.

vishal-biyani:
@life1347 I have fixed the issue and updated PR. I will write a test, but you can definitely start a review.

life1347:
Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[fission/common.go, line 178 at r4](https://reviewable.io/reviews/fission/fission/697#-LDq9gYV1wt4xnNF43-m:-LDq9gYV1wt4xnNF43-n:b-3x4ih1) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/fission/common.go#L178)):*
> ```Go
> 		checkErr(err, fmt.Sprintf("upload file %v", fileName))
> 
> 		url := fmt.Sprintf("%s/proxy/svcname?application=fission-storage", client.Url)
> ```

Could we move this to controller/client/xx.go as client function? So that we can reuse the logic to get other svc name as well

---

*[fission/common.go, line 181 at r4](https://reviewable.io/reviews/fission/fission/697#-LDqA2Ry4ODxXgWdM7Hi:-LDqA2Ry4ODxXgWdM7Hj:b-b51ki7) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/fission/common.go#L181)):*
> ```Go
> 		resp := httpRequest("GET", url, "", nil)
> 		storageSvc := "http://storagesvc.fission"
> 		if resp.StatusCode < 400 {
> ```

I don't prefer hardcode a default svc name here, and we should error out when a client gets non-200 http response from API server. How about

```
// in controller/client
if resp.StatusCode != http.StatusOK {
    return error
}

// continue here
return svcname


// in fission/common.go

svcName, err := client.GetSvcUrl("storagesvc") 
checkErr(err, "xxxx")

// continue here
```

---

*[storagesvc/client/client.go, line 115 at r4](https://reviewable.io/reviews/fission/fission/697#-LDqBMFu-lq-MqG48-af:-LDqBMFu-lq-MqG48-ag:btsixzk) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/storagesvc/client/client.go#L115)):*
> ```Go
> }
> 
> func (c *Client) GetURL(id string, hostUrl string) string {
> ```

It's really to get confused if there are tow GetUrl functions even they are in different capitalization. We can MakeClient with new svcUrl again and GetUrl from the new storagesvc client just like the previous implementation. https://github.com/fission/fission/blob/a76139381653eaba91ab7420879d6da811d04101/fission/common.go#L194-L195

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/697)*

life1347:
Review status: 7 of 9 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/697)*

life1347:
Review status: 3 of 8 files reviewed, 5 unresolved discussions (waiting on @life1347 and @vishal-biyani)

---

*[controller/client/core.go, line 93 at r6](https://reviewable.io/reviews/fission/fission/697#-LErClkI3XC7Rc5NG2nf:-LErClkI3XC7Rc5NG2ng:b-bdxmgf) ([raw file](https://github.com/fission/fission/blob/0d1b630b9dba77d4986c34a926af093364960c03/controller/client/core.go#L93)):*
> ```Go
> 	storageSvc := string(body)
> 
> 	defer resp.Body.Close()
> ```

Move defer before `body, err := ioutil.ReadAll(resp.Body)`, or we may have the leak problem if ReadAll failed

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/697)*

life1347:
@vishal-biyani LGTM, and just one comment need to be addressed ^

life1347:
Review status: 3 of 8 files reviewed, 5 unresolved discussions (waiting on @life1347 and @vishal-biyani)

---

*[controller/client/core.go, line 93 at r6](https://reviewable.io/reviews/fission/fission/697#-LErClkI3XC7Rc5NG2nf:-LErEDaN9TubtxUtWAZU:b-6kgdxc) ([raw file](https://github.com/fission/fission/blob/0d1b630b9dba77d4986c34a926af093364960c03/controller/client/core.go#L93)):*

Move defer before `body, err := ioutil.ReadAll(resp.Body)`, or we may have the leak problem if ReadAll failed
</blockquote></details>

and please add a check to ensure that resp is not nil

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/697)*

vishal-biyani:
Review status: 2 of 8 files reviewed, 5 unresolved discussions (waiting on @life1347)

---

*[controller/client/core.go, line 93 at r6](https://reviewable.io/reviews/fission/fission/697#-LErClkI3XC7Rc5NG2nf:-LEue9MC9z0HCiQWXuHk:b-896fix) ([raw file](https://github.com/fission/fission/blob/0d1b630b9dba77d4986c34a926af093364960c03/controller/client/core.go#L93)):*

and please add a check to ensure that resp is not nil
</blockquote></details>

Done.

---

*[fission/common.go, line 178 at r1](https://reviewable.io/reviews/fission/fission/697#-LD6ptpuc1FRogUVIVX1-r1-178:-LEueDj757JQaYnURXjo:b-896fix) ([raw file](https://github.com/fission/fission/blob/aba3842579c730d00d9e778b588cf8782e07fe25/fission/common.go#L178)):*

Oh right, good catch, I checked the package I created and indeed it has a URL which points to `127.0.0.1` - which won't work. I will re-work and update the PR.
</blockquote></details>

Done.

---

*[fission/common.go, line 178 at r4](https://reviewable.io/reviews/fission/fission/697#-LDq9gYV1wt4xnNF43-m:-LEueAam23kRp5E2Qbdy:b-896fix) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/fission/common.go#L178)):*

Could we move this to controller/client/xx.go as client function? So that we can reuse the logic to get other svc name as well
</blockquote></details>

Done.

---

*[fission/common.go, line 181 at r4](https://reviewable.io/reviews/fission/fission/697#-LDqA2Ry4ODxXgWdM7Hi:-LEueBnoEQ38Fpr9lyRS:b-896fix) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/fission/common.go#L181)):*

I don't prefer hardcode a default svc name here, and we should error out when a client gets non-200 http response from API server. How about

```
// in controller/client
if resp.StatusCode != http.StatusOK {
    return error
}

// continue here
return svcname


// in fission/common.go

svcName, err := client.GetSvcUrl("storagesvc") 
checkErr(err, "xxxx")

// continue here
```
</blockquote></details>

Done.

---

*[storagesvc/client/client.go, line 115 at r4](https://reviewable.io/reviews/fission/fission/697#-LDqBMFu-lq-MqG48-af:-LEueFMO-voYLon0ujx4:b-896fix) ([raw file](https://github.com/fission/fission/blob/e2b42e1938d067124e6f5755c8c498ee2b8c9f28/storagesvc/client/client.go#L115)):*

It's really to get confused if there are tow GetUrl functions even they are in different capitalization. We can MakeClient with new svcUrl again and GetUrl from the new storagesvc client just like the previous implementation. https://github.com/fission/fission/blob/a76139381653eaba91ab7420879d6da811d04101/fission/common.go#L194-L195
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/697)*

PR #700, opened 2018-05-23
fission servers should be configurable

### LINKED ISSUE #699: web server should be configurable
Router, Executor, Controller and others web servers includes some variables and consts that should be configurable by user.

Like dial timeout, keep alive time, and max retries limit, which is according to user's network.

These seems had better to be stored on configuration file (not k8s config spec) and be loaded when server starting up.

I draft it on https://github.com/xiekeyang/fission/commit/af28396049fc62fbdf9fb9eb6d0107ee4d04d523

This set up a `config.yaml` file for server loading, and record some patameters in context, to keep global consistency.

@soamvasani  What do you think?

xiekeyang:
Fix #699 .

This allow user to configurate server start-up parameters by config.yaml.
    
It just open the interface and template, to move a little variables from hard code to config file.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/700)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,34 @@
+package config
+
+import (
+	"io/ioutil"
+	"time"
+
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
+	Router Router `yaml:"router,omitempty"`
+}
+
+type Router struct {
+	DialTimeout  time.Duration `yaml:"dialtimeout,omitempty"`
+	AliveTimeout time.Duration `yaml:"alivetimeout,omitempty"`
+	MaxRetries   int           `yaml:"maxretries,omitempty"`
+}
+
+func LoadConfig(path string) (*Config, error) {
+	cfg := &Config{}
+
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	err = yaml.Unmarshal(b, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg, err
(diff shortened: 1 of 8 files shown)


soamvasani:
Yes to making these configurable.  But the change can be much simpler, with just environment variables.  We can then specify them through helm values, and the user can also edit the Kubernetes resource.  No need for new YAML parsing, no need for a new context package.

xiekeyang:
@soamvasani Thanks for your comment.

However, I think env config seems not to be a very good way. 

* ENV seems not to support structure? It is not easy to manage big variable structure.

* ENV has to be parsed to variable type every where we want to use, like 

```go
pruneInterval, err := strconv.Atoi(os.Getenv("PRUNE_INTERVAL"))
```

* It is risky to over write env. One has to search and collect all exist envs, to avoid name  confliction. During server running, ENV may be over written by any other process incorrectly. And we will get the wrong value.

There are open packages to parse a bunch of ENVs to the right type variable structure. But it return back to original issue, why not do it by `yaml` config? 

In consideration of some popular projects, Kubernetes seems to configure mainly on container resource, but not hard code parameters inside it. After write yaml config, we need just mount it into container path. I feel it not very complex. 

Yaml management is formatted and more safety. With `context` supports global handlers the life-cycle controllable. The self-defined `context` package wrap golang context, and help to parse k/v item.

Please reconsider it, thanks!

xiekeyang:
Close it, and move the configuration to container environment.

PR #701, opened 2018-05-23
Router liveness

smruthi2187:
This solves the bug that causes the issue https://github.com/fission/fission/issues/641

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/701)


router/functionHandler.go
```diff
@@ -153,6 +153,11 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 		overhead := time.Since(startTime)
 
+		// if transport.RoundTrip succeeds and it was a cached entry, then tapService
```
soamvasani: comment needs update?

Issue #662, opened 2018-05-02
build status should not be set to pending for a deploy pkg


smruthi2187:
steps to reproduce 
1. fission env create --name <env_name> --image <image_name>
2. fission fn create --name <func_name> --env <env_name> --deploy <deploy_pkg_1>
3. fission fn update --env <env_name> --name <func_name> --deploy <deploy_pkg_2>
at this point, updatePackage function https://github.com/fission/fission/blob/master/fission/package.go#L138  sets needToBuild=true
and if needToBuild is set to true, we end up setting BuildStatus = pending. 

as a fix, in addition to checking if needToBuild = true, we should also check if the package is a source pkg and then set BuildStatus=pending

sanketsudake:
This has been inactive for 1653 days. We're scrubbing the backlog and closing stale items to keep it actionable. If this is still relevant, please reopen or comment and we'll pick it back up.

PR #702, opened 2018-05-23
Fix for #662: avoid unnecessary builds

smruthi2187:
When a fission function/package update is invoked with the same env reference,  we set a flag to let builder know that the source archive has to be rebuilt into a deployment archive.

so originally i thought it was enough to just check if the above flag is set and if the srcArchive was passed to the function here : https://github.com/fission/fission/blob/master/fission/package.go#L170
but, realized that that is not enough.

there may be a case where an existing function may be updated with a brand new environment reference (which has builder image too) and the user doesnt provide --src while making the update,
ex : fn update <name> --env new-env.
in this case, we should still set the buildStatus to pending so that the source is re-built into a deploy archive. 

hence, the best solution for this issue is to not call updatePackage with env parameter if the env reference hasnt changed from existing spec even though the user may pass this parameter during update.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/702)


soamvasani:
I'm pretty confused.  I thought the bug #662 is about the user specifying no source package but a build being attempted anyway.  Is it also about environments being "updated" without being changed?  Wouldn't that be a separate issue?

smruthi2187:
so, the bug #662 is about user doing a `fn update` for a function with `--env <same_env_ref_as_existing_one>. 
due to env being updated (although not in reality), we set needToBuild = true here : https://github.com/fission/fission/blob/master/fission/package.go#L170

if the user just does a `fn update <name> --deploy <new_deploy_code>`, we dont set the above flag to true. so build wont be triggered.

looking at the code : if you trace `fnUpdate` code here[]https://github.com/fission/fission/blob/master/fission/function.go#L370 and then step into `updatePackage` here[https://github.com/fission/fission/blob/master/fission/package.go#L170], you can see that we end up setting needToBuild = true even though there's no new env referenced as part of this update.

fission/function.go
```diff
@@ -392,6 +392,11 @@ func fnUpdate(c *cli.Context) error {
 
 	envName := c.String("env")
 	envNamespace := c.String("envNamespace")
+	// if the new env specified is the same as the old one, no need to update package with envname
```
soamvasani: Isn't this true for all update params?  E.g. package, build command, etc.?

fission/package.go
```diff
@@ -116,6 +116,11 @@ func pkgUpdate(c *cli.Context) error {
 	})
 	checkErr(err, "get package")
 
+	// if the new env specified is the same as the old one, no need to update package with envname
```
soamvasani: Same comment as above

fission/package.go
```diff
@@ -147,7 +152,21 @@ func updatePackage(client *client.Client, pkg *crd.Package, envName, envNamespac
 	if len(envName) > 0 {
 		pkg.Spec.Environment.Name = envName
 		pkg.Spec.Environment.Namespace = envNamespace
-		needToBuild = true
+
+		// we need to fetch the env to find out if it has a builder image, only then set needToBuild to true
```
soamvasani: This is a separate problem, and it's bigger than this flag.  Simply avoiding a build doesn't really help the user here at all.  A package that has only source and now way to build it isn't useful.  Can you track a separate issue for this?  We should catch this case in function/package creation.

smruthi2187: yeah i agree it isn't really useful to not build. but this happens only on fn/pkg update with just `--env` flag. 
if the user has provided a `--src` flag, then the needToBuild gets set to true in `if len(srcArchive)` condition. 

but we cant catch it during creation right? this happens only on updates. but what we can do is do the GET operation on the env in fnUpdate and pkgUpdatem functions and send another param to this updatePackage function denoting if the env reference passed has a builder image or is just a deploy image.

what do you think?

fission/package.go
smruthi2187: same as above.

fission/function.go
soamvasani: Ok.  Can you put this in a comment?

fission/package.go
soamvasani: If it's not useful, let's not do it.  Let's figure out something that actually helps the user, and do that.

We could add some sort of validation error for functions that have source being associated with envs that have no builder.  This validation will have to be run during function and package creation, and during updates of functions, packages, and environments.

fission/function.go
smruthi2187: ok, 👍

fission/package.go
smruthi2187: ok, makes sense.

fission/function.go
```diff
@@ -392,6 +392,13 @@ func fnUpdate(c *cli.Context) error {
 
 	envName := c.String("env")
 	envNamespace := c.String("envNamespace")
+	// if the new env specified is the same as the old one, no need to update package
+	// same is true for all update parameters, but, for now, we dont check all of them - because, its ok to
+	// re-write the object with same old values, we just end up getting a new resource version for the object.
+	if envName == function.Spec.Environment.Name && envNamespace == function.Spec.Environment.Namespace {
+		envName = ""
```
life1347: Since we adopted multi-tenancy, the the namepsace of evnironment may be changed. For example, user create a same name environment obj but in different namespace.
So please set namespace empty here as well.
https://github.com/fission/fission/blob/80fadcc7b871cb410c50b465c40ec0d20f79dd11/fission/package.go#L148-L152
I also noticed there is a tiny bug in `package.go`, we need to check `envName ` and `envNamespace ` separately in case that user only changed the namespace. I am fine to fix this in later PR, but it will be great if we can solve it in this PR as well.

fission/package.go
```diff
@@ -116,6 +116,13 @@ func pkgUpdate(c *cli.Context) error {
 	})
 	checkErr(err, "get package")
 
+	// if the new env specified is the same as the old one, no need to update package
+	// same is true for all update parameters, but, for now, we dont check all of them - because, its ok to
+	// re-write the object with same old values, we just end up getting a new resource version for the object.
+	if envName == pkg.Spec.Environment.Name && envNamespace == pkg.Spec.Environment.Namespace {
+		envName = ""
```
life1347: same above

smruthi2187:
hi @life1347 thanks for reviewing this! I had completely forgotten about this _old_ PR.
I'll take a look at your review comments and fix them in a day or two!

smruthi2187:
hi @life1347 , i've addressed the review comments. i'll wait for Green CI.
In the meanwhile, i notice  that issue #808 is referenced above :) may be by mistake.. hoping it causes no harm.

life1347:
LGTM

smruthi2187:
Will close this PR to prevent unexpected changes to commit history.
Opened this PR instead : https://github.com/fission/fission/pull/866

PR #705, opened 2018-05-26
Fix go env vendor and add glide support

LittleFox94:
This change fixes the vendor directory for the go environment builder
and adds support for dependency management with glide.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/705)

### DIFF
--- a/environments/go/builder/build.sh
+++ b/environments/go/builder/build.sh
@@ -4,6 +4,16 @@ if [ -d ${SRC_PKG} ]
 then
     echo "Building in directory ${SRC_PKG}"
     cd ${SRC_PKG}
+
+    if [ -f glide.yaml ]; then
+        go get -u github.com/Masterminds/glide
+        glide install
+    fi
+
+    if [ -d vendor ]; then
+        cp -r vendor/* /usr/src/
+    fi
+
     go build -buildmode=plugin -i -o ${DEPLOY_PKG} .
 elif [ -f ${SRC_PKG} ]
 then


environments/go/builder/build.sh
```diff
@@ -4,6 +4,16 @@ if [ -d ${SRC_PKG} ]
 then
     echo "Building in directory ${SRC_PKG}"
     cd ${SRC_PKG}
+
+    if [ -f glide.yaml ]; then
+        go get -u github.com/Masterminds/glide
```
soamvasani: Can we download glide in the Dockerfile itself, that way the build doesnt have to wait for it?

environments/go/builder/build.sh
```diff
@@ -4,6 +4,16 @@ if [ -d ${SRC_PKG} ]
...
+        glide install
+    fi
+
+    if [ -d vendor ]; then
+        cp -r vendor/* /usr/src/
```
soamvasani: I'm a bit confused by this copying, why do we need this?

environments/go/builder/build.sh
LittleFox94: I can change this, yes :)

The idea was not to bloat the image with all possible vendoring tools but only download on demand (builders are reused, right?)

environments/go/builder/build.sh
LittleFox94: Golang ignores the vendor folder for some reason. Since I needed it, I tried things and this worked ^^"

Might be because the usercode is not stored gopath (a friend suggested it ignores vendor then) or because we compile a plugin. Could not find a documented reason yet :/

LittleFox94: https://github.com/golang/go/issues/14566

Not having the usercode in gopath seems to be the actual problem.

vishal-biyani:
Can we close this in favor of https://github.com/fission/fission/issues/1130

PR #709, opened 2018-05-30
Check spec directory exists before reading spec files

life1347:
### Root cause
`filepath.Walk` does nothing if spec directory does not exist

### Solution
1. Use `os.Stat` to check spec directory exists before reading spec files
2. Fix spec example

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/709)

### DIFF
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -523,7 +523,7 @@ func readSpecs(specDir string) (*FissionResources, error) {
 
 	// make sure spec directory exists before continue
 	if _, err := os.Stat(specDir); os.IsNotExist(err) {
-		fatal(fmt.Sprintf("Spec directory %v does not exist", specDir))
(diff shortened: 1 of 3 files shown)


fission/spec.go
```diff
@@ -520,6 +520,12 @@ func (fr *FissionResources) parseYaml(b []byte, loc *location) error {
 // readSpecs reads all specs in the specified directory and returns a parsed set of
 // fission resources.
 func readSpecs(specDir string) (*FissionResources, error) {
+
+	// make sure spec directory exists before continue
+	if _, err := os.Stat(specDir); os.IsNotExist(err) {
+		fatal(fmt.Sprintf("Spec directory %v does not exist", specDir))
```
soamvasani: Whenever add a suggestion for the user to fix the problem, something like:

Spec directory ___ doesn't exist.  Run "fission spec init" to create it.

life1347: done

soamvasani:
LGTM, thanks!

PR #712, opened 2018-05-30
Fix CLI failed to set up port-forwarding when multiple controller pods exist in the same namespace

life1347:
### Root cause
`runPortForward ` only checks amount of controllers not the namespace

### Solution
Use map to filter same namespace and check length of namespace list.

### DIFF
--- a/fission/portforward.go
+++ b/fission/portforward.go
@@ -68,12 +68,18 @@ func runPortForward(kubeConfig string, labelSelector string, localPort string, f
 
 	// make a useful error message if there is more than one install
 	if len(podList.Items) > 1 {
-		namespaces := make([]string, 0)
+		namespaces := make(map[string]struct{})
+		nsList := make([]string, 0)
 		for _, p := range podList.Items {
-			namespaces = append(namespaces, p.Namespace)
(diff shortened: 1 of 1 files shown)


fission/portforward.go
```diff
@@ -68,12 +68,18 @@ func runPortForward(kubeConfig string, labelSelector string, localPort string, f
...
+		namespaces := make(map[string]struct{})
+		nsList := make([]string, 0)
 		for _, p := range podList.Items {
-			namespaces = append(namespaces, p.Namespace)
+			if _, ok := namespaces[p.Namespace]; !ok {
+				namespaces[p.Namespace] = struct{}{}
+				nsList = append(nsList, p.Namespace)
+			}
+		}
+		if len(nsList) > 1 {
```
soamvasani: Don't you have to handle the case where there are more pods but only one namespace?  For some reason?  In that case you have to talk to any pod that's in a healthy state...

life1347: Thank for pointing that out! Just push a new commit to fix the problem.

vishal-biyani:
LGTM

PR #713, opened 2018-05-31
Allow router round-trip to be configurable

xiekeyang:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/713)

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -33,18 +33,24 @@ import (
 	executorClient "github.com/fission/fission/executor/client"
 )
 
+type tsRoundTripperParams struct {
+	timeout         time.Duration
+	timeoutExponent int
+	keepAlive       time.Duration
+	maxRetries      int
+}
+
 type functionHandler struct {
-	fmap        *functionServiceMap
-	executor    *executorClient.Client
-	function    *metav1.ObjectMeta
-	httpTrigger *crd.HTTPTrigger
+	fmap                 *functionServiceMap
+	executor             *executorClient.Client
+	function             *metav1.ObjectMeta
+	httpTrigger          *crd.HTTPTrigger
+	tsRoundTripperParams *tsRoundTripperParams
 }
 
 // A layer on top of http.DefaultTransport, with retries.
 type RetryingRoundTripper struct {
-	maxRetries     int
-	initialTimeout time.Duration
-	funcHandler    *functionHandler
+	funcHandler *functionHandler
 }
 
 // RoundTrip is a custom transport with retries for http requests that forwards the request to the right serviceUrl, obtained
@@ -92,7 +98,6 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 	}
 
 	// set the timeout for transport context
-	timeout := roundTripper.initialTimeout
 	transport := http.DefaultTransport.(*http.Transport)
 	// Disables caching, Please refer to issue and specifically comment: https://github.com/fission/fission/issues/723#issuecomment-398781995
 	transport.DisableKeepAlives = true
@@ -104,7 +109,9 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 		needExecutor = true
 	}
 
-	for i := 0; i < roundTripper.maxRetries-1; i++ {
+	executingTimeout := roundTripper.funcHandler.tsRoundTripperParams.timeout
+
+	for i := 0; i < roundTripper.funcHandler.tsRoundTripperParams.maxRetries-1; i++ {
 		if needExecutor {
 			log.Printf("Calling getServiceForFunction for function: %s", roundTripper.funcHandler.function.Name)
 
@@ -149,8 +156,8 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 		// over-riding default settings.
 		transport.DialContext = (&net.Dialer{
-			Timeout:   timeout,
-			KeepAlive: 30 * time.Second,
+			Timeout:   executingTimeout,
+			KeepAlive: roundTripper.funcHandler.tsRoundTripperParams.keepAlive,
 		}).DialContext
 
 		overhead := time.Since(startTime)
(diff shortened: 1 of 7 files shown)


xiekeyang:
Rebased, PTAL, thanks!

xiekeyang:
This allows customer to configure router round-trip, on keep alive time, dial timeout, and max retries. These relay on customer network, so should be configurable.

xiekeyang:
This seems ready to be merged.

soamvasani:
LGTM, thanks @xiekeyang!

Would you please rebase, and also remove the Signed-off line from your PRs?  We don't follow that convention in this project.

xiekeyang:
> remove the Signed-off line from your PRs? We don't follow that convention in this project.

Ah, sorry that I didn't notice this before. I removed it and in other PRs, and would follow this project convention in future.

xiekeyang:
@soamvasani I removed the `Signed-off line`, and rebased code, PTAL, thanks!

router/functionHandler.go
```diff
@@ -92,7 +92,11 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 	}
 
 	// set the timeout for transport context
-	timeout := roundTripper.initialTimeout
+	timeout, err := time.ParseDuration(os.Getenv("ROUTER_ROUND_TRIP_TIMEOUT"))
```
soamvasani: This parses an env var on every single HTTP request; it's probably not necessary.  Can you move it to the initialization of the roundtripper?

router/functionHandler.go
```diff
@@ -102,7 +106,17 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 		needExecutor = true
 	}
 
-	for i := 0; i < roundTripper.maxRetries-1; i++ {
+	keepAlive, err := time.ParseDuration(os.Getenv("ROUTER_ROUND_TRIP_KEEP_ALIVE_TIME"))
```
soamvasani: same with these, move to init?

xiekeyang:
@soamvasani The structure `RetryingRoundTripper` and `functionHandler` are deeply coupled. 

They build reference each other, and includes some nested structure. That seems a little confused.

I would like to decouple them firstly.

xiekeyang:
@soamvasani 

Here I setup new structure `tsRoundTripperParams` embedded in `functionHandler`. This is used to transfer the round tripper parameters. It is initialized from environment variables only ones when router service starting up.

I keep mainly the previous structures in router package. 

Later, we might need figure out to reorganize them (likely use context), which may be a little big effort.

PTAL, thanks so much!

router/functionHandler.go
```diff
@@ -183,9 +187,8 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 		// just retry after backing off for timeout period.
 		if serviceUrlFromExecutor {
 			log.Printf("request to %s errored out. backing off for %v before retrying",
-				req.URL.Host, timeout)
-			timeout *= time.Duration(2)
-			time.Sleep(timeout)
+				req.URL.Host, roundTripper.funcHandler.tsRoundTripperParams.timeout)
+			time.Sleep(roundTripper.funcHandler.tsRoundTripperParams.timeout * time.Duration(2))
```
soamvasani: This is actually not the same as timeout*=2; you now have a constant sleep through all iterations of the loop.  What you want is something like:

```
<before the loop>
timeout := roundTripper.funcHandler.roundTripperParams.timeout

<in the loop>
...
   timeout *= time.Duration(2)
   time.Sleep(timeout)
```

soamvasani: BTW if you make the "2" configurable as ROUNDTRIP_TIMEOUT_EXPONENT or something, that timeout-doubling behaviour also becomes configurable, so you could set it to 1 for much shorter timeouts (i.e. quicker failures, in case that's what you want).

xiekeyang: Done.

xiekeyang:
@soamvasani changed [here](https://github.com/fission/fission/pull/713/files#diff-d641ab51fb0b026fc1285a83613d1e11R180) and rebased. PTAL, thanks.

router/functionHandler.go
```diff
@@ -149,8 +156,8 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 		// over-riding default settings.
 		transport.DialContext = (&net.Dialer{
-			Timeout:   timeout,
-			KeepAlive: 30 * time.Second,
+			Timeout:   roundTripper.funcHandler.tsRoundTripperParams.timeout,
```
soamvasani: this should be `executingTimeout`

router/functionHandler.go
```diff
@@ -183,9 +190,11 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 		// just retry after backing off for timeout period.
 		if serviceUrlFromExecutor {
 			log.Printf("request to %s errored out. backing off for %v before retrying",
-				req.URL.Host, timeout)
-			timeout *= time.Duration(2)
-			time.Sleep(timeout)
+				req.URL.Host, roundTripper.funcHandler.tsRoundTripperParams.timeout)
```
soamvasani: This should be `executingTimeout`

router/httpTriggers.go
```diff
@@ -49,17 +49,20 @@ type HTTPTriggerSet struct {
 	functions         []crd.Function
 	funcStore         k8sCache.Store
 	funcController    k8sCache.Controller
+
+	tsRoundTripperParams tsRoundTripperParams
 }
 
 func makeHTTPTriggerSet(fmap *functionServiceMap, fissionClient *crd.FissionClient,
-	kubeClient *kubernetes.Clientset, executor *executorClient.Client, crdClient *rest.RESTClient) (*HTTPTriggerSet, k8sCache.Store, k8sCache.Store) {
+	kubeClient *kubernetes.Clientset, executor *executorClient.Client, crdClient *rest.RESTClient, params tsRoundTripperParams) (*HTTPTriggerSet, k8sCache.Store, k8sCache.Store) {
```
soamvasani: Very minor comment, but it's better to pass a `*tsRoundTripperParams` (pointer instead of copy of struct)

router/functionHandler.go
xiekeyang: Wow, my fault.

xiekeyang:
@soamvasani Thanks lots for your review.

I fixed the inappropriate points on your comment.

* timeout in round tripper is all increased with loop.

* round tripper parameters is wrapped and transferred by pointer.

PTAL, thanks so much!

soamvasani:
Thanks @xiekeyang, merged! 🎉

PR #714, opened 2018-05-31
Logger daemonset's update strategy

vishal-biyani:
The default update strategy for DaemonSet is `OnDelete` (Meaning the user has to go and manually delete the pod). This has been the default strategy and is still there for backward compatibility. We should instead switch to `RollingUpdate` so that users don't have to intervene in case logger changes.

https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#daemonset-update-strategy

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/714)


soamvasani:
Yes, LGTM, thanks!

PR #715, opened 2018-05-31
Fix for Windows CLI Port Forwarding

thejosephstevens:
So the root cause here was a bit obscure, but there was an error being swallowed by the `findFreePort()` function in `fission/portforward.go`. The error was `file tcp [::]:65394: not supported by windows`, which was thrown at line 25 when a file open is attempted, generating a file with colon's in the name, which is illegal in Windows. The reason for this file is because of an idiosyncrasy in Go where when you close a TCP Listener, it asynchronously frees the actual port it was consuming, resulting in behavior where you could attempt to use that same port immediately afterward and result in a panic because it's still bound in the OS. This issue is discussed at length [here](https://github.com/golang/go/issues/21856), and at this point has been taken on as an official proposal. It looks as though a fix is supposed to be implemented in Go 1.10, if it hasn't already. The file open that was previously performed here creates a file that I believe is connected to the I/O buffer associated with the TCP listener, and when it closes it is closed synchronously. The logic here was that closing that file can only occur after the write buffer is flushed and therefore the listener must necessarily have unbound once the file close completes.

I have built the CLI in Go 1.10 and done hello world testing with these changes, and have run into no issues with port binding. I'm going to trace the code path here and check to be sure that useful errors are thrown if this turns out to not work properly, so don't merge this yet.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/715)

### DIFF
--- a/fission/install-windows-cli.ps1
+++ b/fission/install-windows-cli.ps1
@@ -1,3 +1,5 @@
 go install
-rm $Env:GOPATH\bin\fission-test.exe
+if (Test-Path $Env:GOPATH\bin\fission-test.exe -PathType Leaf) {
+    rm $Env:GOPATH\bin\fission-test.exe
+}
 mv $Env:GOPATH\bin\fission.exe $Env:GOPATH\bin\fission-test.exe
--- a/fission/install-windows-cli.ps1
+++ b/fission/install-windows-cli.ps1
@@ -1,5 +1,0 @@
-go install
-if (Test-Path $Env:GOPATH\bin\fission-test.exe -PathType Leaf) {
-    rm $Env:GOPATH\bin\fission-test.exe
-}
-mv $Env:GOPATH\bin\fission.exe $Env:GOPATH\bin\fission-test.exe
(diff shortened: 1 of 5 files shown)


thejosephstevens:
Okay, I've now tried to force a break by re-writing `findFreePort()` as this, and I couldn't get it to throw an error (inspired by the test [here](https://gist.github.com/slingamn/3b0f81169de6578c732ec279828c4866)), so I'm confident that this is safe, at least in Go 1.10. @soamvasani can you take a look so we can move this through?
```func findFreePort() (string, error) {
	listener, err := net.Listen("tcp", ":0")
	if err != nil {
		return "", err
	}

	port := strconv.Itoa(listener.Addr().(*net.TCPAddr).Port)

	err = listener.Close()
	if err != nil {
		return "", err
	}

	for {
		newlistener, err := net.Listen("tcp", fmt.Sprintf(":%s", port))
		if err != nil {
			fmt.Println(err)
			continue
			//panic(err)
		}
		newlistener.Close()
		break
	}

	return port, nil
}```

File: fission/install-windows-cli.ps1 [hunk 1]
```diff
@@ -0,0 +1,3 @@
+go install
```
soamvasani: Can we remove this file?  You could test simply by having your gopath bin ahead of the rest of the path.

PR #720, opened 2018-06-01
Push NATS error messages into error queue

Amusement:
New PR with better commit history.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/720)

### DIFF
--- a/fission/mqtrigger.go
+++ b/fission/mqtrigger.go
@@ -72,7 +72,10 @@ func mqtCreate(c *cli.Context) error {
 
 	maxRetries := c.Int("maxretries")
 
-	if maxRetries < 1 || maxRetries > 10 {
+	// If max retries is not given (i.e. is 0), default to 1; otherwise validate the value
+	if maxRetries == 0 {
+		maxRetries = 1
+	} else if maxRetries < 1 || maxRetries > 10 {
 		fatal("Maximum number of retries must be between 1 and 10")
 	}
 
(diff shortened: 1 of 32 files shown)


mqtrigger/messageQueue/nats.go
```diff
@@ -107,35 +107,66 @@ func msgHandler(nats *Nats, trigger *crd.MessageQueueTrigger) func(*ns.Msg) {
...
+				publishErr := nats.nsConn.Publish(trigger.Spec.ErrorTopic, body)
+				if publishErr != nil {
+					log.Error("Failed to publish error to error topic: %v", err)
+					// Do not ack message, quit
+					return
+				}
+			}
+		}
+
+		// Trigger acks message only if a request was processed successfully
```
soamvasani: After the max retries, we should always ack the message.  Otherwise there's no maximum, we'll just retry infinitely.

But, that means the message might be lost -- for this reason, the message itself can be placed on another queue.  That other queue can be then be handled manually by users.  Can you file this part as a separate issue?

fission/mqtrigger.go
```diff
@@ -68,6 +68,14 @@ func mqtCreate(c *cli.Context) error {
 		fatal("Listen topic should not equal to response topic")
 	}
 
+	errorTopic := c.String("errortopic")
+
+	maxRetries := c.Int("maxretries")
+
+	if maxRetries < 1 || maxRetries > 5 {
```
soamvasani: Why limit it?  0 should be allowed.  As should be any maximum.

mqtrigger/messageQueue/nats.go
Amusement: https://github.com/fission/fission/issues/726

PR #725, opened 2018-06-05
Environment warning message bugfix

soamvasani:
This simplifies the message when there are multiple environments in the same namespace.  It also moves it into the verbose category.  Otherwise we'd see it too often, I think.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/725)


smruthi2187:
LGTM.

PR #727, opened 2018-06-06
[Not to be merged] Testing for update package issue

vishal-biyani:
Temp PR

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/727)

PR #730, opened 2018-06-07
Use short gracePeriod during test for function with newdeploy manager as executor type

life1347:
### Root cause
The requests sent to old function pod during the function update process and caused test failure.

### Solution
As a temp workaround, use short terminalGracePeriod so that the traffic can be sent to new function pod in short time.

But the behavior of traffic redirection conflicts with the steps described in [Termination of Pods](https://kubernetes.io/docs/concepts/workloads/pods/pod/). At step 5, the function pod is removed from endpoints list for service and should not redirect any requests after that. So even a longer terminalGracePeriod should not affect the traffic redirection. Any comment or thought is welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/730)

### DIFF
--- a/executor/cleanup.go
+++ b/executor/cleanup.go
@@ -70,13 +70,6 @@ func cleanup(client *kubernetes.Clientset, namespace string, instanceId string) 
 		return err
 	}
 
-	// Pods might still be running user functions, so we give them
-	// a few minutes before terminating them.  This time is the
-	// maximum function runtime, plus the time a router might
-	// still route to an old instance, i.e. router cache expiry
-	// time.
-	time.Sleep(6 * time.Minute)
-
 	err = cleanupPods(client, namespace, instanceId)
 	if err != nil {
 		return err
--- a/executor/newdeploy/newdeploy.go
+++ b/executor/newdeploy/newdeploy.go
@@ -147,7 +147,7 @@ func (deploy *NewDeploy) getDeploymentSpec(fn *crd.Function, env *crd.Environmen
 
 	targetFilename := "user"
 
-	gracePeriodSeconds := int64(6 * 60)
+	gracePeriodSeconds := fission.DefaultTerminationGracePeriod
 	if env.Spec.TerminationGracePeriod > 0 {
 		gracePeriodSeconds = env.Spec.TerminationGracePeriod
 	}
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -479,7 +479,7 @@ func (gp *GenericPool) createPool() error {
 
 	// Use long terminationGracePeriodSeconds for connection draining in case that
 	// pod still runs user functions.
-	gracePeriodSeconds := int64(6 * 60)
+	gracePeriodSeconds := fission.DefaultTerminationGracePeriod
 	if gp.env.Spec.TerminationGracePeriod > 0 {
 		gracePeriodSeconds = gp.env.Spec.TerminationGracePeriod
 	}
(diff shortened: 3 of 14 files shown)


vishal-biyani:
@life1347 Great, I think @smruthi2187 's analysis and your test confirms that the problem is indeed with old pod being active for 30 seconds. But instead of the user specifying the grace period/fixing tests, should we have some smaller value of grace period for new deploy executor. Currently, newdeploy.go has:

```
	gracePeriodSeconds := int64(6 * 60)
	if env.Spec.TerminationGracePeriod > 0 {
		gracePeriodSeconds = env.Spec.TerminationGracePeriod
	}
```

Should we default to a lower value and the user can override if he wants to?

life1347:
@vishal-biyani 
The 360s is come from the comment here, so I think long graceperiod is still better for serving function from the aspect of user.
https://github.com/fission/fission/blob/8a097abf9b3a7ca553dd9692c39847a33d0e4db5/poolmgr/cleanup.go#L56-L62

vishal-biyani:
Ok, I agree that default `graceperiod` should be higher than 1 but 360 seconds (6*60) seems a bit too high to me. The Kubernetes default is 30 seconds seems lot reasonable. Also in case of clean up we are checking if function was accessed for X minutes and only then deleting so that should not affect the termination grace period.

Termination grace period to be short is important because after the function is updated, user should not have to wait too long to access new function.

life1347:
I agree that short grace period is important for the user to access new function, but also the length of grace period should depend on how long we want a function to run after function update. 
Be honest, I am not sure what value is more appropriate for such cases.

vishal-biyani:
Hi @life1347 Here is summary of our quick zoom call:

1) The change you are doing in below place is not in idleObjectReaper - but in the one time clean up when Fission starts/upgrades

```
         // maximum function runtime, plus the time a router might
	// still route to an old instance, i.e. router cache expiry
	// time.
	time.Sleep(6 * time.Minute)
```

2) For idleObjectReaper - it does not make sense that we increase the grace period time or rely entirely on the grace period for deletion. Because the idleObjectReaper actually checks if the function was used in last idlePodReapTime minutes and then deletes it. In this case it is best to have minimum grace period.

3) In case of an update to a function again longer grace period is not desirable as the user may want to use updated function sooner than later. 

4) Which means we stick to a small default value of grace period and if a user wants run functions which continue far beyond the grace period then they should explicitly input that when creating environments/functions. That also means users are aware of the delayed updates.

5) Lastly as part of experimentation to find out ideal grace period, I tried to reduce grace period to like sub 30 second times, but for some reason, neither pool manager nor executor picked it up (It still defaulted to 360). I will test it a bit more and find out how to configure it. This can be tracked here: https://github.com/fission/fission/pull/727

smruthi2187:
The changes LGTM for our tests. for now, leaving `terminationGracePeriod` to 360s might be ok, but i feel a follow up commit can default it to 30s because Kubernetes defaults it to 30s for a reason.

For point 3 @vishal-biyani  mentioned above
```
In case of an update to a function again longer grace period is not desirable as the user may want to use updated function sooner than later.
```
I agree with it too. Users might prefer it sooner than later. even if there are a few users that might prefer otherwise, they do have an option of specifying the `gracePeriod` during `env create`.

soamvasani:
The non-test part of this PR, which makes the grace period more configurable, looks good to me.  That's a good thing to have.

But don't set the grace period low just because it makes this bug disappear.  I think it's fine to simply disable the newdeploy update test to unblock other merges, while we continue to investigate the failures.

vishal-biyani:
We probably don't need this now that #742 verifies the issue. Also you should probably comment on potential performance impact due to that change

soamvasani:
@life1347 I'm not super clear on this, is this still a change we want (assuming we remove the test changes from this PR)?

life1347:
@soamvasani I will open another PR for the non-test part, so close it for now.

PR #731, opened 2018-06-07
ArchiveLiteralSizeLimit: Use Constant Instead Hard Code

xiekeyang:
This use constant to instead hard code for ArchiveLiteralSizeLimit. That ensure if the limitation
value is changed in cli, the controller server can follow it. Otherwise, the condition will return
error, and make error HTTP response.

Later, we should delete it in fission-cli, and make it configurable it chart, to work in fission-
servers only.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/731)


life1347:
LGTM

PR #732, opened 2018-06-07
Bug Fix: Fission Client should exit 1 when creating function with environment not found

xiekeyang:
The older mechanism, in the case of environment not exist, allows program to go on, and create a bad function.

This let customer get a incorrect result, and set up a unworkable function.
    
Actually we should stop the program when environment not found, and return failed result, with the error message.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/732)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -178,7 +178,7 @@ func fnCreate(c *cli.Context) error {
 		})
 		if err != nil {
 			if e, ok := err.(fission.Error); ok && e.Code == fission.ErrorNotFound {
-				fatal(fmt.Sprintf("Environment \"%v\" does not exist. Please create the environment before executing the function. \nFor example: `fission env create --name %v --image <image>`\n", envName, envName))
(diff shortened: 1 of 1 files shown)


soamvasani:
Hi @xiekeyang, thanks for the contribution!

This was intentional.  We don't want to force users to create resources in a certain order.  This follows patterns in Kubernetes (and other systems as well).  For example, you can create a Kubernetes service that points to a certain deployment, and create the deployment later, and Kubernetes will happily have those two find each other.  Relaxing the ordering constraint allows more flexible configuration changes, so you can delete/update/create functions/environments/triggers independently.  (Finally, since we're Kubernetes CRDs, we don't fully control the lifecycle of resources through our CLI; the user could use K8s YAMLs or even `helm install` a function or env.)

We do have a validator in the `fission spec` stuff that checks for missing dependencies and unused resources.  I think any sufficiently complex deployment will end up using specs.  The spec validator is strict by default, because there users are expected to fully specify the app's config upfront, and not change it without a spec change.

I think I'd be ok with adding an option like `--strict-validate` that would run the validator in the non-spec case too, but I'm a bit opinionated that the default behavior should not enforce strict ordering.

xiekeyang:
@soamvasani Understood. I would open another PR to implement it.

PR #733, opened 2018-06-07
Support annotations in environment specs

erwinvaneyk:
Annotations are pretty often used to augment or trigger some external service for a given pod. For example, we annotate our environment pods for Istio. To avoid having to add this manually in the source code, I propose that we provide the ability to the users to alter the annotations of an environment. These user-provided annotations are used in both the runtime and the builder pod of the environment. 

We could also expose the entire PodSpec, but I am not sure if (1) that makes things more convenient for users (2) that might open up a range of non-standard/untested behaviour when users can alter everything of the pod template. So instead I now went for the option to just add the annotations map to the environmentspec (but I am not particularly in favor or against any of the two options).

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/733)

### DIFF
--- a/buildermgr/envwatcher.go
+++ b/buildermgr/envwatcher.go
@@ -498,7 +498,10 @@ func (envw *environmentWatcher) createBuilderDeployment(env *crd.Environment, ns
 	sel := envw.getLabels(env.Metadata.Name, ns, env.Metadata.ResourceVersion)
 	var replicas int32 = 1
 
-	podAnnotation := make(map[string]string)
+	podAnnotation := env.Spec.Runtime.Annotations
+	if podAnnotation == nil {
+		podAnnotation = make(map[string]string)
+	}
 	if envw.useIstio && env.Spec.AllowAccessToExternalNetwork {
 		podAnnotation["sidecar.istio.io/inject"] = "false"
 	}
--- a/buildermgr/envwatcher.go
+++ b/buildermgr/envwatcher.go
@@ -498,12 +498,12 @@ func (envw *environmentWatcher) createBuilderDeployment(env *crd.Environment, ns
 	sel := envw.getLabels(env.Metadata.Name, ns, env.Metadata.ResourceVersion)
 	var replicas int32 = 1
 
-	podAnnotation := env.Spec.Runtime.Annotations
-	if podAnnotation == nil {
-		podAnnotation = make(map[string]string)
+	podAnnotations := env.Metadata.Annotations
+	if podAnnotations == nil {
+		podAnnotations = make(map[string]string)
 	}
 	if envw.useIstio && env.Spec.AllowAccessToExternalNetwork {
-		podAnnotation["sidecar.istio.io/inject"] = "false"
+		podAnnotations["sidecar.istio.io/inject"] = "false"
 	}
 
 	deployment := &v1beta1.Deployment{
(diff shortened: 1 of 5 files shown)


soamvasani:
As we discussed offline, simply copying the annotations from the environment CRD might make sense.  Then you don't have to add an annotations map to the specs.

soamvasani:
Thanks @erwinvaneyk , looks great to me.  @life1347 can you take a quick look too, please?

life1347:
Review status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @erwinvaneyk)

---

*[executor/poolmgr/gp.go, line 487 at r2](https://reviewable.io/reviews/fission/fission/733#-LEgjzIk8pRWJup6WgxM:-LEgjzIk8pRWJup6WgxN:b-tor5wv) ([raw file](https://github.com/fission/fission/blob/7aa67a9676ef979aef17018fb7e0c525c8ca5a21/executor/poolmgr/gp.go#L487)):*
> ```
> podAnnotations
> ```
Need to do the same thing to Newdeploy too.

---

*[test/tests/test_annotations.sh, line 91 at r2](https://reviewable.io/reviews/fission/fission/733#-LEgje1m9kHnYrIFBH1_:-LEgje1n1j9j8oa-A7v4:b-b4ankb) ([raw file](https://github.com/fission/fission/blob/7aa67a9676ef979aef17018fb7e0c525c8ca5a21/test/tests/test_annotations.sh#L91)):*
> ```
> annotations
> ```
Need to add annotations fields to EnvironmentSpec

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/733)*

erwinvaneyk:
Review status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @erwinvaneyk and @life1347)

---

*[executor/poolmgr/gp.go, line 487 at r2](https://reviewable.io/reviews/fission/fission/733#-LEgjzIk8pRWJup6WgxM:-LEszVJA5i4nvxXdPpCs:b-896fix) ([raw file](https://github.com/fission/fission/blob/7aa67a9676ef979aef17018fb7e0c525c8ca5a21/executor/poolmgr/gp.go#L487)):*

> ```
> podAnnotations
> ```
Need to do the same thing to Newdeploy too.
</blockquote></details>

Done.

---

*[test/tests/test_annotations.sh, line 91 at r2](https://reviewable.io/reviews/fission/fission/733#-LEgje1m9kHnYrIFBH1_:-LEszYv8Eue_dE7TAHQU:bo33h4g) ([raw file](https://github.com/fission/fission/blob/7aa67a9676ef979aef17018fb7e0c525c8ca5a21/test/tests/test_annotations.sh#L91)):*

> ```
> annotations
> ```
Need to add annotations fields to EnvironmentSpec
</blockquote></details>

Actually, you caught the opposite problem. The test should define the annotations on the crd metadata.

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/733)*

life1347:
@erwinvaneyk looks like there's an error in test script
```
> kubectl -n f-func-4b496a get pod python-1529323861-ab343a7e-72f0-11e8-91db-42010a80010f-jp062fpm -o go-template={{range $key,$value := .metadata.annotations}}{{$key}}: {{$value}}{{"\n"}}{{end}}
error: error parsing template {{range, template: output:1: unexpected unclosed action in range
[FAILED]: /test_annotations.sh
```

erwinvaneyk:
@life1347 yeah, looking into that. Thanks! :)

erwinvaneyk:
@life1347 fixed it, please merge if you agree ;)

PR #734, opened 2018-06-08
envns should be availabe in message line

xiekeyang:
This line warn customer that dependent environment doesn’t exist, and present the command line
how to create env.

For variable `envNamespace` is specified, here the `envns` option should be available as well.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/734)


smruthi2187:
yes, makes sense. LGTM.

soamvasani:
Merged. Thanks @xiekeyang !

PR #735, opened 2018-06-08
instead hard code by variable in error message

xiekeyang:
`ArchiveLiteralSizeLimit` has constant value. When encountering the package error here, we should warn the exact `ArchiveLiteralSizeLimit` value, but not `256KB` hard code.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/735)

### DIFF
--- a/controller/packageApi.go
+++ b/controller/packageApi.go
@@ -18,9 +18,11 @@ package controller
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 
+	"github.com/dustin/go-humanize"
 	"github.com/gorilla/mux"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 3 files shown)


controller/packageApi.go
```diff
@@ -64,12 +66,14 @@ func (a *API) PackageApiCreate(w http.ResponseWriter, r *http.Request) {
 
 	// Ensure size limits
 	if len(f.Spec.Source.Literal) > int(fission.ArchiveLiteralSizeLimit) {
-		err := fission.MakeError(fission.ErrorInvalidArgument, "Package literal larger than 256K")
+		err := fission.MakeError(fission.ErrorInvalidArgument,
+			fmt.Sprintf("Package literal larger than 0x%s", strconv.FormatInt(fission.ArchiveLiteralSizeLimit, 16)))
```
soamvasani: The error is for humans, plain old decimal is probably better.  Also, please include units (bytes/kb/etc).

xiekeyang:
@soamvasani I introduced the package `github.com/dustin/go-humanize` and formatted the message value to human readable size. Re-Committed, PTAL, thanks a lot!

soamvasani:
LGTM.

Issue #723, opened 2018-06-05
NewDeployment - Package update issue in 0.8.0


vishal-biyani:
After 0.8.0 release there is an issue with package update for new deployment executor functions which was seen consistently after multiple retries. (Log of failure: https://api.travis-ci.org/v3/job/388178258/log.txt) 

The same test has worked in immediately before builds (https://api.travis-ci.org/v3/job/387779126/log.txt)

This issue is to investigate & track the same

Currently following tests are failing consistently:

```
--- PASSED /test_fn_update/test_resource_change.sh
*** FAILED /test_fn_update/test_nd_pkg_update.sh
*** FAILED /test_fn_update/test_configmap_update.sh
*** FAILED /test_fn_update/test_secret_update.sh
--- PASSED /test_fn_update/test_poolmgr_nd.sh
--- PASSED /test_logging/test_function_logs.sh
```

smruthi2187:
I think this is why tests are failing on our CI cluster. Referring to this log : https://api.travis-ci.org/v3/job/389024752/log.txt for the explanation.
1. First, a function, (in this log): hellopython-360269385 is created with executor type NDM. The resource version  ends up being : 41632529 . Also a service, hellopython-360269385-usbzmidi gets created with this selector :  
```"selector": {
                    "environmentName": "python-358862774",
                    "environmentUid": "5494865e-69e6-11e8-91db-42010a80010f",
                    "executorInstanceId": "usbzmidi",
                    "executorType": "newdeploy",
                    "functionName": "hellopython-360269385",
                    "functionUid": "54a90c0e-69e6-11e8-91db-42010a80010f"
                },```
2. Now, when we invoke the function at route : /hellopython-360269385, everything works as expected. 
3. Next, Function gets updated to have new code ( from hello world to hello fission ). The resource version ends up being 41632617.
4.When we invoke the function at route /hellopython-360269385, router does recognize the different resource version and requests executor for a new service. New pod is spun up with the latest code. the old pod is still terminating. But the service url returned to router remains the same : hellopython-360269385-usbzmidi(as expected). 
5. Now, router makes a http request to this service url hellopython-360269385-usbzmidi. But since the old pod is still terminating, the service _is_ _still_ routing to old pod ip that is terminating. So the response received is “hello world” instead of “hello fission”.

I ran a few tests on CI, where I waited for old pod to get deleted fully and then invoked the function with the http request, everything worked correctly. I.e. I did get “Hello Fission”

**Two solutions i can think of**, 
[...]

PR #742, opened 2018-06-12
Testing with keep alive settings for connections

vishal-biyani:
This fix tries to address final finding from the issue https://github.com/fission/fission/issues/723 (The issue with router caching the POD IP address: https://github.com/fission/fission/issues/723#issuecomment-395952964 )

As can be seen from builds for commit - with the fix the tests pass and without it, they fail. Of course, the impact of this on the performance etc. is not clear here. Also, this issue has been seen only in CI cluster so far.

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/742)

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -94,7 +94,7 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 	// set the timeout for transport context
 	timeout := roundTripper.initialTimeout
 	transport := http.DefaultTransport.(*http.Transport)
-	transport.DisableKeepAlives = true
+	//transport.DisableKeepAlives = true
 
 	// cache lookup to get serviceUrl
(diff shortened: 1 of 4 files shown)


vishal-biyani:
Request @soamvasani @smruthi2187 to take a look and decide if we want to make this part of master.

soamvasani:
Given your update about benchmarks in the issue, I think we should merge this for now, and come back to optimize later.

router/functionHandler.go
```diff
@@ -94,6 +94,7 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 	// set the timeout for transport context
 	timeout := roundTripper.initialTimeout
 	transport := http.DefaultTransport.(*http.Transport)
+	transport.DisableKeepAlives = true
```
soamvasani: Please add a comment pointing at the issue, so we remember why we did this.

PR #743, opened 2018-06-12
Extensible Fission CLI

erwinvaneyk:
This PR introduces extensibility to the Fission CLI. The principle behind the design of this extensibility is taken from the approach git uses: the main binary (`fission`) calls other binaries (`fission-workflows`) - which it discovers using simple prefix-rules - based on specific commands (`fission workflow <command>`).

The plugin package contains four important structs:
- `plugin.Metadata` contains all the metadata of plugin (name, aliases, path, version...)
- `plugin.Cache` is a in-memory memorization of plugins (with optional file-based persistence) 
- `plugin.Registry` is a simple representation (currently just a string) of a (remote) registry. The manager uses these to search for plugins to suggest to the user to install.
- `plugin.Manager` ties everything together, providing the API user to find, search, list, and execute plugins.

Unless you want to modify specific behavior of the plugin functionality, you simply can use the public package-level functions that are mapped to the functions of a default plugin.Manager.

To summarize; the main changes of this PR:
- Introduces the plugin functionality as a standalone package (`fission/plugin`); allowing other projects (like workflows) to import and use the plugin metadata struct and other functionality.
- `fission --help` includes the commands available through plugins.
- `fission --version` uses a structured representation and includes version info of all the plugins as well.
- `fission plugins list` shows the discovered/installed plugins
- `fission plugins cache-clear` clears the caches used for listing plugins, which should not happen (but might be useful in case there is something weird happening).
- The logrus logging level is coupled to the fission verbosity flag.
- Fission now creates an (optional) folder in the user home called `.fission` to store optional configuration and caches. For the plugins, `.fission/plugin-cache.json` is created to maintain a cache of the current, detected plugins.
- If a command is unknown to fission it will first check registries (by default `https://github.com/fission/fission/blob/master/plugins.yaml`) to see if it is present there. If so fission will provide instructions how to download and install the cli plugin. If not, the default 'command not found' is displayed.


Documentation/wip/cli-extensibility.md
```diff
@@ -0,0 +1,128 @@
...
+  fission-workflows: 0.3.0
+```
+Again, versioning info for fission-workflows is taken from the plugin info of the commands. 
+Note: a related issue is to have some more formalized plugin support/discoverability on the server-side, 
+but that is out of the scope of this issue.
+
+### Other (optional) extensions and notes
+- Like git we could setup a preferred binary path, where fission looks first when searching for the subcommand. 
+This could optionally be defined with a `FISSION_EXEC_PATH`.
+- With the current approach we cannot have aliases for commands---fission will not be able to find fission-workflows
```
soamvasani: the --plugin output can contain aliases

Documentation/wip/cli-extensibility.md
```diff
@@ -0,0 +1,128 @@
...
+- Like git we could setup a preferred binary path, where fission looks first when searching for the subcommand. 
+This could optionally be defined with a `FISSION_EXEC_PATH`.
+- With the current approach we cannot have aliases for commands---fission will not be able to find fission-workflows 
+when the user calls `fission wf`. This might be UX issue, with these long path names. One option is let the user fix 
+it themselves by symlinking `fission-wf` to `fission-workflows`; using the plugin info Fission can recognize and 
+merge aliases together.
+- To help detect versioning conflicts (old version of fission, too new version of fission workflows). We could add 
+a `requires` field to the fission-workflows plugin info. Then we could throw a warning or error, when two out of sync 
+versions are being used. 
+- To avoid unhelpful errors to the user when they have not installed a plugin, we could add a heuristic to check
```
soamvasani: I think we should compile-in a list of known plugins and URLs for where to download them, and let the user handle the rest for now.  For unknown plugins you can output the generic text.  Later we can get fancier and fetch the plugin list from a server, download and install plugins, etc.

Documentation/wip/cli-extensibility.md
```diff
@@ -0,0 +1,128 @@
...
+GLOBAL OPTIONS:
+   --server value  Fission server URL (default: "http://127.0.0.1:65356")
+   --help, -h      show help
+   --version, -v   print the version
+```
+Of course Fission needs to be able to find all plugins for this. There are several ways in which we can provide discoverability. The simplest one is for Fission to look in the path for all binaries starting with the `fission-*` prefix. Optionally, fission could invoke a specific command on the subcommand to get info about the plugin (such as version, help text, aliases...)
+
+With Fission Workflows this info would look something like this: 
+```bash
+$ fission-workflows --plugin
```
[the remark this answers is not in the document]
erwinvaneyk: The current implementation outputs and parses the plugin metadata as YAML. I can change it to JSON or leave it as it is.

Documentation/wip/cli-extensibility.md
erwinvaneyk: Like the previous comment, this part is a bit outdated. The current flow:
1. Check if command is present locally; if so, execute it as a subcommand.
2. Use a hard-coded url (currently `https://github.com/fission/fission/blob/master/plugins.yaml`) to fetch plugin suggestions. If the requested command is in there, print a suggestion how to download and install it.
3. Otherwise, print the current "command %v not found" warning.

Documentation/wip/cli-extensibility.md
erwinvaneyk: Yes, this document was just a direct copy-paste of the original proposal in issue #743. This part is outdated, plugin metadata can indeed contain aliases. 

I'll update this document before merging.

soamvasani:
Reviewed 3 of 7 files at r1.
Review status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @soamvasani and @erwinvaneyk)

---

*[Documentation/wip/cli-extensibility.md, line 101 at r1](https://reviewable.io/reviews/fission/fission/743#-LEqBJ_fCHtDvXkhJ2AI-r1-101:-LGCbKfa8dcjpAuod5DZ:bw7lqjz) ([raw file](https://github.com/fission/fission/blob/4eb5d9d85ed5036c6baff81397c09882393e6206/Documentation/wip/cli-extensibility.md#L101)):*

Like the previous comment, this part is a bit outdated. The current flow:
1. Check if command is present locally; if so, execute it as a subcommand.
2. Use a hard-coded url (currently `https://github.com/fission/fission/blob/master/plugins.yaml`) to fetch plugin suggestions. If the requested command is in there, print a suggestion how to download and install it.
3. Otherwise, print the current "command %v not found" warning.
</blockquote></details>

Don't reach into the master branch directly.  In fact maybe we should just include the plugin registry object into the cli build itself; no YAML, just a literal object.

---

*[fission/main.go, line 81 at r2](https://reviewable.io/reviews/fission/fission/743#-LGCbjhL1uqB1LarCoS4:-LGCbjhMBnNh9HEhUtTN:b-9m20xm) ([raw file](https://github.com/fission/fission/blob/dee00bbfd705f2362a76a03b6fada6d32fa253df/fission/main.go#L81)):*
> ```Go
> 	fissionConfigPath := os.Getenv("FISSION_CONFIG_PATH")
> 	if len(fissionConfigPath) == 0 {
> 		fissionConfigPath = os.ExpandEnv("$HOME/.fission")
> ```

use filepath.Join for paths (no hardcoding '/' (even though it works pretty much everywhere))

---

*[fission/main.go, line 91 at r2](https://reviewable.io/reviews/fission/fission/743#-LGCfCl3AfB6rE_mc6w1:-LGCfCl4D0eZjmVrNGBR:baxbfhz) ([raw file](https://github.com/fission/fission/blob/dee00bbfd705f2362a76a03b6fada6d32fa253df/fission/main.go#L91)):*
> ```Go
> 	plugin.SetPrefix("fission-")
> 	plugin.AddRegistry(plugin.Registry(path.Join(fissionConfigPath, "plugins-registry.yaml")))
> 	plugin.AddRegistry("https://github.com/fission/fission/blob/master/plugins.yaml")
> ```

Don't link to master branch; too unstable.  If you really must have a URL, point at a released YAML.  Even better if you can do this without an internet connection.

What happens when the URL 404s or we fail to connect?

---

*[fission/plugin/plugin.go, line 1 at r2](https://reviewable.io/reviews/fission/fission/743#-LGChDzAA75EnDBdRceB:-LGChDzB21oeMu3Wo2DS:b-dk2qna) ([raw file](https://github.com/fission/fission/blob/dee00bbfd705f2362a76a03b6fada6d32fa253df/fission/plugin/plugin.go#L1)):*
> ```Go
> // Package plugins provides support for creating extensible CLIs
> ```

I feel like this package could be much simpler.  All we want is:

When user does `fission foo <args>`, we invoke `fission-foo <args>`.  

Do we really need a registry, cache, manager, default manager, refresh function, cache expiration logic, ...

What if we just have one structure: ```
type plugins struct {
       plugins map[string]string  // "workflow" -> fission-workflow, "wf"->fission-workflow, etc.
}
```

This in-memory map is loaded from (a) ~/.fission/plugins.yaml if it exists (b) gathered by searching for all `fission-` commands.

If user invokes a plugin not found in the YAML, we re-do (b) above.

You can hard code the logic for dealing with known plugins in a built-in data structure, something like a "known plugins map".
```


---

*[fission/plugin/plugin.go, line 29 at r2](https://reviewable.io/reviews/fission/fission/743#-LGCgWTm74_6Yys_EFqr:-LGCgWTm74_6Yys_EFqs:b-9f45m) ([raw file](https://github.com/fission/fission/blob/dee00bbfd705f2362a76a03b6fada6d32fa253df/fission/plugin/plugin.go#L29)):*
> ```Go
> 
> 	"github.com/sirupsen/logrus"
> 	"gopkg.in/yaml.v2"
> ```

We use ghodss/yaml in other places, I think.  I'd like to avoid having two YAML libraries

---


*Comments from [Reviewable](https://reviewable.io/reviews/fission/fission/743)*

soamvasani:
Also, the CLI shouldn't be logging.  All its messages should be designed for human users.  Use `verbose()` for things you don't want to output unless there's a --verbose flag (same comments apply to workflows CLI as well).

erwinvaneyk:
Discussed offline: remove for now:
- caching 
- registry

erwinvaneyk:
@soamvasani addressed your comments, please review

life1347:
LGTM, except some minor comments above^.

life1347:
@erwinvaneyk what's the status of this PR? It looks already in good shape except for one left comment.

life1347:
LGTM

PR #751, opened 2018-06-15
Fix executor wrongly passes loop variable reference to function

life1347:
### Root cause

When investigating issues 695, found the bug in router. The reference of loop variable was being passed to function. (Not the root cause of 695)

https://github.com/fission/fission/blob/1ab1c3264ec8b68f4da225bef51db4ffe17c5659/router/httpTriggers.go#L112
https://github.com/fission/fission/blob/1ab1c3264ec8b68f4da225bef51db4ffe17c5659/router/httpTriggers.go#L127

And wrong metrics data will be collected.

https://github.com/fission/fission/blob/1b86c367a04aa07f0bedea96786201fe1948b164/router/functionHandler.go#L89-L92

### Solution

Get reference from slice element.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/751)

### DIFF
--- a/router/httpTriggers.go
+++ b/router/httpTriggers.go
@@ -101,7 +101,8 @@ func (ts *HTTPTriggerSet) getRouter() *mux.Router {
 
 	// HTTP triggers setup by the user
 	homeHandled := false
-	for _, trigger := range ts.triggers {
+	for i := range ts.triggers {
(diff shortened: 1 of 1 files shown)


soamvasani:
🤦🏽 LGTM

PR #764, opened 2018-07-02
Fix spec command overrides existing archive's url of a package

life1347:
### Description

The command `fission spec` overrides existing archive's URL of a package.

### Root cause

The variable `a` is the copy of `archiveFiles[name]`, so the value assignment is not write to map.
```
a := archiveFiles[name]
a.URL = url
```

### Solution

Assign the original map with the copy of archive instance.


soamvasani:
Argh, don't know what I was thinking.  LGTM and thanks.

Issue #771, opened 2018-07-03
Newdeploy prints "fsvc not found in cache" when idle function was deleted


life1347:
The root cause is the fsvc was deleted at idle object cleanup process.

https://github.com/fission/fission/blob/032d1a8b9a49045033bebd18ece83a569cdb3fb4/executor/cleanup.go#L142-L150

Thus, newdeploy cannot get fsvc to delete fn related objects.

https://github.com/fission/fission/blob/032d1a8b9a49045033bebd18ece83a569cdb3fb4/executor/newdeploy/newdeploymgr.go#L501-L507

Issue #757, opened 2018-06-26
Don't delete deployments when idle, scale them to zero instead


soamvasani:
When a function goes idle today, executor deletes the deployment, replicaset, service, etc.  Instead we should just scale the deployment down to zero, only deleting the deployment when the function itself is deleted.

PR #775, opened 2018-07-03
Scale deployment to zero when function is in idle state

life1347:
This PR aims to solve following issues:
* #757 (Don't delete deployments when idle, scale them to zero instead) 
* #771 (Newdeploy prints "fsvc not found in cache" when idle function was deleted)

Now, each exector type has its own `idleObjectReaper` function instead of global one. It reduces the complexity of idleObjectReaper and makes code clean.

Newdeploy scales down the deployment when idle, and scale up to 1 replica when exectuor tries to validate the function address.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/775)


executor/cleanup/cleanup.go
```diff
@@ -23,23 +23,20 @@ import (
...
 
 var (
 	deletePropagation = meta_v1.DeletePropagationBackground
 	delOpt            = meta_v1.DeleteOptions{PropagationPolicy: &deletePropagation}
 )
 
-// cleanupObjects cleans up resources created by old executortype instances
-func cleanupObjects(kubernetesClient *kubernetes.Clientset,
+// CleanupObjects cleans up resources created by old executortype instances
+func CleanupObjects(kubernetesClient *kubernetes.Clientset,
```
erwinvaneyk: You might want to make this more canonical go by taking the package name into account when naming your functions: `cleanup.Objects` instead of `cleanup.CleanupObjects`

executor/executor.go
```diff
@@ -233,17 +234,18 @@ func StartExecutor(fissionNamespace string, functionNamespace string, envBuilder
 	fsCache := fscache.MakeFunctionServiceCache()
 
 	poolID := strings.ToLower(uniuri.NewLen(8))
-	cleanupObjects(kubernetesClient, functionNamespace, poolID)
-	go idleObjectReaper(kubernetesClient, fissionClient, fsCache, time.Minute*2)
-	go cleanupRoleBindings(kubernetesClient, fissionClient, functionNamespace, envBuilderNamespace, time.Minute*30)
+	cleanup.CleanupObjects(kubernetesClient, functionNamespace, poolID)
+	go cleanup.CleanupRoleBindings(kubernetesClient, fissionClient, functionNamespace, envBuilderNamespace, time.Minute*30)
+
+	idlePodReapTime := 2 * time.Minute
```
erwinvaneyk: Move this to package level constant

executor/poolmgr/gpm.go
```diff
@@ -241,3 +248,59 @@ func (gpm *GenericPoolManager) IsValidPod(kubeObjects []apiv1.ObjectReference, p
...
 	return false
 }
+
+// idleObjectReaper reaps objects after certain idle time
+func (gpm *GenericPoolManager) idleObjectReaper(kubeClient *kubernetes.Clientset,
+	fissionClient *crd.FissionClient,
+	fsCache *fscache.FunctionServiceCache,
+	idlePodReapTime time.Duration) {
+
+	pollSleep := time.Duration(2 * time.Minute)
```
erwinvaneyk: Move to package level constant

executor/poolmgr/gpm.go
```diff
@@ -241,3 +248,59 @@ func (gpm *GenericPoolManager) IsValidPod(kubeObjects []apiv1.ObjectReference, p
...
+	for {
```
erwinvaneyk: It looks like this loop cannot be stopped, maybe add a close channel or something like that, to allow this to be stopped gracefully.

executor/poolmgr/gpm.go
```diff
@@ -241,3 +248,59 @@ func (gpm *GenericPoolManager) IsValidPod(kubeObjects []apiv1.ObjectReference, p
...
+		time.Sleep(pollSleep)
+
+		envs, err := fissionClient.Environments(metav1.NamespaceAll).List(metav1.ListOptions{})
+		if err != nil {
+			log.Fatalf("Failed to get environment list: %v", err)
```
erwinvaneyk: Missing a `continue`?

executor/poolmgr/gpm.go
```diff
@@ -241,3 +248,59 @@ func (gpm *GenericPoolManager) IsValidPod(kubeObjects []apiv1.ObjectReference, p
...
+			continue
+		}
+
+		for _, fsvc := range funcSvcs {
+			if fsvc.Executor != fscache.POOLMGR {
+				continue
+			}
+
+			if _, ok := envList[fsvc.Environment.Metadata.UID]; !ok {
+				log.Printf("Environment %v for function %v no longer exists",
```
erwinvaneyk: missing a `continue`?

executor/cleanup/cleanup.go
```diff
@@ -86,81 +83,8 @@ func cleanup(client *kubernetes.Clientset, namespace string, instanceId string)
...
-			for _, kubeobj := range fsvc.KubernetesObjects {
-				deleteKubeobject(kubeClient, &kubeobj)
-			}
-		}
-	}
-}
-
-func deleteKubeobject(kubeClient *kubernetes.Clientset, kubeobj *apiv1.ObjectReference) {
+// DeleteKubeObject deletes given kubernetes object
+func DeleteKubeObject(kubeClient *kubernetes.Clientset, kubeobj *apiv1.ObjectReference) {
```
erwinvaneyk: Just wondering, why is this called `DeleteX` when all other functions in this pkg are `CleanupX`?

executor/cleanup/cleanup.go
life1347: Rename package from `cleanup` to `reaper` instead.

executor/poolmgr/gpm.go
life1347: It's intent to do so, so that the executor will clean up idle function and prints log to notify users that the environment for the function was no longer exists.

executor/poolmgr/gpm.go
life1347: I think its intent to do so, the executor will fatal(crash) directly when it's not able to get environments objects from the API server (maybe due to the connection error, or resources not defined).

executor/poolmgr/gpm.go
life1347: The loop will stop only when container exited just like other loops
https://github.com/fission/fission/blob/e977c7caca4167eb1c6d61792b6cd71a658675b2/executor/poolmgr/gpm.go#L112-L115
I think it should be enough.

executor/cleanup/cleanup.go
life1347: I didn't notice it before. Rename to `CleanupKubeObject`

executor/newdeploy/newdeploy.go
```diff
@@ -58,6 +58,15 @@ func (deploy *NewDeploy) createOrGetDeployment(fn *crd.Function, env *crd.Enviro
 
 	existingDepl, err := deploy.kubernetesClient.ExtensionsV1beta1().Deployments(deployNamespace).Get(deployName, metav1.GetOptions{})
 	if err == nil {
+		// If minScale is 0 then scale to 1, otherwise scale up to minScale
```
erwinvaneyk: I am not sure what the API works right now. But given that the kubernetes API is out of our control, it might be good  to be a bit more defensive here, and handle the case `existingDepl == nil && err == nil`. Currently that would result in a NPE

erwinvaneyk:
@life1347 can you add a test or two to verify the behaviour? Or, If it is not possible or feasible, could explain why that is the case here for future reference?

PR #779, opened 2018-07-04
Fix “rm: missing operand” in release script

life1347:
1. Fix “rm: missing operand” in release script
2. Remove duplicate build step

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/779)


smruthi2187:
LGTM!

PR #784, opened 2018-07-05
Fix go env plugin

life1347:
### Description
go env `0.9.0` failed to load plugin.

```
Listening on 8888 ...
Specializing ...
loading plugin from /userfunc/00ddd189-8044-11e8-b734-080027773752/hello-go-eoeo-hvifwp-sjpfzx
2018/07/05 11:10:28 http: panic serving 172.17.0.14:42940: plugin: not implemented
```

### Root cause
Run go binary in Alpine image required `CGO_ENABLED=0` for static link, however, the implementation of go plugin requires the dynamic link `CGO_ENABLED=1`(https://github.com/golang/go/issues/19569).

### Solution
The change of go-env in 0.9.0 is to reduce the size of env image, so I change the image of go-env from alpine to ubuntu which allow to run go binary that built with `CGO_ENABLED=1` while keeping image small.

The size of go-env grows to ~100MB, but still smaller than previous. (700MB)

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/784)

### DIFF
--- a/test/tests/test_environments/test_go_env.sh
+++ b/test/tests/test_environments/test_go_env.sh
@@ -40,7 +40,7 @@ export -f test_fn
 cd $ROOT/examples/go
 
 log "Creating environment for Golang"
-fission env create --name go --image gcr.io/fission-ci/go-env:test --builder gcr.io/fission-ci/go-builder:test --period 5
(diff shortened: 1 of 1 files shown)


smruthi2187:
LGTM!

PR #787, opened 2018-07-05
Change flag name to KeepArchive for backward compatibility

life1347:
### Description

A new flag `ExtractArchive` was adopted In 0.9.0, which is used by fetcher to see whether a archive needs to extract or not. However, if an environment was created with old CLI the value will be false due to golang default value for the empty field and make the breaking change.

### Solution

Rename the flag for backward compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/787)

### DIFF
--- a/buildermgr/common.go
+++ b/buildermgr/common.go
@@ -54,10 +54,10 @@ func buildPackage(fissionClient *crd.FissionClient, envBuilderNamespace string,
 	builderC := builderClient.MakeClient(fmt.Sprintf("http://%v:8001", svcName))
 
 	fetchReq := &fetcher.FetchRequest{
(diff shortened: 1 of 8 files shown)


smruthi2187:
Fix looks great, i think will solve the issue. i have just one comment though, i think we can call the flag `KeepArchive` instead of `NoExtractArchive`. what do you think?

life1347:
`KeepArchive` make sense to me, I'll change it!

soamvasani:
Thanks for fixing this quickly, LGTM.

vishal-biyani:
LGTM, thanks for taking this up and quick fix.

PR #790, opened 2018-07-11
Break & Stop the build immediately if a non-zero exit code was returned

life1347:
### Description
Sometimes the build is failed due to a non-zero exit code was returned in the early stage, however, the build keeps running and still failed in the end. This behavior costs lots of time to wait for the build result.

### Solution

Follow the instruction here: https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build
`If before_install, install or before_script returns a non-zero exit code, the build is errored and stops immediately.`
Move some of script from `scripts` to `before_script` stage so that minor problems can be found before long running build.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/790)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,13 +31,17 @@ install:
 env:
   - KUBECONFIG=${HOME}/.kube/config PATH=$HOME/k8scli:${PATH}
 
-script:
+before_script:
   - cd ${TRAVIS_BUILD_DIR}
-  - glide install -v
-  - ./fission-bundle/build.sh
   - hack/verify-gofmt.sh
   - hack/verify-govet.sh
+  - glide install -v
+  - go build -o fission/fission fission/*.go
+  - ./fission-bundle/build.sh
   - hack/runtests.sh
+
+script:
+  - cd ${TRAVIS_BUILD_DIR}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,13 +31,17 @@ install:
...
-  - ./fission-bundle/build.sh
   - hack/verify-gofmt.sh
   - hack/verify-govet.sh
+  - glide install -v
+  - go build -o fission/fission fission/*.go
+  - ./fission-bundle/build.sh
   - hack/runtests.sh
+
+script:
+  - cd ${TRAVIS_BUILD_DIR}
```
erwinvaneyk: You probably tested this, but just to be sure: do you need to change the current directory again in the script stage? Doesn't this propagate from the `before_script`?

erwinvaneyk:
If the CI succeeds: LGTM 👍

.travis.yml
life1347: Not sure will the path propagate from the previous stage but looks like it works. 
I think another advantage is that it helps people to know the working directory to `${TRAVIS_BUILD_DIR}` in current stage without checking the configs in previous stages.

PR #791, opened 2018-07-11
Delete namespace in background to reduce build time

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/791)

### DIFF
--- a/test/test_utils.sh
+++ b/test/test_utils.sh
@@ -154,10 +154,7 @@ build_fission_cli() {
     popd
 }
 
-clean_tpr_crd_resources() {
-    # clean tpr & crd resources to avoid testing error (ex. no kind "HttptriggerList" is registered for version "fission.io/v1")
-    # thirdpartyresources part should be removed after kubernetes test cluster is upgrade to 1.8+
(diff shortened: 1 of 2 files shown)


erwinvaneyk:
I had a similar issue with Workflows --- do you ensure that subsequent tests are not affected by 'still deleting' namespaces?

life1347:
Yes, it won't affect following tests since the deleted namespace are only used in this test script.
And this PR is more like a workaround for now to reduce the job time lower than 50mins cause I noticed some of the build was failed due to exceeded time limitation. 
```
https://travis-ci.org/fission/fission/builds/402665319
The job exceeded the maximum time limit for jobs, and has been terminated.
```
We need to find out some ways to lower job time in future.

soamvasani:
Does this work?  Does the travis worker wait for background jobs to finish?

life1347:
Yes, it works. The build time is lower than previous about 8mins. (49 -> 41).
I think Travis worker won't wait for the background process, but the deletion will be finished before the end of the whole build process.

Issue #765, opened 2018-07-02
fluentd config as configmap


soamvasani:
Today we build the fluentd configuration file into the docker image.  We should use a configmap instead, so that users can easily override it (for example to send logs to a service like splunk).

erwinvaneyk:
This might be worth exploring: https://github.com/vmware/kube-fluentd-operator

vishal-biyani:
As an alternative to enabling a custom configmap - may be there should be a way to disable Fission FluentD and instead use existing FluentD/other forwarders. One side effect of that is the `fission fn logs` potentially not working unless you explicitly add influxDB forwarding. Also, some users want their own FluentD so that they can collate logs from other namespaces/applications and not just Fission function namespaces.

life1347:
Solved by #792

PR #792, opened 2018-07-13
Add fluentd.conf as a configmap

erwinvaneyk:
This PR adds fluentd.conf as a config map to the `fission-all` deployment. Functionally nothing has changed to the fluentd.conf in the chart vs the original one.

Furthermore, I propose to keep the old config in the dockerfile as a 'sensible default. 

This PR does not yet include changes needed to handle proper fluentd parsing of function-logs.

This PR solves #765 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/792)


charts/fission-all/templates/fluentd.yaml
```diff
@@ -0,0 +1,81 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-fission-fluentd
+data:
+{{ if .Files.Get "config/fluentd.conf" }}
+  td-agent.conf: |
+{{ .Files.Get "config/fluentd.conf" | indent 3 }}
+{{ else }}
+{{ fail "you need to supply a fluentd config file! add 'fluentd.conf' to your chart 'config' directory." }}
```
soamvasani: Just "invalid chart".  User isn't going to be able to fix this by themselves, if it happens it's simply a bug in our chart.

soamvasani:
@erwinvaneyk do you think we can bring this PR back to life for 0.12?

PR #793, opened 2018-07-15
change image pull policy of builder manager

xiekeyang:
This changes image pull policy from `Always` to `IfNotPresent`. The previous
policy forbid customer to use the exact image existing on K8S Node.

And it should be OK to use fixed policy here. Customer need not set policy by
their own.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/793)


soamvasani:
Sounds fine to me. @life1347 do you see anything wrong with doing this?  I can't remember if `PullAlways` was chosen for some specific reason.

life1347:
@xiekeyang Thanks for this PR! LGTM
@soamvasani  No, I think there was no specific reason at that time. I guess I wrote that only for my personal testing purpose.

PR #794, opened 2018-07-16
add gevent based Python server to benchmark test cases

xiekeyang:
This add the benchmark test cases, which is using gevent based on wsgi
framework.


test/benchmark/tests/concurrency-level/run.sh
```diff
@@ -4,90 +4,93 @@ set -euo pipefail
 
 ROOT=$(dirname $0)/../../../..
 
-for executorType in poolmgr newdeploy
+for env in python_bjoern python_gevent
```
life1347: Just a nitpick, would you mind to change `env` to `framework`?

test/benchmark/tests/concurrency-level/run.sh
```diff
@@ -4,90 +4,93 @@ set -euo pipefail
...
+	        do
+
+	            # Create a hello world function in nodejs, test it with an http trigger
+	            echo "Pre-test cleanup"
+	            fission env delete --name python || true
+
+	            echo "Creating python env"
+	            # Use short grace period time to speed up resource recycle time
+	            # Use high min/max CPU so that K8S will distribute pod in different nodes
+				fission spec apply --specdir "../../../../../test/envs/$env"
```
life1347: Maybe fix the indent here.

test/envs/python_bjoern/README
```diff
@@ -0,0 +1,42 @@
+
```
life1347: It's really a good idea to use spec here, since it's for benchmarking, how about putting these files to `$GOPATH/src/github.com/fission/fission/test/benchmark/assets`?

life1347:
@xiekeyang Thanks for PR and it's LGTM, just some comments above need to be addressed.

xiekeyang:
@life1347 All comments are address, and this is rebased to latest commit. PTAL, thanks so much!

test/benchmark/tests/concurrency-level/run.sh
```diff
@@ -4,90 +4,93 @@ set -euo pipefail
...
+	            trap "fission env delete --name python" EXIT
+
+	            sleep 30
+
+	            fn=python-hello-$(date +%s)
+
+	            echo "Creating package"
+	            rm -rf pkg.zip pkg/ || true
+	            mkdir pkg
+	            cp ../../../assets/hello.py pkg/hello.py
```
life1347:
```
cp $ROOT/benchmark/assets/hello.py pkg/hello.py
```

test/benchmark/tests/concurrency-level/run.sh
```diff
@@ -4,90 +4,93 @@ set -euo pipefail
...
+	        do
+
+	            # Create a hello world function in nodejs, test it with an http trigger
+	            echo "Pre-test cleanup"
+	            fission env delete --name python || true
+
+	            echo "Creating python env"
+	            # Use short grace period time to speed up resource recycle time
+	            # Use high min/max CPU so that K8S will distribute pod in different nodes
+	            fission spec apply --specdir "../../assets/envs/$framework"
```
life1347: I ran the manual test and get following error 
```
Creating python env
Spec directory ../../assets/envs/python_bjoern doesn't exist. Please check directory path or run "fission spec init" to create it.
```

The path should be 
```
../../../assets/envs/$framework
```

But as you know, the relative path is not ideal for code tracing, let's change it to
```
fission spec apply --specdir "$ROOT/benchmark/assets/envs/$framework"
```
which should be easier for people to understand

xiekeyang: Agree, I also feel the absolute path is better. I had kept the same as previous writing conservatively.  So I would fix it sooner.

xiekeyang: Relative path is so unreliable when you run the script under different dir.

xiekeyang:
@life1347 The problem of error path is fixed. Actually the path by `dirname $0` is relative. So previous `$ROOT` became relative also.

Next line call `pushd` to enter another path (https://github.com/fission/fission/blob/master/test/benchmark/tests/concurrency-level/run.sh#L18), that makes it difficult to find correct path by `$ROOT` or by `../../../`.

So I introduce `readlink -f` to instead `dirname $0`, that return absolute path anyway and store in `$ROOT`.

The bench test pass on my local side. PTAL, thanks a lot!

PR #808, opened 2018-07-19
Add retry subcommand to pkg command

life1347:
Currently, CLI doesn't have a convenient way to retry on a failed package. You have to update one of flag to trigger build process. That's really not an ideal way for retrying failed package.

This PR add `retry` subcommand to `pkg` command, it helps user to do retry without updating anything. 

Turn this

```
$ fission pkg update --name restapi-go-pkg --env go
```

into this

```
$ fission pkg retry --name restapi-go-pkg
Set package 'restapi-go-pkg' to pending state
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/808)

### DIFF
--- a/fission/main.go
+++ b/fission/main.go
@@ -241,6 +241,7 @@ func main() {
 	pkgSubCommands := []cli.Command{
 		{Name: "create", Usage: "Create new package", Flags: []cli.Flag{pkgNamespaceFlag, pkgEnvironmentFlag, envNamespaceFlag, pkgSrcArchiveFlag, pkgDeployArchiveFlag, pkgBuildCmdFlag}, Action: pkgCreate},
 		{Name: "update", Usage: "Update package", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag, pkgEnvironmentFlag, envNamespaceFlag, pkgSrcArchiveFlag, pkgDeployArchiveFlag, pkgBuildCmdFlag, pkgForceFlag}, Action: pkgUpdate},
+		{Name: "retry", Usage: "Try to rebuild failed package", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag}, Action: pkgRetry},
 		{Name: "getsrc", Usage: "Get source archive content", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag, pkgOutputFlag}, Action: pkgSourceGet},
 		{Name: "getdeploy", Usage: "Get deployment archive content", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag, pkgOutputFlag}, Action: pkgDeployGet},
 		{Name: "info", Usage: "Show package information", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag}, Action: pkgInfo},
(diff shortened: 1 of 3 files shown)


erwinvaneyk:
Should this really be a concern to the user (with yet another subcommand that they need to understand)? 

If you want to offer the user the option to force a retry, maybe just let `fission pkg update --name restapi-go-pkg --env go` force a retry (even if nothing has changed)

life1347:
I think the retry command is more intuitive than updating the package. And since its a simple command with no flags required, I think it should not be hard to remember. Also, the "update env" way will be no more available after https://github.com/fission/fission/pull/702 is merged.

fission/package.go
```diff
@@ -387,3 +387,30 @@ func pkgDelete(c *cli.Context) error {
...
+	checkErr(err, "find package")
+
+	if pkg.Status.BuildStatus != fission.BuildStatusFailed {
+		log.Fatal(fmt.Sprintf("Package %v is not in %v state.",
+			pkg.Metadata.Name, fission.BuildStatusFailed))
+	}
+
+	updatePackage(client, pkg, "", "", "", "", "", true)
+
+	fmt.Printf("Set package '%v' to %v state\n", pkg.Metadata.Name, fission.BuildStatusPending)
```
soamvasani: How about simply "Retrying build for pkg __.  Use `fission pkg info --name ___` to view status."

fission/package.go
```diff
@@ -387,3 +387,30 @@ func pkgDelete(c *cli.Context) error {
 
 	return nil
 }
+
+func pkgRetry(c *cli.Context) error {
```
soamvasani: Might be nice to have a --wait option.  (But no need to block this pr for it, it could be done later)

fission/main.go
```diff
@@ -240,6 +240,7 @@ func main() {
 	pkgSubCommands := []cli.Command{
 		{Name: "create", Usage: "Create new package", Flags: []cli.Flag{pkgNamespaceFlag, pkgEnvironmentFlag, envNamespaceFlag, pkgSrcArchiveFlag, pkgDeployArchiveFlag, pkgBuildCmdFlag}, Action: pkgCreate},
 		{Name: "update", Usage: "Update package", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag, pkgEnvironmentFlag, envNamespaceFlag, pkgSrcArchiveFlag, pkgDeployArchiveFlag, pkgBuildCmdFlag, pkgForceFlag}, Action: pkgUpdate},
+		{Name: "retry", Usage: "Try to rebuild failed package", Flags: []cli.Flag{pkgNameFlag, pkgNamespaceFlag}, Action: pkgRetry},
```
soamvasani: Can you change `retry` to `rebuild`?  It's not obvious that "retry" is referring to the build.  You could call it "retry-build" but "rebuild" is shorter.

soamvasani:
LGTM assuming above comments are addressed.

fission/package.go
life1347: Just opened an issue https://github.com/fission/fission/issues/848 to track this. I will also address https://github.com/fission/fission/issues/816 in same PR as well.

soamvasani:
LGTM, thanks

Issue #810, opened 2018-07-19
Router panic when trying to update route


germancho14:
On fission 0.9.1 I've noticed the following error in the Router Pod:
```
.
.
.
2018/07/19 22:43:55 Skipping creation of ingress for trigger: d8f1d53f-9831-4faf-8641-6ce7241df698
2018/07/19 22:43:57 Calling getServiceForFunction for function: fn-96a377a1-8dc9-4f2b-96a7-c370fc75d478
2018/07/19 22:43:58 Skipping creation of ingress for trigger: 2276415b-33c7-44e4-acc0-273d46b0878c
E0719 22:43:58.499186       1 runtime.go:66] Observed a panic: "index out of range" (runtime error: index out of range)
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/local/go/src/runtime/asm_amd64.s:573
/usr/local/go/src/runtime/panic.go:502
/usr/local/go/src/runtime/panic.go:28
/go/src/github.com/fission/fission/router/httpTriggers.go:105
/go/src/github.com/fission/fission/router/httpTriggers.go:251
/go/src/github.com/fission/fission/router/httpTriggers.go:221
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/controller.go:202
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/controller.go:309
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/delta_fifo.go:444
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/delta_fifo.go:444
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/controller.go:150
/go/src/github.com/fission/fission/vendor/k8s.io/client-go/tools/cache/controller.go:124
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
/go/src/github.com/fission/fission/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
[...]

PR #811, opened 2018-07-20
Fix router panic when trying to update route

life1347:
### Description
This PR aims to solve the issue #810 . The root cause is the size of `ts.triggers` changed when multiple go-routine called `ts.syncTriggers()`. 

https://github.com/fission/fission/blob/d11ab717076ef8c9911d887383ad41809ef1efb6/router/httpTriggers.go#L109-L111

### Solution
There are two ways to solve the problem:

1. Pass triggers & functions slice to `ts.getRouter()`
This solution will encounter another problem when the router under heavy load. Since we cannot make sure when will a go-routine was brought up to execute. The slice may contain old slices of triggers & functions.

2. Use the unbuffered channel to ensure the update sequence
This solution ensures that there is at most one call to update `ts.triggers` and `ts.functions`, thus that we can avoid the index out of range problem.

Conclusion: Use solution 2.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/811)


germancho14:
I wonder why ts.triggers and ts.functions are needed when we could just get the triggers and functions from the ts.triggerStore and ts.functionStore directly on the getRouter function. It seems that the stores are thread-safe, so the extra channel wouldn't be needed.

life1347:
Yeah, I also noticed that `ts.triggers` & `ts.functions` are not necessary at first glance. But after digging into the source code, notice that its necessary for unit test to do dependencies injection.
https://github.com/fission/fission/blob/ec495fbd1d5b10c4f04bab63a61f62d07c34c3e5/router/router_test.go#L70
If getRouter rely on triggerStore, then it will not be able to set routes during the unit test.

germancho14:
I also so that. Perhaps the store can be mocked for the purpose of unit testing. 
I changed the code to use the store directly and it works. I am creating lots of different functions and routes in a short time span. The change seems to be working at first glance.

life1347:
I personally think the extra benefit of a channel that it helps to reduce the active `update route` go-routine since most of the go-routine are waiting to execute. Since `latestTriggers := ts.triggerStore.List()` always get the latest objects from the store, we don't really lots of `update` go-routine in active state simultaneously.

For the test part, maybe we can mock the store but it also makes the test and changes more complicated.

germancho14:
Wouldn't the amount of work be the same if you are pushing on a channel in ResourceEventHandlerFuncs or just update the router directly using the stores?  I think the important thing here is that when modifying the router routers it should be thread-safe. If you are using ts.triggers and ts.functions you can use a channel (your solution), or use a thread-safe store (what I suggest).  

Perhaps your approach would be better since there would be lots of contention happening on the store using the approach I suggest. With your approach there wouldn't be any contention (process one thing at a time).

smruthi2187:
LGTM!

Issue #759, opened 2018-06-28
Controllers should create the necessary rolebindings on re-sync if they're missing


smruthi2187:
Currently, the role-bindings are created on `add` and `update` events on fission objects ( packages, functions ). But we should ideally do the right thing in the re-sync period by constantly checking if any of the rolebindings are missing and create them if they are.

xiekeyang:
# Reproduce

## Deploy Fission:

Use `fission-all` dir of master branch to deploy:

```
helm --debug install --name fission-all --namespace fission --set "imageTag=latest,fetcherImageTag=latest,serviceType=NodePort,persistence.enabled=false" ./charts/fission-all
```

Fission bundle and fetcher images come from compiling the latest master branch open code.


## Create Python V2 env:

The env spec is below:

```yaml
apiVersion: fission.io/v1
kind: Environment 
metadata:
  name: python
  namespace: fission-function  
spec:
  version: 2
  poolsize: 3                  
  TerminationGracePeriod: 3    
  resources:
    limits:
      cpu: 80m
      memory: 128Mi            
    requests:                  
      cpu: 50m                 
      memory: 64Mi             
  runtime:
    image: fission/python-env:3.6.5 
    imagePullPolicy: IfNotPresent   
    container:
      name: python
```
The base image `fission/python-env:3.6.5` comes from compiling the open source code: https://github.com/fission/fission/blob/master/environments/python/Dockerfile

Exec command to create the env:

```
$ kubectl apply -f ./python.yaml
```

## Create the function:

The function content is:

```python
def main():
    return "hello"
```

The command to create the function:

```
$ fission fn create --name hello --fns fission-function --env python --envns fission-function --code hello.py --entrypoint hello.main --method GET --url hello
```

Notice this use version 2 of Python env. Only ver 2 encounter this error because it will use fetcher to resolve the package.

## Send Request Successfully

At this moment the request to hello trigger URL is OK:

```
$ curl -v http://$FISSION_ROUTER/hello
*   Trying 192.168.99.100...
* Connected to 192.168.99.100 (192.168.99.100) port 31187 (#0)
> GET /hello HTTP/1.1
> Host: 192.168.99.100:31187
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 5
< Content-Type: text/html; charset=utf-8
< Date: Tue, 31 Jul 2018 02:31:28 GMT
< 
* Connection #0 to host 192.168.99.100 left intact
[...]

PR #820, opened 2018-07-26
Chart Fix: fission-fetcher have no access to packages

xiekeyang:
The service account fission-fetcher should be allocated by roll binding
with admin privilege under fission function namespace.

Otherwise, after long time idle, the fetcher link seems to be removed
from cache, and the trigger invocation which need access package will
be failed, with the error:

```
sevice account fission-function:fission-fetcher" cannot get packages.fission.io in the namespace "fission-function"
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/820)

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -90,6 +90,35 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: fission-fetcher
+  namespace: {{ .Release.functionNamespace }}
+subjects:
+  - kind: ServiceAccount
+    name: fission-fetcher
+    namespace: {{ .Values.functionNamespace }}
+roleRef:
+  kind: ClusterRole
+  name: admin
+  apiGroup: rbac.authorization.k8s.io
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: fission-fetcher
+subjects:
+- kind: ServiceAccount
+  name: fission-fetcher
+  namespace: {{ .Values.functionNamespace }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
+
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:


life1347:
@smruthi2187 Could you take a look on this PR? 
From the source code below, fission service should set role for fetcher to fetch package.

https://github.com/fission/fission/blob/f3857abdf55c4fa06b3da73be8fd2c13fcb0cf6d/executor/poolmgr/packagewatcher.go#L50-L56

Does it a bug or we just missed the SA setup in charts?

smruthi2187:
hi @life1347 and @xiekeyang we dont need cluster admin rolebinding for fetcher service account anymore in the charts because the package watcher assigns package-getter rolebinding as and when a package gets created.

@xiekeyang, can you tell me how and when you encountered this error message so I can try to repro the issue.

xiekeyang:
## Reproduce

### Install Fission Services

Use the master branch code.

### Create the Python environment

I used `$ kubectl apply -f ` to create the environment, the environment version should select 2, then the request will invoke fetcher container.

The env spec is as (the base image is compiled by latest code):

```
apiVersion: fission.io/v1
kind: Environment 
metadata:
  name: python
  namespace: fission-function  
spec:
  version: 2
  poolsize: 3                  
  TerminationGracePeriod: 3    
  resources:
    limits:
      cpu: 80m
      memory: 128Mi            
    requests:                  
      cpu: 50m                 
      memory: 64Mi             
  runtime:
    image: fission/python-env:3.6.5 
    imagePullPolicy: IfNotPresent   
    container:
      name: python
      env:
      - name: WSGI_FRAMEWORK   
        value: "GEVENT" 
```

### Create the hello function

Create a function just print hell, with a HTTP trigger for function.

```
$ fission fn create --name hello --fns fission-function --env python --envns fission-function --code hello.py --entrypoint hello.main --method GET --url hello
```

Then the function trigger can be requested successfully.

### Restart Fission Server

**Restart fission servers (I did it by restarting minikube)**, and then request fall in failed, and can’t be retrieved.

**Failed Request Command**

```
$ curl -v http://$FISSION_ROUTER/hello
*   Trying 192.168.99.100...
* Connected to 192.168.99.100 (192.168.99.100) port 30877 (#0)
> GET /hello HTTP/1.1
> Host: 192.168.99.100:30877
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 502 Bad Gateway
< Date: Mon, 30 Jul 2018 03:07:50 GMT
< Content-Length: 0
< 
* Connection #0 to host 192.168.99.100 left intact
```

**The error log from router service :**

```
2018/07/30 03:07:31 Calling getServiceForFunction for function: hello
2018/07/30 03:07:50 http: proxy error: Internal error - Failed to get package: packages.fission.io "hello-py-etii" is forbidden: User "system:serviceaccount:fission-function:fission-fetcher" cannot get packages.fission.io in the namespace "fission-function"
```

Then delete the old function and recreate new but same name function, the request will run successfully.

Actually, not only restart the whole fission server, but also edit specific (like executor), the request will fall in failure.

I used to remove and recreate env, and then request is failed as well, but this is not able to be always reproduced.

I’m not sure if this should be fixed by adding role binding in chart. If you prefer to some other correct way, please just close this PR freely, thanks.

xiekeyang:
ping @smruthi2187

smruthi2187:
yes, i have created an issue for this. https://github.com/fission/fission/issues/759
we dont need to add the rolebinding in chart. i'll fix it soon. thanks though!

xiekeyang:
Close this according to above discuss.

PR #825, opened 2018-07-27
Specs for JVM example

soamvasani:
This shows users that they can do `fission spec apply --wait` instead of manually dealing with zip files.

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/825)


vishal-biyani:
@soamvasani For sake of completeness may we should have environment also in specs? I don't see it right now and package build fails due to lack of environment.

vishal-biyani:
LGTM, I will merge this for now.

Issue #823, opened 2018-07-27
Glide Build fails


vishal-biyani:
The changes added in https://github.com/fission/fission/pull/781 cause `glide up -v` to fail:

```
[ERROR]    Failed to retrieve a list of dependencies: Error resolving imports
[ERROR]    Error looking for go/vendordir/test: Cannot detect VCS
```

And debug logs around it:

```
[INFO]    --> Fetching updates for github.com/pierrec/lz4.
[DEBUG]    Package github.com/mholt/archiver imports github.com/ulikunitz/xz
[DEBUG]    In vendor: github.com/ulikunitz/xz
[DEBUG]    Marking github.com/ulikunitz/xz to be scanned.
[INFO]    --> Fetching updates for github.com/ulikunitz/xz.
[DEBUG]    Trying to open github.com/satori/go.uuid (/Users/vishal/.glide/cache/src/https-github.com-satori-go.uuid)
[DEBUG]    Trying to open go/vendordir/test (/Users/vishal/.glide/cache/src/https-go-vendordir-test)
[DEBUG]    ImportDir error on /Users/vishal/.glide/cache/src/https-go-vendordir-test: open /Users/vishal/.glide/cache/src/https-go-vendordir-test: no such file or directory
[INFO]    --> Fetching go/vendordir/test.
[WARN]    Unable to checkout go/vendordir/test
[ERROR]    Error looking for go/vendordir/test: Cannot detect VCS
[DEBUG]    Trying to open github.com/prometheus/client_golang/prometheus/promhttp (/Users/vishal/.glide/cache/src/https-github.com-prometheus-client_golang/prometheus/promhttp)
```
The problem is that Glide tries to fetch `go/vendordir/test` - which is used in test example but since it is a URL which does not exist, the glide fails to import and fails.

soamvasani:
This seems important.  If the build is still failing let's keep it in this milestone.

life1347:
merged

PR #828, opened 2018-07-29
Uses a real go project to showcase vendor example so glide works

vishal-biyani:
The issue for this PR is explained in #823 

Right now a standard go lang project is used but I am ok with creating a small simple Go project in Fission repo and then use that as a dependency for Go vendor example. As long as the project used resolves to an appropriate path by glide - it works. 

@life1347 Do you mind taking a look at this please.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/828)


life1347:
Ooops, sorry I didn't know that will break `glide up -v`.
I think its good idea to add a real go dependency and make it more realistic, let's keep it. But since its a dependency for example,  I suggest adding following lines into `glide.yaml` to avoid adding the unnecessary dependencies in `glide.yaml`.

```
excludeDirs:
- examples
- hack
- test
```

vishal-biyani:
I tried `excludeDirs` but that could not fix Glide issue. I will check once more if I made some mistake in adding that and confirm here.

life1347:
I tested with glide version: 10.3.1

```
# glide.yaml
package: github.com/fission/fission
excludeDirs:
- examples
- hack
- test
import:
...
```

### Before adding
```
[WARN]	Unable to checkout go/vendordir/test
[ERROR]	Error looking for go/vendordir/test: Cannot detect VCS
...
[ERROR]	Failed to retrieve a list of dependencies: Error resolving imports
```

### After
`glide up -v` without error.

vishal-biyani:
I will update this PR and close other one.

vishal-biyani:
@life1347 updated the PR with `excludeDirs` added

life1347:
LGTM

Issue #807, opened 2018-07-19
Implement v2 interface for environment & environment docs 


life1347:
Currenlty, fission environment has two interface:  v1, v2

* v1
  * Support to load function from one single file.

* v2
  * Support to load function from one file or from directory
  * Support `--entrypoint` to specify which function need to be load in.
  * Support pkg builder to build pkg from source package

Following is the interface implementation status

Environment | v1 | v2
------------ | --- | ---
binary | v | v
go | v | v
jvm | v | v
python | v | v
dotnet | v | x
dotnet20 | v | x
nodejs | v | x
perl | v | x
php7 | v | x
ruby | v | x

To support v2 interface, the environment need to implement:

* API exposed in url `/v2/specialize` with following request json body.

https://github.com/fission/fission/blob/ec495fbd1d5b10c4f04bab63a61f62d07c34c3e5/types.go#L72-L90

* Environment package builder 

The job for environment builder is to build deploy package from source package. To implement it, all you is one Dockerfile that contains necessary build environment and one default build script. 

There are two build envrironment variable will be accessiable during the build process

1. `$SRC_PKG`: The path to source package.
2. `DEPLOY_PKG `: The path to store final articfact (a.k.a deploy package)

For more information of default build script and builder dockerfile, please refer:
* https://github.com/fission/fission/blob/master/environments/go/builder
* https://github.com/fission/fission/blob/master/environments/binary/builder
* https://github.com/fission/fission/blob/master/environments/python/builder

paraspatidar:
in case anyone looking for Support for V2 as well as builder for `dotnet 2.0` , it  is available at following PR : 
https://github.com/fission/fission/pull/1001

PR #836, opened 2018-08-01
Add v2 interface support for nodes env

garyyeap:
Support all the v2 new features(#807)
The examples has been updated as well, please check the usage from the example.

P/S: Documentation in progress.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/836)


soamvasani:
Hey @garyyeap this is great!  Lots of people are going to find this useful!  

Quick comment before I get into the review -- did you test the existing examples?  We need to make sure they still work with the new environment.

garyyeap:
hi @soamvasani 
All tests passed except `stock.js` -- due to the outdated code: 
1. Wasn't sync with README  
2. Google finance API has changed its response

I would suggest removing this example as it relies on non-officially public  API and there is a `weather.js` which has the same functionality as `stock.js` does.

environments/nodejs/server.js
```diff
@@ -7,56 +7,73 @@ const express = require('express');
...
+        return true;
     } catch(e) {
         console.error(`user code load error: ${e}`);
-        res.status(500).send(JSON.stringify(e));
-        return;
+        return false;
+    }
+}
+
+function withMakesureGeneric(func) {
```
soamvasani: call this `WithEnsureGeneric` instead

environments/nodejs/server.js
```diff
@@ -7,56 +7,73 @@ const express = require('express');
...
+function loadFunction(modulepath, funcname) {
     // Read and load the code. It's placed there securely by the fission runtime.
     try {
         var startTime = process.hrtime();
-        userFunction = require(argv.codepath);
+        // support v1 codepath and v2 entrypoint like 'foo', '', 'index.hello'
+        userFunction = funcname ? require(modulepath)[funcname] : require(modulepath);
         var elapsed = process.hrtime(startTime);
         console.log(`user code loaded in ${elapsed[0]}sec ${elapsed[1]/1000000}ms`);
+        return true;
```
soamvasani: Sugestion: return userFunction instead of true/false, and set the global from the specialize functions

environments/nodejs/server.js
```diff
@@ -7,56 +7,73 @@ const express = require('express');
...
+
+        func(req, res);
     }
-    res.status(202).send();
+}
+
+function specializeV2(req, res) {
+    // for V2 entrypoint, 'filename.funcname' => ['filename', 'funcname']
+    const entrypoint = req.body.functionName ? req.body.functionName.split('.') : [];
+    // for V2, filepath is dynamic path
```
soamvasani: check entrypoint.length and make sure it's what you expect, otherwise return an informative error

environments/nodejs/server.js
```diff
@@ -7,56 +7,73 @@ const express = require('express');
...
+    const modulepath = path.join(req.body.filepath, entrypoint[0] || '');
+
+    if(loadFunction(modulepath, entrypoint[1])){
+      res.status(202).send();
+    } else {
+      res.status(500).send(JSON.stringify(e));
```
soamvasani: where is `e` defined?

environments/nodejs/server.js
```diff
@@ -7,56 +7,73 @@ const express = require('express');
...
+    // Node resolves module paths according to a file's location. We load
+    // the file from argv.codepath, but tell users to put dependencies in
+    // the server's package.json; this means the function's dependencies
+    // are in /usr/src/app/node_modules.  We could be smarter and have the
+    // function deps in the right place in argv.codepath; b ut for now we
+    // just symlink the function's node_modules to the server's
+    // node_modules.
+    fs.symlinkSync('/usr/src/app/node_modules', `${path.dirname(modulepath)}/node_modules`);
+
+    loadFunction(modulepath);
```
soamvasani: check errors

environments/nodejs/package.json
```diff
@@ -14,13 +18,10 @@
   },
   "dependencies": {
```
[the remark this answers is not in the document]
garyyeap: I didn't get any issue so far.
I think it will causes confusion for new users if we didn't remove the unused modules and the number may grow up as nobody why those modules existing for.

environments/nodejs/server.js
garyyeap: I'm not sure this is the better way as the entrypoint can be empty(default is index.js) or a file name or `filename.functionname`

environments/nodejs/builder/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cd ${SRC_PKG}
+npm install && npm cache clean --force && cp -r ${SRC_PKG} ${DEPLOY_PKG}
```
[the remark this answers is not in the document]
soamvasani: Oh, that's for making the container smaller.  But in this case we want to keep the cache, to make subsequent builds faster.

environments/nodejs/package.json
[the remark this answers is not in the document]
soamvasani: You wouldn't find an issue unless you were using those dependencies; if you were using them your functions would break.  So we can't remove them without deprecating them first.  Let's get there slowly; keep the old dependencies, then we'll deprecate at some point, and remove them later.

I understand your concern about the module list growing, I don't think it's too big yet but we should watch out for that; and your change certainly helps with that problem.

environments/nodejs/server.js
soamvasani: I see, ok.

life1347:
@soamvasani any review progress on this PR?

environments/nodejs/package.json
```diff
@@ -14,13 +18,10 @@
...
     "body-parser": "*",
-    "co": "~4.6.0",
```
soamvasani: This is confusing... I see that you added a commit for adding these dependencies back, but the final diff still shows the dependencies removed.  Maybe some merge error?

garyyeap: My bad, the file updated from the commit was `examples/nodejs/package.json`, I forgot to update `environments/nodejs/package.json` as well. 

Please check the latest commit. Sorry

soamvasani:
LGTM, thanks!  Could you please rebase to master?

life1347:
test passed: https://travis-ci.org/fission/fission/builds/428466767
merged. 🎉

soamvasani:
👏 @garyyeap !  Thanks for your effort!

PR #838, opened 2018-08-01
Fix newdeploy not updates deployment after function's entrypoint changed

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/838)


vishal-biyani:
LGTM

Issue #796, opened 2018-07-16
HTTP trigger should be to handle wildcard path & multiple HTTP methods


life1347:
1. Handle wildcard path

Router will consider following two paths as two HTTP trigger path. But for RESTful API these two URL paths are considered to map to one function.
```
$ http://$FISSION_ROUTER/account/
$ http://$FISSION_ROUTER/account/foo
```

This can be fixed by `r.PathPrefix("/products/")` describe in https://github.com/gorilla/mux/blob/master/README.md#matching-routes

```
ht := muxRouter.PathPrefix(trigger.Spec.RelativeURL).HandlerFunc(fh.handler)
```

2. Handle multiple HTTP methods

We should allow to bind multiple HTTP methods to one http trigger instead of creating tons of triggers.

Example:
```
$ fission ht create --method GET,POST,PUT --url "/account/" --function h91
```

soamvasani:
We actually do handle wildcard paths... it's an undocumented feature.

You can set the URL to `/account/{name}` and the function will be invoked with a `X-Fission-Param-name: <value>` header.

life1347:
After some tests,

```
$ fission route list
NAME                                 METHOD HOST URL                     INGRESS FUNCTION_NAME
d652aaba-df43-4522-adb0-ebaf569f2e6d GET         /account/{name}/{nnZ}   false   h1
```

The path will map to Header to `X-Fission-Params-{path name}`, like

```
X-Fission-Params-Name:[qwe]
X-Fission-Params-Nnz:[gggg]
```

One thing worth to notice is that the name of parameters will be transform to `[A-Z][a-z0-9]+`. For example, `aaA` will be transform to `Aaa`.
Then, the function can access the parameter through HTTP header `X-Fission-Params-Aaa`.

BTW, both url `/account/{name}/{NnZ}` and `/account/{name}/{nnZ}` are allow to create. We need to add addtional url validation on controller.

```
$ fission ht list
NAME                                 METHOD HOST URL                     INGRESS FUNCTION_NAME
3e02503c-a402-41fe-994f-55b2cbbfb41a GET         /account/{name}/{NnZ}   false   h1
d652aaba-df43-4522-adb0-ebaf569f2e6d GET         /account/{name}/{nnZ}   false   h1
```

PR #839, opened 2018-08-02
Handle multiple http methods in one http trigger

life1347:
Address issue https://github.com/fission/fission/issues/796
Consider the backward compatibility of spec files and existing trigger so not to rename fields from `Method` to `Methods` in the end.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/839)

### DIFF
--- a/router/httpTriggers.go
+++ b/router/httpTriggers.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


soamvasani:
Can you remind me why we need this?  And also, it should be structured in the CRD as a list, not a comma separated string.

soamvasani:
Sorry, I still don't get it -- is this a convenience thing, so users don't have to deal with multiple custom resources per http path, or is it something else?  In that case we could use the CLI to output multiple triggers (e.g. you could pass multiple `--method` params and the cli could create multiple trigger objects).

PR #855, opened 2018-08-09
Configurable namespace creation

### LINKED ISSUE #853: Namespace creation optional
In case I want/have to create the fission namespaces manuel, there should be an option to not create them.

My usecase is the tiller which I use does not have permissions to create namespaces, but to install in all namespaces

michaelgaida:
Added condition with value to make the creation of namespaces optional.

Default: true (no change in default behaviour)

Fix #853

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/855)


charts/fission-all/values.yaml
```diff
@@ -33,6 +33,9 @@ routerPort: 31314
 ## Port at which NATS streaming service should be exposed
 natsStreamingPort: 31316
 
+## Set to false if you create the namespaces manually
+createNamespace: false
```
soamvasani: You said it defaults to true, so this should be `true`, right?

michaelgaida: Good point, will fix. Cheers

vishal-biyani:
Thanks @michaelgaida for this change.

PR #856, opened 2018-08-09
Router optimization

life1347:
In 0.9.2, the router sends multiple GetServiceForFunction requests to executor to get the service URL. However, the response time of executor will increase under high-concurrency situation due to too many requests are waiting for processing.

To solve the problem, an update lock was added to the router. All of goroutines belongs to the same function need to grab the update lock before sending the request. Only the first goroutine which gets the update lock is allowed to send request. In this way, we reduce the burden of executor and lower the failure rate.

Test result:
Fission router 0.9.2: https://gist.github.com/life1347/ab655252c45411b010a220efe50c1281
Fission router with optimization: https://gist.github.com/life1347/461063c12fd1c01e828fef39dd17db76

From the test results, we can see that in most of different concurrency level, the optimize router gives shorter response time than 0.9.2 router. Also, the optimize router is able to handle up to 1600 concurrency while 0.9.2 router is not able to response anything.


router/functionServiceMap.go
```diff
@@ -80,3 +82,24 @@ func (fmap *functionServiceMap) remove(f *metav1.ObjectMeta) error {
 	mk := keyFromMetadata(f)
 	return fmap.cache.Delete(*mk)
 }
+
+func (fmap *functionServiceMap) grabUpdateLock(f *metav1.ObjectMeta, wg *sync.WaitGroup) (*sync.WaitGroup, error) {
```
life1347: Not sure where to put these function is better, for now just put these two here.

soamvasani:
Can you describe the test environment for the benchmark?  Those numbers are really unusual (7 req/sec??)

router/functionServiceMap.go
```diff
@@ -80,3 +82,24 @@ func (fmap *functionServiceMap) remove(f *metav1.ObjectMeta) error {
...
+	mk := f.Namespace + f.Name
```
soamvasani: use cacheKey

router/functionHandler.go
```diff
@@ -102,17 +105,109 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
 	}
 
-	executingTimeout := roundTripper.funcHandler.tsRoundTripperParams.timeout
+	// cache lookup to get serviceUrl
+	serviceUrl, err = roundTripper.funcHandler.fmap.lookup(roundTripper.funcHandler.function)
 
 	for i := 0; i < roundTripper.funcHandler.tsRoundTripperParams.maxRetries-1; i++ {
-		if needExecutor {
+
+		if serviceUrl != nil {
```
soamvasani: can you add err==nill &&

router/functionHandler.go
```diff
@@ -102,17 +105,109 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
+			}
+
+			// if transport.RoundTrip returns a non-network dial error, then relay it back to user
+			if !fission.IsNetworkDialError(err) {
+				err = errors.Wrapf(err, "Error sending request to function %v",
+					roundTripper.funcHandler.function.Name)
+				return resp, err
+			}
+
+			// timeout or dial error goes here
```
soamvasani: need more explanation of failcounter

router/functionHandler.go
```diff
@@ -102,17 +105,109 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
+
+				time.Sleep(10 * time.Millisecond)
+
+				if serviceUrlFromExecutor {
+					continue
+				}
+			} else {
+				// if transport.RoundTrip returns a network dial error and serviceUrl was from cache,
+				// it means, the entry in router cache is stale, so invalidate it.
+				// also set needExecutor to true so a new service can be requested for function.
```
soamvasani: since you removed needExecutor please update the comment

router/functionServiceMap.go
soamvasani: Yes we should add a new cache `fmap.waitGroups` or something like that

life1347:
@smruthi2187 After some testing, I choose to keep changes in router instead of executor client cause it's easier to control/limit total requests at the source(router).

life1347:
Merged 🎉

PR #861, opened 2018-08-13
Check pod container ready state

life1347:
Notice there is an issue(https://github.com/kubernetes/kubernetes/issues/61376) indicates that `status.phase` of pod is in `Running` even when the pod is in `Terminating` state. 

This PR aims to resolve the problem to prevent the terminating pod was selected by poolmgr.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/861)

### DIFF
--- a/common.go
+++ b/common.go
@@ -108,10 +108,17 @@ func IsNetworkDialError(err error) bool {
 	return false
 }
 
-// IsReadyPod checks that all containers in a pod are ready and returns true if so
+// IsReadyPod checks both all containers in a pod are ready and the .metadata.DeletionTimestamp
+// is nil. Return true if so
 func IsReadyPod(pod *apiv1.Pod) bool {
 	// since its a utility function, just ensuring there is no nil pointer exception
 	if pod == nil {
+		return false
+	}
+
+	// pod is in "Terminating" status if deletionTimestamp is not nil
+	// https://github.com/kubernetes/kubernetes/issues/61376
+	if pod.ObjectMeta.DeletionTimestamp != nil {
 		return false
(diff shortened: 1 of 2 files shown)


common.go
```diff
@@ -108,13 +108,20 @@ func IsNetworkDialError(err error) bool {
 	return false
 }
 
-// IsReadyPod checks that all containers in a pod are ready and returns true if so
+// IsReadyPod checks both all containers in a pod are ready and the .metadata.DeletionTimestamp
+// is nil. Return true if so
```
erwinvaneyk: Remove the last sentence; it is not very useful

executor/poolmgr/gp.go
```diff
@@ -228,21 +226,14 @@ func (gp *GenericPool) _choosePod(newLabels map[string]string) (*apiv1.Pod, erro
 		for i := range podList.Items {
 			pod := podList.Items[i]
 
-			// If a pod has no IP it's not ready
-			if len(pod.Status.PodIP) == 0 || string(pod.Status.Phase) != POD_PHASE_RUNNING {
+			// If a pod has no IP or any of pod's containers
+			// is not in ready state, it's not ready.
+			if len(pod.Status.PodIP) == 0 || !fission.IsReadyPod(&pod) {
```
erwinvaneyk: Shouldn't the PodIP check be in the `IsReadyPod` function if it is part of the decision of whether a pod is ready?

life1347: I think we can remove pod ip check here since a pod definitely has IP when its ready

PR #862, opened 2018-08-13
Fix newdeploy fail to update HPA, deployment of a function after function update

life1347:
After adopting multi-tenancy, newdeploy no longer able to update HPA, deployment due to the wrong namespace was used. This PR aims to resolve this by adding backward compatibility code to get correct function obj namespace.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/862)


vishal-biyani:
LGTM, thanks for the fix.

PR #864, opened 2018-08-15
Replay recorded requests by ReqUID

Amusement:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/864)


replayer/replay.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/fission/fission/redis/build/gen"
+)
+
+func ReplayRequest(routerUrl string, request *redisCache.Request) ([]string, error) {
+	path := request.URL["Path"] // Includes slash prefix
+	payload := request.URL["Payload"]
```
life1347: I suddenly found some problem I didn't notice for the record PR. You file an issue to address these problem later.

1. Lost some of the structure fields in original request.
For example, `RawPath` in url is not included.
https://golang.org/src/net/url/url.go?s=9021:9537#L323
```
type URL struct {
	Scheme     string
	Opaque     string    // encoded opaque data
	User       *Userinfo // username and password information
	Host       string    // host or host:port
	Path       string    // path (relative paths may omit leading slash)
	RawPath    string    // encoded path hint (see EscapedPath method)
	ForceQuery bool      // append a query ('?') even if RawQuery is empty
	RawQuery   string    // encoded query values, without '?'
	Fragment   string    // fragment for references, without '#'
}
```
We better to embed the whole url.Url object in the `redisCache.Request` directly.
Also, we lost some of fields in the original request structure like `RemoteAddr ` `Trailer ` . https://github.com/golang/go/blob/master/src/net/http/request.go#L106-L320

2.  Payload should be in `redisCache.Request ` not `redisCache.Request.URL`

smruthi2187:
From @Amusement and @life1347 's discussion, this is good to merge! 🎉

PR #866, opened 2018-08-16
Fix for #662: avoid unnecessary builds

smruthi2187:
Closing the other PR and opened this instead.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/866)

### DIFF
--- a/fission/package.go
+++ b/fission/package.go
@@ -117,6 +117,13 @@ func pkgUpdate(c *cli.Context) error {
 	})
 	checkErr(err, "get package")
 
+	// if the new env specified is the same as the old one, no need to update package
+	// same is true for all update parameters, but, for now, we dont check all of them - because, its ok to
+	// re-write the object with same old values, we just end up getting a new resource version for the object.
+	if envName == pkg.Spec.Environment.Name && envNamespace == pkg.Spec.Environment.Namespace {
+		envName = ""
+	}
+
 	fnList, err := getFunctionsByPackage(client, pkg.Metadata.Name, pkg.Metadata.Namespace)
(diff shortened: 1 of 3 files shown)


fission/package.go
```diff
@@ -145,8 +156,12 @@ func updatePackage(client *client.Client, pkg *crd.Package, envName, envNamespac
 	var srcArchiveMetadata, deployArchiveMetadata *fission.Archive
 	needToBuild := false
 
-	if len(envName) > 0 {
+	if len(envName) > 0 && len(srcArchiveName) > 0 {
```
life1347: Remove `len(srcArchiveName) > 0` as we discussed.

fission/package.go
```diff
@@ -117,6 +117,17 @@ func pkgUpdate(c *cli.Context) error {
 	})
 	checkErr(err, "get package")
 
+	// if the new env specified is the same as the old one, no need to update package
+	// same is true for all update parameters, but, for now, we dont check all of them - because, its ok to
+	// re-write the object with same old values, we just end up getting a new resource version for the object.
+	if len(envName) > 0 && envName == pkg.Spec.Environment.Name && envNamespace == pkg.Spec.Environment.Namespace {
```
life1347: `if len(envName) > 0 && envName == pkg.Spec.Environment.Name` should be enough

fission/package.go
smruthi2187: makes sense. thanks for pointing this out.

fission/package.go
smruthi2187: oops. sure.

life1347:
LGTM

PR #889, opened 2018-09-10
Fix fetcher not close file descriptor correctly

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/889)


soamvasani:
Ouch, can't believe this bug was in there this long. LGTM.

Issue #876, opened 2018-08-29
add X-Forwarded-Host header 


soamvasani:
Fission overwrites the Host header, so users lose information about what Host the original request was for.  We can add this information into X-Forwarded-Host.  While we're at it we should also set the `Forwarded` header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded).

PR #890, opened 2018-09-11
Add X-Forwarded-Host to request header

life1347:
This PR aims to solve #876

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/890)

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -19,8 +19,6 @@ package router
 import (
 	"bytes"
 	"fmt"
-	"github.com/gorilla/mux"
-	"github.com/satori/go.uuid"
 	"io/ioutil"
 	"math/rand"
 	"net"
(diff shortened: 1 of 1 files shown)


router/functionHandler.go
```diff
@@ -305,3 +308,34 @@ func (fh *functionHandler) handler(responseWriter http.ResponseWriter, request *
...
+	if isIPv6 {
+		// For the "Forwarded" header, if a host is an IPv6 address
+		// it should be quoted and enclosed in square brackets
+		host = fmt.Sprintf("\"[%v]\"", req.Host)
+	} else {
+		host = req.Host
+	}
+
+	req.Header.Set("Forwarded", "host="+host)
+	req.Header.Set("X-Forwarded-Host", req.Host)
```
soamvasani: Don't add this header if it already exists.

soamvasani: (Both headers I mean. I think existing Forwarded/XFH headers should just be passed on unmodified... as far as I can understand.)

PR #897, opened 2018-09-18
Check CRD creation error instead of doing return directly

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/897)


soamvasani:
LGTM, thanks

PR #898, opened 2018-09-19
Fixed the spec validation UX issue

### LINKED ISSUE #809: Spec experience needs improvement
When using `fission spec` if you create a object that depends on another object (For ex. a function depends on an environment) - the message shown on CLI seems like an error to user whereas it might just be warning. This is super confusing to someone new to Fission and needs to be improved.

```
$ fission spec init
Creating fission spec directory 'specs'

$ fission env create --spec --name nodejs --envns default --image fission/node-env
$ fission fn create --spec --name hello --env nodejs --code hello.js
Environment "nodejs" does not exist. Please create the environment before executing the function. 
For example: `fission env create --name nodejs --envns default --image <image>`
```
**HERE** user gets confused and thinks something is broken

```
$ fission route create --spec --method GET --url /gh/hello --function hello
function 'hello' does not exist, use 'fission function create --name hello ...' to create the function

$fission spec apply
$fission fn test --name hello
Hello, Fission!
```

vishal-biyani:
This fixes #809 

For normal CLI creation, added `[WARNING]` so that it more clear to user. 

```
$ fission fn create  --name testfn --env gotest --code hello.js 
[WARNING] Environment "gotest" does not exist. Please create the environment before executing the function. 
For example: `fission env create --name gotest --envns default --image <image>`

function 'testfn' created

$ fission ht create --function gotest1 --url /gotest1 
function 'gotest1' does not exist, use 'fission function create --name gotest1 ...' to create the function
trigger '2ae7aa10-0629-4c65-97eb-3d249ac5227d' created
```

For specs - don't give a warning when creating the spec, instead, warn when validating spec (The route error for spec exists already and no code change for that):

```
$ fission fn create --spec --name testfn --env gotest --code hello.js 
$ fission spec validate
[WARNING] Environment gotest is referred in function testfn but not declared in specs

$ fission ht create --spec --function httest --url /goty 
$ fission spec validate
[WARNING] Environment gotest is referred in function testfn but not declared in specs
Error validating specs: 1 error occurred:
	* specs/route-5353a8c0-1483-4974-b469-2425b58cc18f.yaml:1: HTTPTrigger '5353a8c0-1483-4974-b469-2425b58cc18f' references unknown function 'httest'
```

### DIFF
--- a/fission/log/log.go
+++ b/fission/log/log.go
@@ -32,6 +32,10 @@ func Fatal(msg interface{}) {
 }
 
 func Warn(msg interface{}) {
+	os.Stderr.WriteString(fmt.Sprintf("[WARNING] %v\n", msg))
+}
+
+func Log(msg interface{}) {
 	os.Stderr.WriteString(fmt.Sprintf("%v\n", msg))
 }
 
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -388,6 +388,19 @@ func (fr *FissionResources) validate() error {
 	// we do not error on unreferenced functions (you can call a function through workflows,
 	// `fission function test`, etc.)
 
+	// Index envs, warn on functions referencing an environment for which spes does not exist
+	environments := make(map[string]bool)
+	for _, e := range fr.environments {
+		environments[fmt.Sprintf("%s:%s", e.Metadata.Name, e.Metadata.Namespace)] = false
+	}
+
+	functions = make(map[string]bool)
+	for _, f := range fr.functions {
+		if _, ok := environments[fmt.Sprintf("%s:%s", f.Spec.Environment.Name, f.Spec.Environment.Namespace)]; !ok {
+			log.Warn(fmt.Sprintf("Environment %s is referred in function %s but not declared in specs", f.Spec.Environment.Name, f.Metadata.Name))
+		}
+	}
+
 	// (ErrorOrNil returns nil if there were no errors appended.)
 	return result.ErrorOrNil()
 }
@@ -961,7 +974,7 @@ func localArchiveFromSpec(specDir string, aus *ArchiveUploadSpec) (*fission.Arch
 		absGlob := rootDir + "/" + relativeGlob
 		f, err := filepath.Glob(absGlob)
 		if err != nil {
-			log.Warn(fmt.Sprintf("Invalid glob in archive %v: %v", aus.Name, relativeGlob))
+			log.Log(fmt.Sprintf("Invalid glob in archive %v: %v", aus.Name, relativeGlob))
 			return nil, err
 		}
 		files = append(files, f...)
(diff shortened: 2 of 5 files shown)


fission/log/log.go
```diff
@@ -32,6 +32,10 @@ func Fatal(msg interface{}) {
 }
 
 func Warn(msg interface{}) {
+	os.Stderr.WriteString(fmt.Sprintf("[WARNING] %v\n", msg))
+}
+
+func Log(msg interface{}) {
```
erwinvaneyk: Might be better to name this `Info`, which is more in line with naming in common log libraries

fission/spec.go
```diff
@@ -388,6 +388,19 @@ func (fr *FissionResources) validate() error {
...
+	// Index envs, warn on functions referencing an environment for which spes does not exist
+	environments := make(map[string]bool)
+	for _, e := range fr.environments {
+		environments[fmt.Sprintf("%s:%s", e.Metadata.Name, e.Metadata.Namespace)] = false
+	}
+
+	functions = make(map[string]bool)
+	for _, f := range fr.functions {
+		if _, ok := environments[fmt.Sprintf("%s:%s", f.Spec.Environment.Name, f.Spec.Environment.Namespace)]; !ok {
+			log.Warn(fmt.Sprintf("Environment %s is referred in function %s but not declared in specs", f.Spec.Environment.Name, f.Metadata.Name))
```
erwinvaneyk: Nice! 😁

fission/spec.go
```diff
@@ -388,6 +388,19 @@ func (fr *FissionResources) validate() error {
 	// we do not error on unreferenced functions (you can call a function through workflows,
 	// `fission function test`, etc.)
 
+	// Index envs, warn on functions referencing an environment for which spes does not exist
+	environments := make(map[string]bool)
+	for _, e := range fr.environments {
+		environments[fmt.Sprintf("%s:%s", e.Metadata.Name, e.Metadata.Namespace)] = false
```
erwinvaneyk: You can also add a warning, basically the inverse of the one below: `Environment X is unused.`

fission/spec.go
```diff
@@ -388,6 +388,19 @@ func (fr *FissionResources) validate() error {
 	// we do not error on unreferenced functions (you can call a function through workflows,
 	// `fission function test`, etc.)
 
+	// Index envs, warn on functions referencing an environment for which spes does not exist
+	environments := make(map[string]bool)
```
erwinvaneyk: There is a go convention which says that if you are using a map just for the keys. You should use `map[string]struct{}` since that makes it explicit that the value is not significant.

fission/log/log.go
vishal-biyani: Done

fission/spec.go
vishal-biyani: I am not sure we need this. The functions depend on environment hence warning, but environment do not depend on anything specific.

PR #911, opened 2018-09-30
Fixes #758, uses v2 specialize for env versions 2 or higher

### LINKED ISSUE #758: Specialize URL in fetcher error prone
The env version check in github.com/fission/fission/environments/fetcher/cmd/main.go can cause an issue for environment versions 3 and higher. It leads to `/specialize` URL if env version is 1 or 3 & higher. Ideally for env versions greater than 2, the URL should be `/v2/specialize`

```
if envVersion == 2 {
        contentType = "application/json"
        specializeURL = "http://localhost:8888/v2/specialize"
        reader = bytes.NewReader([]byte(*loadPayload))
    } else {
        contentType = "text/plain"
        specializeURL = "http://localhost:8888/specialize"
        reader = bytes.NewReader([]byte{})
    }
```

vishal-biyani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/911)

### DIFF
--- a/environments/fetcher/cmd/main.go
+++ b/environments/fetcher/cmd/main.go
@@ -114,7 +114,7 @@ func specializePod(f *fetcher.Fetcher, fetchPayload *string, loadPayload *string
 	var specializeURL string
 	var reader *bytes.Reader
 
-	if envVersion == 2 {
+	if envVersion >= 2 {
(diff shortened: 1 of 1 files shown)


soamvasani:
LGTM, thanks.

PR #912, opened 2018-10-01
Replace sleep with wait in several test scripts

soamvasani:
1. Allow CLI to wait for router controllers

When a script or a human user creates some resources (functions,
triggers), the router notices these changes and updates its config.
However the user or script has no way of knowing when the router has
completed this task.

This change adds: (1) a router API endpoint that returns the highest
among the ResourceVersions of all functions and httptriggers it has
seen, and (2) A cli change that writes the RV of any created or
updated function or trigger into a file (3) A cli change that waits for the
router's reported RV to match the one written into the file.

[This thing is a violation of abstraction layers.  It assumes that
RVs are monotonically increasing uint64s while the K8s API only
guarantees that they are unique strings (not even numbers).  However
K8s gets these numbers from the etcd index, which is (I think) the Raft
log index, which does increase monotonically (I think).]

2. Use the new CLI in tests to reduce unconditional sleeps

3. Use the new (K8s 1.11) `kubectl wait` feature to remove a few more sleeps from tests

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/912)

### DIFF
--- a/fission/admin.go
+++ b/fission/admin.go
@@ -47,21 +47,15 @@ func updateSessionRV(newRVstr string) error {
 		// nothing to update, "new" isn't new enough
 		return nil
 	}
-	fmt.Printf("setting rv to %v", newRVstr) //xxx
 
 	// write to temp and rename file (file renames are usually atomic)
 	fnTemp := fn + ".tmp"
 	err = ioutil.WriteFile(fnTemp, []byte(newRVstr), 0600)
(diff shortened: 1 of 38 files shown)


life1347:
It's a bit old PR and the problem (long CI test time) we met at that time is solved. Close it for now. Feel free to reopen it.

Issue #908, opened 2018-09-28
fission fn test returns unhelpful 502 on failed specializations


erwinvaneyk:
Triggering functions that failed to specialise results in the 502 error on invoking. This is not very helpful to the user as they will have to dig through the logs to uncover the cause of the error. I would expect that, since the executor knows that building has failed, it should return the error message of environment pod the instead of 502.

### Example
Executor logs:
```
2018/09/28 12:37:28 [inputs] Checking for cached function service
2018/09/28 12:37:28 [inputs] No cached function service found, creating one
2018/09/28 12:37:28 [inputs] getting environment for function
2018/09/28 12:37:28 [&ObjectMeta{Name:inputs,GenerateName:,Namespace:default,SelfLink:/apis/fission.io/v1/namespaces/default/functions/inputs,UID:3faf9893-c31b-11e8-9a55-42010a800070,ResourceVersion:29700741,Generation:1,CreationTimestamp:2018-09-28 12:37:23 +0000 UTC,DeletionTimestamp:<nil>,DeletionGracePeriodSeconds:nil,Labels:map[string]string{},Annotations:map[string]string{},OwnerReferences:[],Finalizers:[],ClusterName:,Initializers:nil,}] getting env
2018/09/28 12:37:28 [inputs] getting function service from pool
2018/09/28 12:37:28 [inputs] Choosing pod from pool
2018/09/28 12:37:28 [map[unmanaged:true functionName:inputs functionUid:3faf9893-c31b-11e8-9a55-42010a800070 executorInstanceId:exdoo13v]] found 1 ready pods of 1 total
2018/09/28 12:37:28 Chosen pod: workflow-3b1f7faa-c31b-11e8-9a55-42010a800070-3z56lxrr-99fjvvq8 (in 5.081866ms)
2018/09/28 12:37:28 [inputs] calling fetcher to copy function with fetcher url: http://10.20.0.150:8000/
2018/09/28 12:37:28 [inputs] specializing pod
2018/09/28 12:37:28 specialize url: http://10.20.0.150:8888/v2/specialize
2018/09/28 12:37:28 Failed to specialize pod: Internal error - failed to specialize package: failed to parse bytes into WorkflowSpec: json: cannot unmarshal string into Go value of type map[string]*json.RawMessage
[...]

PR #917, opened 2018-10-02
Return the error on failed specializations with `fn test --debug` 

smruthi2187:
Fix for https://github.com/fission/fission/issues/908

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/917)


charts/fission-all/values.yaml
```diff
@@ -99,3 +99,8 @@ pruneInterval: 60
 
 ## Fission pre-install/pre-upgrade checks live in this image
 preUpgradeChecksImage: fission/pre-upgrade-checks
+
+## the environment where fission is installed.
+## If its set to any other value other than production, then debug information is returned in http response for
+## fn test cli in case of fission internal errors.
+envType: production
```
soamvasani: I think the production/debug flag should be a per-function. Ok to have a global default, though.

executor/client/client.go
```diff
@@ -47,30 +46,20 @@ func MakeClient(executorUrl string) *Client {
...
 
 	body, err := json.Marshal(metadata)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	resp, err := http.Post(executorUrl, "application/json", bytes.NewReader(body))
 	if err != nil {
-		return "", err
```
soamvasani: I think the client code should handle the HTTP response, and not return it out of the interface.  You could perhaps add a custom error type if you want to return more information in the error.

fission/main.go
```diff
@@ -103,6 +103,7 @@ func main() {
 	fnLogCountFlag := cli.StringFlag{Name: "recordcount", Usage: "the n most recent log records"}
 	fnForceFlag := cli.BoolFlag{Name: "force", Usage: "Force update a package even if it is used by one or more functions"}
 	fnExecutorTypeFlag := cli.StringFlag{Name: "executortype", Usage: "Executor type for execution; one of 'poolmgr', 'newdeploy' defaults to 'poolmgr'"}
+	fnTestDebugFlag := cli.BoolFlag{Name: "debug", Usage: "return internal error details as part of http response for debugging"}
```
soamvasani: If you have a per-function "dev/production" flag, you can remove the header and this flag, and just unconditionally return errors in dev mode

router/functionHandler.go
```diff
@@ -154,23 +156,37 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
-				log.Printf("Err from GetServiceForFunction : %v", err)
+				log.Debugf("Err from GetServiceForFunction : %v", err)
 				// We might want a specific error code or header for fission failures as opposed to
 				// user function bugs.
 				return nil, err
 			}
 
+			if resp.StatusCode != http.StatusOK {
+				if roundTripper.funcHandler.envType != "production" && len(req.Header.Get(fission.FissionDebugHeader)) != 0 {
+					return resp, nil
```
soamvasani: So this is just the executor error, right?  Can we return something more useful to the user, like say the last N lines of the pod's logs?

charts/fission-all/values.yaml
erwinvaneyk: ^ agreed - since " any other value other than production, then debug information is returned", maybe it is clearer to just make it a boolean debug flag `debug: true` or `env.debug: true`. Or do we plan on having more env types?

fission/main.go
erwinvaneyk: @soamvasani, just my opinion on this, I like this per-request level of debugging; it makes debugging a one-of request quick and easy. Use cases: demos, explanations, and troubleshooting a production-level function (you don't want all workload to suddenly be in debug mode). 

Isn't updating the function spec before and after a request to set and unset debug mode a bit verbose from a user's perspective, or am I missing a faster workflow to do this?

router/functionHandler.go
```diff
@@ -154,23 +156,37 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
 			if err != nil {
-				log.Printf("Err from GetServiceForFunction : %v", err)
+				log.Debugf("Err from GetServiceForFunction : %v", err)
 				// We might want a specific error code or header for fission failures as opposed to
 				// user function bugs.
 				return nil, err
 			}
 
+			if resp.StatusCode != http.StatusOK {
+				if roundTripper.funcHandler.envType != "production" && len(req.Header.Get(fission.FissionDebugHeader)) != 0 {
```
erwinvaneyk: If you are keeping the "production" and "test" values, you might want to extract them to package-level constants.

router/functionHandler.go
```diff
@@ -154,23 +156,37 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
+					return resp, nil
+				}
+
+				return nil, fmt.Errorf("Internal Error")
```
erwinvaneyk: Internal error is not really helpful when debugging. Maybe add the actual HTTP error (status code + msg/body) to the error here? Or log it?

router/functionHandler.go
```diff
@@ -154,23 +156,37 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
+
 			if err != nil {
-				log.Printf("Err from GetServiceForFunction : %v", err)
+				log.Debugf("Err from GetServiceForFunction : %v", err)
 				// We might want a specific error code or header for fission failures as opposed to
 				// user function bugs.
 				return nil, err
 			}
 
+			if resp.StatusCode != http.StatusOK {
```
erwinvaneyk: I understand that the executor server only returns 200 or 4xx or 5xx, but it might be future proof to not reject other 'succesful' status types here. Maybe: `resp.StatusCode >= 400`?

router/functionHandler.go
```diff
@@ -154,23 +156,37 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
...
+				if roundTripper.funcHandler.envType != "production" && len(req.Header.Get(fission.FissionDebugHeader)) != 0 {
+					return resp, nil
+				}
+
+				return nil, fmt.Errorf("Internal Error")
+			}
+
+			svcName, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				return nil, err
```
erwinvaneyk: The errors of the `io` package are not very contextual. I would be in favour of just wrapping the `err` with an error explaining that it failed to read the body of the GetService request?

charts/fission-all/values.yaml
smruthi2187: I was thinking that there's one installation of fission per each env - dev, qa, production etc.
However, if there is a possibility that one fission instance can handle a few functions that are dev grade and few that are production grade, then it makes sense to have a per function debug flag.

router/functionHandler.go
smruthi2187: yes, this error has a summary of what failed during pod specialization. definitely helps to have pod's logs. but i was thinking `fission fn logs` can be useful for that.

pkg/apis/fission.io/v1/const.go
```diff
@@ -96,3 +96,8 @@ const (
 	// set a max number for iterations to prevent infinite processing of canary config
 	MaxIterationsForCanaryConfig = 10
 )
+
+const (
+	// FissionDebugHeader is added to the http request to the function when debugging information is requested
+	FissionDebugHeader = "X-Fission-Debug"
```
soamvasani: While the other comments can be addressed in later PRs, we'll have to remove this header before merging this PR -- this isn't something that the function's users should be able to set, since it exposes information about the function that the fission function authors may not want to expose to their users.

executor/client/client.go
```diff
@@ -57,16 +57,20 @@ func (c *Client) GetServiceForFunction(metadata *metav1.ObjectMeta) (string, err
 
 	resp, err := http.Post(executorUrl, "application/json", bytes.NewReader(body))
 	if err != nil {
+		log.Printf("Returning from http post, err : %v", err)
```
soamvasani: We've tried to follow a rule of having no printfs in clients.  One reason is that the client may be called from a cli context where you're not allowed to log.  Another reason is that this is a library and you want all this information to be available to the caller, not just the person reading the logs.  If you think these two errors are important enough to distinguish, use `errors.Wrap` or new error types.

Issue #769, opened 2018-07-03
Provide an easier way to create archive for pkg


life1347:
It will be nice to provide additional flags to create archive when creating function/pkg.

Solution 1. Add new flags
```
--sourcearchive value, --src value     Local path or URL for source archive
--deployarchive value, --deploy value  Local path or URL for binary archive
--sourcedir value, --srcdir value     Local path for source directory
--deploydir value, --deploydir value     Local path for deploy directory
```

Solution 2. Change behavior of current flag, and retired some flag alias.
```
--src value     Local path to source directory/archive or URL for source archive
--deploy value  Local path to deployment directory/archive or URL for deployment archive
```

erwinvaneyk:
Agree fully with solving this issue. Though I am a bit worried that we degrade user experience if we keep adding parameters to the command. 

What about allowing users to simply input a directory for `--src` or `--deploy`, and let the CLI handle both cases based on whether the input is a file, glob or a directory?

erwinvaneyk:
`--src hello.go` => build single file
`--src hello.go --src world.go` or `--src *.go` => build package from multiple files
`--src myFnDir/` => build package from directory

life1347:
@erwinvaneyk Oh! You're absolutely right! In this way, we don't need to add more flags to CLI

life1347:
But looks like we need to retire `sourcearchive` and `deployarchive` as well to avoid confusion.

life1347:
Sorry, I missed the thread notification. Originally, I thought that the flag can only accept one parameter, but I noticed we can use `StringSlice` to solve the problem so the constraint I mentioned is no longer exists.

vishal-biyani:
While working on this, I realized a few more cases/variations and I am documenting them here for future reference:

Following variations will work

```
--src hello.go => build single file 
--src hello.go --src world.go or --src "*.go" => build package from multiple files
--src myFnDir/ => build package from directory
--src pkg.zip => Source archive - should work as it works today

--deploy somedir/ => deployment package from a directory
--deploy pkg.zip => Deploy archive - should work as it works today

--src/--deploy http://someurl/somefile.js - this should work but no tests  currently
```

Following is not supported as of now:

```
--src dir1/ --src dir2/ => Multiple directory combinations
--deploy dir1/ --deploy dir2/ => Currently only single flag supported for deploy flag
```

life1347:
Just wondering why `--src dir1/ --src dir2/` is not support.
[...]

PR #927, opened 2018-10-10
Archive package user experience

vishal-biyani:
Solves #769

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/927)


vishal-biyani:
Hi @life1347 There are some specific cases for source & deploy archive (Please check comment: https://github.com/fission/fission/issues/769#issuecomment-428554520). Also source archive can have multiple inputs and deployment archive can have only one - do you think it still makes sense to simplify the above?

vishal-biyani:
Hi @life1347  I stand slightly corrected here, I think the multiple directories will work, with some riders, check examples below. I also verified this with a simple program using `filepath.Glob`

```
$ ls dir1/
__init__.py	build.sh

$ ls dir2/
requirements.txt	user.py

$ fission package create --env python --src dir1/ --src dir2/
globs= [dir1/ dir2/]
files= [dir1/ dir2/]
Package 'python-1539691799-6x6b' created

$ fission package create --env python --src dir1 --src dir2
globs= [dir1 dir2]
files= [dir1 dir2]
Package 'python-1539691811-nm71' created

$ fission package create --env python --src dir1/* --src dir2/*
globs= [dir1/__init__.py]
files= [dir1/__init__.py]
Package 'python-1539691817-sp66' created

$ fission package create --env python --src "dir1/*" --src "dir2/*"
globs= [dir1/* dir2/*]
files= [dir1/__init__.py dir1/build.sh dir2/requirements.txt dir2/user.py]
Package 'python-1539691843-0czk' created
```
As for the question:

> What specific cases will break things if we do filepath.Glob first?

The http based file links won't work

But yes I believe that the change can be simplified a lot more. I will update my PR shortly.

vishal-biyani:
@life1347 This is ready for review now. Some points to note:

To support `--code` flag a special exception had to be made. The `--code` flag is used to provide a single file and is not archived into a package, whereas `--deploy` creates an archive after this PR. But to keep backward compatibility we don't create an archive if the code is passed with --code flag. Making this consistent will need some changes in fetcher potentially and we could also remove redundant options such as `--sourcearchive` & `--deployarchive` as a separate change after one release.

life1347:
@vishal-biyani I'm confused with "remove redundant options such as --sourcearchive & --deployarchive". It means to remove the alias or remove whole subcommand?

soamvasani:
0.12 is too close, looks like there is still a bit of work here. Let's get to this in the next release.

vishal-biyani:
@life1347 After the discussion yesterday, a summary of the discussion and some more new findings:

- We will keep `--code` flag - as it is useful to the user
- Next release, we will give a warning for `--sourcearchive` and `--deployarchive` flags and remove one release later. Flags `--src` and `--deploy` will serve the function - of both files/directories and archive being passed to function
- We should warn the user to use `--deploy "multifile/*"` - so all files from the multifile dir are picked up

Findings:

- While technically we can use `--deploy "multifile/"` instead og `--deploy "multifile/*"` - it causes the directory `multifile` to be extracted inside the dir `/userfunc/user` and hence function fails.

- While we used the `fission.GetTempDir()` - it broke the usage of deploy flag with spec. I am working on that and a test for specs with packages which I will push shortly.

I have also filed #946 and #948 which were found while testing/fixing this one.

life1347:
LGTM, we can merge it after rebasing master.
Please file an issue for adding warning & removing `--sourcearchive` and `--deployarchive` flag as you mentioned above.

vishal-biyani:
Thanks, @life1347 I am going to test this a bit more on the spec side and ensure I am not introducing breaking changes before merging this to master. Might update test case too.

life1347:
As we discussed check the length of args and show message if a user tries to pass path with wildcard`*` without quoting it.

Issue #903, opened 2018-09-21
Kafka integration test 


vishal-biyani:
A test needs to be written for #831

PR #944, opened 2018-10-19
Kafka tests

vishal-biyani:
#903 A basic and simple test for Kafka integration. 

The producer function produces a message on a fixed topic and consumer function gets the message.  The logs of mqtrigger-kafka are grepped for successful invocation of consumer function.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/944)


test/tests/mqtrigger/kafka/test_kafka.sh
```diff
@@ -0,0 +1,88 @@
...
+
+nodeenv="node-kafka"
+goenv="go-kafka"
+producerfunc="producer-func"
+consumerfunc="consumer-func"
+
+log() {
+    echo $1
+}
+export -f log
```
erwinvaneyk: I did not know you could do this, cool! However, why is it needed? It looks like the functions are only used within this file?

test/tests/mqtrigger/kafka/kafka_pub/kafka-pub.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	_, _, err = producer.SendMessage(&sarama.ProducerMessage{
+		Topic: "testtopic",
+		Value: sarama.StringEncoder("{\"name\": \"testvalue\"}"),
+	})
+
+	if err != nil {
+		w.Write([]byte(fmt.Sprintf("Failed to publish message to topic %s: %v", "testtopic", err)))
+	}
+	w.Write([]byte("Successfully sent to testtopic"))
```
erwinvaneyk: If you haven't, you could use this test also as an example. I am sure it would be useful for users to have an example of Kafka in the examples directory.

test/tests/mqtrigger/kafka/kafka_pub/kafka-pub.go
```diff
@@ -0,0 +1,31 @@
...
+	if err != nil {
+		panic(err)
+	}
+	_, _, err = producer.SendMessage(&sarama.ProducerMessage{
+		Topic: "testtopic",
+		Value: sarama.StringEncoder("{\"name\": \"testvalue\"}"),
+	})
+
+	if err != nil {
+		w.Write([]byte(fmt.Sprintf("Failed to publish message to topic %s: %v", "testtopic", err)))
```
erwinvaneyk: Missing a return/panic here. Every message will also result in "successfully sent to testtopic"

test/tests/mqtrigger/kafka/test_kafka.sh
```diff
@@ -0,0 +1,88 @@
...
+      sleep 1
+    done
+}
+export -f test_mqmessage 
+
+waitBuild() {
+    log "Waiting for builder manager to finish the build"
+    
+    while true; do
+      kubectl --namespace default get packages $1 -o jsonpath='{.status.buildstatus}'|grep succeeded
```
erwinvaneyk: use `kubectl wait` instead of this custom wait/retry functionality

test/tests/mqtrigger/kafka/test_kafka.sh
vishal-biyani: I did this to make it work on mac (Like other scripts), but can be removed for now

vishal-biyani: This was a fallback for using log on Mac, can be removed though.

test/tests/mqtrigger/kafka/kafka_pub/kafka-pub.go
vishal-biyani: Agreed, will send a separate PR.

test/tests/mqtrigger/kafka/kafka_pub/kafka-pub.go
vishal-biyani: Fixed!

test/tests/mqtrigger/kafka/test_kafka.sh
vishal-biyani: Wait would mean we have to constantly watch the IO, would try implementing in another PR

PR #956, opened 2018-10-30
Update influxdb image

vannrt:
The old image tutum/influxdb does not exist on DockerHub anymore.
Update template to point to the latest official influxdb image.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/956)

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -386,7 +386,7 @@ spec:
     spec:
       containers:
       - name: influxdb
-        image: tutum/influxdb
+        image: influxdb
         env:
           - name: PRE_CREATE_DB
             value: fissionFunctionLog


vannrt:
The tests failed because of this error: `helm: command not found`

soamvasani:
Thanks, @vannrt ! 

@life1347 would you mind testing this?  Could be urgent given the old image is gone from dockerhub.

life1347:
After testing, the official influxdb image is not compatible with fission deployment and require some changes. Before we fully adopting official image and for backward compatibility issue, I've pushed a influxdb image `fission/influxdb` to fission docker repo and will open a PR later to fix charts issue.

life1347:
Open PR https://github.com/fission/fission/pull/957 to fix the issue instead. 
Thanks @vannrt for reporting the issue :)

Issue #707, opened 2018-05-29
http: proxy error: http: invalid Read on closed Body


soamvasani:
From slack user thenamly:

Fission workflows call the function and fail with router returning 502. This is sometimes happening but not always:
```
Invoking Fission function: 'http://router.fission/fission-function/datainput'." component=fnenv.fission fn="runtime:\"fission\" ID:\"datainput\" "
[datainput][status]: 502" component=fnenv.fission fn="runtime:\"fission\" ID:\"datainput\" "
[datainput][Content-Type]:  " component=fnenv.fission fn="runtime:\"fission\" ID:\"datainput\" "
[datainput][output]: ''" component=fnenv.fission fn="runtime:\"fission\" ID:\"datainput\" "
msg="[%!s(int=502)] Failed : %!v(MISSING)" component=fnenv.fission fn="runtime:\"fission\" ID:\"datainput\" "
```
Then rest of the invocation completely fail:
```
"Backend client appending event." aggregate.id=DataInput aggregate.type=function event.id= event.type=TASK_FAILED nats.Subject=invocation.wi-bb481bba-9470-46a1-b5d8-8be973e688db
"Handling invocation notification!" component=controller-wi labels="map[aggregate.type:function parent.type:invocation parent.id:wi-bb481bba-9470-46a1-b5d8-8be973e688db event.id:5 event.type:TASK_FAILED aggregate.id:DataInput]" notification=TASK_FAILED
msg="Scheduler evaluating..." component=scheduler wfi=wi-bb481bba-9470-46a1-b5d8-8be973e688db
msg="Applying action: fail" component=controller-wi
msg="Backend client appending event." aggregate.id=wi-bb481bba-9470-46a1-b5d8-8be973e688db aggregate.type=invocation event.id= event.type=INVOCATION_FAILED nats.Subject=invocation.wi-bb481bba-9470-46a1-b5d8-8be973e688db
msg="Handling invocation notification!" component=controller-wi labels="map[aggregate.id:wi-bb481bba-9470-46a1-b5d8-8be973e688db aggregate.type:invocation parent.type: parent.id: event.id:6 event.type:INVOCATION_FAILED]" notification=INVOCATION_FAILED
I  10.16.0.16 - - [28/May/2018:11:40:30 +0000] "POST / HTTP/1.1" 500 7
msg="Invocation not successful, was 'FAILED'"
 ```
fission router gives these two after that:
```
[...]

PR #963, opened 2018-10-31
Fix Read on Closed body error 

smruthi2187:
Issue : https://github.com/fission/fission/issues/707

### DIFF
--- a/router/functionHandler.go
+++ b/router/functionHandler.go
@@ -168,6 +168,14 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 	executingTimeout := roundTripper.funcHandler.tsRoundTripperParams.timeout
 
+	var reqBody []byte
+	if req.Body != nil {
+		reqBody, err = ioutil.ReadAll(req.Body)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	for i := 0; i < roundTripper.funcHandler.tsRoundTripperParams.maxRetries-1; i++ {
 
 		// cache lookup to get serviceUrl
@@ -216,6 +224,9 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 			if serviceUrlFromCache {
 				go roundTripper.funcHandler.tapService(serviceUrl)
 			}
+
+			// reset the body before every retry
+			req.Body = ioutil.NopCloser(bytes.NewReader(reqBody))
 
 			// forward the request to the function service
 			resp, err = transport.RoundTrip(req)


router/functionHandler.go
```diff
@@ -217,6 +225,9 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 				go roundTripper.funcHandler.tapService(serviceUrl)
 			}
 
+			// reset the body before every retry
+			req.Body = ioutil.NopCloser(bytes.NewReader(reqBody))
```
soamvasani: Using nopcloser is a great idea!  But can we directly wrap the req body in a nopcloser instead of wrapping a buffer?

Also, what closes the request body now?  Does it never get closed?  Is that a problem?

router/functionHandler.go
```diff
@@ -168,6 +168,14 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 	executingTimeout := roundTripper.funcHandler.tsRoundTripperParams.timeout
 
+	var reqBody []byte
+	if req.Body != nil {
+		reqBody, err = ioutil.ReadAll(req.Body)
```
soamvasani: This puts the full request in the router's memory, which has a couple of downsides -- router memory usage increases (req body could be large), and it prevents us from supporting streaming (starting a function before the full request has come in).

router/functionHandler.go
```diff
@@ -89,32 +90,24 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
-// RoundTrip is a custom transport with retries for http requests that forwards the request to the right serviceUrl, obtained
```
soamvasani: This comment looks useful, why remove it?

router/functionHandler.go
```diff
@@ -168,6 +161,11 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (resp *htt
 
 	executingTimeout := roundTripper.funcHandler.tsRoundTripperParams.timeout
 
+	// wrap the req.Body with another ReadCloser interface.
+	if req.Body != nil {
+		req.Body = &fakeCloseReadCloser{req.Body}
+	}
```
soamvasani: Can we do `defer req.Body....RealClose()` here instead of calling it each place that returns?

life1347:
LGTM, but I think we should add whole [removed comments](https://github.com/fission/fission/blob/3676602fa233ec7ccffbfc2dd4944f3fd9b23da6/router/functionHandler.go#L92-L117) back since its quite useful. After rebasing master and adding comments, the PR is ready to merge.

PR #964, opened 2018-10-31
Keep prometheus and canary deploy set to false in fission-core

smruthi2187:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/964)


soamvasani:
LGTM, this makes sense, fission-core should be as small as possible.

PR #969, opened 2018-11-03
Properly render Helm charts

sdake:
Fixes: https://github.com/fission/fission/issues/968

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -401,8 +401,9 @@ spec:
                 name: influxdb
                 key: password
 
----
 {{- if .Values.heapster }}
+
+---
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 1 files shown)


life1347:
Hi @sdake thanks for the PR! After applying the code changes and run the command again, it generate another issue like following.
```
              secretKeyRef:
                name: influxdb
                key: password---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  name: logger
```

After some invesgating, I think the root cause is `{{- end -}}` at L446
https://github.com/fission/fission/blob/585d86448f9437b99995f6f08cb139b74a9f6b5f/charts/fission-all/templates/deployment.yaml#L446

Helm removes all following space when it sees `{{- end -}}`. Remove the second `-` should be able to fix the problem. Would you mind to address it? thanks!

sdake:
@life1347 I have made a minor modification to the suggestion you made.  The trailing stripping requested by the helm template's gotpl isn't correct.  Tthere was one other minor syntax problem with the output of the manifest.  Taking a look overall at the rendered manifests holistically, there are many yaml syntax errors.  Fortunately Kubernetes is fairly forgiving of yaml syntax problems.

Cheers
-steve

life1347:
LGTM. And yes, I also noticed some minor syntax problem in rendered manifests and will take a look to fix them.

Issue #955, opened 2018-10-29
Newdeploy should scales deployment instead of creating it when minscale is 0


life1347:
Now if a function's minScale is 0, newdeploy only creates deployment/service/hpa when the first request come to the function. It increases the cold-start time for newdeploy.

With PR #775 , newdeploy are able to scale deployment from 0. 

https://github.com/fission/fission/blob/29aabaabda23c50fb7c22240d45e745c9ece7945/executor/newdeploy/newdeploy.go#L61-L62

To shorten the cold-start time, I propse to change how newdeploy acts when a function's minScale is 0. Newdeploy creates all k8s objects no matter what value of minScale is given, then scales the deployment to 1 when the first request comes to router.

In this way, we can
1. Remove legacy code & unify behavior of newdeploy 
2. Shorten cold-start time

--
(UPDATE): set default minscale to 1 in CLI

PR #970, opened 2018-11-05
Fix issues when specifying resources/scales during updating/creation process

life1347:
To fix part of https://github.com/fission/fission/issues/955 (minscale to 1) and some other issues when specifying resources/scales during updating/creation process.

1. default minscale to 1
2. fix failed to update min/max resources in one time.
3. fix user can create env with improper req/limits resource.
4. add `[FATAL]` to CLI log message

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/970)

### DIFF
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -323,10 +323,17 @@ func getResourceReq(c *cli.Context, resources v1.ResourceRequirements) v1.Resour
 		limitResources[v1.ResourceMemory] = memLimit
 	}
 
-	if limitResources[v1.ResourceCPU].Cmp(requestResources[v1.ResourceCPU]) < 0 {
+	limitCPU := limitResources[v1.ResourceCPU]
+	requestCPU := requestResources[v1.ResourceCPU]
+
+	if limitCPU.Cmp(requestCPU) < 0 {
 		log.Fatal(fmt.Sprintf("MinCPU (%v) cannot be greate than MaxCPU (%v)", limitResources[v1.ResourceCPU], requestResources[v1.ResourceCPU]))
 	}
-	if limitResources[v1.ResourceMemory].Cmp(requestResources[v1.ResourceMemory]) < 0 {
+
+	limitMem := limitResources[v1.ResourceMemory]
+	requestMem := requestResources[v1.ResourceMemory]
+
+	if limitMem.Cmp(requestMem) < 0 {
 		log.Fatal(fmt.Sprintf("MinMemory (%v) cannot be greate than MaxMemory (%v)", limitResources[v1.ResourceMemory], requestResources[v1.ResourceMemory]))
 	}
 
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -330,7 +330,7 @@ func getResourceReq(c *cli.Context, resources v1.ResourceRequirements) v1.Resour
 	if limitCPU.IsZero() {
 		limitResources[v1.ResourceCPU] = requestCPU
 	} else if limitCPU.Cmp(requestCPU) < 0 {
-		log.Fatal(fmt.Sprintf("MinCPU (%v) cannot be greate than MaxCPU (%v)", requestCPU.String(), limitCPU.String()))
+		log.Fatal(fmt.Sprintf("MinCPU (%v) cannot be greater than MaxCPU (%v)", requestCPU.String(), limitCPU.String()))
 	}
 
 	limitMem := limitResources[v1.ResourceMemory]
@@ -339,7 +339,7 @@ func getResourceReq(c *cli.Context, resources v1.ResourceRequirements) v1.Resour
 	if limitMem.IsZero() {
 		limitResources[v1.ResourceMemory] = requestMem
 	} else if limitMem.Cmp(requestMem) < 0 {
-		log.Fatal(fmt.Sprintf("MinMemory (%v) cannot be greate than MaxMemory (%v)", requestMem.String(), limitMem.String()))
+		log.Fatal(fmt.Sprintf("MinMemory (%v) cannot be greater than MaxMemory (%v)", requestMem.String(), limitMem.String()))
 	}
 
 	resources = v1.ResourceRequirements{
--- a/fission/environment.go
+++ b/fission/environment.go
@@ -219,7 +219,7 @@ func envUpdate(c *cli.Context) error {
 	env.Spec.AllowAccessToExternalNetwork = envExternalNetwork
 
 	if c.IsSet("mincpu") || c.IsSet("maxcpu") || c.IsSet("minmemory") || c.IsSet("maxmemory") || c.IsSet("minscale") || c.IsSet("maxscale") {
-		log.Fatal("Currently it's not support to update resources limits/requests after creating the environment, please recreate it instead.")
+		log.Fatal("Updating resource limits/requests for existing environments is currently unsupported; re-create the environment instead.")
 	}
 
 	_, err = client.EnvironmentUpdate(env)
@@ -299,6 +299,7 @@ func getResourceReq(c *cli.Context, resources v1.ResourceRequirements) v1.Resour
 	}
 
 	var limitResources map[v1.ResourceName]resource.Quantity
+
 	if len(resources.Limits) == 0 {
 		limitResources = make(map[v1.ResourceName]resource.Quantity)
 	} else {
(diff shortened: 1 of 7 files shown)


fission/environment.go
```diff
@@ -218,6 +218,10 @@ func envUpdate(c *cli.Context) error {
 
 	env.Spec.AllowAccessToExternalNetwork = envExternalNetwork
 
+	if c.IsSet("mincpu") || c.IsSet("maxcpu") || c.IsSet("minmemory") || c.IsSet("maxmemory") || c.IsSet("minscale") || c.IsSet("maxscale") {
+		log.Fatal("Currently it's not support to update resources limits/requests after creating the environment, please recreate it instead.")
```
soamvasani: 1. No logs in the CLI -- it should simply be a printf and exit.

2. "Updating resource limits/requests for existing environments is currently unsupported; re-create the environment instead.  (Issue NNN)".

And let's file an issue to track allowing these updates.

fission/environment.go
```diff
@@ -319,6 +323,20 @@ func getResourceReq(c *cli.Context, resources v1.ResourceRequirements) v1.Resour
 		limitResources[v1.ResourceMemory] = memLimit
 	}
 
+	limitCPU := limitResources[v1.ResourceCPU]
+	requestCPU := requestResources[v1.ResourceCPU]
+
+	if limitCPU.Cmp(requestCPU) < 0 {
+		log.Fatal(fmt.Sprintf("MinCPU (%v) cannot be greate than MaxCPU (%v)", limitResources[v1.ResourceCPU], requestResources[v1.ResourceCPU]))
```
soamvasani: The problem here is if that the user specifies a request but not a limit, this errors out.  How about just setting max=min if max is zero.

fission/function.go
```diff
@@ -71,6 +71,11 @@ func printPodLogs(c *cli.Context) error {
 
 func getInvokeStrategy(minScale int, maxScale int, executorType string, targetcpu int) fission.InvokeStrategy {
 
+	// set default minScale to 1
+	if minScale == 0 {
```
soamvasani: With this change, how will a user specify minScale = 0?  If you want to default to zero you have to do it where you have the cli object, then you can check `c.IsSet("minscale")` or something like that to see if the user specified a minscale

fission/function.go
```diff
@@ -207,12 +212,14 @@ func fnCreate(c *cli.Context) error {
 	}
 
 	invokeStrategy := getInvokeStrategy(c.Int("minscale"), c.Int("maxscale"), c.String("executortype"), getTargetCPU(c))
-	resourceReq := getResourceReq(c, apiv1.ResourceRequirements{})
+
 	if (c.IsSet("mincpu") || c.IsSet("maxcpu") || c.IsSet("minmemory") || c.IsSet("maxmemory")) &&
 		invokeStrategy.ExecutionStrategy.ExecutorType == fission.ExecutorTypePoolmgr {
-		log.Warn("CPU/Memory specified for function with pool manager executor will be ignored in favor of resources specified at environment")
+		log.Fatal("To limit CPU/Memory for function with executor type \"poolmgr\", please specify resources limits when creating environment")
```
soamvasani: I think a warning is better here.  We can still execute the function, we just can't provide the requested resources.  Also, in the future if K8s supports changing resources on existing pods, we could potentially support this with poolmgr.

fission/function.go
```diff
@@ -527,50 +534,54 @@ func fnUpdate(c *cli.Context) error {
...
 		case fission.ExecutorTypeNewdeploy:
 			fnExecutor = fission.ExecutorTypeNewdeploy
 		default:
 			log.Fatal("Executor type must be one of 'poolmgr' or 'newdeploy', defaults to 'poolmgr'")
 		}
-		if (c.IsSet("mincpu") || c.IsSet("maxcpu") || c.IsSet("minmemory") || c.IsSet("maxmemory")) &&
+		if (c.IsSet("mincpu") || c.IsSet("maxcpu") || c.IsSet("minmemory") || c.IsSet("maxmemory") || c.IsSet("minscale") || c.IsSet("maxscale")) &&
 			fnExecutor == fission.ExecutorTypePoolmgr {
-			log.Warn("CPU/Memory specified for function with pool manager executor will be ignored in favor of resources specified at environment")
+			log.Fatal("To limit CPU/Memory for function with executor type \"poolmgr\", please specify resources limits when creating environment")
```
soamvasani: Same comment, warning is better.

fission/log/log.go
```diff
@@ -27,7 +27,7 @@ var (
 )
 
 func Fatal(msg interface{}) {
-	os.Stderr.WriteString(fmt.Sprintf("%v\n", msg))
+	os.Stderr.WriteString(fmt.Sprintf("[FATAL] %v\n", msg))
```
soamvasani: These messages are not logs; they're meant for humans.  So something like `"Fatal error: ..."` is better than `"[FATAL] ..."`

fission/environment.go
life1347: 1. It's not the built-in log package, it prints message and exit. (fission/fission/log)
issue to track resource change to env: https://github.com/fission/fission/issues/972

fission/function.go
```diff
@@ -69,40 +74,87 @@ func printPodLogs(c *cli.Context) error {
 	return nil
 }
 
-func getInvokeStrategy(minScale int, maxScale int, executorType string, targetcpu int) fission.InvokeStrategy {
-
-	if maxScale == 0 {
```
soamvasani: I'm confused about this old code.  Were we preventing users from setting minScale=0?

fission/function.go
```diff
@@ -69,40 +74,87 @@ func printPodLogs(c *cli.Context) error {
...
+		strategy = &fission.InvokeStrategy{
+			StrategyType: fission.StrategyTypeExecution,
+			ExecutionStrategy: fission.ExecutionStrategy{
+				ExecutorType: fission.ExecutorTypePoolmgr,
+			},
+		}
+	} else {
+		// set default value
+		targetCPU := DEFAULT_TARGET_CPU_PERCENTAGE
+		minScale := DEFAULT_MIN_SCALE
```
soamvasani: Ok, I see, just making sure I understand:

1. If user sets nothing except executor=newdeploy, minscale=1
2. if users sets executor=newdeploy and minscale=0, they get minscale=0

And if executor is not set, it defaults to poolmgr.

So I was wondering: if minscale=1, and executor is not specified, should we default executor=newdeploy?  Or is that too confusing?

fission/main.go
```diff
@@ -74,13 +78,13 @@ func main() {
...
-	minCpu := cli.StringFlag{Name: "mincpu", Usage: "Minimum CPU to be assigned to pod (In millicore, minimum 1)"}
-	maxCpu := cli.StringFlag{Name: "maxcpu", Usage: "Maximum CPU to be assigned to pod (In millicore, minimum 1)"}
-	minMem := cli.StringFlag{Name: "minmemory", Usage: "Minimum memory to be assigned to pod (In megabyte)"}
-	maxMem := cli.StringFlag{Name: "maxmemory", Usage: "Maximum memory to be assigned to pod (In megabyte)"}
-	minScale := cli.StringFlag{Name: "minscale", Usage: "Minimum number of pods (Uses resource inputs to configure HPA)"}
-	maxScale := cli.StringFlag{Name: "maxscale", Usage: "Maximum number of pods (Uses resource inputs to configure HPA)"}
-	targetcpu := cli.IntFlag{Name: "targetcpu", Value: 80, Usage: "Target average CPU usage percentage across pods for scaling"}
+	minCpu := cli.IntFlag{Name: "mincpu", Usage: "Minimum CPU to be assigned to pod (In millicore, minimum 1)"}
+	maxCpu := cli.IntFlag{Name: "maxcpu", Usage: "Maximum CPU to be assigned to pod (In millicore, minimum 1)"}
+	minMem := cli.IntFlag{Name: "minmemory", Usage: "Minimum memory to be assigned to pod (In megabyte)"}
```
soamvasani: Hmm, I have a slight preference for the strings because the user can specify the units instead of checking the help to see which units are needed.  Also if no units are specified you could always default to megabytes.

life1347: I think we should use IntFalg instead of StringFlag for couple reasons
1. We use `Int` to get value internally. And since the flags are public for a long time, if we change the input format from number to string it may break users code.
https://github.com/fission/fission/blob/b456dec1385838279a0917b6542b9e0416a6d796/fission/environment.go#L280-L285
2. For most of users, they don't actually know what is the correct unit to specify with, also kubernetes provides to many way to specify the resource unit and may make more confusion to users. For example, cpu can be represent as`100m` or `0.1` while memory is `Ei, Pi, Ti, Gi, Mi, Ki`. It also brings difficulity for CLI to validate what is a valid value for resource.

fission/function.go
life1347: no we should not do that, its too confusing. I add a warning message here to tell users to specify executor type explicitly instead.
https://github.com/fission/fission/pull/970/files#diff-17e3b56f4e0c4afc96fc4653126df7f8R105

PR #971, opened 2018-11-05
Show builder image when list all envs

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/971)


vishal-biyani:
LGTM

PR #975, opened 2018-11-07
Shorten poolmgr deployment name

life1347:
PR https://github.com/fission/fission/pull/182 is a little bit old, create a new one instead.

The field `GenerateName` will cause newdeploy not able to find the correct hpa, svc and deploy to update/delete due to 3 resource's suffix are different. 

Since the original purpose of the PR is to shorten the name of deployment, so use `Name` but without `gp.env.Metadata.UID` in pool deployment name. Also, add environment information to annotation for troubleshooting when dumping the CRDs.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/975)

### DIFF
--- a/executor/newdeploy/newdeploy.go
+++ b/executor/newdeploy/newdeploy.go
@@ -203,8 +203,12 @@ func (deploy *NewDeploy) getDeploymentSpec(fn *crd.Function, env *crd.Environmen
 
 	deployment := &v1beta1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels: deployLabels,
 			Name:   deployName,
+			Labels: deployLabels,
+			Annotations: map[string]string{
+				"env-name": env.Metadata.Name,
+				"env-uuid": string(env.Metadata.UID),
+			},
 		},
 		Spec: v1beta1.DeploymentSpec{
 			Replicas: &replicas,
@@ -405,9 +409,8 @@ func (deploy *NewDeploy) createOrGetHpa(hpaName string, execStrategy *fission.Ex
 	if err != nil && k8s_err.IsNotFound(err) {
 		hpa := asv1.HorizontalPodAutoscaler{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      hpaName,
-				Namespace: depl.ObjectMeta.Namespace,
-				Labels:    depl.Labels,
+				Name:   hpaName,
+				Labels: depl.Labels,
 			},
 			Spec: asv1.HorizontalPodAutoscalerSpec{
 				ScaleTargetRef: asv1.CrossVersionObjectReference{
--- a/executor/newdeploy/newdeploy.go
+++ b/executor/newdeploy/newdeploy.go
@@ -205,10 +205,6 @@ func (deploy *NewDeploy) getDeploymentSpec(fn *crd.Function, env *crd.Environmen
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   deployName,
 			Labels: deployLabels,
-			Annotations: map[string]string{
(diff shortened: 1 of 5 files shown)


executor/newdeploy/newdeploy.go
```diff
@@ -203,8 +203,12 @@ func (deploy *NewDeploy) getDeploymentSpec(fn *crd.Function, env *crd.Environmen
 
 	deployment := &v1beta1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels: deployLabels,
 			Name:   deployName,
+			Labels: deployLabels,
+			Annotations: map[string]string{
+				"env-name": env.Metadata.Name,
```
soamvasani: we have "envName" defined as a constant somewhere?  can we either use that or define a new constant

executor/newdeploy/newdeploymgr.go
```diff
@@ -548,9 +548,7 @@ func (deploy *NewDeploy) fnDelete(fn *crd.Function) (*fscache.FuncSvc, error) {
 }
 
 func (deploy *NewDeploy) getObjName(fn *crd.Function) string {
-	return fmt.Sprintf("%v-%v",
-		fn.Metadata.Name,
-		deploy.instanceID)
+	return strings.ToLower(fmt.Sprintf("newdeploy-%v-%v", strings.ToLower(fn.Metadata.Name), deploy.instanceID))
```
soamvasani: Why add "newdeploy"?

executor/poolmgr/gp.go
```diff
@@ -459,8 +459,7 @@ func (gp *GenericPool) specializePod(pod *apiv1.Pod, metadata *metav1.ObjectMeta
 // A pool is a deployment of generic containers for an env.  This
 // creates the pool but doesn't wait for any pods to be ready.
 func (gp *GenericPool) createPool() error {
-	poolDeploymentName := fmt.Sprintf("%v-%v-%v",
-		gp.env.Metadata.Name, gp.env.Metadata.UID, strings.ToLower(gp.poolInstanceId))
+	poolDeploymentName := strings.ToLower(fmt.Sprintf("poolmgr-%v-%v", gp.env.Metadata.Name, gp.poolInstanceId))
```
soamvasani: Does this work if two different namespaces have an env with the same name?

executor/newdeploy/newdeploymgr.go
soamvasani: [Resolved because I noticed you added it for poolmgr as well, sounds good to me.]

executor/newdeploy/newdeploymgr.go
```diff
@@ -547,19 +548,19 @@ func (deploy *NewDeploy) fnDelete(fn *crd.Function) (*fscache.FuncSvc, error) {
 }
 
 func (deploy *NewDeploy) getObjName(fn *crd.Function) string {
-	return fmt.Sprintf("%v-%v",
-		fn.Metadata.Name,
-		deploy.instanceID)
+	return strings.ToLower(fmt.Sprintf("newdeploy-%v-%v-%v", fn.Metadata.Name, fn.Metadata.Namespace, deploy.instanceID))
```
soamvasani: I'm just nitpicking, but an object "c" in namespace "a-b" will have the same name as an object "b-c" in namespace "a".  You could solve it by adding a small random string to the name.

executor/util/util.go
```diff
@@ -23,6 +23,16 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+const (
```
soamvasani: "util" packages tend to just grow without limit...  can these constants be simply placed in executor.go or something like that?

PR #976, opened 2018-11-07
Pre-create kubernetes resources for function with minScale=0

### LINKED ISSUE #955: Newdeploy should scales deployment instead of creating it when minscale is 0

life1347:
Fix #955

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/976)

### DIFF
--- a/executor/newdeploy/newdeploymgr.go
+++ b/executor/newdeploy/newdeploymgr.go
@@ -72,6 +72,7 @@ type (
 		reqType         requestType
 		fn              *crd.Function
 		responseChannel chan *fnResponse
+		precreate       bool
 	}
 
 	fnResponse struct {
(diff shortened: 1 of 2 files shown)


executor/newdeploy/newdeploymgr.go
```diff
@@ -384,7 +384,7 @@ func (deploy *NewDeploy) fnUpdate(oldFn *crd.Function, newFn *crd.Function) {
 		if oldFn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType != fission.ExecutorTypeNewdeploy &&
 			newFn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType == fission.ExecutorTypeNewdeploy {
 			log.Printf("function type changed to new deployment, creating resources: %v", newFn)
-			_, err := deploy.fnCreate(newFn)
+			_, err := deploy.fnCreate(newFn, false)
```
vishal-biyani: Should we not precreate in this case since the function is changing type from poolmanager to newdeploy?

vishal-biyani:
LGTM

Issue #606, opened 2018-04-04
Trap aggregation in test scripts


vishal-biyani:
Some tests currently call `trap` more than once, but the way `trap` behaves only last one is effective ([Check this for reference](https://unix.stackexchange.com/questions/170409/bash-issuing-trap-twice)). We should have a `cleanup()` function which is called as a trap and does all the cleanup activity.

soamvasani:
I had no idea!  So many of our tests do this 😱

life1347:
Solved by #1128

PR #986, opened 2018-11-13
Mac test utility

vishal-biyani:
A small util for enabling testing on local dev machines specifically for Mac.

We can extend this to all test scripts at some point.


vishal-biyani:
All scripts now have a cleanup function and that is called by `trap` which fixes issue https://github.com/fission/fission/issues/606

soamvasani:
This is old but I finally got around to reviewing -- LGTM, thanks Vishal!

Issue #954, opened 2018-10-27
Add new interface to router to simplify svcAddrUpdateLock operation


life1347:
https://github.com/fission/fission/pull/856#pullrequestreview-169003755

> I see what you're doing with this, but I'm wondering if it would be simpler to have an interface like RunOnceOrWait(key string, handler func) which runs handler only once for each value of key and waits for func if it is already running. Then you wont have to return locks across the interface, ableToUpdate flags, etc. Right? [This is a bigger change and we're close to the release so it's fine to do it later, assuming the idea makes sense.]

PR #991, opened 2018-11-15
Refactor RoundTrip function for code reading

life1347:
Refactor RoundTrip function into multiple small functions for better code reading and understanding.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/991)


soamvasani:
Still reading this, but while you're here do you want to try addressing #954 as well?

PR #996, opened 2018-11-19
Added warning to fix #946

### LINKED ISSUE #946: Old package with new function causes function to refer to old env
Create a package with an environment

```
$fission package create --env python --deploy "multifile/*"
Package 'python-1540268954-9jdn' created
```

Now create a new environment and a function which wants to use this package:

```
$ fission env create --name pythonb --image fission/python-env
environment 'pythonb' created
```

```
$ fission fn create --name py98 --env pythonb --pkg python-1540268954-9jdn --entrypoint main.main --executortype newdeploy --minscale 1
function 'py98' created

$ fission fn list
NAME UID                                  ENV    EXECUTORTYPE MINSCALE MAXSCALE MINCPU MAXCPU MINMEMORY MAXMEMORY TARGETCPU
py98 8a25984c-d685-11e8-9b4e-42010aa0001c python newdeploy    1        1        0      0      0         0         80

```

The function still refers to environment `python` because package refers to old environment. Ideally the function should point to `pythonb` environment.

vishal-biyani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/996)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -502,7 +502,7 @@ func fnUpdate(c *cli.Context) error {
 			log.Fatal("Package is used by multiple functions, use --force to force update")
 		}
 
-		pkgMetadata, err = updatePackage(client, pkg, envName, envNamespace, srcArchiveName, deployArchiveName, buildcmd, false)
+		pkgMetadata, err = updatePackage(client, pkg, pkg.Spec.Environment.Name, pkg.Spec.Environment.Namespace, srcArchiveName, deployArchiveName, buildcmd, false)
 		util.CheckErr(err, fmt.Sprintf("update package '%v'", pkgName))
 
 		fmt.Printf("package '%v' updated\n", pkgMetadata.GetName())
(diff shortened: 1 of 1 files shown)


fission/function.go
```diff
@@ -165,6 +165,9 @@ func fnCreate(c *cli.Context) error {
 		util.CheckErr(err, fmt.Sprintf("read package in '%v' in Namespace: %s. Package needs to be present in the same namespace as function", pkgName, fnNamespace))
 		pkgMetadata = &pkg.Metadata
 		envName = pkg.Spec.Environment.Name
+		if envName != c.String("env") {
```
life1347: We should add same code to fnUpdate as well

vishal-biyani: @life1347 - for fnUpdate this is not needed and in fact the behavior there is different but that is for another issue!

fission/function.go
```diff
@@ -525,6 +528,12 @@ func fnUpdate(c *cli.Context) error {
 		ResourceVersion: pkgMetadata.ResourceVersion,
 	}
 
+	if function.Spec.Environment.Name != pkg.Spec.Environment.Name {
+		log.Warn("Function's environment is different than package's environment, package's environment will be used for creating function")
```
life1347: ... for updating function

Issue #940, opened 2018-10-17
Canary config spec: rename `funcn` and `funcn-1` 


soamvasani:
The names "funcn" and "funcn-1" are a bit unusual -- for one, we shouldn't abbreviate "function" to "func", and secondly it isn't clear that we're talking about versions.  Suggested better names:

oldVersion / newVersion
oldFunction / newFunction
stableFunction / unstableFunction

smruthi2187:
Ok. however, changing the spec now will lead to a problem. 
if there is an existing canary config created (from 0.11.0 release) on the cluster and now someone upgrades fission to new version( which has new fields ), then canary deployment will not work.  
do you suggest we still go ahead and make the change?

soamvasani:
Since it's a very very new feature we could just break compatibility.  For what it's worth, it would be pretty unusual to upgrade Fission in the middle of doing an actual canary deployment.

smruthi2187:
ok 👍

PR #1003, opened 2018-11-26
Rename canary flag name from funcN/funcN-1 to newfunc/oldfunc

life1347:
Address #940

### DIFF
--- a/fission/main.go
+++ b/fission/main.go
@@ -277,13 +277,13 @@ func main() {
 	// canary configs
 	canaryConfigNameFlag := cli.StringFlag{Name: "name", Usage: "Name for the canary config"}
 	triggerNameFlag := cli.StringFlag{Name: "httptrigger", Usage: "Http trigger that this config references"}
-	funcNFlag := cli.StringFlag{Name: "funcN", Usage: "New version of the function"}
-	funcNminus1Flag := cli.StringFlag{Name: "funcN-1", Usage: "Old stable version of the function"}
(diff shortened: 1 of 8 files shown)


pkg/apis/fission.io/v1/typefields.go
```diff
@@ -338,8 +338,8 @@ type (
 	// Canary Config Spec
 	CanaryConfigSpec struct {
 		Trigger                 string      `json:"trigger"`
-		FunctionN               string      `json:"funcn"`
-		FunctionNminus1         string      `json:"funcn-1"`
+		NewFunction             string      `json:"newfunc"`
```
soamvasani: can we change `newfunc` to `newfunction` in the json?

fission/main.go
```diff
@@ -277,13 +277,13 @@ func main() {
 	// canary configs
 	canaryConfigNameFlag := cli.StringFlag{Name: "name", Usage: "Name for the canary config"}
 	triggerNameFlag := cli.StringFlag{Name: "httptrigger", Usage: "Http trigger that this config references"}
-	funcNFlag := cli.StringFlag{Name: "funcN", Usage: "New version of the function"}
-	funcNminus1Flag := cli.StringFlag{Name: "funcN-1", Usage: "Old stable version of the function"}
+	newFunc := cli.StringFlag{Name: "newfunc", Usage: "New version of the function"}
```
soamvasani: Can we use the full word "function" in the cli?  (we do this for the triggers too)

PR #1005, opened 2018-11-26
[WIP] Fix requests are sent to unready function pod (newdeploy)

### LINKED ISSUE #998: Refactor newdeploy function pod specialization process
The executor creates deployment for function with executor type newdeploy and return service url once number of replicas is equal or higher than minScale. However, in some cases (e.g. large binary) the function pod is still in specializing process and it's not ready for serving, causing requests failure.

Kubernetes provides readiness probe to detect whether a pod is ready to serve requests. To adopt the mechanism, fetcher needs a add new HTTP API that returns correspond status code base on status flag of specialization.

life1347:
Fix #998


vishal-biyani:
LGTM after rebase

Issue #951, opened 2018-10-24
Spec command should check existence of a resource in spec dir before writing to yaml file


life1347:
1. duplicate pkg declaration 
```
$ fission function create --spec --name calc-eval --env python --src demo-src-pkg.zip --entrypoint calc.main
$ fission function create --spec --name calc-form --env python --src demo-src-pkg.zip --entrypoint form.main

$ fission spec apply --wait
Failed to read specs: 2 errors occurred:
	* specs/function-calc-form.yaml:1: Duplicate ArchiveUploadSpec 'demo-src-pkg-zip', first defined in specs/function-calc-eval.yaml:1
	* specs/pkg.yaml:1: Duplicate ArchiveUploadSpec 'demo-src-pkg-zip', first defined in specs/function-calc-eval.yaml:1
```

2. If `--spec` is specified, should we check local package first?
```
$ fission function create --spec --name calc-form --env python --pkg python-pkg-zip --entrypoint form.main
Failed to read package in 'python-pkg-zip' in Namespace: default. Package needs to be present in the same namespace as function: Resource not found - NotFound
```

PR #1018, opened 2018-12-05
handle duplicate archive and package specs; handle multifile archives better

soamvasani:
3 changes:

1. Archives with multiple files were being handled with only one input file (when multiple --src flags were provided)
2. Avoid creating a zip file when we don't need to
3. Check for duplicate ArchiveUploadSpecs and Package.Specs before writing them to the spec dir 

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1018)

### DIFF
--- a/common.go
+++ b/common.go
@@ -28,12 +28,10 @@ import (
 	"strings"
 	"syscall"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/gorilla/handlers"
 	"github.com/imdario/mergo"
 	"github.com/mholt/archiver"
-	uuid "github.com/satori/go.uuid"
+	"github.com/satori/go.uuid"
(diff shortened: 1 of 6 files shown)


soamvasani:
Fixes #951 and #1018

soamvasani:
Should be ready for review.

Known problems:

1. We read and parse the specs twice during the creation of a new function.  Only affects performance for now, I'll come back to this later and clean up
2. Package and archive naming could be improved.  Right now we just use the first file name and append a short unique string to it.
3. The duplicate check could be improved (different file order, comparison of equivalent globs, etc).  I'll come back to this later.  Right now since names are unique, at least the commands won't error out if we fail to detect two identical archives as duplicates.

life1347:
LGTM

PR #1032, opened 2018-12-19
Use Header.Set() to override the existing header value

### LINKED ISSUE #1027: Router should override the X-Fission-Function header values
Currently the `router/util.go` uses [`Header.Add()`](https://golang.org/pkg/net/http/#Header.Add) to add values to `X-Fission-Function-*` headers. 
https://github.com/fission/fission/blob/34b9be0bd33928a0682d65af19f392a307339d5a/router/util.go#L16-L19
If the request coming to function have these headers set already to some value, Fission will append own values to the Header key. This can break the things depending on `HeadersToMetadata` function as it uses [`Header.Get()`](https://golang.org/pkg/net/http/#Header.Get) which just returns first value.  
It would be better if we use [`Header.Set()`](https://golang.org/pkg/net/http/#Header.Set) to set the values so that the values for the Fission specific headers will always be correct ones.

Can be reproduced like this,
```python
from flask import request, make_response
import logging
logger = logging.getLogger('simple.fn')


def main():
    logger.info("received headers: %s", request.headers)
    print(request.headers)
    response = make_response("It worked!")
    response.headers = request.headers
    return response
```
```console
$ curl -v -H "X-Fission-Function-Namespace: bugspace" <loadBalancerIP>/simple-fn
*   Trying <loadBalancerIP>...
* TCP_NODELAY set
* Connected to <loadBalancerIP> (<loadBalancerIP>) port 80 (#0)
> GET /simple-fn HTTP/1.1
> Host: <loadBalancerIP>
> User-Agent: curl/7.61.1
> Accept: */*
> X-Fission-Function-Namespace: bugspace
> 
< HTTP/1.1 200 OK
< Accept: */*
< Accept-Encoding: gzip
< Content-Length: 10
< Forwarded: host=<loadBalancerIP>;
< Host: 10.36.8.25:8888
< User-Agent: curl/7.61.1
< X-Fission-Function-Name: simple
< X-Fission-Function-Namespace: bugspacedefault
< X-Fission-Function-Resourceversion: 1184133
< X-Fission-Function-Uid: 6eda599e-02ac-11e9-bc54-42010aa000ee
< X-Forwarded-For: 10.160.0.3
< X-Forwarded-Host: <loadBalancerIP>
< Date: Tue, 18 Dec 2018 10:38:54 GMT
< Content-Type: text/plain; charset=utf-8
< 
* Connection #0 to host <loadBalancerIP> left intact
It worked!
```

life1347:
Header.Add() appends string if there are same header key exists. Use Header.Set() to prevent this.
Fix #1027

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1032)

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -142,7 +142,7 @@ func (builder *Builder) reply(w http.ResponseWriter, pkgFilename string, buildLo
 		statusCode = http.StatusInternalServerError
 	}
 
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Set("Content-Type", "application/json")
 	// should write header before writing the body,
 	// or client will receive HTTP 200 regardless the real status code
 	w.WriteHeader(statusCode)
--- a/environments/fetcher/fetcher.go
+++ b/environments/fetcher/fetcher.go
@@ -482,7 +482,7 @@ func (fetcher *Fetcher) UploadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	log.Println("Completed upload request")
-	w.Header().Add("Content-Type", "application/json")
+	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
 	w.Write(rBody)
 }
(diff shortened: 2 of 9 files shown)


vishal-biyani:
Hi @life1347 Are there use cases where we might want to append headers and not set them?

life1347:
As far as I know, we don't have such use cases for appending headers.

PR #1039, opened 2018-12-30
[python-env] Add OpenTracing support

bhavin192:
This makes it possible to trace the calls made to Fission functions created with Python environment.
In addition to things discussed in #200 this makes it possible to trace things if the function is making calls to things which are outside the Fission platform (with the help of headers). Like some request to external API end point.

#### SUMMARY OF CHANGES
- Uses global tracer which won't be closed
  - As the `Tracer.close()` is not synchronous we can't create and close
    it for every request as it does not wait for the UDP sender to flush
    all the packets correctly. Instead use same tracer while function is
    alive is the better option for now
- Takes reporting_host and reporting_port from env variable
  - This needs `jaeger_client >= 3.13.0`
  - Uses the environment variable `JAEGER_AGENT_HOST` and
    `JAEGER_AGENT_PORT` to get the value
  - This must be set to correct value while creating the fission
    function environment
- Saves the `span` and `generated_headers` in `g`
  - This way user can use the span and set tags or use the
    `generated_headers` while sending external request from within the
    function code
- Decorates the `userfunc` in specialize call
  - If the value of environment variable `ENABLE_TRACING` is set then
    decorate the userfunc with `initialize_tracing` from `lib.tracing`

#### TODO
- [ ] Update the Python environment docs 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1039)

### DIFF
--- a/environments/python/lib/tracing.py
+++ b/environments/python/lib/tracing.py
@@ -1,0 +1,73 @@
+from jaeger_client import Config
+from flask import request, make_response, g
+from opentracing.propagation import Format
+import logging
+
+logger = logging.getLogger("lib.tracing")
+# Tracer is created globally and initialized only once because of
+# following bug.
+# https://github.com/jaegertpracing/jaeger-client-python/issues/50
+# We won't be calling tracer.close()
+tracer = None
+
+
+def initialize_tracing(func):
+    """
+    Decorator which initializes the tracing related stuff.
+    It creates tracer and a new span.
+    Also extracts and injects the headers required for Jaeger tracing
+    to work across multiple functions
+
+    :param func: Function object
+    :returns: decorated function
+    """
+    def inner():
+        global tracer
+        fission_func_name = request.headers.get("X-Fission-Function-Name",
+                                                "name")
+        span_name = fission_func_name + "-span"
+        if tracer is None:
+            tracer = _init_tracer(fission_func_name)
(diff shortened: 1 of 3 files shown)


soamvasani:
Hi @bhavin192, thanks for working on this!  Why implement this in the Python environment?  If you do this in the router (somewhere [in this neighbourhood](https://github.com/fission/fission/blob/master/router/functionHandler.go#L268)), we'll have function-level tracing for all languages instead of just Python.  

In general we like to implement as much as possible outside the environments and within the fission language-independent core, keeping environments simple so that most features apply to all languages.

vishal-biyani:
Hi @soamvasani The context of implementing this was specific to a customer need. They were passing a business event as message payload to the function. The business event had an event ID in UUID format which is what they wanted to use as trace ID. The event would pass through multiple functions and Kafka queues in between. The idea is that for a given event - you should be able to trace all functions that were executed in the tree. Also, the function code can add additional baggage and tags to the trace. 

Since we could not find an easy way to accommodate this in the router, we implemented this in the environment. Also, they plan to use the only Python so support for additional language was not an issue as well.

I agree that this is better placed in the router, I request @bhavin192 to make this change in the router and send an updated PR.

bhavin192:
@soamvasani @vishal-biyani yes, doing changes in router itself would be better thing to do. Closing this PR.

PR #1047, opened 2019-01-07
Replace router svcAddrUpdateLocks with new throttler package for code readability&reusability

life1347:
Address followup comments in #991 and add new throttler package.

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1047)

### DIFF
--- a/throttler/throttler.go
+++ b/throttler/throttler.go
@@ -40,9 +40,9 @@ type (
 		timeExpiry time.Duration
 	}
 
-	// Throttler is a simple throttling mechanism implementation provides
-	// ability to limit the total amount of requests to do the same thing
-	// at the same time.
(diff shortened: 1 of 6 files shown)


throttler/throttler.go
```diff
@@ -0,0 +1,229 @@
...
+type (
+	// actionLock is the lock that will be used when
+	// gorountine tries to update service address entry.
+	actionLock struct {
+		wg         *sync.WaitGroup
+		ctimestamp time.Time // creation time of lock
+		timeExpiry time.Duration
+	}
+
+	// Throttler is a simple throttling mechanism implementation provides
```
soamvasani: ... is a simple throttling mechanism that provides the ability to...

PR #1059, opened 2019-01-14
Bumped up default CPU for fetcher, fixes #1058

### LINKED ISSUE #1058: increase fetcher maxcpu
Via Slack -- our default fetcher maxcpu causes the unzip operation to be pretty slow.  Since this is the only time that fetcher uses any non-trivial CPU, it seems safe to bump up the default to about 1000m.

vishal-biyani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1059)


soamvasani:
Thanks, fission-core chart would need a change as well?

soamvasani:
LGTM, thanks

PR #1064, opened 2019-01-18
Update helm charts README

darkgerm:
The README for helm charts is out-of-date. Update it based on `fission-all/values.yaml` and `fission-core/values.yaml`.
An undocumented variable `persistence.existingClaim` is also added in both values.yaml.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1064)

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -62,8 +62,8 @@ Parameter | Description | Default
 `pruneInterval` | The frequency of archive pruner (in minutes) | `60`
 `preUpgradeChecksImage` | Fission pre-install/pre-upgrade checks live in this image | `fission/pre-upgrade-checks`
 `debugEnv` | If there are any pod specialization errors when a function is triggered and this flag is set to true, the error summary is returned as part of http response | `true`
-`prometheusDeploy` | Set to true if prometheus needs to be deployed along with fission | `true`
-`canaryDeployment.enabled` | Set to true if you need canary deployment feature | `true`
+`prometheusDeploy` | Set to true if prometheus needs to be deployed along with fission | `true` in `fission-all`, `false` in `fission-core`
+`canaryDeployment.enabled` | Set to true if you need canary deployment feature | `true` in `fission-all`, `false` in `fission-core`
 
 ### Extra configuration for `fission-all`
 
--- a/charts/README.md
+++ b/charts/README.md
@@ -85,7 +85,7 @@ Parameter | Description | Default
 `kafka.version` | Kafka broker version | `nil`
 `heapster` | Enable Heapster (only enable this in clusters where heapster does not exist already) | `false`
 
-* - Please note that deploying of Azure Storage Queue or Kafka is not done by Fission chart and you will have to explicitly deploy them.
+Please note that deploying of Azure Storage Queue or Kafka is not done by Fission chart and you will have to explicitly deploy them.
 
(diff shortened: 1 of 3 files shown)


charts/fission-all/values.yaml
```diff
@@ -80,14 +80,25 @@ kafka:
...
+  ## If true, fission will create/use a Persistent Volume Claim
+  ## If false, use emptyDir
+  ##
   enabled: true
+
+  ## A manually managed Persistent Volume Claim name
+  ## Requires persistence.enabled: true
+  ## If defined, PVC must be created manually before volume will be bound
+  ##
+  # existingClaim:
```
life1347: It would be better to copy default value to here for reference.

charts/README.md
```diff
@@ -37,34 +37,53 @@ $ helm delete my-release
...
+`natsStreamingPort` | Nats streaming service port | `31316`
+`azureStorageQueue.enabled` | Azure storage account name | `false`
+`azureStorageQueue.key` | Azure storage account name | `""`
+`azureStorageQueue.accountName` | Azure storage access key | `""`
+`kafka.enabled` | Kafka trigger enabled | `false`
+`kafka.brokers` | Kafka brokers uri | `broker.kafka:9092`
+`kafka.version` | Kafka broker version | `nil`
+`heapster` | Enable Heapster (only enable this in clusters where heapster does not exist already) | `false`
 
 * - Please note that deploying of Azure Storage Queue or Kafka is not done by Fission chart and you will have to explicitly deploy them.
```
life1347: Remove the dash `* -`

charts/README.md
```diff
@@ -37,34 +37,53 @@ $ helm delete my-release
...
+`pullPolicy` | Image pull policy | `IfNotPresent`
+`fetcherImage` | Fission fetcher repository | `fission/fetcher`
+`fetcherImageTag` | Fission fetcher image tag | `1.0-rc2`
+`controllerPort` | Fission Controller service port | `31313`
+`routerPort` | Fission Router service port | ` 31314`
+`functionNamespace` | Namespace in which to run fission functions (this is different from the release namespace) | `fission-function`
+`builderNamespace` | Namespace in which to run fission builders (this is different from the release namespace) | `fission-builder`
+`enableIstio` | Enable istio integration | `false`
+`persistence.enabled` | If true, persist data to a persistent volume | `true`
+`persistence.existingClaim` | Provide an existing PersistentVolumeClaim instead of creating a new one | `nil`
```
life1347: default is "fission-storage-pvc"

charts/fission-all/values.yaml
darkgerm: It is undefined by default.
if existingClaim is undefined: a PVC named "fission-storage-pvc" will be created.
if existingClaim == "fission-storage-pvc": no PVC will be created. One needs to create a PVC named "fission-storage-pvc" in advance or the chart will fail to install.

The meaning of undefined and set it to "fission-storage-pvc" is different. If it really needs a default value, it will be `null`. (leave it blank also represents null in yaml)

I think it's better to leave it value blank and commented out.
I will add document about default PVC name but I think it will not be this line.

life1347: You're right, I missed the pvc creation in pvc.yaml.

life1347:
🎉 LGTM

PR #1067, opened 2019-01-20
Adding check for requirements file,

### LINKED ISSUE #1020: Python builder fails if requirements.txt does not exist
If `--src` flag is used and `requirements.txt` is not used then build fails for Python builder environment.

vishal-biyani:
In case the user is simply using a source file and --src flag

Fixes #1020


environments/python/builder/defaultBuildCmd
```diff
@@ -1,2 +1,6 @@
 #!/bin/sh
-pip3 install -r ${SRC_PKG}/requirements.txt -t ${SRC_PKG} && cp -r ${SRC_PKG} ${DEPLOY_PKG}
\ No newline at end of file
+if [ -f ${SRC_PKG}/requirements.txt ]
```
life1347: add `set -euxo pipefail` let the build failed if any of error occurs

vishal-biyani: Done @life1347

PR #1068, opened 2019-01-20
Spec archive optimisation

vishal-biyani:
Spec archive has issues with single files (Will file issue in a bit) and can be optimized to use the Glob method already available.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1068)

### DIFF
--- a/fission/spec.go
+++ b/fission/spec.go
@@ -960,11 +960,32 @@ func applyResources(fclient *client.Client, specDir string, fr *FissionResources
 // localArchiveFromSpec creates an archive on the local filesystem from the given spec,
 // and returns its path and checksum.
 func localArchiveFromSpec(specDir string, aus *ArchiveUploadSpec) (*fission.Archive, error) {
-	files := make([]string, 0)
+	// get root dir
+	var rootDir string
+	if len(aus.RootDir) == 0 {
+		rootDir = filepath.Clean(specDir + "/..")
+	} else {
+		rootDir = aus.RootDir
+	}
 
-	files, err := fission.FindAllGlobs(aus.IncludeGlobs)
-	if err != nil {
-		util.CheckErr(err, "finding all globs")
+	// get a list of files from the include/exclude globs.
+	//
+	// XXX if there are lots of globs it's probably more efficient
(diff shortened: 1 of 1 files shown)


vishal-biyani:
Fixes https://github.com/fission/fission/issues/1070

fission/spec.go
```diff
@@ -960,31 +960,11 @@ func applyResources(fclient *client.Client, specDir string, fr *FissionResources
...
-
-	// get a list of files from the include/exclude globs.
-	//
-	// XXX if there are lots of globs it's probably more efficient
-	// to do a filepath.Walk and call path.Match on each path...
 	files := make([]string, 0)
-	if len(aus.IncludeGlobs) == 1 && archiver.Zip.Match(aus.IncludeGlobs[0]) {
-		files = append(files, aus.IncludeGlobs[0])
-	}
-	for _, relativeGlob := range aus.IncludeGlobs {
```
life1347: The root cause of the issue is here. We can solve the issue by adding an `else` here.

fission/spec.go
```diff
@@ -960,31 +960,11 @@ func applyResources(fclient *client.Client, specDir string, fr *FissionResources
 // localArchiveFromSpec creates an archive on the local filesystem from the given spec,
 // and returns its path and checksum.
 func localArchiveFromSpec(specDir string, aus *ArchiveUploadSpec) (*fission.Archive, error) {
-	// get root dir
-	var rootDir string
-	if len(aus.RootDir) == 0 {
```
life1347: No, we cannot do this. If we didn't add root dir to the file path as an absolute path, then we are not able to create the archive if the spec dir is not in the current path.
```
$ fission spec apply --specdir $GOPATH/src/github.com/fission/fission-bank-sample/specs
Fatal error: Failed to apply specs: Archive 'bank-go-zip' is empty
```

fission/spec.go
vishal-biyani: Thanks - good catch - I have fixed and updated the PR @life1347

life1347:
LGTM

PR #1069, opened 2019-01-20
Clear message in case of function/pod failure

### LINKED ISSUE #952: Router should reply meaningful message back to CLI when encountering environment error
If an environment is not ready, then executor may not be able to reply message to the router before timeout. Router will send following error message to the client which make no sense to them. We should provide more useful messages instead.

```
$ fission fn test --name calc-form
Error calling function calc-form: 500 Post http://executor.fission/v2/getServiceForFunction: dial tcp: i/o timeoutFailed to parse the base URL: parse 192.168.99.100:31313: first path segment in URL cannot contain colon
```

vishal-biyani:
Fixes #952 

This is not perfect but good for review fix. There are mainly two things that need to be fixed separately:

1) Right now the error is so long is because it is wrapped and passed on at every layer. This is great for tracing error but not for the end user. The challenge is determing which part of multiple errors is most useful for users. I don't know the solution yet, but the problem statement is fairly clear.

2) Currently, we are checking one pod since none is ready - there is room for future addition!

```
$ fission fn test --name hello
Error calling function hello: 500; Please try again or fix the error: Error updating service address entry for function hello_default: Internal error - Timeout: waited too long for pod of deployment python-poolmgr-default in namespace fission-function to be ready, error: ImagePullBackOffBack-off pulling image "fission/python"
```

The [error in article](https://medium.com/@bluszcz/serverless-microk8s-kubernetes-fcd6b875cd33) that user is facing is due to wrong image while creating environment. I think there is a separate bug for validating that image is accessible/exists before creating environment.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1069)

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -172,7 +172,6 @@ func (executor *Executor) createServiceForFunction(meta *metav1.ObjectMeta) (*fs
 	}
 
 	if fsvcErr != nil {
-		fsvcErr = errors.Wrap(fsvcErr, fmt.Sprintf("[%v] Error creating service for function", meta.Name))
 		log.Print(fsvcErr)
 	}
 
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -172,6 +172,7 @@ func (executor *Executor) createServiceForFunction(meta *metav1.ObjectMeta) (*fs
 	}
 
 	if fsvcErr != nil {
+		fsvcErr = errors.Wrap(fsvcErr, fmt.Sprintf("[%v] Error creating service for function", meta.Name))
 		log.Print(fsvcErr)
 	}
 
--- a/executor/poolmgr/gp.go
+++ b/executor/poolmgr/gp.go
@@ -574,9 +574,26 @@ func (gp *GenericPool) waitForReadyPod() error {
 		}
 
 		if time.Since(startTime) > gp.podReadyTimeout {
+			podList, err := gp.kubernetesClient.CoreV1().Pods(gp.namespace).List(metav1.ListOptions{
+				LabelSelector: labels.Set(
+					gp.deployment.Spec.Selector.MatchLabels).AsSelector().String(),
+			})
+			if err != nil {
+
+			}
+
+			// Since even single pod is not ready, choosing the first pod to inspect is a good approximation. In future this can be done better
+			pod := podList.Items[0]
+			var podErr string
+			for _, cStatus := range pod.Status.ContainerStatuses {
+				if cStatus.Ready != true {
+					podErr = cStatus.State.Waiting.Reason + cStatus.State.Waiting.Message
+				}
+			}
+
 			return errors.Errorf(
-				"Timeout: waited too long for pod of deployment %v in namespace %v to be ready",
-				gp.deployment.ObjectMeta.Name, gp.namespace)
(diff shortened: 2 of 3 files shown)


executor/poolmgr/gp.go
```diff
@@ -574,9 +574,26 @@ func (gp *GenericPool) waitForReadyPod() error {
...
+			if err != nil {
+				log.Printf("Error getting pod list after timeout waiting for ready pod: %v", err)
+			}
+
+			// Since even single pod is not ready, choosing the first pod to inspect is a good approximation. In future this can be done better
+			pod := podList.Items[0]
+			var podErr string
+			for _, cStatus := range pod.Status.ContainerStatuses {
+				if cStatus.Ready != true {
+					podErr = cStatus.State.Waiting.Reason + cStatus.State.Waiting.Message
```
life1347: It's better to use multierror package ("github.com/hashicorp/go-multierror") here to collect information for all unready contains. Also, it's better to have an error output format something like`fmt.Sprintf("Reason: %v, message: %v")`.

executor/executor.go
```diff
@@ -172,7 +172,6 @@ func (executor *Executor) createServiceForFunction(meta *metav1.ObjectMeta) (*fs
 	}
 
 	if fsvcErr != nil {
-		fsvcErr = errors.Wrap(fsvcErr, fmt.Sprintf("[%v] Error creating service for function", meta.Name))
```
life1347: Why remove this line? It's useful for troubleshooting if the log contains metadata.

executor/poolmgr/gp.go
vishal-biyani: @life1347 - I agree that this mechanism is not perfect - but collecting information for all containers might add too much information. For ex. if the pool size is 10 or 20 as an example - then we will have that many repeat messages. Since not a single pod was ready - it is a good approximation that all pods have the same error.

executor/executor.go
vishal-biyani: It was same information being printed by the caller and the called function - this making overall message to user pretty verbose. Shall we just log meta?

life1347: I'm doubt by removing this line can help us to eliminate verbose log entirelly cause this line just adds some extra information here. In fact, I think we should improve the log here. 

As you can see, the newdeploy print and return same error log back to caller.
https://github.com/fission/fission/blob/1dbb4f4fb73ed1a3be97f87b059c8a5e1ebc9536/executor/newdeploy/newdeploymgr.go#L301-L304
And for poolmanager, some of errors are returned directly.
https://github.com/fission/fission/blob/1dbb4f4fb73ed1a3be97f87b059c8a5e1ebc9536/executor/executor.go#L159-L167

How about we leave this line as the same, and open other PR to address verbose log issue here?

vishal-biyani:
After last change:

```
$ fission fn test --name hello
Error calling function hello: 500; Please try again or fix the error: Error updating service address entry for function hello_default: Internal error - [hello] Error creating service for function: Timeout: waited too long for pod of deployment python-poolmgr-default in namespace fission-function to be ready, error: 1 error occurred:
	* ImagePullBackOff: Back-off pulling image "fission/python"
```

life1347:
LGTM

PR #1075, opened 2019-01-22
Use official fluentd-gcp image instead of fork it from source.

darkgerm:
Originally the fluentd Dockerfile is fork from Kubernetes GitHub repo. It's out-of-sync and hard to upgrade.

I use the official fluentd-gcp image (gcr.io/google-containers/fluentd-gcp) as base image to build. After this PR if one need to upgrade fluentd, he only needs to change base image tag instead of merge Dockerfile from upstream manually.

This PR should also fix the CI build fail issue.
The root cause is the new version of fluent-plugin-influxdb can't be installed in the old version of fluentd image. I solve this issue by upgrading the fluentd image to let new version work. To prevent this happened again, I fixed the version of plugins in Dockerfile.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1075)


logger/fluentd/Dockerfile
```diff
@@ -1,70 +1,19 @@
...
+    && apt-get purge -y --auto-remove \
+                     -o APT::AutoRemove::RecommendsImportant=false \
+                     $BUILD_DEPS \
+    && rm -rf /tmp/* \
+              /var/lib/apt/lists/* \
+              /usr/lib/ruby/gems/*/cache/*.gem \
+              /var/log/* \
+              /var/tmp/*
+
+COPY fluent.conf /etc/fluent/fluent.conf
```
life1347: Now, the mount path of config file of logger is at /etc/td-agent/
https://github.com/fission/fission/blob/1dbb4f4fb73ed1a3be97f87b059c8a5e1ebc9536/charts/fission-all/templates/fluentd.yaml#L63
you need to change the path in both charts as well.

life1347: Sorry, this is an important PR and affect the CI build, I addressed the issue to make CI back to live again.

life1347:
Resolved by #1076  Close this for now, feel free to reopen it.

PR #1079, opened 2019-01-24
OpenTracing for Fission

vishal-biyani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1079)

### DIFF
--- a/buildermgr/common.go
+++ b/buildermgr/common.go
@@ -17,6 +17,7 @@ limitations under the License.
 package buildermgr
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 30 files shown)


vishal-biyani:
Solves #200

vishal-biyani:
Hi @life1347 I have addressed most of the comments except for the health endpoints exclusion from trace. ATM it looks like OpenCensus does not support the same based on issues below:

- https://github.com/census-instrumentation/opencensus-go/pull/881
- https://github.com/census-instrumentation/opencensus-specs/issues/160
- https://github.com/census-instrumentation/opencensus-specs/issues/151

We are not using the Jaeger client directly so that is not relevant for us (https://github.com/jaegertracing/jaeger/issues/814)

So I suggest we raise an issue for health/readiness check endpoints and then track it.

vishal-biyani:
We should explore reducing the sampling rate

vishal-biyani:
@life1347 The sampling rate is now configurable via Helm chart values. I have not taken up removing the readiness/liveness probes - will take that up in a separate PR after 1.0 if you ok to approve this after review.

Issue #625, opened 2018-04-16
Logger Warning: Unknown placeholder `${kubernetes_labels_functionUid}` found


life1347:
```
2018-04-16 07:40:23 +0000 [info]: #0 disable filter chain optimization because [Fluent::KubernetesMetadataFilter] uses `#filter_stream` method.
2018-04-16 07:40:23 +0000 [warn]: #0 record_reformer: unknown placeholder `${kubernetes_labels_functionUid}` found
```

vishal-biyani:
Closing as a duplicate of #1053

Issue #1053, opened 2019-01-11
fission fn logs produces no output


soamvasani:
via slack -- `fission fn logs` for a python function showed no logs even though the function is definitely logging.  This shows up in the fluentd pod's logs:

```
2019-01-11 22:13:27 +0000 [warn]: #0 record_reformer: unknown placeholder `${kubernetes_labels_functionUid}` found
```

life1347:
duplicate #625

soamvasani:
Closed by #1086

PR #1086, opened 2019-01-30
Switch from fluentd to fluentbit for log forwarding

soamvasani:
This removes fluentd in favor of using fluentbit, which is lighter (in
memory usage) and seems to be more actively maintained.

Fluentbit's config file format is different from fluentd's.  It also
doesn't support the same record modification stuff that fluentd
supports, so we have to change the influxdb query slightly.  This
means that after an upgrade, the new CLI won't work for querying older
logs.  Hopefully, this slight breakage is acceptable; if users really
need older logs they can use the older CLI.

This change also filters out fetcher logs in the influx query.  They
probably were never very useful to users debugging their functions, so
this reduces the noise.

Also, remove unnecessary stacktraces from fetcher logs.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1086)


soamvasani:
This closes #625 . And hopefully #1053.

life1347:
Notice some logs with empty `kubernetes_labels_functionUid` were sent to influxDB. After some experiments, this problem only happens with poolmgr. I think the root cause is fission changes pod labels when the pod is selected. Before metadatas sync, fluent-bit kubernetes plugin uses old metadata to decorate the log entry hence no `kubernetes_labels_functionUid` value. Because of this issue, we are not able to query logs from influxdb.

```
// use `kubernetes_labels_functionUid`, return nothing
select * from /^kube.var.log.containers.*/ where "kubernetes_container_name" != 'fetcher' AND "kubernetes_labels_functionUid" = '"4e8e4994-2639-11e9-815b-080027fb9776"'


// use `kubernetes_pod_id`
select * from /^kube.var.log.containers.*/ where "kubernetes_container_name" != 'fetcher' AND "kubernetes_pod_id" = '4e9acc23-2639-11e9-815b-080027fb9776'

name: kube.var.log.containers.nodejs-poolmgr-default-7b8785dbd5-8pz9t_fission-function_nodejs-6e66a9e13b55d3d89e2c5207daa8517a8b15edffb63c56e3861c4741fd61955d.log
time                _seq kubernetes_container_name kubernetes_docker_id                                             kubernetes_host kubernetes_labels_environmentName kubernetes_labels_environmentNamespace kubernetes_labels_environmentUid     kubernetes_labels_environmentUid_1 kubernetes_labels_executorInstanceId kubernetes_labels_executorType kubernetes_labels_functionName kubernetes_labels_functionNamespace kubernetes_labels_functionUid kubernetes_labels_pod-template-hash kubernetes_labels_unmanaged kubernetes_namespace_name kubernetes_pod_id                    kubernetes_pod_name                     log                                                                                                                  stream
----                ---- ------------------------- --------------------                                             --------------- --------------------------------- -------------------------------------- --------------------------------     ---------------------------------- ------------------------------------ ------------------------------ ------------------------------ ----------------------------------- ----------------------------- ----------------------------------- --------------------------- ------------------------- -----------------                    -------------------                     ---                                                                                                                  ------
1549036323709873137 196  nodejs                    6e66a9e13b55d3d89e2c5207daa8517a8b15edffb63c56e3861c4741fd61955d minikube        nodejs                            default                                4e84b1ac-2639-11e9-815b-080027fb9776                                    vbbtgtqg                             poolmgr                                                                                                                         7b8785dbd5                                                      fission-function          4e9acc23-2639-11e9-815b-080027fb9776 nodejs-poolmgr-default-7b8785dbd5-8pz9t \n                                                                                                                   stdout
1549036323709923454 197  nodejs                    6e66a9e13b55d3d89e2c5207daa8517a8b15edffb63c56e3861c4741fd61955d minikube        nodejs                            default                                4e84b1ac-2639-11e9-815b-080027fb9776                                    vbbtgtqg                             poolmgr                                                                                                                         7b8785dbd5                                                      fission-function          4e9acc23-2639-11e9-815b-080027fb9776 nodejs-poolmgr-default-7b8785dbd5-8pz9t \u003e fission-nodejs-runtime@0.1.0 start /usr/src/app\n                                                             stdout
1549036323709930276 198  nodejs                    6e66a9e13b55d3d89e2c5207daa8517a8b15edffb63c56e3861c4741fd61955d minikube        nodejs                            default
```

soamvasani:
Yeah :(  Sounds like this is too unreliable for 1.0.  One idea for a fix is to populate a separate database table with a function->container id mapping from fission (not fluentbit), and use that mapping to query the logs reliably using container ID.

life1347:
Sounds good to me. But I'm afraid we are not able to ship this PR in 1.0. Maybe move to 1.1?

soamvasani:
Yeah, makes sense.

charts/fission-all/templates/fluentbit.yaml
```diff
@@ -27,9 +33,11 @@ spec:
         svc: logger
     spec:
       containers:
-        - name: fluentd
-          image: "{{ .Values.repository }}/{{ .Values.logger.fluentdImage }}:{{ .Values.logger.fluentdImageTag }}"
+        - name: fluentbit
+          image: fluent/fluent-bit:0.13
```
darkgerm: fluent-bit v1.0.4 is out. We should consider using the latest version.

charts/fission-all/config/fluentbit.conf
```diff
@@ -0,0 +1,56 @@
...
+    Log_Level info
+    Parsers_File parsers.conf
+
+[INPUT]
+    Name tail
+    Tag kube.*
+    Path ${LOG_PATH}
+    Mem_Buf_Limit 5MB
+    Parser docker
+    DB /tmp/flb_kube.db
```
darkgerm: The `DB` is used to keep track of monitored files and offsets. If we don't keep this file, the log will be doubled if the fluentbit pod is restarted. A solution is to keep this file in node's filesystem by mounting a hostPath volume.

charts/fission-all/values.yaml
```diff
@@ -53,8 +53,6 @@ enableIstio: false
 ## Logger config
 logger:
   influxdbAdmin: "admin"
-  fluentdImage: fission/fluentd
-  fluentdImageTag: 1.0-rc2
```
vishal-biyani: I think it would be good to keep these and give users the flexibility to use their image/tag?

charts/fission-all/config/fluentbit.conf
life1347: good catch!

soamvasani:
LGTM, and thanks for taking over this patch @life1347.

PR #1087, opened 2019-02-04
Adds environment variables for NATS cluster id and NATS client id

zeromberto:
Adds possibility to configure a NATS streaming cluster different from the default cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1087)


soamvasani:
Hmm, this looks like a good change, why did you close it @zeromberto ?

PR #1091, opened 2019-02-05
Add configurable timeout to fission function test

erwinvaneyk:
Currently the implementation of the CLI (or any Fission component) specifies any timeout on
the `fission fn test` functionality. If a user function or environment does not specify any
timeout the command will stall. Nor is there any parameter to set a timeout if the user 
wanted to. 

Besides that I propose to set a sensible default for the timeout to 30 seconds, to avoid new
users having to figure this out the hard way. The alternative could be to set the timeout to 
0 (no timeout).

For a future PR/discussion: extend this timeout parameter to a global flag, and support 
timeouts on all CLI commands.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1091)

### DIFF
--- a/fission/function.go
+++ b/fission/function.go
@@ -790,9 +790,16 @@ func fnTest(c *cli.Context) error {
 		functionUrl.RawQuery = query.Encode()
 	}
 
+	ctx := context.Background()
+	if deadline := c.Duration("timeout"); deadline > 0 {
+		var closeCtx func()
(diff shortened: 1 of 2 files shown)


soamvasani:
LGTM, thanks!

vishal-biyani:
@erwinvaneyk I think there is a `go vet` error in build: `./fission/function.go:795: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak` can you please fix it and update?

PR #1098, opened 2019-02-11
Adding annotations for prometheus scraping to fission-core

### LINKED ISSUE #1096: Expose Prometheus metrics in fission-core
I'd love to get fission to export its metrics but Prometheus annotations have been added to `fission-all` chart only. Is there any chance these gets included in the `fission-core` chart?

If you don't feed good with adding these annotations the the core chart, another way would be to only expose the metrics port in the service so it can be referenced manually with a ServiceMonitor (prometheus-operator). 

Related:
- https://github.com/fission/fission/pull/677
- https://github.com/fission/fission/issues/201

vishal-biyani:
Fixes #1096

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1098)

### DIFF
--- a/charts/fission-core/templates/deployment.yaml
+++ b/charts/fission-core/templates/deployment.yaml
@@ -179,6 +179,10 @@ spec:
       labels:
         application: fission-router
         svc: router
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "8080"
     spec:
       containers:
       - name: router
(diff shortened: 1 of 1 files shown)


erwinvaneyk:
I see that we explicitly add the metrics ports to the deployment spec of the router and executor in the fission-all chart:  https://github.com/fission/fission/blob/b324d2c38ec3ace19f6ddad1eff3742dea4c85c4/charts/fission-all/templates/deployment.yaml#L226

We should probably either add it to the fission-core template as well. Or, if it is legacy configuration, remove it from the fission-all chart.

Other than that, LGTM 👍

soamvasani:
LGTM, thanks @vishal-biyani

vishal-biyani:
@erwinvaneyk The ports in the spec are only for additional information. I will not make the change you suggested -  but when we unify core & all chart - we will consolidate changes.

PR #1105, opened 2019-02-21
Fix canary config manager creation error in controller

life1347:
If no valid prometheus server url provided in configmap, controller will try to retrieve related information from environment variables.
Due to  controller uses only host address without proper protocol scheme/port for prometheus client to connect with and makes canary config manager failed to start.
To fix this, this PR adds extra lines to get necessary information and append those infos into prometheus service url.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1105)


soamvasani:
LGTM, thanks @life1347

Issue #644, opened 2018-04-25
Support for tolerations on function


vishal-biyani:
[Virtual Kubelet](https://github.com/virtual-kubelet/virtual-kubelet) opens up some interesting possibilities for scheduling containers on targets like ACI, Hyper.sh, Fargate etc. One of the things needed for pods to be scheduled on virtual kubelet provider is tolerations that match the provider's taints. Fission should enable tolerations on functions.

There are other aspects too for example what to do with services, HPA etc. but tolerations are a good first starting point.

vishal-biyani:
There could be other use cases too, for example, running certain pods on nodes which have taints for the availability of GPUs

life1347:
Solved with PodSpec. https://docs.fission.io/docs/spec/podspec/toleration/

Issue #640, opened 2018-04-24
Ability to init function with some data on volume


medzin:
**Summary:**
I need to run some code that will pre-download some confidential data for function, preferably once (something similar to k8s init containers). It would be cool if the data could be shared on volume.

**Context:**
I'm looking for a simple way to integrate Vault KV with functions. Using the Secrets API increases the complexity of the solution because it requires the synchronization of data between Kubernetes and Vault and adds the related problems, i.e. potential differences in stored configuration in both services and maintaining additional code. Ability to execute a simple init script adds a powerful integration possibilities.

life1347:
Solved by #1106

PR #1106, opened 2019-02-21
Pod specs

vishal-biyani:
A basic tested working version of pod specs in action. Will update with tests and documentation shortly.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1106)


vishal-biyani:
For CLI we could add an error if both are provided. 

For CRD - in future, we could add error, but for this release, something simpler would be nice.

vishal-biyani:
This PR is now ready for review. Two working examples of tolerations (https://github.com/fission/fission/issues/644) and init container (https://github.com/fission/fission/issues/640) with a volume are shown in draft post [here](https://github.com/platform9/blog.fission.io/pull/48)

There are some limitations though and these need to be addressed in the future:

- Right now the validation of duplication of container spec and pod spec os handled on client side only. On server side if we handle this then there is no feedback loop to user about the function not working - except at looking for logs of the pods. I believe we should implement this in future when we introduce status for Fission CRDs

- Although you can create volumes - you cannot mount them on the Fetcher or function container as of now. If you define the volume mounts in podspec - then the spec creates duplicate values or a completely different container - none of which is useful. We would need an ability to add a mountpoint with name to function/fetcher container

- More to be updated as more trial is done as part of the blog post.

vishal-biyani:
I tested this some more and I think that the `podspec` and `containerspec` need not be mutually exclusive and in fact, both need to be supported. Let me start with some problems faced with podspec and then come back to working examples of both in tandem.

## Scenario 1 - Use only pod spec

If we have to mount a volume on the function container using pod spec - we will write a env spec something like this:

```
metadata:
  creationTimestamp: null
  name: python
  namespace: default
spec:
  TerminationGracePeriod: 360
  builder: {}
  keeparchive: false
  poolsize: 3
  resources: {}
  runtime:
    functionendpointport: 0
    image: fission/python-env 
    loadendpointpath: ""
    loadendpointport: 0
    podspec:
      containers:
      - name: python
        image: fission/python-env
        volumeMounts:
          - name: infopod
            mountPath: /etc/infopod
            readOnly: false
      volumes:
        - name: infopod
          downwardAPI:
            items:
              - path: "labels"
                fieldRef:
                  fieldPath: metadata.labels
```

What this does is merges the container from pod spec into deployment's containers but fails with `Duplicate value: "python"]` - since we are using same name for container (And changing name means it will create a new container altogether). With `mergo.WithOverride` option while merging also it does not produce a valid YAML and causes failure.

## Scenario 2 - Pod Spec and Container Spec

Here we use pod spec to define 1) A volume 2) A init container that mounts the volume 2) An additional container that runs (Say sidecar of some sort)

We use container spec to mount the same volume on to function container and this all works (After not checking in CLI if both pod and container spec exist): 

```
  runtime:
    functionendpointport: 0
    image: fission/python-env 
    loadendpointpath: ""
    loadendpointport: 0
    container:
      volumeMounts:
      - name: infopod
        mountPath: /etc/infopod
        readOnly: false
    podspec:
      containers:
      - name: third-container
        image: spaster/alpine-sleep
      initContainers:
      - name: init-py
        image: fission/python-env 
        command: ['sh', '-c', 'cat /etc/infopod/labels']
        volumeMounts:
          - name: infopod
            mountPath: /etc/infopod
            readOnly: false
      volumes:
        - name: infopod
          downwardAPI:
            items:
              - path: "labels"
                fieldRef:
                  fieldPath: metadata.labels
```

## Some thoughts

- PodSpec allows modifying the containers in functions - which is not most desirable. We at most want someone to modufy the function container and may be not touch fetcher container at all. Of course the current merging does not support this at all.

- It is a good idea to restrict container changes to only function container, while being able to add additional container. 

- We can check if any of containers in pod spec have same name as the function/fetcher container - and throw the error in CLI/validation.

- We keep podspec and container spec both, where 
  - Podspec is for all things outside of containers - except for adding a new container to function pod
  - Container spec for adding things to function container

What do you think @soamvasani @erwinvaneyk

soamvasani:
I'm somewhat confused about what exactly you're proposing -- if you're saying there will be no function container modification and no fetcher container modification, that might be too restrictive.  For example you wouldn't be able to add a volume, right?

#1150 changed the way we add fetcher from a literal containerspec in the executor to a transformation of a pod spec done in the fetcherConfig packge.  Similarly, we could view even the addition of a function also as a transformation of a pod spec.  So how about this:

 1. Executors are viewed as a transformation of a pod spec, possibly an empty one
 1. The executor transformation is to add a fetcher container and a function container
 1. Each of these could be either an addition or a merge
 1. If a container being added (function or fetcher) already exists, it is merged with the one that exists and that merged container is put into the podspec.  So if we're adding a function container and the podspec already has one, we'd merge the two container specs (with mergo, or without it if it doesn't work) and assign that into the podspec.

laurence-hudson-mindfoundry:
This functionally would be really valuable in clusters using cluster-autoscaler, where it could be used to control which node pool / gcp instance group / aws autoscaling group, that is used by the pool managers/executors via affinity/nodeSelector/tolerations (to control scale up) along with the `cluster-autoscaler.kubernetes.io/safe-to-evict` attribute (to control scale down).

If I'm reading the changes/discussion correctly then either scenario 1 or 2 would work for this use-case, although I dont think the pool-managers are covered here.

vishal-biyani:
Build might be green but the spec tests haven't run - so consider this still not good for merge until a further update.

vishal-biyani:
@life1347 I have addressed review comments and tests will pass after the current build. Can you please take a look?

life1347:
LGTM

PR #1110, opened 2019-03-01
PHP 7.3 v2 Specialization

AlbertoLopezBenito:
Noticed in #807

This PR providers support for the v2 specialization for PHP. I based it off the python package to do it.
Features added:
 - Fission v2 support
 - PHP 7.3 from 7.1
 - Removed `ZendServer` usage (deprecated)
 - Removed usage of built-in PHP server, is not recomended to use it in production.

NOTE: Fission v1 is still working, all changes are backward compatible

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1110)


life1347:
Hi @AlbertoLopezBenito Thanks for this PR! I've finished the review and think it's almost in good shaper, however, after some tests I have couple questions want to confirm:

1. Is it possible for php-builder to do code validation during package build stage?

I tried with the following error source code (no `<?$php` at the beginning of file), and the function returns whole file content directly instead of an error. Is that a normal behavior for php? If not, could we enforce code validation during build stage (or in code loading stage)?

2. No function error logs

If there is any syntax errors in source code, for example missing `;`, the php server doesn't print any error logs. Is that normal or it's intentional?

3. A simple example for php package usage (Optional)

It would be great if there is a simple example in `fission/examples/php7` to demonstrate how to create a php package with composer.json. It helps people to understand how it actually works.

Thanks again for your work.

AlbertoLopezBenito:
Hi @life1347 
I just uploaded a new commit checking syntax errors in source files and adding an example using the builder. It's the same example that python provides :wink: 

Related to
> Is it possible for php-builder to do code validation during package build stage?

PHP renders the code and will execute the parts with `<?php`. It was designed for that purpose, so it's normal behavior. 

Thanks for your review!

life1347:
Looks like there are error in php server
Use example to test, and function works for the first time. Then, it crashed for the second time. Following is test logs.
```
poolmgr-php-default-ncnwzvk9-5657758556-pnsbr fetcher 2019/03/07 07:59:13 Successfully placed at /userfunc/e257b7da-40ae-11e9-b759-08002778d774
poolmgr-php-default-ncnwzvk9-5657758556-pnsbr fetcher 2019/03/07 07:59:13 Elapsed time in fetch request = 123.037967ms
poolmgr-php-default-ncnwzvk9-5657758556-pnsbr php [2019-03-07 07:59:13] Function.DEBUG: File read: example.txt [] []
poolmgr-php-default-ncnwzvk9-5657758556-pnsbr php Fatal error: Cannot redeclare execute() (previously declared in /userfunc/e257b7da-40ae-11e9-b759-08002778d774/handlers/FileReader.php:8) in /userfunc/e257b7da-40ae-11e9-b759-08002778d774/handlers/FileReader.php on line 16

```

environments/php7/server.php
```diff
@@ -0,0 +1,91 @@
...
+
+        $parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+        try {
+            $parser->parse(file_get_contents($codePath));
+        } catch (Throwable $throwable) {
+            $logger->error($codePath . ' - ' . $throwable->getMessage());
+            return new Response(500, [], $codePath . ' - ' . $throwable->getMessage());
+        }
+
+        require $codePath;
```
life1347: I guess it's the root cause because server requires user function twice.

life1347: maybe add a flag to not require again if we already did that before. Or, any better suggestion?

life1347: After moving `require` to specialize stage it works perfectly. Once this problem is addressed then we can merge it! 💯

AlbertoLopezBenito: @life1347 Fixed! I changed `require` to `require_once`. This prevents that behavior :)

life1347:
Thanks for all your works @AlbertoLopezBenito , merged! 🎉 💯 🥇

vishal-biyani:
Great job and thanks a lot @AlbertoLopezBenito 🎉

PR #1111, opened 2019-03-04
Support --plugin parameter in Fission CLI

### LINKED ISSUE #1109: Fission CLI infinitely fork
To reproduce:
```
# 1.  download fission cli
curl ...

# 2. duplicate it with name "fission-*" (cp and ln both work)
mv fission fission-1.0.0
ln -s fission-1.0.0 fission

# 3. make sure the binary can be found in PATH
export PATH=$PWD:$PATH

# 4. run fission
fission
```
It will infinitely fork `fission-1.0.0 --plugin`.
The way to clean it up without reboot: delete or rename fission-1.0.0. (kill may not work for fork bomb)

erwinvaneyk:
This PR fixes #1109 where using the fission CLI as a plugin would result in indefinite recursive calls when renaming fission  to fission-XYZ would cause a fork bomb, due to the CLI recursively calling itself to discover plugins.

To fix this I added support for the `--plugin` parameter to the Fission CLI itself.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1111)


darkgerm:
It's a little strange to regard itself as a plugin. How about:
- return something (e.g. `{"name":"fission-cli"}`) to tell it's cli itself when `--plugin` is given.
- do not regard as a plugin if it is a fission cli.

erwinvaneyk:
Hey @darkgerm - thanks for taking look at this!  I don't have strong preference, but I don't see what is so strange about it. This PR keeps the plugin rule simple: if the binary matches `fission-*`, it will show up as a plugin. What do you think? 😁

darkgerm:
I just found that if both Action and CommandNotFound are given, CommandNotFound will never be executed. It results the plugin will not be called anymore.
There is a closed issue here https://github.com/urfave/cli/issues/354

In this case, I think we should use Action only in order to catch the no-argument case.

erwinvaneyk:
@darkgerm I fixed the issue. Could you give it a second look?

soamvasani:
LGTM, as long as `fission` without arguments still prints the help text and no weird errors.

fission/main.go
```diff
@@ -323,9 +325,34 @@ func newCliApp() *cli.App {
 
 	app.Before = cliHook
 	app.CommandNotFound = handleCommandNotFound
```
darkgerm: Since this line is useless, it's better to remove it in order not to generate confuse.

darkgerm:
> This PR keeps the plugin rule simple: if the binary matches `fission-*`, it will show up as a plugin.

Hi @erwinvaneyk , I think your idea is better. Let's skip my proposal.

fission/main.go
erwinvaneyk: Good point 👍 will fix

Issue #929, opened 2018-10-10
Updating environment doesn't update function deployment when using newdeploy


soamvasani:
Updating an env should cause any existing functions deployed on the old env to be re-deployed onto the new ones, within some reasonable amount of time.

vishal-biyani:
Hi, @soamvasani I noticed that even for pool manager if a function is already unmanaged/ specialized - then it points to the old version of the image even after updating the environment.  Do we want to consider that as part of change too?

soamvasani:
With minscale the problem is that the function pod is more permanent... Poolmgr pods usually recycle at some point. But yes the problem does exist in both backends. I wonder if we could just kill the pod? Then the solution would be same for both backends?  Router already has retries for the case where the pod is dead.

PR #1116, opened 2019-03-11
Function update after change in env

vishal-biyani:
This fixes: https://github.com/fission/fission/issues/929

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1116)


vishal-biyani:
Summary of change:

1) Added a new controller to newdeploy. Every time an env updates - it will check for all related functions and simply update the resource version of fn with the latest resource version of env.

The actual update of newdeploy deployment is done by an existing controller for functions.

2) From a CLI - when you create a function - the resource version of the environment is retrieved and updated to function. I think is going to break scenarios where someone creates a function first and env later (And I think this anyway does not work for newdeploy as of now)

3) For Specs - similar update to function resourceversion

I think this change is going to break the existing deployments and we will have to update existing functions to get the env version of environments as a one-time upgrade activity!

life1347:
For some environment like Go it's dangerous to do automatically upgrade since the compiled plugin may not fit for the new image. How about exposing a new flag in CLI and let users choose whether updating or not? (By default, we don't do upgrade automatically.)

vishal-biyani:
Hi @life1347 I am not sure I understand the flag - will flag differ the update of the image for existing pods or what exactly it will do?

Secondly, as of now the poolmanager immediately re-creates the pool after the image is updated for an environment so I think keeping the behavior consistent is probably a better idea.

I am worried about cross resource updates and the right design for that. Currently, we have a tight dependency between env, package, and function in terms of creation - which is not desirable. I am still exploring Kubernetes docs to see if this is a solved pattern and we should use that instead of current solution in the code of this PR.

life1347:
Yeah, you're right for the consistent behavior for executors. Or how about we can add some description text to CLI help to explain potential problem if update an environment?

```
NAME:
   fission environment update - Update environment

USAGE:
   fission environment update [command options] [arguments...]

DESCRIPTION:
   some text here
```

You can add a description with like the following 
```
{Name: "update", Usage: "Update environment", Description: "some text here" .....
```

vishal-biyani:
@ajbouh & @soamvasani This was PR we were discussing for comments in the call.

soamvasani:
I wonder if you can put this same controller into `newdeploymgr`, and call `deploy.updateFunction` on all functions for that environment whenever an environment changes?

For finding the functions of an env you could either iterate over them all in the store or keep a env->fn map and update it from the function controller that's already there.

ajbouh:
I didn't see anything in this change that invalidates the function cache in router. What happens to incoming requests when old function environments are being "updated"?

soamvasani:
> I didn't see anything in this change that invalidates the function cache in router. What happens to incoming requests when old function environments are being "updated"?

I think the idea is that by updating the function CRD itself, all of the logic around function updates gets triggered.  The router cache (functionServiceMap) uses the resource version in the cache key, so updates get handled properly, in this diff.

But, now I realize that my proposal to leave the CRD unchanged and just redeploy the pod(s) is not so simple, because the router cache would also need changing.

ajbouh:
There's also a race condition, isn't there? Where incoming request gets routed to a function that's shut(ting) down?

soamvasani:
> There's also a race condition, isn't there? Where incoming request gets routed to a function that's shut(ting) down?

The functionServiceMap cache has an expiry of 1 minute, and the pods are killed after 6 minutes.  So as long as requests are shorter than 5 minutes, we're ok.

vishal-biyani:
@soamvasani and @ajbouh - changed PR to not break the API. 

As far `The router cache (functionServiceMap) ` - the router uses Service URL for newdeploy executor so cache does not need an update.

Lastly - We would have to add a rolling upgrade and maxUnavailable for newdeploy deployment - with some defaults , will do that in a separate PR

PR #1120, opened 2019-03-14
Fix unit test testCRD random failure problem

life1347:
Kubernetes watch returns a sequence changes of CRD resources. In some cases, the older changes will be returned and break the test. To prevent this, we wait for a couple of seconds to ensure we get all changes from the server and then verify the content of the object.
https://travis-ci.org/fission/fission/builds/505781138#L1337-L1371

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1120)

### DIFF
--- a/crd/crd_test.go
+++ b/crd/crd_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 The Fission Authors.
+Copyright 2018 The Fission Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


crd/crd_test.go
```diff
@@ -104,37 +102,45 @@ func functionTests(crdClient *rest.RESTClient) {
...
-		if !ok {
-			log.Panicf("Can't cast to Function")
-		}
-		if wf.Spec.Environment.Name != function.Spec.Environment.Name {
-			log.Panicf("Bad object from watch: %#v", wf)
+	var watchFunc *Function
+	for {
+		select {
+		case <-time.NewTimer(3 * time.Second).C:
+			if watchFunc.Spec.Environment.Name != function.Spec.Environment.Name {
```
soamvasani: 1. check for watchFunc == nil?
 2. I'm confused about the change, overall.  Are you trying to ignore stale changes?  If so why error out on them?  Also what do we gain from changing the timer from 1 sec to 3 sec?

life1347:
close this PR for now, will open it later once figure out a better way to fix the problem.

Issue #1104, opened 2019-02-20
Allow creating environments with the newdeploy executor type


TimJones:
I have a workflow where latency isn't important but resource use is, so all my functions are declared with `--executortype newdeploy --minscale 0`.
When starting my fission project I had to create the environment which by default is a `poolmgr` which is unneeded in this use case and wastes resources.

I would like to see setting up of the environment allowing a selection for the executor type.

vishal-biyani:
Set version of env to 3 if poolsize is specified by user

PR #1121, opened 2019-03-15
Configurable zero pool size in case of newdeploy function

vishal-biyani:
Fixes https://github.com/fission/fission/issues/1104

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1121)


soamvasani:
Thanks, the diff LGTM.  The title of the PR is a bit misleading though, there's no automatic setting of poolsize to zero anywhere, right?

Issue #821, opened 2018-07-27
Multiple files in $KUBECONFIG trigger "Error loading config file"


nea:
Hi

* Platform Windows

I am using a multiple config file approach and have set up all files as semicolon separated list under `KUBECONFIG`. Now, fission says

```
Failed to connect to Kubernetes: Error loading config file
```

with the list of files as string. Of course, it cannot read it as it is not just one single file. Is this a known error? Are there workarounds Or do I have to merge all cluster configurations into one file?

Thanks a lot in advance

vishal-biyani:
@nea Can you check if you can run `kubectl get pods` successfully?

nea:
Hi @vishal-biyani 

Yeah, that was my first thought, too. But that works. For example:

```
NAME                              READY     STATUS    RESTARTS   AGE
buildermgr-d86d68dcf-rrrgj        1/1       Running   2          1h
controller-7fbcc99f54-s9mz6       1/1       Running   0          1h
executor-d869645ff-5cdch          1/1       Running   3          1h
influxdb-6799f75947-7p825         1/1       Running   0          1h
kubewatcher-5784b586f4-9f5zs      1/1       Running   2          1h
logger-5lx4t                      1/1       Running   0          1h
mqtrigger-57775c8766-p97cn        1/1       Running   3          1h
nats-streaming-7686cd7ff4-stbgk   1/1       Running   0          1h
router-669f5cf765-vgjsc           1/1       Running   2          1h
storagesvc-66c447d67-nbrmr        1/1       Running   0          1h
timer-678f7b8884-mp7kz            1/1       Running   2          1h
```

for ` kubectl get pods -n fission`. 

I also did some context switching as explained at the Kubernetes doc to verify that my two clusters are working.


Thanks

vishal-biyani:
Thanks for confirming that, I think I am able to reproduce it the way you mentioned. For now I suggest either merge kubeconfigs into one or export only one file in Kubeconfig in that shell. I will leave this issue open so we can fix it later.

nea:
Hey @vishal-biyani 

Thanks for the heads up. Will do until further notice.


Best

soamvasani:
Ok, so I think are supposed to use [`ClientConfigLoadingRules`](https://godoc.org/k8s.io/client-go/tools/clientcmd#ClientConfigLoadingRules) with the `Precedence` field set to the list of paths in the KUBECONFIG.

[...]

PR #1126, opened 2019-03-18
support KUBECONFIG with multiple kube config files

grounded042:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1126)

### DIFF
--- a/fission/util/util.go
+++ b/fission/util/util.go
@@ -53,20 +53,6 @@ func GetFissionNamespace() string {
 	return fissionNamespace
 }
 
-func GetKubeConfigPath() string {
-	kubeConfig := os.Getenv("KUBECONFIG")
-	if len(kubeConfig) == 0 {
-		home := os.Getenv("HOME")
-		kubeConfig = filepath.Join(home, ".kube", "config")
-
-		if _, err := os.Stat(kubeConfig); os.IsNotExist(err) {
-			log.Fatal("Couldn't find kubeconfig file. " +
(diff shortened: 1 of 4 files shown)


grounded042:
this fixes https://github.com/fission/fission/issues/821

fission/util/util.go
```diff
@@ -128,10 +113,31 @@ func KubifyName(old string) string {
...
-	config, err := clientcmd.BuildConfigFromFlags("", kubeConfig)
+// GetKubernetesClient builds a new kubernetes client. If the KUBECONFIG
+// environment variable is empty or doesn't exist, ~/.kube/config is used for
+// the kube config path
+func GetKubernetesClient() (*restclient.Config, *kubernetes.Clientset) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+
+	kubeConfigPath := os.Getenv("KUBECONFIG")
+	if len(kubeConfigPath) == 0 {
+		home := os.Getenv("HOME")
```
life1347: To support windows better, I suggest here we can use `os/user` to get the correct home directory path.
reference: https://stackoverflow.com/a/13004756

grounded042: Thanks! I'll take a look.

grounded042: @life1347 I changed things up as per the link. Seems like a better cross-platform implementation.

life1347:
LGTM

Issue #653, opened 2018-04-29
Function logs CI setup issue


vishal-biyani:
This commands in sequence work fine from local machine pointing to a GKE Fission setup, but fail in the CI setup every time.
```
$ fission env create --name node --image fission/node-env
$ fission fn create --name hello --env node --code badhello.js 
$ fission fn test --name hello
Error calling function hello: 502 
> fission-nodejs-runtime@0.1.0 start /usr/src/app
> node server.js

Codepath defaulting to  /userfunc/user
Port defaulting to 8888
user code load error: SyntaxError: Unexpected token function
::ffff:10.8.4.17 - - [29/Apr/2018:03:40:34 +0000] "POST /specialize HTTP/1.1" 500 2 "-" "Go-http-client/1.1"

```

The error in CI is:

```
goroutine 2609 [running]:
runtime/debug.Stack(0xc4202cf390, 0x10, 0x10)
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/runtime/debug/stack.go:24 +0x79
runtime/debug.PrintStack()
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/runtime/debug/stack.go:16 +0x22
github.com/fission/fission/controller.(*API).respondWithError(0xc4203d0900, 0x7f231f491a38, 0xc420704ff0, 0x20fe6a0, 0xc4202cf390)
	src/github.com/fission/fission/controller/api.go:98 +0x29
github.com/fission/fission/controller.(*API).FunctionPodLogs(0xc4203d0900, 0x7f231f491a38, 0xc420704ff0, 0xc420210d00)
	src/github.com/fission/fission/controller/functionApi.go:241 +0x71b
github.com/fission/fission/controller.(*API).FunctionPodLogs-fm(0x7f231f491a38, 0xc420704ff0, 0xc420210d00)
	src/github.com/fission/fission/controller/api.go:196 +0x48
net/http.HandlerFunc.ServeHTTP(0xc4202ce5b0, 0x7f231f491a38, 0xc420704ff0, 0xc420210d00)
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/net/http/server.go:1942 +0x44
github.com/fission/fission/vendor/github.com/gorilla/handlers.loggingHandler.ServeHTTP(0x2105060, 0xc42000c018, 0x2106d20, 0xc4202ce5b0, 0x2114da0, 0xc4202b1dc0, 0xc420210d00)
[...]

PR #1128, opened 2019-03-24
Refactor test framework

darkgerm:
----
edit on 5/15

## Overview

The purpose of this PR is refactoring the testing framework:
- Let CI can run tests in parallel.
- Make it more easy to use at local.

For how to use the new framework, I've written a simple readme here: `test/README.md`

The main reason that CI builds in this PR easily to fail is the insufficient resources of the CI cluster. I reduced the parallel jobs and separate the tests into 2 groups (one run in parallel while another in serial) as a workaround. A possible solution is temporary scaling up the cluster when CI is running. I will use another PR to do this.

### tests that remain disabled
`mqtrigger/kafka/test_kafka.sh` by #944
`test_function_test/test_fn_test.sh` by #653
`test_environments/test_java_env.sh` by #957

### test that enabled by this PR
`test_fn_update/test_idle_objects_reaper.sh`

### test that disabled by this PR
`test_obj_create_in_diff_ns.sh`: This test is too large. Need to split into multiple tests.

----

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1128)


vishal-biyani:
Looks interesting @darkgerm, just curious how do logs from parallel tests appear in the console?

darkgerm:
Hi @vishal-biyani 

I planned to save the output in different log files. 
For example:
- `test/tests/test_archive_pruner.sh` output saves to `test/logs/test_archive_pruner.log`
- `test/tests/test_specs/test_spec.sh` output saves to `test/logs/test_specs/test_spec.log`

After the test is finished, check each log files and print it one by one (or only print the error one) so the logs will not mix together.

Currently, this PR has something to be fixed and I'm still working on it:
- The version of parallel in Ubuntu is too old that `--results` does not support template yet. (it's 20161222 in ubuntu14.04/18.04 while on mac is 20190322). I need to use another way to do it. 
- `run_test.sh` is not used in CI and I want to make it more useful (not only be used in CI but also can be used for local testing. Can merge with `run_test_mac.sh`)
- Let the test can run in parallel (need to fix it one by one)

By the way, travis will kill the build if there is no output for a long time (since all output are redirect to files, it looks like it is frozen). `travis_wait` directive can solve this by simply increasing the waiting time.

vishal-biyani:
Down from ~50 minutes to 20 minutes - I am so excited about this!

test/test_utils.sh
```diff
@@ -425,55 +431,89 @@ dump_logs() {
...
+        $ROOT/test/tests/test_node_hello_http.sh \
+        $ROOT/test/tests/test_package_command.sh \
+        $ROOT/test/tests/test_pass.sh \
+        $ROOT/test/tests/test_router_cache_invalidation.sh \
+        $ROOT/test/tests/test_specs/test_spec.sh \
+        $ROOT/test/tests/test_specs/test_spec_multifile.sh
+    FAILURES=$?
+
+    # FIXME: run tests with newdeploy one by one.
+    export JOBS=1
```
vishal-biyani: Minor: It would be nice to create an issue to separately investigate why these tests fail in parallel

test/utils.sh
```diff
@@ -0,0 +1,82 @@
...
+    echo $(cat /dev/urandom | tr -dc 'a-z' | fold -w 8 | head -n 1)
+}
+
+clean_resource_by_id() {
+    test_id=$1
+    KUBECTL="kubectl --namespace default"
+    set +e
+
+    crds=$($KUBECTL get crd | grep "fission.io" | awk '{print $1}')
+    crds="$crds configmaps secrets"
```
vishal-biyani: Also add `ingress` - we do create ingress in some test I think

vishal-biyani: Never mind - you are deleting route later and I think it will take care of this

vishal-biyani:
Except for 1/2 minor comments, this change LGTM. @life1347 not sure if you got a chance to review?

life1347:
LGTM

test/test_utils.sh
darkgerm: Sure. I'll create one.

test/utils.sh
darkgerm: `ingress`, which is created by fission, will be deleted later by deleting `*.fission.io` crds or`fission route delete`.
`configmaps` and `secrets` are the only resources that will be created "manually" by kubectl.

life1347:
Merged. 🇹🇼 💯 🎉

PR #1134, opened 2019-03-29
Fix misleading log when setup portforward

life1347:
Remove misleading log in portforward functions.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1134)


soamvasani:
Oof. LGTM, thanks.

PR #1136, opened 2019-03-29
Migrate from glide to official dependencies management tool: Go Module

life1347:
Go module as an official Go dependencies management tool was first introduced in 1.11 as an experimental opt-in feature. This PR aims to migrate from glide to go module for better development experience for contributors.

```
# activate go module, you may need to:
export GO111MODULE=on
```

For more details, please visit: https://github.com/golang/go/wiki/Modules

Problem found so far:
1. CanaryConfigManager needs to upgrade prometheus client_golang to newer veriosn. (0.8.0 -> 0.9.2)  to address package code change.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1136)


soamvasani:
Wow, nice!

Should we upgrade our build to use Go 1.12?

life1347:
Looks like there are `go vet` problem for 1.12.1. Let's stay in 1.11 for now.

Issue #1132, opened 2019-03-27
PayloadTooLargeError: request entity too large


JannikZed:
We are actively using the mqt-trigger system with fission and everything is running well (normally). Today I was manually looking around the function execution pods of fission and saw a super large amount of errors like this: 
```
PayloadTooLargeError: request entity too large
at readStream (/usr/src/app/node_modules/raw-body/index.js:155:17)
at getRawBody (/usr/src/app/node_modules/raw-body/index.js:108:12)
at read (/usr/src/app/node_modules/body-parser/lib/read.js:77:3)
at jsonParser (/usr/src/app/node_modules/body-parser/lib/types/json.js:135:5)
at Layer.handle [as handle_request] (/usr/src/app/node_modules/express/lib/router/layer.js:95:5)
at trim_prefix (/usr/src/app/node_modules/express/lib/router/index.js:317:13)
at /usr/src/app/node_modules/express/lib/router/index.js:284:7
at Function.process_params (/usr/src/app/node_modules/express/lib/router/index.js:335:12)
at next (/usr/src/app/node_modules/express/lib/router/index.js:275:10)
at urlencodedParser (/usr/src/app/node_modules/body-parser/lib/types/urlencoded.js:100:7)
::ffff:10.42.5.126 - - [27/Mar/2019:11:55:19 +0000] "POST / HTTP/1.1" 413 1109 "-" "Go-http-client/1.1"
PayloadTooLargeError: request entity too large
at readStream (/usr/src/app/node_modules/raw-body/index.js:155:17)
at getRawBody (/usr/src/app/node_modules/raw-body/index.js:108:12)
at read (/usr/src/app/node_modules/body-parser/lib/read.js:77:3)
at jsonParser (/usr/src/app/node_modules/body-parser/lib/types/json.js:135:5)
at Layer.handle [as handle_request] (/usr/src/app/node_modules/express/lib/router/layer.js:95:5)
at trim_prefix (/usr/src/app/node_modules/express/lib/router/index.js:317:13)
at /usr/src/app/node_modules/express/lib/router/index.js:284:7
at Function.process_params (/usr/src/app/node_modules/express/lib/router/index.js:335:12)
at next (/usr/src/app/node_modules/express/lib/router/index.js:275:10)
at urlencodedParser (/usr/src/app/node_modules/body-parser/lib/types/urlencoded.js:100:7)
[...]

PR #1149, opened 2019-04-03
#1132 nodejs environment, increase body size

JannikZed:
This will increase the maximum body-size in the nodejs environment to 1mb, to reflect the NATS maximum message size as written in #1132 
Still some more testing needed.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1149)


soamvasani:
LGTM.  (What's the default limit?  I tried googling but couldn't find it easily.)

JannikZed:
Well I understand the documentation of the body-parser like this:

The maximum request body size is 100kb: 

``` 
limit
Controls the maximum request body size. If this is a number, then the value specifies the number of bytes; if it is a string, the value is passed to the bytes library for parsing. Defaults to '100kb'.
``` 
https://www.npmjs.com/package/body-parser#limit

JannikZed:
@soamvasani what do you think? Let's merge? :)

soamvasani:
Thanks for the reminder, @JannikZed.  Yes, I'll merge this.

PR #1152, opened 2019-04-05
Enable go module support for go environment

### LINKED ISSUE #1130: Support go module for go env
Looks like go module become more stable and an official way to manage dependencies in go, it seems good timing for us to adopt it in go env.

life1347:
Fix #1130

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1152)


life1347:
Find an [issue](https://github.com/golang/go/issues/31951) related to go plugin & go mod, we should not merge this PR before the problem is resolved/answered.

Issue #1155, opened 2019-04-09
Setting affinity/tolerations on fission core deployments


davidsmf:
There's already issues (#604, #644) for adding affinity and tolerations on the functions.
However, we'd be interested in all of the core-functionality from fission having affinity support. That way we can run it on a node without so much load.

Would people want it to be configured once for all of the deployments, or separately for each so you could have separate tolerations or affinity for the executor and the controller?

vishal-biyani:
Closed by https://github.com/fission/fission/pull/1170

PR #1156, opened 2019-04-09
Add support for setting nodeSelector/affinity/tolerations on all depl…

davidsmf:
A simple solution to #1155 all of the deployments will have the same affinity etc.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1156)

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -163,6 +163,18 @@ spec:
       - name: config-volume
         configMap:
           name: feature-config
+{{- if .Values.controller.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.controller.nodeSelector | indent 8 }}
+    {{- end }}
+    {{- if .Values.controller.tolerations }}
+      tolerations:
+{{ toYaml .Values.controller.tolerations | indent 8 }}
+    {{- end }}
+    {{- if .Values.controller.affinity }}
+      affinity:
+{{ toYaml .Values.controller.affinity | indent 8 }}
+    {{- end }}
 ---
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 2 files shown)


life1347:
Hi @davidsmf , thanks for the PR.

Helm failed due to there is no variable definition in values.yaml

```
Error: render error in "fission-all/templates/deployment.yaml": template: fission-all/templates/deployment.yaml:166:14: executing "fission-all/templates/deployment.yaml" at <.Values.controller.n...>: can't evaluate field nodeSelector in type interface {}
```

And since for most users don't need to use affinity, here are some suggestions to address the problem

1. Adding the following config to `values.yaml` (change `controller` to `coreComponmentPodConfig`)

```
#coreComponmentPodConfig:
#  affinity:
#  tolerations:
#  nodeSelector:
```

2. Add some comment to explain the setting would be very helpful for users. Then users will know what the setting is for and uncomment setting when needed.
3. Add example setting for user to follow with

For the `render error` problem, you can check the `coreComponmentPodConfig` exists first so that helm won't break even the variables are not existing.

```
{{- if .Values.coreComponmentPodConfig }}
  {{- if .Values.coreComponmentPodConfig.nodeSelector }}
  ......
{{- end}}
```

soamvasani:
@davidsmf thanks for the PR!  Will you be addressing @life1347 's comments?  Please let us know if you have any questions.

davidsmf:
Sorry, been busy with a newborn, I should be able to address them, but not
for a couple of weeks.

David

soamvasani:
Congrats on the newborn :smile: :baby:

life1347:
After #1170 we are able to set affinity on all core components. Close this one.
Still thanks for all your contribution and congrats on the newborn 🎉

PR #1163, opened 2019-04-15
Change log level in executor for better log reading/troubleshooting

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1163)


commonrbacutil.go
```diff
@@ -122,14 +122,14 @@ func AddSaToRoleBindingWithRetries(logger *zap.Logger, k8sClient *kubernetes.Cli
 	for i := 0; i < MaxRetries; i++ {
 		_, err = k8sClient.RbacV1beta1().RoleBindings(roleBindingNs).Patch(roleBinding, types.JSONPatchType, patchJson)
 		if err == nil {
-			logger.Info("patched rolebinding",
+			logger.Error("patched rolebinding",
```
vishal-biyani: If err == nil - then logging it as error seems a bit odd to me

life1347: oops

life1347: fixed.

PR #1166, opened 2019-04-22
Fix the logger not working

darkgerm:
#### Problem
- logger can't create symlink because `/var/log/fission` does not exist. The test can pass because `/var/log` is a `hostPath` volume and `/var/log/fission` already exists on all nodes.
- fluentbit cannot open database `/var/log/fission/flb_kube.db`

#### Solution
- Use an init container to create `/var/log/fission`.
- Create `fissionSymlinkPath` if it does not exist when logger start. It's in case if the log path changed in the future or someone run logger directly.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1166)

### DIFF
--- a/charts/fission-all/templates/fluentbit.yaml
+++ b/charts/fission-all/templates/fluentbit.yaml
@@ -32,6 +32,14 @@ spec:
       labels:
         svc: logger
     spec:
+      initContainers:
+        - name: init
+          image: busybox
+          command: ['mkdir', '-p', '/var/log/fission']
+          volumeMounts:
+            - name: container-log
+              mountPath: /var/log/
+              readOnly: false
       containers:
         - name: logger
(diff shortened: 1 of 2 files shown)


charts/fission-all/templates/fluentbit.yaml
```diff
@@ -32,6 +32,14 @@ spec:
       labels:
         svc: logger
     spec:
+      initContainers:
```
soamvasani: Is this really needed?  Is the change in logger.go not enough?

darkgerm: Yes, because fluentbit also needs it.
`charts/fission-all/config/fluentbit.conf` Line 13
```
DB /var/log/fission/flb_kube.db
```

vishal-biyani:
As discussed in weekly meeting - merging this.
Thanks @darkgerm for the fix!

PR #1167, opened 2019-04-25
Add simple anonymous usage metrics

soamvasani:
This change reports total function call count to an analytics url once a day.
It's done behind the existing analytics flag.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1167)


vishal-biyani:
@soamvasani Can you please fix below:

```
$ hack/verify-gofmt.sh
!!! 'gofmt -s' needs to be run on the following files: 
./router/analytics.go
./router/metrics.go
The command "hack/verify-gofmt.sh" failed and exited with 1 during .
```

soamvasani:
@vishal-biyani can we make sure we put a note in the release so people are informed about this? And as usual, `--set analytics=false` will switch it off.

Issue #1174, opened 2019-05-08
Dotnet arguments example fails 


CanerPatir:
syntax error on https://github.com/fission/fission/blob/master/examples/dotnet/arguments.cs

vishal-biyani:
Hi, @CanerPatir Thanks - can you please paste the error?

CanerPatir:
Fatal error: Failed to execute HTTP request: Get http://127.0.0.1:52850/fission-function/hello2: context deadline exceeded

life1347:
Fix by #1175

PR #1175, opened 2019-05-08
#1174

CanerPatir:


life1347:
@CanerPatir Thanks for the PR. Would you mind also addressing the same problem in dotnet20 [example](https://github.com/fission/fission/blob/master/examples/dotnet20/arguments.cs)?

CanerPatir:
Yes, this is also addressing dotnet20 too. I will be fixing that in an additional commit as well

PR #1180, opened 2019-05-10
router analytics -- close http response body

soamvasani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1180)

### DIFF
--- a/router/analytics.go
+++ b/router/analytics.go
@@ -3,6 +3,7 @@ package router
 import (
 	"bytes"
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


router/analytics.go
```diff
@@ -56,6 +57,12 @@ func (a *Analytics) run() {
 			continue
 		}
 
-		_, _ = http.Post(a.url, "application/json", bytes.NewReader(msgbytes))
+		resp, err := http.Post(a.url, "application/json", bytes.NewReader(msgbytes))
+		if err != nil && resp != nil && resp.Body != nil {
+			// read body to end and close it; otherwise we
+			// leak some resources in the http transport
+			_, _ = ioutil.ReadAll(resp.Body)
```
life1347: actually, Close() drains data automatically. So maybe we can remove ReadAll here?

soamvasani: Didn't know that, thanks.  When is doEarlyClose set/unset?  (also, ioutil.Discard seems like a better way to write this)

life1347: It happens at the server side for connection reuse. https://github.com/golang/go/blob/2fd97ee25e8a623e460aff26b4c422dec9532654/src/net/http/server.go#L1007-L1009

router/analytics.go
```diff
@@ -56,6 +56,10 @@ func (a *Analytics) run() {
 			continue
 		}
 
-		_, _ = http.Post(a.url, "application/json", bytes.NewReader(msgbytes))
+		resp, err := http.Post(a.url, "application/json", bytes.NewReader(msgbytes))
+		if err == nil && resp != nil && resp.Body != nil {
```
vishal-biyani: Nitpick, from Go source (https://golang.org/src/net/http/client.go) `When err is nil, resp always contains a non-nil resp.Body.` - so checking `resp.Body` is not nil is redundant.

PR #1195, opened 2019-05-30
Analytics bugfix

soamvasani:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1195)

### DIFF
--- a/pkg/router/analytics.go
+++ b/pkg/router/analytics.go
@@ -17,7 +17,7 @@ type (
 		url string
 	}
 	AnalyticsData struct {
-		id                string
+		ID                string `json:"ID"`
 		FunctionCallCount uint64 `json:"FunctionCallCount"`
 	}
 )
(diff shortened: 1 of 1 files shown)


router/analytics.go
```diff
@@ -49,7 +49,7 @@ func (a *Analytics) run() {
 	ticker := time.NewTicker(24 * time.Hour)
 	for range ticker.C {
 		msg := a.gatherData()
-		msg.id = a.id
+		msg.ID = a.id
```
life1347: You may also want to handle err at L59.

soamvasani: I don't understand? The http post error is already handled?

life1347:
```
resp, err := http.Post(a.url, "application/json", bytes.NewReader(msgbytes))
```
No, right now we only handle the resp not err. If the err is not so critical, maybe we can ignore it?
```
resp, _ := xxxxx
```

soamvasani: Oh I see, good point and thanks for fixing it.

PR #1203, opened 2019-06-04
Fallback to get user home directory from env

life1347:
A user reported there is an error while trying to use fission CLI
```
$ fission env create --name nodejs --image fission/node-env:1.2.1
Fatal error: Could not get the current users directory: user: Current not implemented on darwin/amd64
```
This problem may be related to the cross-compilation problem as described in comments. So in PR tries to get home directory path from env as fallback method.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1203)


soamvasani:
Thanks, LGTM

PR #1206, opened 2019-06-14
modify code to make log collection comprehensive

cocoifly10:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1206)

### DIFF
--- a/pkg/publisher/webhookPublisher.go
+++ b/pkg/publisher/webhookPublisher.go
@@ -6,7 +6,7 @@ you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
-    
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/pkg/publisher/webhookPublisher.go
+++ b/pkg/publisher/webhookPublisher.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2017 The Fission Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+    
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


cocoifly10:
@life1347 Hope you review the code when you have time, thank you!

pkg/publisher/webhookPublisher.go
```diff
@@ -80,52 +80,53 @@ func (p *WebhookPublisher) svc() {
...
+		if ce := p.logger.Check(level, msg); ce != nil {
+			ce.Write(fields...)
+		}
+	}()
 
 	var buf bytes.Buffer
 	buf.WriteString(r.body)
 
 	// Create request
 	req, err := http.NewRequest("POST", url, &buf)
```
life1347: Though it's not included in changes, would you mind to change `POST` into `http.MethodPost`?

pkg/publisher/webhookPublisher.go
```diff
@@ -80,52 +80,53 @@ func (p *WebhookPublisher) svc() {
...
 		req.Header.Set(k, v)
 	}
 
 	// Make the request
 	resp, err := http.DefaultClient.Do(req)
-
-	// All done if the request succeeded with 200 OK.
-	if err == nil && resp.StatusCode == 200 {
+	var body []byte
+	if err == nil {
```
life1347: Though I can understand the whole code flow, however, 

1. two `if err == nil` here may confuse people when code reading.
2. people may miss the err of `ioutil.ReadAll` inside block. 

How about we check the `err != nil` first, so that we can move `ioutil.ReadAll` out from the block?

pkg/publisher/webhookPublisher.go
cocoifly10: I have change POST int http.MethodPost in the latest commit

pkg/publisher/webhookPublisher.go
cocoifly10: Code has been changed as you suggested, hope u review again, thx!

pkg/publisher/webhookPublisher.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2017 The Fission Authors.
-
```
life1347: Before the CI pass (https://travis-ci.org/fission/fission/builds/549017388), could you restore the blank line accidentally deleted here?

pkg/publisher/webhookPublisher.go
```diff
@@ -80,50 +77,51 @@ func (p *WebhookPublisher) svc() {
...
 	resp, err := http.DefaultClient.Do(req)
-
-	// All done if the request succeeded with 200 OK.
-	if err == nil && resp.StatusCode == 200 {
-		resp.Body.Close()
-		return
-	}
-
-	// Log errors
+	var body []byte
```
life1347: How about moving this line jsut right before `body, err = ioutil.ReadAll(resp.Body)`? 
```
} else {
    var body []byte
    body, err = ioutil.ReadAll(resp.Body)
    ...
```

pkg/publisher/webhookPublisher.go
cocoifly10: Sorry, I just saw your mesage. Restored already!

pkg/publisher/webhookPublisher.go
cocoifly10: Have done as you suggest.

life1347:
LGTM

life1347:
Merged. Thanks for the PR 🎉

PR #1207, opened 2019-06-19
Reuse docker maven build cache in JVM environment build process

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1207)

### DIFF
--- a/environments/jvm/Dockerfile
+++ b/environments/jvm/Dockerfile
@@ -1,7 +1,13 @@
 FROM maven:3.5-jdk-8 as BUILD
-COPY src /usr/src/myapp/src
-COPY pom.xml /usr/src/myapp
-RUN mvn -f /usr/src/myapp/pom.xml clean package
+WORKDIR /usr/src/myapp/
+
+# To reuse the build cache, here we split maven dependency
+# download and package into two RUN commands to avoid cache invalidation.
+COPY pom.xml .
+RUN mvn dependency:go-offline
+
+COPY src /usr/src/myapp/src/
+RUN mvn package
 
 FROM openjdk:8-jdk-alpine
 VOLUME /tmp


environments/jvm/Dockerfile
```diff
@@ -1,7 +1,13 @@
 FROM maven:3.5-jdk-8 as BUILD
-COPY src /usr/src/myapp/src
-COPY pom.xml /usr/src/myapp
-RUN mvn -f /usr/src/myapp/pom.xml clean package
+WORKDIR /usr/src/myapp/
+
+# To reuse the build cache, here we split maven dependency
+# download and package into two RUN commands to avoid cache invalidation.
+COPY pom.xml .
+RUN mvn dependency:go-offline
```
vishal-biyani: I am sorry I don't understand how this will avoid cache invalidation. AFAIK - go-offline is noly downloading the libraries offline

life1347:
The idea behind the scenes is that when we modify the Server.java the checksum of the file changed and caused the docker image re-build the image layer and all subsequent layers.

```
L1: COPY src /usr/src/myapp/src
L2: COPY pom.xml /usr/src/myapp	
L3: RUN mvn -f /usr/src/myapp/pom.xml clean package
```

Docker checks the file checksum in layer L1, since we modified the file, the existing build layer (L1) plus all subsequent layers (L2, L3) are considered out-of-date and being invalidated. That's why docker downloads the maven dependencies every time even no dependencies changed.

Since most of the time we don't change the `pom.xml`, as long as we download dependencies before coping the server.java then we can re-use the cache layer (L1, L2). The file checksum in L3 changed, docker will rebuild L3 & L4.

```
L1: COPY pom.xml .
L2: RUN mvn dependency:go-offline

L3: COPY src /usr/src/myapp/src/
L4: RUN mvn package
```

Here are some ref links 
1. how to reuse maven build cache: https://medium.com/@nieldw/caching-maven-dependencies-in-a-docker-build-dca6ca7ad612
2. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache

vishal-biyani:
LGTM - thanks this is useful!

PR #1210, opened 2019-06-19
Fix fetcher of newdeploy pod error out when starting up.

life1347:
1. Retry to fetch package instead of error out directly
2. fix readiness probe path

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1210)


soamvasani:
What kind of errors benefit from retries here?

life1347:
Mostly for the scenario that Istio is enabled, it takes a longer time for fetcher to restart than simply do retry.

soamvasani:
As discussed in the meeting, it would be good to retry only on the errors we get in that case (e.g. connection refused)

PR #1211, opened 2019-06-19
Fix executor panic problem when specialization timeout

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1211)


soamvasani:
LGTM, thanks!

Issue #1147, opened 2019-04-03
applying fission spec should warn if the config maps that are referenced in spec, are not present


viveksinghggits:
In current scenario if we apply spec, to lets say create a function, and that spec is referencing some config maps that are not present in the cluster. We should have a warning that referenced config maps are not present like we get if the fission environment that is being referenced is not present.

In current implementation it becomes tedious to debug because even the POD for that function doesn't get into `running` state and because of that we can not even see the logs of that POD. The POD fails, and describing it returns below response
```
Warning  Unhealthy  5m24s (x2 over 5m26s)  kubelet, <host>  Readiness probe failed: Get http://<IP>:8000/healthz: dial tcp <IP>:8000: connect: connection refused
  Normal   Pulled     4m41s (x4 over 5m28s)  kubelet, <host>  Container image "fission/fetcher:0.12.0" already present on machine
  Normal   Created    4m40s (x4 over 5m28s)  kubelet, <host>  Created container
  Normal   Started    4m40s (x4 over 5m27s)  kubelet, <host>  Started container
  Warning  BackOff    28s (x31 over 5m23s)   kubelet, <host>  Back-off restarting failed container
```

hd40910:
@vishal-biyani  @mehuls Can we have this in 1.3 ?

PR #1214, opened 2019-06-26
Configmaps/secrets in function exist check

vishal-biyani:
Fixes https://github.com/fission/fission/issues/1147

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1214)

### DIFF
--- a/pkg/fission-cli/spec.go
+++ b/pkg/fission-cli/spec.go
@@ -34,6 +34,7 @@ import (
 	"github.com/pkg/errors"
 	uuid "github.com/satori/go.uuid"
 	"github.com/urfave/cli"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	fv1 "github.com/fission/fission/pkg/apis/fission.io/v1"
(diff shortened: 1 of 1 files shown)


pkg/fission-cli/spec.go
```diff
@@ -343,6 +343,27 @@ func (fr *FissionResources) validate() error {
 			packages[mapKey(pkgMeta)] = true
 		}
 
+		client := util.GetApiClient(c.GlobalString("server"))
+		for _, cm := range f.Spec.ConfigMaps {
+			c, err := client.ConfigMapGet(&metav1.ObjectMeta{
```
life1347: Since the parameter pass-in is already named `c`, how about changing the variable name here to prevent confusion?

pkg/fission-cli/spec.go
```diff
@@ -343,6 +343,27 @@ func (fr *FissionResources) validate() error {
...
+				Name:      cm.Name,
+				Namespace: cm.Namespace,
+			})
+			if c == nil && err != nil {
```
life1347: The error may not equal to non-exists error, we need to check it before printing log.
```
if k8serrors.IsNotFound(err) {
   .....
```

pkg/fission-cli/spec.go
```diff
@@ -343,6 +343,27 @@ func (fr *FissionResources) validate() error {
...
+				log.Warn(fmt.Sprintf("Configmap %s is referred in the spec but not present in the cluster", cm.Name))
+			}
+		}
+
+		for _, s := range f.Spec.Secrets {
+			c, err := client.SecretGet(&metav1.ObjectMeta{
+				Name:      s.Name,
+				Namespace: s.Namespace,
+			})
+			if c == nil && err != nil {
```
life1347: same above

vishal-biyani:
@life1347 Fixed review comments.

life1347:
LGTM

PR #1223, opened 2019-07-07
Fix nil pointer when CLI unable to get server version

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1223)


soamvasani:
LGTM, thanks

Issue #1164, opened 2019-04-15
Update secret/configmaps in function pod if source changed


life1347:
It would be helpful for fetcher to check and pull changes of secrets/configmaps periodically.
#1161

soamvasani:
We'd have to be careful not to race with user code that's reading the config or secret.

vishal-biyani:
Just updating the conversation I had with Soam here - we should do something similar to PR https://github.com/fission/fission/pull/1116 which updates a function when an image in the environment changes. We can watch all secrets & configmaps which are related to a function (Do we use labels to identify configmaps related to a function?) and then update function if there is a change in configmap/secret.

hd40910:
@mehuls #1159 is related .
@vishal-biyani  Do we have release date planned for 1.3 ?

PR #1224, opened 2019-07-08
Function Update if config/secret changes

vishal-biyani:
If a secret/configmap value changes - the function should update with the new value. 

In the case of pool manager, this works as the specialized pod always picks up the new value from configmap/secret. 

In the case of newdeploy - this controller terminates the old pod which in turn creates new pods with new values of configmap/secret.

Fixes https://github.com/fission/fission/issues/1164 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1224)


pkg/executor/newdeploy/newdeploymgr.go
```diff
@@ -223,6 +224,37 @@ func (deploy *NewDeploy) GetFuncSvc(ctx context.Context, metadata *metav1.Object
...
+		LabelSelector: labels.Set(funcLabels).AsSelector().String(),
+	})
+
+	if err != nil {
+		return err
+	}
+
+	for _, po := range podList.Items {
+		err := deploy.kubernetesClient.CoreV1().Pods(po.ObjectMeta.Namespace).Delete(po.ObjectMeta.Name, &metav1.DeleteOptions{})
+		if err != nil {
```
life1347: Need to check the error type here, if it's a non-exist error then we should not return it.

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,185 @@
...
+						err := ndm.RecycleFuncPods(logger, f)
+						if err != nil {
+							logger.Error("Failed to recycle pods for function after configmap changed",
+								zap.Error(err),
+								zap.Any("function", f))
+						}
+					}
+
+					if f.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType == fv1.ExecutorTypePoolmgr {
+						// Future implementation
```
life1347: We need to document the different behavior of newdeploy/poolmgr when facing secret/configmap update in `https://docs.fission.io/usage/access-secret-cfgmap-in-function/` so that users won't confuse about this.

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+	})
+	return store, controller
+
+}
+
+func getSecretRelatedFuncs(logger *zap.Logger, m *metav1.ObjectMeta, fissionClient *crd.FissionClient) []fv1.Function {
+	funcList, err := fissionClient.Functions(metav1.NamespaceAll).List(metav1.ListOptions{})
+	if err != nil {
+		logger.Error("Error getting functions while updating secrets", zap.Error(err), zap.Any("configmap", m))
```
life1347: if error happened we should return here

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,185 @@
...
+
+		},
+	})
+	return store, controller
+}
+
+func getConfigmapRelatedFuncs(logger *zap.Logger, m *metav1.ObjectMeta, fissionClient *crd.FissionClient) []fv1.Function {
+	funcList, err := fissionClient.Functions(metav1.NamespaceAll).List(metav1.ListOptions{})
+	if err != nil {
+		logger.Error("Error getting functions while updating configmap", zap.Error(err), zap.Any("configmap", m))
```
life1347: if error happened we should return here

vishal-biyani:
Thinking more about the implementation and after looking at [issue](https://github.com/kubernetes/kubernetes/issues/13488) I think deleting pods like this is a pretty bad idea. Especially for new deploy executor - where if all pods are deleted and if it takes time for new pod to come up then the latency is higher than one would expect from new deploy function.

The closest solution I can think of is similar to https://github.com/kubernetes/kubernetes/pull/76062 - but that is only on the client side and something similar would have to be done on the server side within Fission.

life1347:
Yeah, agree that be a problem if under heavy workloads.

Of course, it's better to switch to the new function version. But consider the problem you mentioned, I think we should consider how to migrate traffic from old fn to the new one more smoothly. 

How about we add SHA of secret/cfgmap to deployment, then we simply update that label to trigger a rolling update? Kubernetes will handle the instance rolling update for us.

And in fact, it adds extra protection to users if their newer function won't come up.

What do you think?

vishal-biyani:
I think the migration of traffic from old to a new pod is taken care of by `TerminationGracePeriod` - it will ensure that old traffic is served and routing new traffic to the new pod is taken care by Service (At least in new deployment). 

SHA of configmap - the reason I don't like because we can potentially have multiple configmaps/secrets per function which creates a lot of extra decisions.

I am inclining towards adding a new annotation/ ENV variable - the value will a simple timestamp called `ConfigLastUpdated` or something like that. Anytime we have a configmap/secret update - we simply update the TS and it triggers the rolling update.

life1347:
I like the idea of having `timestamp`. If I remember correctly the annotation update won't trigger an update, we have to add it to ENV.

vishal-biyani:
The PR is now changed and ready for re-review based on the discussion above. The implementation is different for Pool manager vs. new deploy executor and is described below:

1) For new deploy executor - a timestamp ENV has been added to function deployment. This env is updated when a secret/configmap changes. This causes rolling update (Currently with MaxUnavailable 25% and MaxSurge 100%) and the pods in deployment are replenished and hence have new configmaps and secrets when they start.

2) For pool manager, the specialized pod is deleted. Existing requests will be served based on PreStop hook and `TerminationGracePeriod` but all new requests will go to a pod which is specialized from the pool. I want to monitor this more and make changes as we see any needs. We can not do a rolling update here as the pod is not under any deployment and even if it did - we would need to specialize specifically again on start!

Also doc update PR raised: https://github.com/fission/docs.fission.io/pull/140

pkg/executor/poolmgr/gpm.go
```diff
@@ -130,6 +132,41 @@ func (gpm *GenericPoolManager) Run(ctx context.Context) {
...
+	podList, err := gpm.kubernetesClient.CoreV1().Pods(metav1.NamespaceAll).List(metav1.ListOptions{
+		LabelSelector: labels.Set(funcLabels).AsSelector().String(),
+	})
+
+	if err != nil {
+		return err
+	}
+
+	for _, po := range podList.Items {
+		err := gpm.kubernetesClient.CoreV1().Pods(po.ObjectMeta.Namespace).Delete(po.ObjectMeta.Name, &metav1.DeleteOptions{})
```
life1347: We also need to invalidate the cache so that executor will specialize a new pod for function.
One thing I worry about `deleting pod` here is when a configmap/secret is comsumed by many functions, once it get updated poolmanager will specialize tons of new pods, thus cluster runs up all available resource in the end if `old` pods are still in `Terminating` state.

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,180 @@
...
+			if (secret.Name == m.Name) && (secret.Namespace == m.Namespace) {
+				relatedFunctions = append(relatedFunctions, f)
+				break
+			}
+		}
+	}
+	return relatedFunctions, nil
+}
+
+func recyclePods(logger *zap.Logger, funcs []fv1.Function, ndm *nd.NewDeploy, gpm *gpm.GenericPoolManager) {
```
life1347: I don't really like the term `recycle` here since we are just trying to do the rolling update. Any thought? `rolloutNewPods`?

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, f := range funcs {
+		var err error
+		if f.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType == fv1.ExecutorTypeNewdeploy {
+			err = ndm.RecycleFuncPods(logger, f)
+		}
+		if f.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType == fv1.ExecutorTypePoolmgr {
```
life1347: Use switch here would be better

pkg/executor/cms/cmscontroller.go
```diff
@@ -0,0 +1,180 @@
...
+	listWatch := cache.NewListWatchFromClient(kubernetesClient.Core().RESTClient(), "secrets", metav1.NamespaceAll, fields.Everything())
+	store, controller := cache.NewInformer(listWatch, &apiv1.Secret{}, resyncPeriod, cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) {},
+		DeleteFunc: func(obj interface{}) {},
+		UpdateFunc: func(oldObj interface{}, newObj interface{}) {
+			oldS := oldObj.(*apiv1.Secret)
+			newS := newObj.(*apiv1.Secret)
+			if oldS.ObjectMeta.ResourceVersion != newS.ObjectMeta.ResourceVersion {
+				if newS.ObjectMeta.Namespace != "kube-system" {
+					logger.Info("Secret changed",
```
life1347: Info level is a little annoying, how about changing it to `Debug`

pkg/executor/cms/cmscontroller.go
vishal-biyani: I think Rollout is a bit confusing with Kubernetes's rollout and also in case of Pool Manager - we are simply deleting and not rolling out. I want to convey that we are taking old pods - fitting them with new configmaps/secrets and bringing them back to service. WDYT?

life1347:
LGTM

PR #1241, opened 2019-07-18
make forloop without useless index

mjaow:
Use for loop with item (not index)

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1241)

### DIFF
--- a/pkg/executor/poolmgr/gpm.go
+++ b/pkg/executor/poolmgr/gpm.go
@@ -267,11 +267,10 @@ func (gpm *GenericPoolManager) eagerPoolCreator() {
 		// creating pools for envs that are actually used by functions.  Also we might want
 		// to keep these eagerly created pools smaller than the ones created when there are
 		// actual function calls.
-		for _, item := range envs.Items {
-			env := item
+		for _, env := range envs.Items {
 			// Create pool only if poolsize greater than zero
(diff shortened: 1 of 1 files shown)


pkg/executor/poolmgr/gpm.go
```diff
@@ -267,11 +267,11 @@ func (gpm *GenericPoolManager) eagerPoolCreator() {
 		// creating pools for envs that are actually used by functions.  Also we might want
 		// to keep these eagerly created pools smaller than the ones created when there are
 		// actual function calls.
-		for i := range envs.Items {
-			env := envs.Items[i]
+		for _, item := range envs.Items {
```
vishal-biyani: We could directly get env in the loop:

```
    for _, env := range envs.Items {
```

mjaow: right,i'm done

life1347:
No, we cannot remove the index here. This is in intentional for doing that. As the pointer of the environment is the address to the loop variable, however, it's value will be changed as loop iterate to next item.

life1347:
For details, see https://github.com/fission/fission/pull/221

life1347:
Close this as the problem mentioned above. But @mjaow thanks for your PR!

PR #1245, opened 2019-07-24
Add swagger (OpenAPI 2.0) support

### LINKED ISSUE #85: Swagger API spec is incomplete/outdated
Hi,

I noticed in https://github.com/fission/fission/blob/master/Documentation/Architecture.md that a swagger spec is mentioned.
However api/swagger.json does not exist (anymore ?)

In my view the project would benefit from formally specified API's between the components.
Say one would like to build a fission dashboard then having to go through the Go code to find the URL's is rather cumbersome. Its much easier to generate a javascript client from a swagger spec.

And if you would like to take it even further then its possible to generate a the server stubs, the client stubs and even the command-line client (e.g. using http://swagger.io/tools/ or https://goa.design/ )

Last but not least: formally specifying the API's increase security as it stops invalid inputs early in the process :-)

Cheers,
Hans

life1347:
This PR aims to add swagger support to the controller for users to write their own client.

Screen Shot 2019-07-25 at 3 40 21 AM

For those who cannot wait for the PR to be merged: 

1. Go https://editor.swagger.io/
2. File > Import URL > Paste `https://gist.githubusercontent.com/life1347/1fd9b1e55c9897e405f78bac5bcf7c69/raw/45364c36636fd52f2c9f3a97cd926419c2f99604/Fission%2520Swagger%2520API
`

Fix #85

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1245)


pkg/apis/fission.io/v1/tool/swagger_type_docs.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+
+	flag "github.com/spf13/pflag"
+	"k8s.io/klog"
```
soamvasani: Why are we adding this?

life1347: This tool parses all struct fields and its command for go-restful. So that we don't need to write the go-restful description manually.

PR #1249, opened 2019-07-26
Allow to set deployment config uid during initialization

### LINKED ISSUE #1232: Datatype for UID in fission-deployment-config.yaml
We are using specs for doing all the deployments on fission. 
Building the specs on the CI server and then pushing the created specs on PROD for deployment. 
We recreate the specs dynamically for every build, which obviously creates new UID in fission-deployment-config.yaml

With this approach when we deploy the incremental changes to PROD, it fails with duplicate resource error. 

To overcome this, we are replacing the UID in fission-deployment-config.yaml with function name post specs are created. 
In below example replacing **uid: b16b8e4d-cb83-4197-8892-9cdaa7eb5a7f** with **uid: test-function-name**

Currently this approach is working for us.
So just wanted to understand, if there are any validations for UID which may fail in future? 
Any limitations? 
Will this approach has any risks? 

<pre>
$ cat fission-deployment-config.yaml.original
apiVersion: fission.io/v1
kind: DeploymentConfig
name: customer-status-csm
uid: b16b8e4d-cb83-4197-8892-9cdaa7eb5a7f

$ cat fission-deployment-config.yaml
apiVersion: fission.io/v1
kind: DeploymentConfig
name: customer-status-csm
uid: test-function-name

</pre>

life1347:
Fix #1232

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1249)

### DIFF
--- a/pkg/fission-cli/main.go
+++ b/pkg/fission-cli/main.go
@@ -272,11 +272,12 @@ func NewCliApp() *cli.App {
 	// specs
 	specDirFlag := cli.StringFlag{Name: "specdir", Usage: "Directory to store specs, defaults to ./specs"}
 	specNameFlag := cli.StringFlag{Name: "name", Usage: "(optional) Name for the app, applied to resources as a Kubernetes annotation"}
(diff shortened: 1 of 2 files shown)


vishal-biyani:
LGTM

Issue #1258, opened 2019-08-01
fission chart doesn't allow empty repository tag in values.yaml


parauliya:
**Fission/Kubernetes version**
<pre>
$ fission --version
client:
  fission/core:
    gitcommit: 94b157137d45f311dba9affa1ad6ee79b56a73cb
    builddate: "2019-07-29T08:21:45Z"
    version: 1.4.1
server:
  fission/core:
    gitcommit: unknown
    builddate: unknown
    version: unknown

$ kubectl version
{{{
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.1", GitCommit:"4485c6f18cee9a5d3c3b4e523bd27972b1b53892", GitTreeState:"clean", BuildDate:"2019-07-18T09:18:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:32:14Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
}}}

**Describe the bug**
Currently fission is not able to installed with minikube cluster and pods are not in running state and showing an error of InvalidImageName for the image name "index.docker.io/minikube/fission-all".
The above name is being built as per follows 
{{{
image: "{{ .Values.repository }}/{{ .Values.image }}:{{ .Values.imageTag }}"
}}}
The default value of repository is "index.docker.io" in charts/fission-all/values.yaml

Now if there is a case where repository field is not required (like minikube) and left empty the above image name consist "/" in the start of the name of image which makes the image name incorrect. For e.g. while installing fission with minikube repository field is not needed for the image name. Image name in this case is "minikube/fission-all" . But once you removed the default value of repository field in the charts/fission-all/value.yaml file the image name came out as "/minikube/fission-all". 
Because of the incorrect docker image name pods got failed with the error like InvalidImageName.

**To Reproduce**
Try to install fission with minikube and leave the repository field empty in charts/fission-all/values.yaml file.

**Expected result**
Fission should be installed successfully.

**Actual result**
Fission is not being installed. Pods are not in running state.

PR #1259, opened 2019-08-01
Issue #1258: Allow empty repository tag in chart values.yaml

parauliya:
…yaml

This fix allows the empty repository tag in values.yaml file.
The changes are done in the templates of deployment and fluentbit where repository tag will be added
in the docker image name only when specified with some value in case of empty repository tag image name
contain the name of image and tag only.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1259)

### DIFF
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -31,3 +31,13 @@ canary:
   {{- printf "\n" -}}
 {{- end -}}
 
+{{- define "fission-all-bundleImage" -}}
+{{- $repositoryName := .Values.repository -}}
+{{- $imageName := .Values.image -}}
+{{- $imageTag := .Values.imageTag | toString -}}
+{{- if .Values.repository }}
+        {{- printf " %s/%s:%s" $repositoryName $imageName $imageTag -}}
+{{- else }}
+        {{- printf " %s:%s" $imageName $imageTag -}}
+{{- end }}
+{{- end -}}
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -34,7 +34,7 @@ canary:
 {{/*
 This template generates the image name for the deployment depending on the value of "repository" field in values.yaml file.
 */}}
-{{- define "fission-all-bundleImage" -}}
+{{- define "fission-bundleImage" -}}
 {{- if .Values.repository -}}
     {{ .Values.repository }}/{{ .Values.image }}:{{ .Values.imageTag }}
 {{- else -}}
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -31,13 +31,13 @@ canary:
   {{- printf "\n" -}}
 {{- end -}}
 
+{{/*
+This template generates the image name for the deployment depending on the value of "repository" field in values.yaml file.
+*/}}
 {{- define "fission-all-bundleImage" -}}
-{{- $repositoryName := .Values.repository -}}
-{{- $imageName := .Values.image -}}
-{{- $imageTag := .Values.imageTag | toString -}}
-{{- if .Values.repository }}
-        {{- printf " %s/%s:%s" $repositoryName $imageName $imageTag -}}
-{{- else }}
-        {{- printf " %s:%s" $imageName $imageTag -}}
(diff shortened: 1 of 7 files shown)


life1347:
@parauliya Thanks for the PR! I will take a look soon.

charts/fission-all/templates/deployment.yaml
```diff
@@ -128,7 +128,11 @@ spec:
     spec:
       containers:
       - name: controller
+{{ if .Values.repository }}
```
bhavin192: I think it would be more convenient to define a template in `_helpers.tpl` and include it multiple times.

life1347:
@parauliya Overall it looks good! And just like @bhavin192 said, we can use `_helpers.tpl` to replace the duplicate parts ([Reference](https://helm.sh/docs/chart_template_guide/#declaring-and-using-templates-with-define-and-template)). I am fine to merge this PR if you're not available for the further changes. Let me know what do you think :)

parauliya:
@life1347 I think @bhavin192's suggestion is considerable as the code would be more convenient and readable. But I need some time to make this change as right now I am busy with something else.
I'll do this change by tomorrow. is that ok?

charts/fission-all/templates/deployment.yaml
parauliya: sounds good. I'll do that by tomorrow.

life1347:
@paraspatidar Yes, sure of course. Take your time. :)

parauliya:
> Yes, sure of course. Take your time. :)

parauliya:
I have updated the code as per the @bhavin192's review comment.
Please have a look at it and merge this request to the main branch.
In case of any issue please let me know.

charts/fission-all/templates/_helpers.tpl
```diff
@@ -31,3 +31,13 @@ canary:
   {{- printf "\n" -}}
 {{- end -}}
 
+{{- define "fission-all-bundleImage" -}}
```
bhavin192: Have you considered something simpler like this one?
```yaml
{{- define "fission-all-bundleImage" -}}
{{- if .Values.repository -}}
    {{ .Values.repository }}/{{ .Values.image }}:{{ .Values.imageTag }}
{{- else -}}
    {{ .Values.image }}:{{ .Values.imageTag }}    
{{- end }}
{{- end -}}
```

charts/fission-all/templates/deployment.yaml
```diff
@@ -128,7 +128,7 @@ spec:
     spec:
       containers:
       - name: controller
-        image: "{{ .Values.repository }}/{{ .Values.image }}:{{ .Values.imageTag }}"
+        image: {{ template "fission-all-bundleImage" . }}
```
bhavin192: From Helm documentation

> It is considered preferable to use `include` over `template` in Helm templates simply so that the output formatting can be handled better for YAML documents.

https://helm.sh/docs/chart_template_guide/#the-include-function

charts/fission-all/templates/_helpers.tpl
parauliya: Yes, I tried this initially but this wasn't working but I have just tried it and it is working. May be I was doing something wrong or missing something. Thanks for the suggestion. I'll change it.

charts/fission-all/templates/deployment.yaml
parauliya: Replaced template with include.

parauliya:
I have done the changes as suggested by @bhavin192 .

charts/fission-all/templates/_helpers.tpl
```diff
@@ -31,3 +31,13 @@ canary:
   {{- printf "\n" -}}
 {{- end -}}
 
+{{/*
+This template generates the image name for the deployment depending on the value of "repository" field in values.yaml file.
+*/}}
+{{- define "fission-all-bundleImage" -}}
```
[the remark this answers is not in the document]
life1347: Please add a comment to the two values.yaml
```
## Leave it empty for using existing local image
```
To something like
```
## Image base repository
## Leave it empty for using existing local image
repository: index.docker.io
```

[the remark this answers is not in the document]
parauliya: Above both changes are incorporated

life1347:
LGTM, thanks!

PR #1271, opened 2019-08-13
fix a panic bug caused by err.Error()

cocoifly10:
fix bug: 
funciton errorHandler use err.Error(), that will cause panic, because errors.Wrapf("request returned failure: %v", resp.StatusCode) return nil
<pre>
panic: runtime error: invalid memory address or nil pointer dereference

/go/src/github.com/fission/fission/pkg/mqtrigger/messageQueue/kafka.go:256
</pre>

@life1347

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1271)

### DIFF
--- a/pkg/mqtrigger/messageQueue/kafka.go
+++ b/pkg/mqtrigger/messageQueue/kafka.go
@@ -214,7 +214,7 @@ func kafkaMsgHandler(kafka *Kafka, producer sarama.SyncProducer, trigger *fv1.Me
 	}
 	if resp.StatusCode != 200 {
 		errorHandler(kafka.logger, trigger, producer, url,
-			errors.Wrapf(err, "request returned failure: %v", resp.StatusCode))
+			fmt.Errorf("request returned failure: %v", resp.StatusCode))
 		return false
 	}
 	if len(trigger.Spec.ResponseTopic) > 0 {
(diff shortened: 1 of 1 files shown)


life1347:
https://travis-ci.org/fission/fission/builds/571294726
```
pkg/mqtrigger/messageQueue/kafka.go:217:4: undefined: fmt
The command "hack/runtests.sh" failed and exited with 2 during .
```

cocoifly10:
> https://travis-ci.org/fission/fission/builds/571294726
> 
> ```
> pkg/mqtrigger/messageQueue/kafka.go:217:4: undefined: fmt
> The command "hack/runtests.sh" failed and exited with 2 during .
> ```

@life1347 Pardon my careless! 
By the way, I'm wonderring how can I upgrade fission smoothly, so that client won't be affected.
I try to deploy 2 replicas in kubernetes, but requests only come to the newest one. Dose that means there must be a period  its functions will not act correctly every time fission upgrade?

life1347:
@cocoifly10 

Do you mean to have two fission installs in one kubernetes cluster or two replicas of core components?
```
I try to deploy 2 replicas in kubernetes, but requests only come to the newest one.
```

Currently, the major problem is that executor is a stateless service, so an executor pod has no way to retrieve all function pods information and needs to recreate all function pods. I would suggest upgrading core components one by one manually for now. Follow the order 
```
1. controller 
2. executor (wait for couple mins for new function pods come up and for router to migrate traffic to new function pods)
3. router 
4. other components
``` 
But it's too complex for doing this every time, I'm working on this part to make sure users have good upgrade experience and will publish an upgrade guide at that time.

Issue #1160, opened 2019-04-11
Fission fn create does not validate the provided functionName


hd40910:
While creating a fission function if we can pass a invalid function name as given below.
This should not be allowed. The spec gets applied without any issue and then the issue gets logged only in the var logs.

Invalid function name : some-name (Here - is not allowed) but there is no error

The function gets created and when the function is called it only return internal server error 500 or 504.

In below given example the provided values sqs-publisher is invalid.

API Response:

![image](https://user-images.githubusercontent.com/17830882/55968706-2a8b4500-5c9a-11e9-8353-cfce12bf27cc.png)

Spec file : 

![image](https://user-images.githubusercontent.com/17830882/55968985-adac9b00-5c9a-11e9-9b0d-b7ad510265d1.png)

Logs : 

Apr 11 17:25:36 ilgss0166 journal: [2019-04-11 14:25:36,663] ERROR in app: Exception on /v2/specialize [POST]
Apr 11 17:25:36 ilgss0166 journal: Traceback (most recent call last):
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
Apr 11 17:25:36 ilgss0166 journal:    response = self.full_dispatch_request()
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
Apr 11 17:25:36 ilgss0166 journal:    rv = self.handle_user_exception(e)
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
Apr 11 17:25:36 ilgss0166 journal:    reraise(exc_type, exc_value, tb)
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
Apr 11 17:25:36 ilgss0166 journal:    raise value
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
Apr 11 17:25:36 ilgss0166 journal:    rv = self.dispatch_request()
Apr 11 17:25:36 ilgss0166 journal:  File "/usr/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
[...]

PR #1272, opened 2019-08-13
Issue 1160: Validation of functionName/entrypoint while function creation through CLI or CRD

parauliya:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1272)


parauliya:
> @parauliya dont I suggest we should allow `-` as well, because src files can have `-` in them which will result in `--entrypoint` having `-`.

@viveksinghggits No we can't allow '-' even for source file as well because there are environment like "python" in which if you provide '-' in source file and then try to import it, it will give you the error.
Actually, #1160 is about this only where functionName in the yaml provided as "sqs-publisher.main". There you can see that the '-' in the file name "sqs-publisher" is creating the problem not the function name main.

pkg/fission-cli/spec.go
```diff
@@ -496,6 +497,11 @@ func (fr *FissionResources) parseYaml(b []byte, loc *location) error {
 		if err != nil {
 			return errors.Wrap(err, fmt.Sprintf("Failed to parse %v in %v", tm.Kind, loc))
 		}
+		regex := regexp.MustCompile("^[a-zA-Z0-9_.]*$")
+		if !regex.MatchString(v.Spec.Package.FunctionName) {
```
life1347: Since we are not able to 100% guarantee that no existing spec works after adding this check.
It's better to put the check at `https://github.com/fission/fission/blob/88b534377540995cbaaac0175ad0845378456957/pkg/fission-cli/spec.go#L438` and shows a warning message instead of returning an error here.
It would be great to add a TODO mark to remind us to abort requests after couple releases.
```
// TODO: xxxxxx
```

pkg/fission-cli/function.go
```diff
@@ -206,6 +207,11 @@ func fnCreate(c *cli.Context) error {
 		}
 	}
 	entrypoint := c.String("entrypoint")
+	regex := regexp.MustCompile("^[a-zA-Z0-9_.]*$")
+	if !regex.MatchString(entrypoint) {
+		log.Fatal("Entrypoint should not contain any special character except underscore('_') and dot('.').")
```
life1347: We should add the same check at `fnUpdate` as well. 
Also, please make the check a callable function in `pkg/fission-cli/util/util.go` so that we can reuse it.

pkg/fission-cli/spec.go
parauliya: Sounds good and changes are done.

pkg/fission-cli/function.go
parauliya: Added a function in ```util.go``` and added the check in fnUpdate as well. Thanks for pointing out this missed fnUpdate.

life1347:
Hey @parauliya I will do some more tests and review this change again. I expect to merge it next week. Before that, would you mind to rebase the master and fix the conflict? Thanks.

life1347:
BTW, the PR should only contain the commits for the issue (#1160). You can pull the latest master and checkout to a new branch, then cherry-pick related changes to it and force push the new branch.

parauliya:
@life1347 I have resolved all the conflicts for this pull request manually.
Therefore once you finished your testing around these changes you can merge this request directly to fission/master branch.

life1347:
After checking different language server implementation, I'm now thinking we should not have this limit since it's very dependent what language user used. 

For example, [nodejs](https://github.com/fission/fission/tree/master/examples/nodejs#usage-2) allows to have hyphen in function entrypoint. Also, issue #1160 is no longer exists after we switched the way we import. 

```
$ fission fn update --name py2 --entrypoint "user-test.main"
function 'py2' updated

$ fission fn test --name py2
a: 1
b: {c: 3, d: 4}
```

The PR is helpful in some ways, however, it also limits the way for users to call their function. 

Since this problem is dependent on what language a function run with, I think we should not apply the same policy to all languages. Instead, we better to update environment doc to let users know what are the available symbol they can use.

@parauliya @soamvasani @vishal-biyani @viveksinghggits Any thoughts?

viveksinghggits:
Hi @life1347 
I agree, I think we just thought about python when we looked at the issue that was raised about this. It would be a check for language specific `entrypoint` that doest seem to be much useful.

soamvasani:
Hey @parauliya, after @life1347 's comment above, it sounds like this PR's validation is too restrictive.

I think the core problem is that the set of valid chars in a module+function is too language-specific for a validator like this.  We'd need to design and implement language-specific validators to do this correctly.

For now I'm closing this -- redesigning it would be better handled as a new PR.  If you're interested, please let us know!  Thanks much for your work on this, and apologies for not being able to merge it.

Issue #1280, opened 2019-08-20
Consumer and producer logs are empty for kafka


vadasambar:
zap does not log the created consumer and producer properly. It basically shows empty object in the log.

https://github.com/fission/fission/blob/9c63975497d83c09809e1050dd140bd7e9008417/pkg/mqtrigger/messageQueue/kafka.go#L84

https://github.com/fission/fission/blob/9c63975497d83c09809e1050dd140bd7e9008417/pkg/mqtrigger/messageQueue/kafka.go#L96

```
{"level":"info","ts":1566295240.3187578,"logger":"kafka","caller":"messageQueue/kafka.go:90","msg":"created a new consumer","consumer":{}}

{"level":"info","ts":1566295240.3311443,"logger":"kafka","caller":"messageQueue/kafka.go:112","msg":"created a new producer","producer":{}}
```

Using fmt.Sprintf instead of `zap.Any` fixes this.

vadasambar:
I have added a sample code fix in the PR.

PR #655, opened 2018-04-30
Enabling multi-tenancy for fission objects.

smruthi2187:
I've added a summary of changes in Multi-tenancy.md under Documentation/wip directory. I'd recommend going through that first before reviewing the change.

I have one TODO remaining - with this change, we dont permit users to reference secrets and configMaps in namespaces other than the function namespace. So, I have to write a Job that can be run during `helm install` that checks all functions' secret/configmap references and fail the installation if there are any cross references.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/655)

PR #1281, opened 2019-08-20
Fix kafka producer and consumer logs show empty objects

vadasambar:
Sample code fix for #1280

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1281)


pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -81,7 +81,8 @@ func (kafka Kafka) subscribe(trigger *fv1.MessageQueueTrigger) (messageQueueSubs
 	consumerConfig.Group.Return.Notifications = true
 	consumerConfig.Config.Version = kafka.version
 	consumer, err := cluster.NewConsumer(kafka.brokers, string(trigger.Metadata.UID), []string{trigger.Spec.Topic}, consumerConfig)
-	kafka.logger.Info("created a new consumer", zap.Any("consumer", consumer))
+	kafka.logger.Info(fmt.Sprintf("created a new consumer: %#v", consumer))
```
life1347: 1. I'm wondering is it necessary to print the whole consumer/producer object instead of important fields like `trigger.Spec.Topic`, `trigger.Metadata.UID` and `kafka.version` that are helpful for troubleshooting. 
2. The reason for using `zap.Any` here is that zap prints structured logs so that users are able to search log by fields name. 

How about we change the log output here to something like
```
kafka.logger.Info("created a new consumer", 
     zap.String("topic", trigger.Spec.Topic), 
     zap.String("triggerName", trigger.Metadata.Name),
     .....other fields that are helpful....)
```

In this way, we can find important information easily in lines of log. BTW, it would be great if you can provide the log output after changes.

(comment for both consumer and producer log changes)

vadasambar: Logs for consumer

using zap.Any(),
```
{"level":"info","ts":1566474746.5043716,"logger":"kafka","caller":"messageQueue/kafka.go:85","msg":"created a new consumer","consumer":{}}
```
using Sprintf to print the entire object, 
```
{"level":"info","ts":1566474746.5044372,"logger":"kafka","caller":"messageQueue/kafka.go:86","msg":"created a new consumer: &cluster.Consumer{client:(*cluster.Client)(0xc0003e4000), ownClient:true, consumer:(*sarama.consumer)(0xc000078600), subs:(*cluster.partitionMap)(0xc0003c8200), consumerID:\"\", groupID:\"2c2a6b6e-5b00-4c5f-956b-112999585b69\", memberID:\"\", generationID:0, membershipMu:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, coreTopics:[]string{\"input\"}, extraTopics:[]string(nil), dying:(chan cluster.none)(0xc000084240), dead:(chan cluster.none)(0xc0000842a0), closeOnce:sync.Once{m:sync.Mutex{state:0, sema:0x0}, done:0x0}, consuming:0, messages:(chan *sarama.ConsumerMessage)(0xc000084480), errors:(chan error)(0xc0002318c0), partitions:(chan cluster.PartitionConsumer)(0xc000231920), notifications:(chan *cluster.Notification)(0xc0000844e0), commitMu:sync.Mutex{state:0, sema:0x0}}"}
```
printing only the important fields using zap.String and other types

```
{"level":"info","ts":1566474746.5044491,"logger":"kafka","caller":"messageQueue/kafka.go:87","msg":"created a new consumer","brokers":["my-cluster-kafka-brokers.kafka.svc.cluster.local:9092"],"input topic: ":"input","response/output topic: ":"output","error topic: ":"error","trigger name: ":"kafkatest"}
```

vadasambar: Logs for consumer

using zap.Any(),
```
{"level":"info","ts":1566474746.5073383,"logger":"kafka","caller":"messageQueue/kafka.go:104","msg":"created a new producer","producer":{}}
```

using Sprintf to print the entire object,
```
{"level":"info","ts":1566474746.50739,"logger":"kafka","caller":"messageQueue/kafka.go:105","msg":"created a new producer: &sarama.syncProducer{producer:(*sarama.asyncProducer)(0xc00022eaf0), wg:sync.WaitGroup{noCopy:sync.noCopy{}, state1:[3]uint32{0x0, 0x2, 0x0}}}"}
```

printing only the important fields using zap.String and other types
```
{"level":"info","ts":1566474746.5073981,"logger":"kafka","caller":"messageQueue/kafka.go:106","msg":"created a new producer","brokers":["my-cluster-kafka-brokers.kafka.svc.cluster.local:9092"],"input topic":"input","response/output topic":"output","error topic":"error","trigger name":"kafkatest"}
```

vadasambar: I also noticed that the message that he message here is empty as well
https://github.com/fission/fission/blob/9c3195a84ea20617a88acfc5b432bdd44902e73c/pkg/mqtrigger/messageQueue/kafka.go#L66

Logs

With zap.Any()

```
{"level":"info","ts":1566474746.4592006,"caller":"messageQueue/kafka.go:66","msg":"created kafka queue","kafka":{}}
```

With Sprintf(),

```
{"level":"info","ts":1566474746.4592347,"caller":"messageQueue/kafka.go:67","msg":"created kafka queue: messageQueue.Kafka{logger:(*zap.Logger)(0xc0004c85a0), routerUrl:\"http://router.default\", brokers:[]string{\"my-cluster-kafka-brokers.kafka.svc.cluster.local:9092\"}, version:sarama.KafkaVersion{version:[4]uint{0x0, 0x8, 0x2, 0x0}}}"}
```

vadasambar: Maybe we can decide which fields are important and just log them.

life1347: I think we log the function & kafka server related information should be enough.

vadasambar: So, `brokers`, `input topic`, `output topic`, `error topic` for the consumer/producer logs and `kafka version`, and `brokers` for kafka queue ? 

Would that make sense ?

life1347: > brokers, input topic, output topic, error topic for the consumer/producer logs

Also, function metadata like name & namespace.

Others make sense to me.

vadasambar: Is there any good way to retrieve function namespace? It's not present on MessageQueueTrigger struct.

life1347: I think you can print trigger namespace instead since both of the trigger and function should be in the same namespace.

vadasambar: I may be wrong but doesn't trigger run in the same namespace as executor and router (usually `fission` namespace) while a function runs inside a pod in `fission-function` namespace.

life1347: In fact, that's different things. 

To support multi-tenancy, all CRDs should be created in the same namespace. That means you cannot create a trigger that references a function in the different namespace. (Currently, all CRDs are created in `default` namespace by default). A user could choose which namespace he/she wants to install fission. But to fully utilize resources, we currently shared the `fission-function` and `fission-builder` namespace to create environment pod for compiling and serving. And actually there are clues in CLI, for example:

```
$ fission fn create -h
.....
OPTIONS:
   --fnNamespace value, --fns value  Namespace for function object (default: "default")
```

That's why we can use the namespace of trigger instead (though it's not good). You can have a look at this PR https://github.com/fission/fission/pull/655 .

life1347:
@vadasambar any updates for this?

vadasambar:
> @vadasambar any updates for this?

Hi, sorry I was a little busy this week. I will check this as soon as I get some time. Thanks for the reminder :+1

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: You're right. Thank you so much for explaining this. I did not notice that the CRDs for mqt and functions are created in the same namespace.

vadasambar:
Here's how the logs look like now (I have installed fission in the `default` namespace)
```
2019-09-26T12:43:17.620Z	INFO	kafka	messageQueue/kafka.go:79	brokers set	{"brokers": ["my-cluster-kafka-brokers.kafka:9092"]}
2019-09-26T12:43:18.590Z	INFO	kafka	messageQueue/kafka.go:87	created a new consumer	{"brokers": ["my-cluster-kafka-brokers.kafka:9092"], "input topic": "suraj", "output topic": "suraj-output", "error topic": "error", "trigger name": "kafkatest", "function namespace": "default", "function name": "consumerfunc"}
2019-09-26T12:43:18.636Z	INFO	kafka	messageQueue/kafka.go:106	created a new producer	{"brokers": ["my-cluster-kafka-brokers.kafka:9092"], "input topic": "suraj", "output topic": "suraj-output", "error topic": "error", "trigger name": "kafkatest", "function namespace": "default", "function name": "consumerfunc"}
```

life1347:
LGTM

PR #1284, opened 2019-08-21
 Issue 1229: Function level timeout

parauliya:
This commit added the hardcoded timeout at client side

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1284)


parauliya:
Hi, 
This pull request only containes the changes regarding function timeout for http triggers.
Function timeout for mqtrigger and timetrigger is also being implemented since it's not same as http triggers. I have made this pull request ready for review to get the first hand comments. @life1347 @vishal-biyani @sanketsudake @bhavin192.

life1347:
Hi @parauliya Thanks for the PR.
May I ask the reason that you choose to add the timeout setting to triggers instead of adding it to function?

parauliya:
Hi @life1347,
There are different ways by which the triggers has been implemented and called. For e.g. in case of http trigger it create its own "functionhandler" struct and then make a call to the function and in case of mqt and time trigger they don't really care about the "functionhandler" they simply call the function. In that case we need the timeout while creating the mqt and time trigger otherwise it won't be able to access the timeout while calling the function/trigger. 
What I found was in each trigger is that,  there is a similarity that they all have "FunctionReference" struct in their spec. Therefore I added the timeout in  "FunctionReference" struct so that can be accessible while calling any trigger. 
Just FYI, I will be passing the timeout at the time of mqt or time trigger call in request headers itself so it can be accessible to router easily.

Please let me know in case of any concern, confusion or any missing/incorrect information.

life1347:
>  In that case we need the timeout while creating the mqt and time trigger otherwise it won't be able to access the timeout while calling the function/trigger.

In fact, all triggers call HTTP trigger (router) to get a response from the function instead of calling function pod directly. So the timeout setting of HTTP trigger is the minimum timeout setting in the whole calling path.

> Just FYI, I will be passing the timeout at the time of mqt or time trigger call in request headers itself so it can be accessible to router easily.

I don't know how you gonna achieve this for time trigger to pass header. But I do not suggest doing this as we cannot differentiate whether a request is coming from inside or outside the cluster and router may run out of all resources if receiving malicious requests with forge request headers.

> What I found in each trigger is, there is a similarity that they all have "FunctionReference" struct in their spec.

FunctionReference is the place for function reference, please do not add other configs that are not related to it.

After checking the source code again, I would suggest add timeout setting to `HTTPTriggerSpec`. The reasons as follow:

1. The right place to put the config.
2. Decouple timeout setting from function. 
  * Users can create mulitple HTTP triggers to the same function with different timeout setting.
3. To simplify the way for users to set the timeout setting. 
  * Users don't need set timeout setting in different type of triggers.

pkg/fission-cli/httptrigger.go
```diff
@@ -59,11 +59,12 @@ func getMethod(method string) string {
...
 
-func setHtFunctionRef(functionList []string, functionWeightsList []int) (*fv1.FunctionReference, error) {
+func setHtFunctionRef(functionList []string, functionWeightsList []int, functionTimeout uint64) (*fv1.FunctionReference, error) {
 	if len(functionList) == 1 {
 		return &fv1.FunctionReference{
-			Type: fv1.FunctionReferenceTypeFunctionName,
-			Name: functionList[0],
+			Type:            fv1.FunctionReferenceTypeFunctionName,
+			Name:            functionList[0],
+			FunctionTimeout: functionTimeout,
```
vishal-biyani: FunctionReference seems an odd place for adding a timeout attribute - reference should be that's it - a reference. I haven't looked at whole flow - but I suggest revisiting the decision of placing timeout in reference struct!

parauliya:
Hi @life1347 
Yes,  every request goes through the router but not HTTPTrigger exactly.
Unless HTTPTrigger is created we may not have timeout for functions.

There are two places where it creates functionhandler in `getRouter()` function itself.
1.  While iterating on HTTPtriggers as per the below link:
https://github.com/fission/fission/blob/86b506c9b6345af568bed8b2bd68d7a473ace08c/pkg/router/httpTriggers.go#L152
2. Functions which don't have HTTP triggers. In this case it iterates on functions and populate functionHandler as per the below link:
https://github.com/fission/fission/blob/86b506c9b6345af568bed8b2bd68d7a473ace08c/pkg/router/httpTriggers.go#L211

So if we put the timeout in `HTTPTriggerSpec`, it won't be available for the functions which do not have an `HTTPtrigger`.

There was one another solution that I thought of earlier to pass the `timeout` in `Function Spec`, access it before creating the `functionHandler` and add it to `functionHandler`. 
Currently there is no direct mapping of `HTTPTrigger` and `Function spec`, so I may need to iterate over all functions in 1 ( In case of HTTPTriggers ) or we may create a direct mapping for it to access the `Function Spec`.
Adding timeout to `Function Spec` will work for `TimeTrigger` and `MQTrigger` as well because of the 2 ( second link above ).

Let me know what you think on this.

life1347:
> Yes, every request goes through the router but not HTTPTrigger exactly.....
Ah, I see what you saying. I'm thinking another approach is to add a new CRD that store timeout setting. But let me think about it and will discuss the idea with you and other folks.

life1347:
After discussion yesterday, the idea of adding new CRD resource to store timeout setting is better to wait until we really need to set the different timeout for different triggers. For now, we can add to `functionSpec`. 

> Currently there is no direct mapping of HTTPTrigger and Function spec, so I may need to iterate over all functions in 1

You can create the map at `updateRouter` and pass the map to `getRouter` for later usage.
https://github.com/fission/fission/blob/86b506c9b6345af568bed8b2bd68d7a473ace08c/pkg/router/httpTriggers.go#L349-L370

parauliya:
@life1347,
I have pudated the PR with the new changes. I have discarded the trigger level timeout apporach and implemented the function level timeout that we had thought of initially.
 `FunctionTimeout` is a part of `FunctionSpec` now. 
For HTTPTriggers I have created the map of `HTTPTriggerName` and `FunctionTimeout` in the updateRouter() function as per your suggestion. Thanks for your suggestion.

life1347:
@parauliya did you push your latest commits? I don't see any changes.

parauliya:
> @parauliya did you push your latest commits? I don't see any changes.
@life1347 
No I didn't push any changes as I have some confusion in some of the comments which I have mentioned. I am planning to push all the changes together. Could you please look into my queries for some of your comments and answer my queries.

life1347:
> I added the goroutine and select to capture the context timeout and create HTTP response ...

You might want to take a look at `ReverseProxy > ErrorHandler`. With ErrorHandler we can reply useful information back to user without leaking any detail information in fission cluster. But I think we can do it in another PR since it takes some time to do that. Ref(check the code part): https://tech.mercari.com/entry/2018/12/05/105737

> I am not able to understand this apporach completely. If you create a map in updateRouter() and pass it in getRouter(), how can I access this map ....

I'm thinking something like [this](https://gist.github.com/life1347/7b970270574e703d740800c724f2fd59), what do you think?

parauliya:
@life1347 
Yes, I think it should work. My only concern is like in this case we are changing a lot of existing struct and functions like signature of getRouter() function , RetryingRoundTripper struct and functionHandler struct as well. And in my previous approach we were only changing functionHandler struct and HTTPTriggerset struct only. And Since we just need this change for HTTPTriggers only I did a bit complex but small change as our remaining triggers were not required these changes. But I think your approach is more efficient in the cases where we have lot of functions so I am going with your apporach and will implement asap and update this pull request.

life1347:
I believe there are definitely some other ways to call the `functionTimeoutMap` instead of adding it in the struct. I attached the gist is because it's easier for me to explain the whole idea, it's not the thing you definitely need to follow with. If you have any suggestion, feel free to propose. It's always good to have discussions for a better design. :)

> And in my previous approach we were only changing functionHandler struct and HTTPTriggerset struct only....

https://github.com/fission/fission/blob/b044fced84706fcba9ec99158ef32826f1a9a0ba/pkg/router/httpTriggers.go#L373-L374

The problem in the previous approach is that HTTP trigger does not always point to a function, it may be a map contains two functions for canary deployment. (It's bad I know, I will refactor router in the later milestone.)

https://github.com/fission/fission/blob/b044fced84706fcba9ec99158ef32826f1a9a0ba/pkg/apis/fission.io/v1/types.go#L416

And, you can only know which function the request goes to in functionHandler. 

https://github.com/fission/fission/blob/b044fced84706fcba9ec99158ef32826f1a9a0ba/pkg/router/functionHandler.go#L452

That's why the previous approach won't work under such circumstances.

Also, we decided to set the timeout setting for all requests to a function so it's not so useful to iterate HTTP trigger to create a timeout map since you're able to do it in a much simpler way.

Hope I explain this well.

pkg/error/httperror.go
```diff
@@ -56,6 +56,8 @@ func MakeErrorFromHTTP(resp *http.Response) error {
 		errCode = ErrorNotFound
 	case http.StatusConflict:
 		errCode = ErrorNameExists
+	case http.StatusRequestTimeout:
+		errCode = ErrorRequestContextTimedOut
```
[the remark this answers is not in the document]
life1347: Also, please add error description to `errorDescriptions (L137)`

pkg/apis/fission.io/v1/types.go
```diff
@@ -337,6 +337,10 @@ type (
 
 		// InvokeStrategy is a set of controls which affect how function executes
 		InvokeStrategy InvokeStrategy
+
+		// FunctionTimeout provides a maximum amount of duration wihtin which a request for a particular function execution should be complete.
+		// This is optional. If not specified default value will be taken as 60s
+		FunctionTimeout uint64 `json:"functionTimeout,omitempty"`
```
life1347: I think `Int` should be enough since we wait for second instead of mills.

pkg/router/functionHandler.go
```diff
@@ -42,6 +37,11 @@ import (
 	"github.com/fission/fission/pkg/redis"
 	"github.com/fission/fission/pkg/throttler"
 	"github.com/fission/fission/pkg/types"
+	"github.com/pkg/errors"
```
life1347: Please sort the imports

pkg/router/functionHandler.go
```diff
@@ -289,8 +291,18 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Res
...
+		}
+		roundTripper.logger.Debug("Creating context for request for ", zap.Any("Time", roundTripper.timeout))
+		var closeCtx func()
+		ctx, closeCtx := context.WithTimeout(context.Background(), time.Duration(roundTripper.timeout)*time.Second)
+		defer closeCtx()
+
 		// forward the request to the function service
-		resp, err = ocRoundTripper.RoundTrip(req)
+		resp, err = ocRoundTripper.RoundTrip(req.WithContext(ctx))
+
```
life1347: nitpick: remove the blank line

pkg/router/functionHandler.go
```diff
@@ -319,6 +331,7 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Res
 			return resp, nil
 		} else if i >= roundTripper.funcHandler.tsRoundTripperParams.maxRetries-1 {
 			// return here if we are in the last round
+			//Check for timeout error first
```
life1347: A blank should follow the `//`. (Here and some other places)

pkg/router/httpTriggers.go
```diff
@@ -351,20 +357,26 @@ func (ts *HTTPTriggerSet) updateRouter() {
...
+
+		//Iterate over each HTTPTrigger
 		for _, t := range latestTriggers {
 			triggers = append(triggers, *t.(*fv1.HTTPTrigger))
 		}
 		ts.triggers = triggers
 
 		// get functions
 		latestFunctions := ts.funcStore.List()
+		functionTimeout := make(map[types.UID]uint64)
```
life1347: `make(map[types.UID]uint64)` -> `make(map[types.UID]uint64, len(latestFunctions))` for performance.
btw, please remove the blank line. (L370)

pkg/error/httperror.go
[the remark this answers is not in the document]
parauliya: I'll make `ErrorRequestContextTimeout` to `ErrorRequestTimeout`. Is that ok?
I'll also add the descritpion in `errorDescriptions`

pkg/apis/fission.io/v1/types.go
parauliya: I made it uint so that user can't provide the value in negative integers. In case of uint parsing will itself fail when user will try to provide timeout in negative integers. and If we make it as Int we have to do some extra verification for values of --fnTimeout parameter. I can make it uint only instead of uint64 if it is ok for you.

pkg/router/functionHandler.go
parauliya: Yup will do that.

pkg/router/functionHandler.go
parauliya: This comment was to be removed. Forgot to remove it.

pkg/router/functionHandler.go
parauliya: Done

pkg/router/httpTriggers.go
parauliya: Done

pkg/router/functionHandler.go
```diff
@@ -289,8 +291,18 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Res
 
 		roundTripper.logger.Debug("request headers", zap.Any("headers", req.Header))
 
+		//Creating context for client
+		if roundTripper.timeout == 0 {
```
life1347: oops, I missed this.
`roundTripper.timeout == 0` -> `roundTripper.timeout <= 0`

pkg/apis/fission.io/v1/types.go
life1347: Ah, I see what you mean. But I think it's better to check whether an input value is valid instead of using uint as a way to prevent the problem.
I think it would be good to add validation at https://github.com/fission/fission/blob/0d872020e3a69c042855c03c5feca923541b1907/pkg/apis/fission.io/v1/validation.go#L324

pkg/router/functionHandler.go
life1347: Normally, the order of import follows the rule
```
built-in pkg

3rd pkg

fission pkg
```

pkg/router/functionHandler.go
parauliya: Yes, I add only  `==` instead of `<=` because it was uint so can't be `< 0`. But now when I change `uint64` to `int` this change will also be required.

pkg/router/functionHandler.go
parauliya: Yes, but If I do this `./hack/verify-gofmt.sh` will fail because it requires the imports to be sorted in alphabatical order. I have tried this already.

life1347: Could you try with this? I can pass the `verify-gofmt.sh`
```
import (
	"bytes"
	"context"
	"fmt"
	"io"
	"io/ioutil"
	"math/rand"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"
	"time"

	"github.com/pkg/errors"
	"go.opencensus.io/plugin/ochttp"
	"go.uber.org/zap"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	k8stypes "k8s.io/apimachinery/pkg/types"

	fv1 "github.com/fission/fission/pkg/apis/fission.io/v1"
	"github.com/fission/fission/pkg/crd"
	ferror "github.com/fission/fission/pkg/error"
	"github.com/fission/fission/pkg/error/network"
	executorClient "github.com/fission/fission/pkg/executor/client"
	"github.com/fission/fission/pkg/redis"
	"github.com/fission/fission/pkg/throttler"
	"github.com/fission/fission/pkg/types"
)
```

parauliya: Yes this is working.

pkg/apis/fission.io/v1/types.go
parauliya: I think the check should be added in at below location where Function spec is getting verified:

https://github.com/fission/fission/blob/0d872020e3a69c042855c03c5feca923541b1907/pkg/apis/fission.io/v1/validation.go#L285

pkg/apis/fission.io/v1/validation.go
```diff
@@ -304,6 +304,10 @@ func (spec FunctionSpec) Validate() error {
 		result = multierror.Append(result, spec.InvokeStrategy.Validate())
 	}
 
+	if spec.FunctionTimeout <= 0 {
+		result = multierror.Append(result, MakeValidationErr(ErrorInvalidValue, "FunctionTimeout value", spec.FunctionTimeout, "not a valid value. Should always be more than 0"))
```
life1347: "FunctionTimeout value" -> "FunctionSpec.FunctionTimeout"

Consider the compatibility of old fission spec files which have no `FunctionTimeout` field may be broken by this change, please comment out the validation and prints warnings to users so that they can add miss field. You can take a look at following code snippets as an example.

1. Prints warning when a user uses spec command with old spec files.
https://github.com/fission/fission/blob/e481c234ee038f8ce8c877a4ab7d3fbf8b8b5597/pkg/fission-cli/cmd/spec/spec.go#L489-L492

2. I will let validate function return warning in later PR. Before that, let's comment out the validation part to prevent compatibility problem. 
https://github.com/fission/fission/blob/e481c234ee038f8ce8c877a4ab7d3fbf8b8b5597/pkg/apis/fission.io/v1/validation.go#L350-L353

pkg/error/httperror.go
```diff
@@ -56,6 +56,8 @@ func MakeErrorFromHTTP(resp *http.Response) error {
 		errCode = ErrorNotFound
 	case http.StatusConflict:
 		errCode = ErrorNameExists
+	case http.StatusRequestTimeout:
+		errCode = ErrorRequestTimedOut
```
life1347: Is it intentional to use `TimedOut` instead of `Timeout`?

pkg/fission-cli/function.go
```diff
@@ -224,6 +224,7 @@ func fnCreate(c *cli.Context) error {
 		}
 	}
 	entrypoint := c.String("entrypoint")
+	fnTimeout := c.Int("fntimeout")
```
life1347: Return error when a user specifies the illegal value, just like how we do in specialization timeout.
https://github.com/fission/fission/blob/e481c234ee038f8ce8c877a4ab7d3fbf8b8b5597/pkg/fission-cli/function.go#L155-L160

pkg/fission-cli/function.go
```diff
@@ -575,6 +578,11 @@ func fnUpdate(c *cli.Context) error {
 	if len(entrypoint) > 0 {
 		function.Spec.Package.FunctionName = entrypoint
 	}
+
+	if c.IsSet("fntimeout") {
+		function.Spec.FunctionTimeout = fnTimeout
```
life1347: Return error when a user specifies the illegal value, just like how we do in specialization timeout.
https://github.com/fission/fission/blob/e481c234ee038f8ce8c877a4ab7d3fbf8b8b5597/pkg/fission-cli/function.go#L155-L160

pkg/error/httperror.go
parauliya: Actualy yes, But I think Timeout will be a better option.

pkg/router/functionHandler.go
```diff
@@ -289,8 +292,17 @@ func (roundTripper RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Res
...
 		roundTripper.logger.Debug("request headers", zap.Any("headers", req.Header))
 
+		// Creating context for client
+		if roundTripper.timeout <= 0 {
+			roundTripper.timeout = fv1.DEFAULT_FUNCTION_TIMEOUT
+		}
+		roundTripper.logger.Debug("Creating context for request for ", zap.Any("Time", roundTripper.timeout))
+		var closeCtx func()
+		ctx, closeCtx := context.WithTimeout(context.Background(), time.Duration(roundTripper.timeout)*time.Second)
+		defer closeCtx()
```
life1347: Actually, it's not necessary to declare `closeCtx` here, you can remove it.
Also, a possible resource leak problem: `defer in for loop`.[[reference](https://stackoverflow.com/questions/45617758/defer-in-the-loop-what-will-be-better)]

How about we changing this into
```
// forward the request to the function service
ctx, closeCtx := context.WithTimeout(context.Background(), time.Duration(roundTripper.timeout)*time.Second)
resp, err = ocRoundTripper.RoundTrip(req.WithContext(ctx))
closeCtx()
```

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -490,6 +490,10 @@ func (fr *FissionResources) Validate(c *cli.Context) error {
 		if strategy.ExecutorType == fv1.ExecutorTypeNewdeploy && strategy.SpecializationTimeout < fv1.DefaultSpecializationTimeOut {
 			log.Warn(fmt.Sprintf("SpecializationTimeout in function spec.InvokeStrategy.ExecutionStrategy should be a value equal to or greater than %v", fv1.DefaultSpecializationTimeOut))
 		}
+		fnTimeout := f.Spec.FunctionTimeout
+		if fnTimeout <= 0 {
+			log.Warn(fmt.Sprintf("FunctionTimeout in function spec should be field which should have a value greater than 0"))
```
life1347: `FunctionTimeout in function spec should be a value greater than or equal to 0`

pkg/fission-cli/function.go
```diff
@@ -224,6 +224,15 @@ func fnCreate(c *cli.Context) error {
 		}
 	}
 	entrypoint := c.String("entrypoint")
+
+	var fnTimeout int
+	if c.IsSet("fntimeout") {
+		fnTimeout = c.Int("fntimeout")
+		if fnTimeout <= 0 {
+			log.Fatal("fntimeout must be greater than 0")
```
life1347: `fntimeout must be greater than or equal to 0`

pkg/fission-cli/function.go
```diff
@@ -224,6 +224,15 @@ func fnCreate(c *cli.Context) error {
 		}
 	}
 	entrypoint := c.String("entrypoint")
+
+	var fnTimeout int
+	if c.IsSet("fntimeout") {
+		fnTimeout = c.Int("fntimeout")
+		if fnTimeout <= 0 {
```
life1347: `fnTimeout < 0`

pkg/fission-cli/function.go
```diff
@@ -575,6 +585,15 @@ func fnUpdate(c *cli.Context) error {
 	if len(entrypoint) > 0 {
 		function.Spec.Package.FunctionName = entrypoint
 	}
+
+	if c.IsSet("fntimeout") {
+		fnTimeout := c.Int("fntimeout")
+		if fnTimeout <= 0 {
+			log.Fatal("fntimeout must be greater than 0")
```
life1347: same here (L591 - L592)

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -490,6 +490,10 @@ func (fr *FissionResources) Validate(c *cli.Context) error {
 		if strategy.ExecutorType == fv1.ExecutorTypeNewdeploy && strategy.SpecializationTimeout < fv1.DefaultSpecializationTimeOut {
 			log.Warn(fmt.Sprintf("SpecializationTimeout in function spec.InvokeStrategy.ExecutionStrategy should be a value equal to or greater than %v", fv1.DefaultSpecializationTimeOut))
 		}
+		fnTimeout := f.Spec.FunctionTimeout
+		if fnTimeout <= 0 {
```
life1347: The flag description said ` If set to zero, by default it will wait of 60s for the response.`. So the condition here should be `fnTimeout < 0`.

pkg/apis/fission.io/v1/validation.go
```diff
@@ -304,6 +304,11 @@ func (spec FunctionSpec) Validate() error {
 		result = multierror.Append(result, spec.InvokeStrategy.Validate())
 	}
 
+	// TODO Add below validation warning
+	/*if spec.FunctionTimeout <= 0 {
```
life1347: `spec.FunctionTimeout <= 0` -> spec.FunctionTimeout < 0

`Should always be more than 0` -> `Should always be greater than or equal to 0`

pkg/fission-cli/main.go
```diff
@@ -122,13 +122,15 @@ func NewCliApp() *cli.App {
 	fnLogCountFlag := cli.StringFlag{Name: "recordcount", Usage: "the n most recent log records"}
 	fnForceFlag := cli.BoolFlag{Name: "force", Usage: "Force update a package even if it is used by one or more functions"}
 	fnExecutorTypeFlag := cli.StringFlag{Name: "executortype", Value: types.ExecutorTypePoolmgr, Usage: "Executor type for execution; one of 'poolmgr', 'newdeploy' defaults to 'poolmgr'"}
+	fnExecutionTimeoutFlag := cli.IntFlag{Name: "fntimeout, ft", Value: 60, Usage: "The length of time to wait for the response. If set to zero, by default it will wait of 60s for the response."}
```
life1347: Usage -> 
```
The length of time to wait for the response. If the flag is not provided, by default it will wait of 60s for the response.
```

pkg/router/functionHandler.go
parauliya: Yes, you are right defer can cause a resource leak. Initially it was added since we were processing a request in a goroutine. I will correct it.

pkg/fission-cli/main.go
parauliya: Changed to this `Time duration to wait for the response while executing the function. If the flag is not provided, by default it will wait of 60s for the response.`

pkg/fission-cli/cmd/spec/spec.go
parauliya: Done

pkg/fission-cli/function.go
parauliya: Done

life1347:
That's fair. Let's do it.

parauliya:
> That's fair. But since the PR is in good shape, I want to merge it first. Let's do it in another PR.

Sure.

life1347:
LGTM, once the [CI build](https://travis-ci.org/fission/fission/builds/583200541) passed then we are good to go

life1347:
Merged! @parauliya Thanks for your effort! 💯

Issue #1293, opened 2019-08-29
Add support to connect mqtriggers to TLS secured Kafka cluster 


vadasambar:
Can we add support for MQTriggers to connect to a TLS secured Kafka cluster?

MQTrigger can use the user key-pair and broker certificate to connect to the Kafka cluster using Sarama. Sarama offers out of the box support for TLS.

vishal-biyani:
Related: https://github.com/fission/fission/issues/1162 

Hey @vadasambar - Yes I think TLS is a better option I think. Where will the key pair and certs live in a function? As a secret/configmap?

viveksinghggits:
@vadasambar 
I think we should have a way to configure the client certs, using which fission mqt will be connecting to the secured kafka cluster, at the time of fission installation. Like we manually change the kafka broker if our broker is not running at the default host (`  brokers: 'broker.kafka:9092'`) that is configured in the `values.yaml`. Eventually fission should create a secret containing those certs and mount that secret to the mqt pod.

@vishal-biyani what do you think?

vadasambar:
@vishal-biyani, yes. ConfigMap/Secret is what I was thinking.

vadasambar:
@viveksinghggits , I like the idea of configuring the certificates when installing fission but if I am understanding this correctly, wouldn't that mean adding configmap/secret in the templates ?

viveksinghggits:
@vadasambar 
yes, but eventually what we can do is create a secret with the certs and provide the name of that secret in the `values.yaml`.
Providing secrets in the `values.yaml` doesnt seem to be a good idea.

vadasambar:
@viveksinghggits makes sense. I will try to create a POC with this approach.

vadasambar:
@viveksinghggits I've tried implementing this approach in the above POC.

PR #1300, opened 2019-09-03
Implement TLS authentication for kafka mqt

vadasambar:
details
----------
- uses secret to store keys and certificates

Fix for #1293

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1300)


life1347:
Also, please check the error log in CI build: https://travis-ci.org/fission/fission/builds/580226550#L1172

charts/fission-all/templates/deployment.yaml
```diff
@@ -680,7 +680,53 @@ spec:
...
+          mountPath: /etc/secrets
+        {{- end}}
+        
       serviceAccount: fission-svc
+
+      {{- if .Values.kafka.authentication.tls.enabled}}
+      volumes:
+      - name: kafka-secrets
+        secret:
+          secretName: mqtrigger-kafka-secrets
```
life1347: How about we name it `mqtrigger-kafka-tls-cert`?

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
-		Url:    mqUrl,
+		MQType:  mqType,
+		Url:     mqUrl,
+		Secrets: secrets,
 	}
 	messageQueue.MakeMessageQueueTriggerManager(logger, fissionClient, routerUrl, mqCfg)
 	return nil
 }
+
+func getCurrentNamespace() string {
```
life1347: I think you can remove this function since we don't need it anymore.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+	namespace := strings.TrimSpace(string(data))
+	if err == nil && len(namespace) > 0 {
+		return namespace
+	}
+
+	return "default"
+}
+
+func readSecrets(logger *zap.Logger) (map[string][]byte, error) {
+	secretsPath := "/etc/secrets"
```
life1347: Instead of hard coding the `secretsPath`, I prefer to read it from the environment variable.
So that user can change the path later.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+			filePath := path.Join(secretsPath, fileName)
+			secret, fileReadErr := ioutil.ReadFile(filePath)
+
+			if fileReadErr != nil {
+				return secrets, fileReadErr
+			}
+
+			secrets[fileName] = secret
+		}
+
```
life1347: nitpick: some of blank lines are not really necessary, please remove them. For example,
```
secretFiles, err := ioutil.ReadDir(secretsPath)

if err != nil {
	return secrets, err
}
```

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
 	// Message queue type: nats is the only supported one for now
 	mqType := os.Getenv("MESSAGE_QUEUE_TYPE")
 	mqUrl := os.Getenv("MESSAGE_QUEUE_URL")
+
+	// For authentication with message queue
+	secrets, err := readSecrets(logger)
+
+	if err != nil {
+		panic(err)
```
life1347: return instead of panic

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+	return "default"
+}
+
+func readSecrets(logger *zap.Logger) (map[string][]byte, error) {
+	secretsPath := "/etc/secrets"
+	secrets := make(map[string][]byte)
+
+	// return if no secrets exist
+	if _, err := os.Stat(secretsPath); os.IsNotExist(err) {
+		return secrets, nil
```
life1347: `return nil, err`
I personally prefer to return `nil` for the fields other than error.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+	}
+
+	secretFiles, err := ioutil.ReadDir(secretsPath)
+
+	if err != nil {
+		return secrets, err
```
life1347: same here

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+
+		// /etc/secrets contain some hidden directories (like .data)
+		// ignore them
+		if !secretFile.IsDir() && !strings.HasPrefix(fileName, ".") {
+			logger.Info(fmt.Sprintf("Reading secret from %#v", fileName))
+			filePath := path.Join(secretsPath, fileName)
+			secret, fileReadErr := ioutil.ReadFile(filePath)
+
+			if fileReadErr != nil {
+				return secrets, fileReadErr
```
life1347: `return nil, fileReadErr`

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
 	// Message queue type: nats is the only supported one for now
 	mqType := os.Getenv("MESSAGE_QUEUE_TYPE")
 	mqUrl := os.Getenv("MESSAGE_QUEUE_URL")
+
+	// For authentication with message queue
+	secrets, err := readSecrets(logger)
```
life1347: Since we support different types of message queue you should do this part at `makeKafkaMessageQueue`.

charts/fission-all/templates/deployment.yaml
```diff
@@ -680,7 +680,53 @@ spec:
...
         - name: DEBUG_ENV
           value: {{ .Values.debugEnv | quote }}
+        # TLS authentication is TLS with authentication (2 way)
+        # More info: https://docs.confluent.io/current/kafka/authentication_ssl.html#ssl-overview
+        {{- if .Values.kafka.authentication.tls.enabled}}
+        - name: TLS_ENABLED
+          value: "true"
+        volumeMounts:
+        - name: kafka-secrets
+          mountPath: /etc/secrets
```
life1347: `/etc/fission/secrets` would be better.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +44,68 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+	if err == nil && len(namespace) > 0 {
+		return namespace
+	}
+
+	return "default"
+}
+
+func readSecrets(logger *zap.Logger) (map[string][]byte, error) {
+	secretsPath := "/etc/secrets"
+	secrets := make(map[string][]byte)
```
life1347: I suggest moving this line just right before the place where it's first called, like
```
secrets := make(map[string][]byte)
for _, secretFile := range secretFiles {
```

charts/fission-all/templates/deployment.yaml
```diff
@@ -680,7 +680,53 @@ spec:
...
+        {{- end}}
+
```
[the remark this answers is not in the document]
vadasambar: Let me fix that. :+1:

charts/fission-all/values.yaml
```diff
@@ -104,7 +104,27 @@ azureStorageQueue:
...
+  # brokers: 'my-broker.kafka:9092' # or my-bootstrap-server.kafka:9092/9093
+  # Sample config for authentication
+  #   authentication:
+  #   tls:
+  #     enabled: true
+  #     caCert: 'auth/kafka/ca.crt'
+  #     userCert: 'auth/kafka/user.crt'
+  #     userKey: 'auth/kafka/user.key'
+
+
```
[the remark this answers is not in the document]
vadasambar: Let me fix that. :+1:

pkg/mqtrigger/mqtrigger.go
```diff
@@ -17,17 +17,21 @@ limitations under the License.
...
-
-	"github.com/pkg/errors"
-	"go.uber.org/zap"
+	"path"
+	"strings"
 
 	"github.com/fission/fission/pkg/crd"
 	"github.com/fission/fission/pkg/mqtrigger/messageQueue"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
```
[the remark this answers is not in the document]
vadasambar: I think the imports were automatically add by gofmt. Should I reverse their order?

pkg/mqtrigger/mqtrigger.go
vadasambar: I will remove this. Thanks for pointing this out :)

pkg/mqtrigger/mqtrigger.go
vadasambar: I like this idea. Will make the change. Thank you :)

pkg/mqtrigger/mqtrigger.go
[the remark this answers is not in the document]
life1347: You may need to do it manually. The order of imports follow the rule
```
built-in pkg

3rd pkg

fission pkg
```

pkg/mqtrigger/mqtrigger.go
vadasambar: Will fix this. Thanks for pointing it out :)

pkg/mqtrigger/mqtrigger.go
vadasambar: I think moving this part into `makeKafkaMessageQueue` makes sense but I was thinking of secrets here as more of a generic thing because other message queues might need this in the future. Because we read all the secrets and make it available to any message queue, tt might save us some work in the future (say we want to have authentication for NATS or some other message queue in the future) because we would have a pipeline to read secrets ready. 

What do you think?

charts/fission-all/templates/deployment.yaml
vadasambar: Will do this :+1:

pkg/mqtrigger/mqtrigger.go
vadasambar: Will do this :+1:

pkg/mqtrigger/mqtrigger.go
life1347: Oops, sorry, I forgot to remove this comment last night. You're right for this.

charts/fission-all/templates/deployment.yaml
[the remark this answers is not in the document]
vadasambar: Done

charts/fission-all/values.yaml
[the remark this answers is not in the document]
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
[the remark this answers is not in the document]
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

charts/fission-all/templates/deployment.yaml
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: Done

charts/fission-all/templates/deployment.yaml
```diff
@@ -680,7 +680,50 @@ spec:
           value: {{ .Values.traceSamplingRate | default "0.5" | quote }}
         - name: DEBUG_ENV
           value: {{ .Values.debugEnv | quote }}
+        # TLS authentication is TLS with authentication (2 way)
+        # More info: https://docs.confluent.io/current/kafka/authentication_ssl.html#ssl-overview
+        {{- if .Values.kafka.authentication.tls.enabled}}
```
life1347: nit: Leave empty before `}}`  L685, L693, L695, L700

charts/fission-all/values.yaml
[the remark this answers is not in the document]
life1347: Looks like you missed it.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +45,57 @@ func Start(logger *zap.Logger, routerUrl string) error {
 	// Message queue type: nats is the only supported one for now
 	mqType := os.Getenv("MESSAGE_QUEUE_TYPE")
 	mqUrl := os.Getenv("MESSAGE_QUEUE_URL")
+
+	// For authentication with message queue
+	secrets, err := readSecrets(logger)
```
life1347: We should check whether the `TLS_ENABLED` is true. If not, we should not call the function.

pkg/mqtrigger/mqtrigger.go
```diff
@@ -40,10 +45,57 @@ func Start(logger *zap.Logger, routerUrl string) error {
...
+func readSecrets(logger *zap.Logger) (map[string][]byte, error) {
+	secretsPath := strings.TrimSpace(os.Getenv("MESSAGE_QUEUE_SECRETS"))
+	// return if no secretsPath is present
+	if len(secretsPath) == 0 {
+		return nil, nil
+	}
+
+	// return if no secrets exist
+	if _, err := os.Stat(secretsPath); os.IsNotExist(err) {
+		return nil, nil
```
life1347: We should only run the function when `TLS_ENABLED` is true, so return the error here.
In this way, we can distinguish the situations if it's `TLS_ENABLED=false` or user gave the wrong path to volume.

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -63,6 +68,17 @@ func makeKafkaMessageQueue(logger *zap.Logger, routerUrl string, mqCfg MessageQu
 		brokers:   strings.Split(mqCfg.Url, ","),
 		version:   kafkaVersion,
 	}
+
+	if tls, _ := strconv.ParseBool(os.Getenv("TLS_ENABLED")); tls == true {
+		kafka.tls = true
+
+		authKeys := make(map[string][]byte)
+		authKeys["caCert"] = mqCfg.Secrets["caCert"]
```
life1347: Check wether `mqCfg.Secrets`  is nil before accessing.

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -63,6 +68,17 @@ func makeKafkaMessageQueue(logger *zap.Logger, routerUrl string, mqCfg MessageQu
 		brokers:   strings.Split(mqCfg.Url, ","),
 		version:   kafkaVersion,
 	}
+
+	if tls, _ := strconv.ParseBool(os.Getenv("TLS_ENABLED")); tls == true {
+		kafka.tls = true
+
+		authKeys := make(map[string][]byte)
```
life1347: Check wether `mqCfg.Secrets`  is nil before accessing to it.

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -80,18 +96,34 @@ func (kafka Kafka) subscribe(trigger *fv1.MessageQueueTrigger) (messageQueueSubs
...
+		}
+
+		producerConfig.Net.TLS.Config = tlsConfig
+		consumerConfig.Net.TLS.Config = tlsConfig
+	}
+
+	consumer, err := cluster.NewConsumer(kafka.brokers, string(trigger.Metadata.UID), []string{trigger.Spec.Topic}, consumerConfig)
+	kafka.logger.Info("created a new consumer", zap.Any("consumer", consumer))
+	if err != nil {
+		panic(err)
```
life1347: return error

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -80,18 +96,34 @@ func (kafka Kafka) subscribe(trigger *fv1.MessageQueueTrigger) (messageQueueSubs
...
 	producerConfig.Version = kafka.version
+
+	// Setup TLS for both producer and consumer
+	if kafka.tls {
+		consumerConfig.Net.TLS.Enable = true
+		producerConfig.Net.TLS.Enable = true
+		tlsConfig, err := kafka.getTLSConfig()
+
+		if err != nil {
+			panic(err)
```
life1347: return error

vadasambar: My bad. Will fix this. Thank you!

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: My bad. Will fix this. Thank you!

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: Will do this. Thank you!

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: Will do this. Thank you!

charts/fission-all/templates/deployment.yaml
vadasambar: Oops! Thank you for noticing it. Will fix it. Thank you!

vadasambar: Fixed

pkg/mqtrigger/mqtrigger.go
vadasambar: I think checking for `TLS_ENABLED` makes sense but I was thinking of `readSecrets` as a generic way to read any kind of secrets for a message queue (these secrets can be unrelated to TLS). What do you think?

life1347:
@vadasambar any updates for this?

pkg/mqtrigger/mqtrigger.go
vadasambar: @life1347 , what do you think of this ^ ?

vadasambar:
> @vadasambar any updates for this?

Hi, sorry I was a little busy this week. I will check this as soon as I get some time. Thanks for the reminder :+1:

vadasambar:
@life1347 , I did not get a reply on the review comment. I thought I would just bring it to notice because it is buried in other review comments. 
Link to review comment: https://github.com/fission/fission/pull/1300#discussion_r322142708
> I think checking for TLS_ENABLED makes sense but I was thinking of readSecrets as a generic way to read any kind of secrets for a message queue (these secrets can be unrelated to TLS). What do you think?

pkg/mqtrigger/mqtrigger.go
life1347: make sense

life1347:
sorry.. i missed that comment. that make sense to me.

vadasambar:
I think I have made all the changes requested from my side. If there's anything left, please let me know. :)

life1347:
I will review the PR again before weekend, before that please fix the code conflicts. Thanks!

life1347:
There are still some unresolved comments, please fix them.

charts/fission-all/values.yaml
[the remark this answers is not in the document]
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: If we want to use `readSecrets` as a generic function (i.e., use them for other messaging queues as well), it would make more sense to log that the secretsPath was not found and let it return nil. 

There are two cases which are of importance here:
1. When no secrets are present and they are not required. i.e., there is no need for secrets e.g. connecting to a non TLS secured Kafka cluster would not need secrets.
2. When no secrets are present but they should be present (secrets are required). This would be the case when the Kafka cluster is to be connected using TLS but the secrets were not mounted properly.

Returning nil should handle the first scenario just fine because there will be no error and throwing an error is not really required here. Problem arises in the second case. If the secrets cannot be found at the path but they should be present there, an error should be raised so that we know what went wrong. Returning nil here would be just ignoring this case and it would make it difficult to debug the problem. 

I believe returning error is what will handle the second case. But returning error when secrets will not be required (connecting to non TLS Kafka cluster) and distinguishing this harmless thing with an actual error would be difficult. 

One way to fix this can be, we log the information (maybe as a warning) so that we know that the secrets were not found and hope that the messaging queue would throw an error which together with the warning would help debug the problem. We can word the warning to convey the intensity of the problem. I am not very happy with this fix but this is something which can solve the problem for now. 

What do you think?

vadasambar:
Sorry, I thought I had fixed them. I have 3 pending changes. Thank you for pointing them out. I think I will fix them and then rebase/merge my branch with master.

pkg/mqtrigger/mqtrigger.go
life1347: I don't quite get it. If the secrets are not necessary, why calling this function?

It's the caller's responsibility to decide whether it needs to call the function, instead of calling a function that will silently ignore the error and pretend nothing happen. A generic function should always return the error instead of hiding it (unless the function itself can handle it) and let the caller to decide what to do to handle the error.

In this case, I think we can change the function signature

```
// caller checks whether it need to read secrets
secretsPath := strings.TrimSpace(os.Getenv("MESSAGE_QUEUE_SECRETS"))
if secretsPath > 0 {
    secrets, err := readSecrets(secretsPath, logger) (map[string][]byte, error)
}


// change the function signature to
func readSecrets(path string, logger *zap.Logger) (map[string][]byte, error) {

    // return if no secrets exist
	if _, err := os.Stat(path); os.IsNotExist(err) {
        return nil, err
    }
....
```

It brings a couple of benefits,

1. A more generic function that can accept any path instead of reading from the environment variable.
2. It's easier to write unit tests for the function.

pkg/mqtrigger/messageQueue/messageQueue.go
```diff
@@ -44,6 +44,8 @@ type (
 	MessageQueueConfig struct {
 		MQType string
 		Url    string
+		// Secrets *v1.Secret
```
life1347: What is this for? Maybe we can remove it?

vadasambar: Removed

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: Done

pkg/mqtrigger/messageQueue/kafka.go
vadasambar: Done

pkg/mqtrigger/mqtrigger.go
vadasambar: This makes sense. :+1:

life1347:
@vadasambar thanks for all your effort! Merged!

PR #1324, opened 2019-09-23
support custom private go vendor

ti:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1324)

### DIFF
--- a/environments/go/builder/build.sh
+++ b/environments/go/builder/build.sh
@@ -22,6 +22,15 @@ fi
 
 cd ${srcDir}
 
+# use private vendor is the vendor dir is exsit
+if  [ -d "vendor" ] && [ ! -z ${GOLANG_VERSION} ] && version_ge ${GOLANG_VERSION} "1.11"; then
+  if [ ! -f "go.mod" ]; then
+     go mod init
+     fi
+  go build -mod=vendor -buildmode=plugin -i -o ${DEPLOY_PKG} .
+  exit 0;
+  fi
+
 if [ ! -z ${GOLANG_VERSION} ] && version_ge ${GOLANG_VERSION} "1.12"; then
     if [ -f "go.mod" ]; then
         go mod download


ti:
@life1347  

for bug: golang 1.12 does not support private vendor (the vendor may be not in the public interface, eg:gitlab or somewhere else)

fixed: if the golang have "vendor" dir, then just use the vendor to build.

environments/go/builder/build.sh
```diff
@@ -22,6 +22,15 @@ fi
 
 cd ${srcDir}
 
+# use private vendor is the vendor dir is exsit
+if  [ -d "vendor" ] && [ ! -z ${GOLANG_VERSION} ] && version_ge ${GOLANG_VERSION} "1.11"; then
```
life1347: Fission enables GO111MODULE environment variable after 1.12, so here we can set greater and equal to 1.12
```
version_ge ${GOLANG_VERSION} "1.12"
```

environments/go/builder/build.sh
```diff
@@ -22,6 +22,15 @@ fi
...
+  if [ ! -f "go.mod" ]; then
+     go mod init
+     fi
+  go build -mod=vendor -buildmode=plugin -i -o ${DEPLOY_PKG} .
+  exit 0;
+  fi
```
life1347: Please fix the indent of the newly added code block.

environments/go/builder/build.sh
```diff
@@ -22,6 +22,15 @@ fi
 
 cd ${srcDir}
 
+# use private vendor is the vendor dir is exsit
+if  [ -d "vendor" ] && [ ! -z ${GOLANG_VERSION} ] && version_ge ${GOLANG_VERSION} "1.11"; then
+  if [ ! -f "go.mod" ]; then
```
life1347: I think what following block doing is to prevent `cannot find module` issue.
```
if [ ! -f "go.mod" ]; then
     go mod init
     fi
```
Please add a line of comment to explain why we need it.

life1347:
Hi @ti , this PR is really valuable for go env users and I really want to merge this PR. Since we are close to the release date of 1.6, If you need any help, feel free to ping me. :)

life1347:
Hey @ti , because the release date is 10/10, would you mind to address the comments above? Thanks.

life1347:
Merged by #1346 . All the credits of PR1346 go to @ti . Thanks for your report and effort.

PR #1355, opened 2019-10-14
calling the function that handles kafka messages, asynchronously

### LINKED ISSUE #1354: Kafka MQT message processing delays
When Kafka MQT handles messages from Kafka the loop handles one message at a time and waits for the processing to finish:

https://github.com/fission/fission/blob/e42f81a7a1b65caf66e49931df6833417645c1e1/pkg/mqtrigger/messageQueue/kafka.go#L168-L171

The call to function `kafkaMsgHandler` should be asynchronous so that messages are processed faster 

https://github.com/fission/fission/blob/e42f81a7a1b65caf66e49931df6833417645c1e1/pkg/mqtrigger/messageQueue/kafka.go#L204

viveksinghggits:
This change fixes #1354

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1355)


vishal-biyani:
Hey @viveksinghggits The change looks good to me - were you able to test it with a Kafka setup - if you can post us logs of before and after - that will be super helpful. CC @msshroff

viveksinghggits:
Hi @vishal-biyani ,
To test this I wrote a python that takes 15 seconds to get executed, in the earlier implementation the mqtrigger picked one message from kafka topic and waited for it to get completed and then picked out another message but after the change mqtrigger is picking the messages asynchronously and calling the function. Since the log level in the code has been changed I didnt get any useful logs in the mqtrigger pod, i will change the log level and update the logs here.

viveksinghggits:
Hi @vishal-biyani ,
The test scenario that mentioned are sufficient or I have to cover something else, if not I think we can go ahead with this PR.
cc @life1347

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -201,7 +199,7 @@ func (kafka Kafka) unsubscribe(subscription messageQueueSubscription) error {
 	return subscription.(*cluster.Consumer).Close()
 }
 
-func kafkaMsgHandler(kafka *Kafka, producer sarama.SyncProducer, trigger *fv1.MessageQueueTrigger, msg *sarama.ConsumerMessage) bool {
+func kafkaMsgHandler(kafka *Kafka, producer sarama.SyncProducer, trigger *fv1.MessageQueueTrigger, msg *sarama.ConsumerMessage, consumer *cluster.Consumer) bool {
```
life1347: I think we can remove the return bool from the function signature if we no longer need it.

viveksinghggits: Changed the signature of the function, removed the return type.

pkg/mqtrigger/messageQueue/kafka.go
```diff
@@ -227,7 +225,7 @@ func kafkaMsgHandler(kafka *Kafka, producer sarama.SyncProducer, trigger *fv1.Me
 		kafka.logger.Error("failed to create HTTP request to invoke function",
 			zap.Error(err),
 			zap.String("function_url", url))
-		return false
+
```
life1347: Hey @viveksinghggits , you cannot just remove the return here and other places directly, they still need to return but without the bool...

life1347: Please check the code and its purpose again, thanks.

viveksinghggits: My Bad @life1347, I understand and will make the changes.

PR #1374, opened 2019-11-03
added example for builder podspec

viveksinghggits:
added example for builder podspec, in `examples/spec-example/podspec-example`

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1374)


life1347:
Please add a simple README to describe what and how to use this spec example `examples/spec-example/podspec-example`. Then, we can merge it.

viveksinghggits:
Hi @life1347,
In the https://github.com/viveksinghggits/fission/blob/pod-spec-example/examples/spec-example/podspec-example/specs/README#L43-#L53 I have mentioned in short about why do we have the `podspec` in the environment builder. Do you want me to explain that in more detail.

life1347:
You should move that part to the README under `examples/spec-example/podspec-example`. The README under specs is generated by code and should not change it.

life1347:
LGTM

Issue #107, opened 2017-02-06
Support Openshift 3


yoshiwaan:
Hi,

Great idea and I'm liking the ease of implementation. The company I'm working with at the moment has deployed Openshift 3 rather than pure kubernetes, so I was hoping to have native support for Openshift 3 in Fission.

Openshift is mostly a superset of Kubernetes so it's not not too difficult to get working, I managed to get Fission up and running today with broad strokes, but was hoping to have it done at the application level so there are fewer security concessions.

Openshift has stricter security rules than Kubernetes with a few specifically affecting Fission. That being said the security rules are pretty good, so might be something worth adding to Fission on Kubernetes as well. The main things are:
* Containers run as a random numeric user (so either need a volume or world writable directory to make file system changes)
* Containers run with fewer capabilities
* There is a concept of projects in Openshift. They are directly mapped to namespaces, but you need to explicitly allow permissions for API calls between projects

**Summary**
In short the following needs doing to allow Fission to work on Openshift 3:
1. The fission etcd deployment can't create it's config store directory due to permission. The fix for Openshift would be to specify ETCD_DATA_DIR to either use a volume or create the etcd config dir in something like /tmp, which is world writable
1. The fission controller can't create it's /filestore directory. In Openshift it needs to create the /filestore directory somewhere that is world writable in the container or on a volume. The controller also seem to fail to function correctly unless the filestore directory is created but doesn't panic/error on that failure, so I think it's worth changing the behaviour to panic/exit on /filestore creation failure so that it's easier to debug and more clearly shows as an error (pod restarts)
1. The pool manager doesn't have permission to call cross namespace in Openshift 3, so calls to the fission-function namespace are rejected. The best fix in Openshift would be to create a custom role for the pool manager in the deployment template (with only the API calls to the other project the pool manager needs to make) and assigned it to the correct service account in the fission-function project. An easier but less security fix is to assign the cluster-administrator role in the fission-function project to the default fission service account (system:serviceaccount:fission:default). Due to those missing rules this means that at least the following API calls fail from the pool manager in the current deployment:
   * list extensions.deployments in the fission-function namespace
   * create extensions.deployments in the fission-function namespace
1. It might just be the file permissions above, but until I granted the capability to run as root for the controller (the same set of capabilities as Docker assigns by default) I would get the exception right at the bottom of this ticket. The only way I could fix the issue was to grant those capabilities (as I couldn't change the directory without changing source code like for etcd) so I'm not sure if there's more that needs doing.

[...]

PR #1405, opened 2019-11-12
Fix logger init container has no permission to create directory

life1347:
On the machine that doesn't allow the container to create sub-directory
in the directory that mounts into the container, the logger enters CrashLoopBackOff.

This PR changes the location of the symlink path to prevent the permission
problem when init container trying to create a directory and a new PV
is added for fluent-bit to store its database file.

Fix logger problem mentioned in https://github.com/fission/fission/issues/107

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1405)

### DIFF
--- a/charts/fission-all/config/fluentbit.conf
+++ b/charts/fission-all/config/fluentbit.conf
@@ -10,7 +10,7 @@
     Path ${LOG_PATH}
     Mem_Buf_Limit 5MB
     Parser docker
-    DB /var/log/fission/flb_kube.db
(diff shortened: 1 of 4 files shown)


life1347:
This PR needs more tests and validation on OpenShift. Here are some useful references for users to fix the logger problem before the PR gets merged.
1. https://github.com/code-ready/crc
2. https://github.com/fluent/fluentd-kubernetes-daemonset#running-on-openshift
3. https://kazuhira-r.hatenablog.com/entry/2019/01/27/013423

PR #1411, opened 2019-11-16
Add --spec to package command

life1347:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1411)


life1347:
Fix #1407

PR #1462, opened 2019-12-10
Add the not present cmname while fn create in err message [CLI-UX]

viveksinghggits:
If we are referring more than one configmaps in our command while creating function it fails, but doesnt say the configmap name that is not present.
This change fixes that.
This is how the error message gets displayed now
```
Error: error checking configmap cmtwo: Resource not found - NotFound
```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1462)


life1347:
Could you also address the same problem for the secret at L238?

life1347:
LGTM

Issue #1404, opened 2019-11-12
Remove FISSION_ROUTER from codebase


life1347:

sahil-lakhwani:
Hi @life1347 
Taking a look at this, I see the `FISSION_ROUTER` is used at:

- `pkg/fission-cli/cmd/function/test.go` for `fission function test`
- all the shell files in demo examples

For `fission function test` the `FISSION_ROUTER` is optional, and I think the else case will always work

For all the shell files, I think we should be able to get the cluster IP and the address of the fission `router` service  using `kubectl`. Then use that to replace `FISSION_ROUTER`

Need your thoughts.

life1347:
So far I don't see any potential issue that may break things. And actually, we don't need to use `kubectl`. 

https://github.com/fission/fission/blob/711d7c2c41bd3d1171636026e093fd36ecd30e05/pkg/fission-cli/cmd/function/test.go#L62-L67

This part will create a local port forwarder to the remote server. So you can simply remove the If condition. But please add a warning to notify user the `FISSION_ROUTER` is no longer supported when it's not empty.

sahil-lakhwani:
Right. A warning is good.
So my mention for `kubectl` was for where `FISSION_ROUTER` is being referenced in shell files under `/demos`.
For example:
https://github.com/fission/fission/blob/711d7c2c41bd3d1171636026e093fd36ecd30e05/demos/hello/run.sh#L32-L37

Are we looking to replace references like these as well?

life1347:
I think the `$FISSION_ROUTER` in run.sh should be set by the user before running it. So I think we can leave it for now.

PR #1465, opened 2019-12-12
Remove FISSION_ROUTER for fn test

sahil-lakhwani:
Removing use of environment variable FISSION_ROUTER for `fission fn test`
The check for the variable is in a `if` block. Removing the `if` doesn't harm with some modification.
Added a warning for the user.

Resolves https://github.com/fission/fission/issues/1404

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1465)


life1347:
LGTM! Thanks 👍

PR #1468, opened 2019-12-17
Add spec list feature

### LINKED ISSUE #852: Fission spec list
It would be nice to have `list` like functionality for fission specs - which lists all components in the spec.

```
$ fission spec list
Archive archive://iot-kafka-consumer-1-0-0-jar-with-dependencies-jar
Archive archive://iot-kafka-producer-1-0-0-jar
2 environments: java, python
3 packages: iot-kafka-consumer-1-0-0-jar-with-dependencies-jar-ywpb, iot-kafka-producer-1-0-0-jar-rrvv, web-pkg
7 functions: count, data, fuelavg, iot-web, kc, kp, speedavg
```

anubhav6663:
Fixes #852 
Adding fission spec list functionality. It lists Functions, Environment, Packages, HTTPTrigger, MessageQueueTriggers, Canary Config, TimeTrigger, KubeWatch created by an application specification. 

Here is the screenshot of the result: 
speclistout

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1468)

### DIFF
--- a/pkg/fission-cli/cmd/spec/command.go
+++ b/pkg/fission-cli/cmd/spec/command.go
@@ -62,11 +62,11 @@ func Commands() *cobra.Command {
 
 	listCmd := &cobra.Command{
 		Use:   "list",
-		Short: "List all Fission resources in the application specification",
+		Short: "List all the resources that were created through this spec",
 		RunE:  wrapper.Wrapper(List),
 	}
 	wrapper.SetFlags(listCmd, flag.FlagSet{
-		Optional: []flag.Flag{flag.SpecDir},
+		Optional: []flag.Flag{flag.SpecDeployID, flag.SpecDir},
 	})
 
 	command := &cobra.Command{
@@ -75,7 +75,7 @@ func Commands() *cobra.Command {
 		Short:   "Manage a declarative application specification",
 	}
 
-	command.AddCommand(initCmd, validateCmd, applyCmd, destroyCmd, listCmd)
+	command.AddCommand(initCmd, validateCmd, applyCmd, listCmd, destroyCmd)
 
 	return command
 }
--- a/pkg/fission-cli/cmd/spec/command.go
+++ b/pkg/fission-cli/cmd/spec/command.go
@@ -60,13 +60,22 @@ func Commands() *cobra.Command {
 		Optional: []flag.Flag{flag.SpecDir},
 	})
 
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all Fission resources in the application specification",
+		RunE:  wrapper.Wrapper(List),
+	}
+	wrapper.SetFlags(listCmd, flag.FlagSet{
+		Optional: []flag.Flag{flag.SpecDir},
+	})
+
 	command := &cobra.Command{
 		Use:     "spec",
 		Aliases: []string{"specs"},
 		Short:   "Manage a declarative application specification",
 	}
 
-	command.AddCommand(initCmd, validateCmd, applyCmd, destroyCmd)
+	command.AddCommand(initCmd, validateCmd, applyCmd, destroyCmd, listCmd)
 
 	return command
 }
--- a/pkg/fission-cli/cmd/spec/list.go
+++ b/pkg/fission-cli/cmd/spec/list.go
@@ -47,29 +47,178 @@ func (opts *ListSubCommand) do(input cli.Input) error {
 }
 
 func (opts *ListSubCommand) run(input cli.Input) error {
-	// get specdir
-	specDir := util.GetSpecDir(input)
-
-	// read everything from spec
-	fr, err := ReadSpecs(specDir)
-	if err != nil {
-		return errors.Wrap(err, "error reading specs")
+	deployID := input.String(flagkey.SpecDeployID)
+	if len(deployID) == 0 {
+		// get specdir and read the deployID
+		specDir := util.GetSpecDir(input)
+		fr, err := ReadSpecs(specDir)
+		if err != nil {
+			return errors.Wrap(err, "error reading specs")
+		}
+		deployID = fr.DeploymentConfig.UID
 	}
 
-	fns, err := opts.Client().V1().Function().List("")
+	allfn, err := getAllFunctions(opts)
 	if err != nil {
-		return errors.Wrap(err, "error listing functions")
+		return errors.Wrap(err, "error getting Functions from all namespaces")
 	}
+	specfns := getAppliedFunctions(allfn, deployID)
+	ShowFunctions(specfns)
 
-	flag := true
-	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
-	for _, f := range fns {
-		if f.Metadata.Annotations["fission-uid"] == fr.DeploymentConfig.UID {
-			if flag {
-				fmt.Printf("\nFunctions:\n")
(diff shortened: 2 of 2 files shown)


life1347:
This is awesome! I'm reviewing it now, cannot wait to merge it.

pkg/fission-cli/cmd/spec/command.go
```diff
@@ -60,13 +60,22 @@ func Commands() *cobra.Command {
 		Optional: []flag.Flag{flag.SpecDir},
 	})
 
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all Fission resources in the application specification",
```
life1347: Since the list command only lists applied resources, how about we change it to 
```
List all applied Fission resources in the application specification
```

pkg/fission-cli/cmd/spec/command.go
```diff
@@ -60,13 +60,22 @@ func Commands() *cobra.Command {
...
+	})
+
 	command := &cobra.Command{
 		Use:     "spec",
 		Aliases: []string{"specs"},
 		Short:   "Manage a declarative application specification",
 	}
 
-	command.AddCommand(initCmd, validateCmd, applyCmd, destroyCmd)
+	command.AddCommand(initCmd, validateCmd, applyCmd, destroyCmd, listCmd)
```
life1347: nit pick: `initCmd, validateCmd, applyCmd, listCmd, destroyCmd`
change the order of command to make list command shows before destroy.
```
# before
Available Commands:
  init        Create an initial declarative application specification
  validate    Validate declarative application specification
  apply       Create, update, or delete resources from application specification
  destroy     Delete all Fission resources in the application specification
  list        List all Fission resources in the application specification

# after:
Available Commands:
  init        Create an initial declarative application specification
  validate    Validate declarative application specification
  apply       Create, update, or delete resources from application specification
  list        List all Fission resources in the application specification
  destroy     Delete all Fission resources in the application specification
```

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,266 @@
...
+	// get specdir
+	specDir := util.GetSpecDir(input)
+
+	// read everything from spec
+	fr, err := ReadSpecs(specDir)
+	if err != nil {
+		return errors.Wrap(err, "error reading specs")
+	}
+
+	fns, err := opts.Client().V1().Function().List("")
```
life1347: Let's move the resource list & print code to it's own function instead of putting all of them in run function.
For example, 
```
func showAppliedFunctions(deployID string) {
...
}
```
In this way, we are able to write tests for different resources.

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error listing functions")
+	}
+
+	flag := true
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	for _, f := range fns {
+		if f.Metadata.Annotations["fission-uid"] == fr.DeploymentConfig.UID {
+			if flag {
+				fmt.Printf("\nFunctions:\n")
```
life1347: I'm just wondering why do this in the for loop, any reasons for not printing these two lines before we entering for loop?

life1347: To keep code consistency let change it a bit: `fmt.Fprint("\nFunctions:\n")`

pkg/fission-cli/cmd/spec/command.go
```diff
@@ -60,13 +60,22 @@ func Commands() *cobra.Command {
 		Optional: []flag.Flag{flag.SpecDir},
 	})
 
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all Fission resources in the application specification",
+		RunE:  wrapper.Wrapper(List),
+	}
+	wrapper.SetFlags(listCmd, flag.FlagSet{
+		Optional: []flag.Flag{flag.SpecDir},
```
life1347: How about we add a flag that allows a user to specify the deployID directly instead of moving into spec dir everytime?

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,266 @@
...
+	ws, err := opts.Client().V1().KubeWatcher().List("")
+	if err != nil {
+		return errors.Wrap(err, "Error listing kubewatches")
+	}
+	flag = true
+	w = tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	for _, wa := range ws {
+		if wa.Metadata.Annotations["fission-uid"] == fr.DeploymentConfig.UID {
+			if flag {
+				fmt.Printf("\nKube Watches:\n")
```
life1347: `Kube Watchers`

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,266 @@
...
+	hts, err := opts.Client().V1().HTTPTrigger().List("")
+	if err != nil {
+		return errors.Wrap(err, "Error listing HTTPTrigger")
+	}
+	flag = true
+	w = tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	for _, trigger := range hts {
+		if trigger.Metadata.Annotations["fission-uid"] == fr.DeploymentConfig.UID {
+			if flag {
+				fmt.Printf("\nHttp Triggers:\n")
```
life1347: `HTTP Triggers`

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,266 @@
...
+
+	fns, err := opts.Client().V1().Function().List("")
+	if err != nil {
+		return errors.Wrap(err, "error listing functions")
+	}
+
+	flag := true
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
+	for _, f := range fns {
+		if f.Metadata.Annotations["fission-uid"] == fr.DeploymentConfig.UID {
```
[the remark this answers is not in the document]
life1347: It's not possible for now since the CLI client talks to the controller instead of kubernetes and it hasn't implemented this feature.

pkg/fission-cli/cmd/spec/list.go
anubhav6663: These two line will be printed only if there is at least one function resource applied by the spec. If there is no resource present it will not print the tab.

life1347: ok, that's fair.

anubhav6663:
@viveksinghggits @life1347  I have done the changes suggested by you. Please review the same  and provide you comments.

pkg/fission-cli/cmd/spec/list.go
```diff
@@ -0,0 +1,454 @@
...
+			if f.Metadata.Annotations["fission-uid"] == deployID {
+				kubeWatchTriggerlist = append(kubeWatchTriggerlist, f)
+			}
+		}
+	}
+	return kubeWatchTriggerlist
+}
+
+// ShowFunctions displays info of Functions
+func ShowFunctions(fns []fv1.Function) {
```
life1347: Is it intentional to export the function? If not, how about we change it to lower case `showFunctions`?
(here and other functions.)

anubhav6663: Yes. It is intentional.  If we will need to print any of the resource in same format as above we can use the same function. 
I think we should export print functions of `fission fn list` or others as well. In that case we can use existing code and reduce code duplication.

life1347: fair enough. LGTM

PR #1471, opened 2019-12-19
Add failed/success output to spec validate

### LINKED ISSUE #621: `fission spec validate` should output something on success
Very minor issue, but `spec validate` doesn't output anything when there are no errors.  It would be nice to output a summary of the specs (e.g. number of functions, packages, envs, triggers) on success.

anubhav6663:
Fixes #621  
Adding Failed/Passed status on spec validate. 
On pass it outputs the number of resources present in the spec.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1471)


pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -117,11 +117,11 @@ func (opts *ApplySubCommand) run(input cli.Input) error {
 		}
 
 		// validate
-		err = fr.Validate(input)
+		warning, err := fr.Validate(input)
 		if err != nil {
 			return errors.Wrap(err, "error validating specs")
 		}
-
+		_ = warning
```
viveksinghggits: Hi @anubhav6663 
if we have to swallow the warning we can use `_` at the line 120 itself. But please make sure Validate prints the warnings.

anubhav6663: Yeah `Validate` prints the warnings.

viveksinghggits: then I suggest to remove line `124` and have it in `120`

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -276,9 +276,10 @@ func (fr *FissionResources) validateFunctionReference(functions map[string]bool,
 	return nil
 }
 
-func (fr *FissionResources) Validate(input cli.Input) error {
```
life1347: I understand why returning the bool here. But I don't think it's a good way to do so.
Let's return a list of warning message `[]string` instead of `bool` to indicate there are warnings when applying spec.

life1347: For example:
```
warnings := []string

warnings = append(warnings, <warn message>)

``

File: pkg/fission-cli/cmd/spec/validate.go [hunk 3]
```diff
@@ -56,11 +57,16 @@ func (opts *ValidateSubCommand) run(input cli.Input) error {
 	}
 
 	// this does the rest of the checks, like dangling refs
-	err = fr.Validate(input)
+	warning, err := fr.Validate(input)
 	if err != nil {
 		return errors.Wrap(err, "error validating specs")
 	}
-
+	if warning {
```
life1347: We don't need this if-else actually. We can always print `Spec validation successful\nSpec contains ....` and shows warnings in the end.

pkg/fission-cli/cmd/spec/spec.go
anubhav6663: We are already printing the warning on console. I thought there is no use of saving warnings again. I don't think we are going to use the warning again. @life1347

pkg/fission-cli/cmd/spec/validate.go
anubhav6663: Okay. We can use the warning list here.

pkg/fission-cli/cmd/spec/spec.go
life1347: oh, I mean instead of printing warning in `Validate()` we should return all of them. In this way, we are also able to write tests for the function by check returned msg instead of monitoring the console output.

anubhav6663:
@life1347 Please review the changes.

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -456,30 +455,33 @@ func (fr *FissionResources) Validate(input cli.Input) error {
...
 		if e.Spec.Poolsize != 3 && e.Spec.Version < 3 {
-			console.Warn("Poolsize can only be configured when environment version equals to 3, default poolsize 3 will be used for creating environment pool.")
+			warnings = append(warnings, fmt.Sprintf("Poolsize can only be configured when environment version equals to 3, default poolsize 3 will be used for creating environment pool."))
 		}
 	}
 
 	for _, f := range fr.Functions {
 		if _, ok := environments[fmt.Sprintf("%s:%s", f.Spec.Environment.Name, f.Spec.Environment.Namespace)]; !ok {
-			console.Warn(fmt.Sprintf("Environment %s is referenced in function %s but not declared in specs", f.Spec.Environment.Name, f.Metadata.Name))
+
```
life1347: Please remove the blank line. (here and following)

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -276,8 +276,10 @@ func (fr *FissionResources) validateFunctionReference(functions map[string]bool,
 	return nil
 }
 
-func (fr *FissionResources) Validate(input cli.Input) error {
+//Validate validateSs if there is any
```
life1347:
```
// Validate validates if there is any
```

pkg/fission-cli/cmd/spec/spec.go
anubhav6663: Done

life1347:
LGTM

PR #1474, opened 2019-12-20
Fix hello-spec-example

sahil-lakhwani:
On applying the spec in `hello-spec-example` the executor fails with below logs, and no pods are creator. As the logs suggests, no executor type is specified.

```
"fission_http_error":"Unknown executor type ' ' ''
```

Adding an executor type to the function spec in this PR

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1474)


life1347:
Hi @sahil-lakhwani !
I'm fine with the change but just curios what version will cause this issue? Cause with 1.7.1 I'm able to apply it without any problem. If it shows `Unknown executor type` then we may need to take a look at whether there are any compatibility problems.

life1347:
Looks like it's a very old spec example. You may also want to change the name of `fission-config.yaml` to `fission-deployment-config.yaml` so that spec command will be able to read the deployment config.

sahil-lakhwani:
@life1347 Renamed the file

PR #1480, opened 2020-01-07
Add --dry option to view the spec without saving it

### LINKED ISSUE #895: Add option to dump spec to stdout
For debugging, understanding and code generation reasons it would be useful to have the option to print the generated specs to STDOUT instead of a directory. Often you might simply want to see the spec without having to initialize a spec directory, dump the spec there, retrieve it, and then delete the spec directory - just for viewing the generated spec.

I can currently think of three possible approaches.

**Option 1**: dump specs to stdout if no spec directory is present
```
$ fission env create --name python --image fission/python-env --spec
<spec>
```
Upside: No additional parameters for users to understand/use.
Downside: we lose current warning/error, which might lead to confusion with users how to use the spec workflow.

**Option 2**: use dedicated option for dumping the spec to stdout (for example `--dry` or `--yaml`)
```
$ fission env create --name python --image fission/python-env --dry
<spec>
```
Upside: No changes to existing behaviour. Decoupled from the spec approach
Downside: yet another option, which is a bit ambiguous with the current `--spec` option, since spec can refer both to the spec workflow or getting the k8s spec.

**Option 3**: allow users to provide an optional destination to the `--spec` option
```
$ fission env create --name python --image fission/python-env --spec /dev/stdout
<spec>
```
The default behaviour (no argument provided) will result in the current behaviour.
Upside: No changes to existing behaviour. No additional option
Downside: having it print to the stdout (`--spec /dev/stdout`) is a bit cumbersome for inexperienced users. `--spec 0` achieves the same and is more platform-agnostic, but is a bit more cryptic. Besides, this issue could be addressed in the documentation/usage instructions. 

For the solution, I have a slight preference for the third option, but I am interested in other opinions?

anubhav6663:
Resolves #895 
Adding --dry option, by using this we can view the generated spec without saving it.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1480)

### DIFF
--- a/pkg/fission-cli/cmd/spec/spec.go
+++ b/pkg/fission-cli/cmd/spec/spec.go
@@ -162,6 +162,78 @@ func save(data []byte, specDir string, specFile string) error {
 	return nil
 }
 
+// called from `fission * create --dry`
+func SpecDump(resource interface{}) error {
+	var meta metav1.ObjectMeta
+	var kind string
+
+	// make sure we're writing a known type
+	var data []byte
+	var err error
+	switch typedres := resource.(type) {
+	case types.ArchiveUploadSpec:
+		typedres.Kind = "ArchiveUploadSpec"
+		meta = metav1.ObjectMeta{
+			Name: typedres.Name,
+		}
+		kind = typedres.Kind
+		data, err = yaml.Marshal(typedres)
+	case fv1.Package:
+		typedres.TypeMeta.APIVersion = fv1.CRD_VERSION
+		typedres.TypeMeta.Kind = "Package"
+		meta = typedres.Metadata
+		kind = typedres.TypeMeta.Kind
+		data, err = yaml.Marshal(typedres)
+	case fv1.Function:
+		typedres.TypeMeta.APIVersion = fv1.CRD_VERSION
+		typedres.TypeMeta.Kind = "Function"
+		meta = typedres.Metadata
+		kind = typedres.TypeMeta.Kind
+		data, err = yaml.Marshal(typedres)
+	case fv1.Environment:
+		typedres.TypeMeta.APIVersion = fv1.CRD_VERSION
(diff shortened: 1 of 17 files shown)


anubhav6663:
@life1347  Can you please review this PR.

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -162,6 +162,78 @@ func save(data []byte, specDir string, specFile string) error {
 	return nil
 }
 
+// called from `fission * create --dry`
+func SpecDump(resource interface{}) error {
```
life1347: Please don't duplicate `SpecSave`code directly, it lowers the maintainability of codebase.
We can instead of moving the duplicate part to an independent function and call it in `SpecSave` and `SpecDump`.

pkg/fission-cli/flag/flag.go
```diff
@@ -170,6 +170,7 @@ var (
 	SpecWait     = Flag{Type: Bool, Name: flagkey.SpecWait, Usage: "Wait for package builds"}
 	SpecWatch    = Flag{Type: Bool, Name: flagkey.SpecWatch, Usage: "Watch local files for change, and re-apply specs as necessary"}
 	SpecDelete   = Flag{Type: Bool, Name: flagkey.SpecDelete, Usage: "Allow apply to delete resources that no longer exist in the specification"}
+	SpecDump     = Flag{Type: Bool, Name: flagkey.SpecDump, Usage: "View the generated specs"}
```
life1347: To follow the naming style, how about we change the name to `SpecDry`?

life1347:
@anubhav6663 why closed this PR? I think it's almost in good shape.

anubhav6663:
@life1347  A lot of changes would have been required. Created another PR for the same issue and that is much better than this. 
https://github.com/fission/fission/pull/1483

PR #1483, opened 2020-01-09
Add --dry option to view the generated spec without saving

### LINKED ISSUE #895: Add option to dump spec to stdout

anubhav6663:
Resolves #895 
Adding --dry option to view the generated spec. It can be used with --spec as well. 
We can view and save the spec at same time, if we will provide --dry and --spec in same command. --spec will work same as previously ( specs directory is required) but you can view your generated specs.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1483)

### DIFF
--- a/pkg/fission-cli/cmd/spec/spec.go
+++ b/pkg/fission-cli/cmd/spec/spec.go
@@ -33,6 +33,7 @@ import (
 	"github.com/fission/fission/pkg/fission-cli/cliwrapper/cli"
 	"github.com/fission/fission/pkg/fission-cli/cmd/spec/types"
 	"github.com/fission/fission/pkg/fission-cli/console"
+	flagkey "github.com/fission/fission/pkg/fission-cli/flag/key"
 	"github.com/fission/fission/pkg/fission-cli/util"
 	"github.com/fission/fission/pkg/utils"
 )
@@ -162,8 +163,8 @@ func save(data []byte, specDir string, specFile string) error {
 	return nil
 }
 
-// called from `fission * create --spec`
-func SpecSave(resource interface{}, specFile string) error {
+// called from `fission * create --spec` or `fission * create --dry`
+func SpecSave(resource interface{}, specFile string, input cli.Input) error {
 	var meta metav1.ObjectMeta
 	var kind string
 	var specDir = "specs"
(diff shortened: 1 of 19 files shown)


anubhav6663:
@life1347  Please review this. Thanks.

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -162,8 +163,8 @@ func save(data []byte, specDir string, specFile string) error {
 	return nil
 }
 
-// called from `fission * create --spec`
-func SpecSave(resource interface{}, specFile string) error {
+// called from `fission * create --spec` or `fission * create --dry`
+func SpecSave(resource interface{}, specFile string, input cli.Input) error {
```
life1347: Hi @anubhav6663 , 

Sorry, maybe I didn't explain it well. I mean to move the encoding part to an independent function instead of merging two functions. The reason not to merge them is it makes harder for writing unit test.

Please check this sample code: https://gist.github.com/life1347/829271ea0dfa35e0ad05f18bba4b65ce

pkg/fission-cli/cmd/spec/spec.go
```diff
@@ -222,11 +223,20 @@ func SpecSave(resource interface{}, specFile string) error {
...
+		if input.Bool(flagkey.SpecDry) {
+			return fmt.Errorf("can't display resource %#v spec", resource)
+		}
 		return fmt.Errorf("can't save resource %#v", resource)
 	}
 	if err != nil {
 		return errors.Wrap(err, "Couldn't marshal YAML")
 	}
+	if input.Bool(flagkey.SpecDry) {
+		console.Info(fmt.Sprintf("Generated Spec\n%v", (string(data))))
```
life1347: People may want to do something like `fission create xxx --dry > a.yaml`.
So we can just print the crd YAML directly without adding prepending string.
And, `console.XXX` is to print the message for the user. So if you want to print the content of CRD, use `fmt.Printf(string(data))`.

life1347:
hey, any updates?

PR #1500, opened 2020-01-21
Move to new CI cluster

life1347:
Don't know somehow after moving to the new CI cluster with newer Kubernetes version (13 -> 15) the informer delete event arrives router after ~20secs and causes failure of ingress test. In order to not block the CI, this PR modified the ingress test script to regenerates the TEST_ID before creating a new trigger to avoid the name conflict issue.

This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1500)


vishal-biyani:
🎉

PR #1548, opened 2020-03-03
Charts: Improves custom properties

odelucca:
## ☕ Purpose

During our installation of Fission using `Argo`, we've noticed that, by default, there are some missing custom properties and minor bugs on Fission charts. They are:

- We can't set a custom Prometheus port
- The `post-install` and `post-upgrade` jobs name is the same

We managed to fix those using `sed` after fetching Fission yaml, but it would be highly appreciated if we could contribute by fixing those 2 points.

On this PR I've added the `prometheus.port` to the `core` and `all` charts, and also added a suffix to the jobs, to prevent errors when installing directly with `kubectl apply`

## 🧐 Checklist

- [x] Adds a suffix to prevent duplicated job names
- [x] Adds a custom Prometheus port option

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1548)


vishal-biyani:
Hi @odelucca - thanks for this useful PR. On a side note - we would love to understand a bit more about you using Fission with Argo. Can you please hit me on Fission slack.

odelucca:
@vishal-bihyani sure, I've already explained to @taching but I will get in touch with you

odelucca:
I'm closing this PR since I'm going to create a new one, with the contents of this PR, and also some feature toggles as seen in #1551

Issue #1553, opened 2020-03-09
Question: already have clusterscoped nats streaming


shaunc:
Newbie question (/doc request): I have a cluster with clusterscoped nats streaming, used for another purpose. I would like to try out fission. How can I set it up to use existing nats streaming cluster?

life1347:
You can change the NATS streaming url to the existing NATS streaming service.

https://github.com/fission/fission/blob/4fb1a8fbe8e52c0b0bf4ef3f19b02a79d8639a25/charts/fission-all/templates/deployment.yaml#L588-L589

shaunc:
ok ... hmm... so I have to patch the chart (or role my own) -- would you accept a PR to add a variable to the chart for `nats-streaming` service name?

samoscybercraig:
So you have to have the Cluster ID named to match the one fission expects - see #1549 for some more details on it.

shaunc:
They will be in different namespaces ... (unless there is a way to use an "externalName" service to forward? Never tried that before...) ... so I'd want url with host like -- e.g. -- `nats-streaming.nats-io.service.cluster.local`

vishal-biyani:
Thanks for raising the issue and PR @shaunc I think this can be closed now.

Issue #1549, opened 2020-03-04
Nats streaming integration cluster ID is hardcoded


samoscybercraig:
**Fission/Kubernetes version**

<pre>
Fission client: v1.8.0
Fission server: v1.8.0
</pre>

<pre>
Kubernetes Client Version: v1.17.3
Kubernetes Server Version: v1.17.3
</pre>

**Kubernetes platform (e.g. Google Kubernetes Engine)**

Minikube

**Describe the bug**

When using an "external" nats streaming cluster with fission, it is not possible as the cluster id required is hardcoded in the mqtrigger container which communicates with the nats-streaming element [CODE](https://github.com/fission/fission/blob/e7c0ff7d41b9ef93b60827314ff255c5b685134e/pkg/mqtrigger/messageQueue/nats.go#L35). Also the boolean in values.yaml for nats controls both nats-streaming and mqtrigger, one can't choose false to create their own nats streaming but true for mqtrigger and pass url,port,authentication and cluster id. This however can be achieved by manually adjusting the helm charts, however the main bug exists that setting cluster id is not used in the mqtrigger. 

Cluster IDs in generic clusters should be configurable, this cannot be tested with the nats-streaming operator as it uses the hostname from the FQDN as the cluster. On validation the operator errors if using the "default" fission cluster id, as it is mixed case which is not valid for a hostname. 

**To Reproduce**
Create an external nats-streaming cluster: 

>`$ kubectl apply -f https://github.com/nats-io/nats-operator/releases/download/v0.5.0/00-prereqs.yaml`

>`$ kubectl apply -f https://github.com/nats-io/nats-operator/releases/download/v0.5.0/10-deployment.yaml`

>`$ kubectl apply -f https://raw.githubusercontent.com/nats-io/nats-streaming-operator/master/deploy/default-rbac.yaml`

>`$ kubectl apply -f https://raw.githubusercontent.com/nats-io/nats-streaming-operator/master/deploy/deployment.yaml`

Deploy a cluster using the operator 

<pre>
echo '
---
apiVersion: "nats.io/v1alpha2"
kind: "NatsCluster"
metadata:
  name: "example-nats"
spec:
  size: 3
' | kubectl apply -f -
</pre>

<pre>
echo '
---
apiVersion: "streaming.nats.io/v1alpha1"
kind: "NatsStreamingCluster"
metadata:
  name: "example-stan"
spec:
  size: 3
  natsSvc: "example-nats"
' | kubectl apply -f -
</pre>

[...]

PR #1554, opened 2020-03-16
feat: support external nats streaming

shaunc:
Adds support for an external nats cluster to helm fission-all chart. Also supports configuration of clusterID, clientID and queue group. 

Addresses:
* https://github.com/fission/fission/issues/1553
* https://github.com/fission/fission/issues/1549

The `nats` section of the helm chart adds:
*  external: false
*  url: ''
*  authToken: "defaultFissionAuthToken"
*  clusterID: "fissionMQTrigger"
*  clientID: "fission"
*  queueGroup: "fission-messageQueueNatsTrigger"

`url` empty uses current url template. `external` true removes the nats streaming cluster deployment and the service that exposes it.

In the mqtrigger/messageQueue/nats package, environment variables are consumed:
* MESSAGE_QUEUE_CLUSTER_ID
* MESSAGE_QUEUE_CLIENT_ID
* MESSAGE_QUEUE_QUEUE_GROUP

I wasn't sure where to document these. I made no changes to the other message queue trigger implementations. I did push a public docker image to `factfiber/fission-bundle:latest`, and run fission with values:
```
image: factfiber/fission-bundle
imageTag: latest
nats:
  url: nats://nats-cluster.nats-io.svc.cluster.local:4222
  external: true
  clusterID: "nats-streaming-cluster"
```
This ran successfully against a cluster-scoped streaming server in the `nats-io` namespace. There are no tests in the PR.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1554)

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -594,10 +594,10 @@ spec:
         - name: MESSAGE_QUEUE_CLIENT_ID
           value: {{ .Values.nats.clientID }}
         - name: MESSAGE_QUEUE_URL
-        {{- if not .Values.nats.url }}
-          value: nats://{{ .Values.nats.authToken }}@nats-streaming:4222
+        {{- if .Values.nats.authToken }}
+          value: nats://{{ .Values.nats.authToken }}@{{ .Values.nats.hostaddress }}
         {{- else }}
-          value: {{ .Values.nats.url }}
+          value: nats://{{ .Values.nats.hostaddress }}
         {{- end }}
         - name: TRACE_JAEGER_COLLECTOR_ENDPOINT
           value: "{{ .Values.traceCollectorEndpoint }}"
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -507,6 +507,7 @@ spec:
 #      serviceAccountName: fission-svc
 
 {{- if .Values.nats.enabled }}
+{{- if not .Values.nats.external }}
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -555,6 +556,7 @@ spec:
 {{- if .Values.extraCoreComponentPodConfig }}
 {{ toYaml .Values.extraCoreComponentPodConfig | indent 6 -}}
 {{- end }}
+{{- end }}
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -585,8 +587,18 @@ spec:
         env:
         - name: MESSAGE_QUEUE_TYPE
           value: nats-streaming
+        - name: MESSAGE_QUEUE_CLUSTER_ID
+          value: {{ .Values.nats.clusterID }}
+        - name: MESSAGE_QUEUE_QUEUE_GROUP
+          value: {{ .Values.nats.queueGroup }}
+        - name: MESSAGE_QUEUE_CLIENT_ID
+          value: {{ .Values.nats.clientID }}
         - name: MESSAGE_QUEUE_URL
+        {{- if not .Values.nats.url }}
           value: nats://{{ .Values.nats.authToken }}@nats-streaming:4222
+        {{- else }}
+          value: {{ .Values.nats.url }}
+        {{- end }}
         - name: TRACE_JAEGER_COLLECTOR_ENDPOINT
           value: "{{ .Values.traceCollectorEndpoint }}"
         - name: TRACING_SAMPLING_RATE
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -29,6 +29,7 @@ controllerPort: 31313
 routerPort: 31314
 
 ## Port at which NATS streaming service should be exposed
+## (only if nats enabled and not external)
 natsStreamingPort: 31316
 
 ## Set to false if you create the namespaces manually
@@ -132,8 +133,15 @@ router:
 ### NATS Streaming, enabled by default
 nats:
   enabled: true
+  external: false
+
+  # nats URL for use with external nats. If set will override template;
+  # pass authToken explicitly if necessary
+  url: ''
   authToken: "defaultFissionAuthToken"
   clusterID: "fissionMQTrigger"
+  clientID: "fission"
+  queueGroup: "fission-messageQueueNatsTrigger"
 
 ## Azure-storage-queue: enable and configure the details
 azureStorageQueue:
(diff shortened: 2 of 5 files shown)


pkg/mqtrigger/messageQueue/nats/nats.go
```diff
@@ -34,16 +35,32 @@ import (
...
 
 const (
-	natsClusterID  = "fissionMQTrigger"
-	natsProtocol   = "nats://"
-	natsClientID   = "fission"
-	natsQueueGroup = "fission-messageQueueNatsTrigger"
+	natsProtocol          = "nats://"
+	defaultNatsClusterID  = "fissionMQTrigger"
+	defaultNatsClientID   = "fission"
+	defaultNatsQueueGroup = "fission-messageQueueNatsTrigger"
```
life1347: You can just remove the `defaultNatsQueueGroup` since we never use it.

charts/fission-all/values.yaml
```diff
@@ -132,8 +133,15 @@ router:
 ### NATS Streaming, enabled by default
 nats:
   enabled: true
+  external: false
```
life1347: we can remove `external` and check the url value is empty in the charts.

charts/fission-all/templates/deployment.yaml
```diff
@@ -585,8 +587,18 @@ spec:
...
+          value: {{ .Values.nats.clusterID }}
+        - name: MESSAGE_QUEUE_QUEUE_GROUP
+          value: {{ .Values.nats.queueGroup }}
+        - name: MESSAGE_QUEUE_CLIENT_ID
+          value: {{ .Values.nats.clientID }}
         - name: MESSAGE_QUEUE_URL
+        {{- if not .Values.nats.url }}
           value: nats://{{ .Values.nats.authToken }}@nats-streaming:4222
+        {{- else }}
+          value: {{ .Values.nats.url }}
```
life1347:
```
value: nats://{{ .Values.nats.authToken }}@{{ .Values.nats.url }}
```

life1347: From the example you attached, 
```
url: nats://nats-cluster.nats-io.svc.cluster.local:4222
```
How about we can add a an extra case for 
```
value: nats://{{ .Values.nats.authToken }}@{{ .Values.nats.url }}
```
if `authToken` is not empty. What do you think?

shaunc: One can include in the URL, but you may be right that a user might not think of that.  Hmm... the issue is that currently if you have an external id and don't have auth token (using user/password, for instance), or you have a different auth token, and include your token in the url. `nats.authToken` has a non-empty default. Also seems that "url" should mean entire url.

What about this: have two options: `nats.endpoint` and `nats.url` -- both blank by default. If `nats.enpoint` is set then the url is `nats://{{ .Values.nats.authToken }}@{{ .Values.nats.endpoint }}`; otherwise, if `nats.url` is set, the value is given by it, and if neither are set then as current?

life1347: The endpoint is a bit useless because if the URL is not empty means user want to use external service. 

So it will be like
```
        {{- if not .Values.nats.url }}
          value: nats://{{ .Values.nats.authToken }}@nats-streaming:4222
        {{- else if .Values.nats.authToken }}
          value: nats://{{ .Values.nats.authToken }}@{{ .Values.nats.url }}
        {{- else }}
          value: {{ .Values.nats.url }}
        {{- end }}
```

shaunc: Hmm... so I was thinking:
```
{{- if and (not .Values.nats.url) (not .Values.nats.endpoint) }}
  value: nats://{{ .Values.nats.authToken }}@nats-streaming:4222
{{- else if not .Values.nats.url }}
  value: nats://{{ .Values.nats.authToken }}@{{ .Values.nats.endpoint }}
{{- else }}
   value: {{ .Values.nats.url }}
{{- end }}
```
The rationale being that `url` should always actually be a url; external service could be specified either with endpoint (for default service token) or url (not only auth token, but perhaps proxy ... so not necessarily even `nats://`).

But I'll patch according to what you say if I'm not convincing. Hmm... are you saying you want to get rid of `nats.external` and just use url (per your suggestion) or url or endpoint (per my suggestion) not only for nats url config, but also elsewhere for nats-streaming deployment and service, which are gated by `nats.external` now?

life1347: oh, yes, I mean to get rid of `external`, sorry for that.

About `url` and `endpoint`, I personally think they are similar things and bring confusion to the user. Everyone has their own definition, and in the end they have to check how these two parameters work in the chart. 

How about we change it to `hostaddress` which is normally considered as the format of `<ip-or-dns>:<port>` to avoid the confusion?

The config will become

```
nats:
    hostaddress: nats-cluster.nats-io.svc.cluster.local:4222
    authToken: xxxxxxx
```

We can append the `nats://` in charts instead of putting it at values.yaml.

shaunc: This is good. Will send update this eve.

Hmm... oh -- but then we need to keep `nats.external` (for nats-streaming deployment/service gate), right?

life1347: yes, if we change to `hostaddress` we will need external. Thanks for reminding me.

life1347: btw, please add some comments to values.yaml as well so that people can quickly understand what `external` and other newly add parameter means.

shaunc: Very good

samoscybercraig:
Is this still awaiting testing @life1347 - I'm really looking forward for this feature to hit master! Thanks @shaunc for your work on it.

life1347:
@samoscybercraig this PR is now waiting for follow up commits
@shaunc any progress?

shaunc:
@life1347 @samoscybercraig ... sorry for the delay -- something else urgent came up. I got distracted, but hopefully I've remembered what to do! :) I'll tweak if I've forgotten anything.

shaunc:
BTW -- was thinking that would be good to optionally allow `nats.hostaddress` to be a comma separated list for HA setup... but didn't actually put that in.

life1347:
Wait for CI https://travis-ci.org/github/fission/fission/builds/674073638 passed, we are good to go. About the HA part we can add in other PR later.

vishal-biyani:
Hey, @shaunc I replicated your commit in the PR https://github.com/fission/fission/pull/1576 and build is passing. Thanks a lot for this PR. I will close this and merge the other PR.

PR #1564, opened 2020-03-30
GitPod for Fission

vishal-biyani:
This PR adds [Gitpod](https://www.gitpod.io/) for Fission with a base image that has Go, Docker client, etc. installed. This itself is not sufficient for developing Fission fully on Gitpod though. Some of the following are still needed for making Gitpod fully usable for Fission Development.

- [ ] Ability to build docker images from source from within Gitpod (Dependent on https://github.com/gitpod-io/gitpod/issues/1337) along with Kind installed for cluster

_**OR**_

- [ ] A way to configure a remote Kubernetes cluster and potentially using Kaniko/buildah etc. in the remote cluster

**Additionally** following is needed

- [ ] Kubectl, Helm, VSCode extensions installed in the base image

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1564)

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -69,7 +69,7 @@ RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
 LABEL dazzle/layer=lang-go
 LABEL dazzle/test=tests/lang-go.yaml
 USER gitpod
-ENV GO_VERSION=1.14 \
+ENV GO_VERSION=1.12 \
     GOPATH=$HOME/go-packages \
     GOROOT=$HOME/go
 ENV PATH=$GOROOT/bin:$GOPATH/bin:$PATH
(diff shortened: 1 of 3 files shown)


.gitpod.Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+LABEL dazzle/test=tests/lang-go.yaml
+USER gitpod
+ENV GO_VERSION=1.12 \
+    GOPATH=$HOME/go-packages \
+    GOROOT=$HOME/go
+ENV PATH=$GOROOT/bin:$GOPATH/bin:$PATH
+RUN curl -fsSL https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz | tar xzs
+# user Go packages
+ENV GOPATH=/workspace/go \
+    PATH=/workspace/go/bin:$PATH
```
jankeromnes, line 79: Hello! Super cool that you're configuring Gitpod for Fission. 💯

As a quick note, it seems that you're copying a bunch of stuff from the default `gitpod/workspace-full` image. I think you can simplify this `.gitpod.Dockerfile` by replacing everything above this line with:

```
FROM gitpod/workspace-full
```

vishal-biyani: Hi @jankeromnes I am not completely there yet. Fission's dev workflow requires a Kubernetes cluster.

About the workspace-full image, I didn't want to add Ruby, Java, C++ & PHP to image as I did not need them. But looking back maybe that will make the definition far simpler. 

Also, the workspace full image uses Go 1.14 whereas we need 1.12 for now. But thanks a lot for the suggestion.

vishal-biyani:
Closing till we can revisit this.

PR #1568, opened 2020-04-11
[chart] Add PSP for logger

LittleFox94:
This pull requests adds a flag to generate a PodSecurityPolicy for the logger.

I have PodSecurityPolicies enabled in my cluster and run the logger pods in privileged mode. They were not able to start before this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1568)


charts/fission-all/templates/fluentbit.yaml
```diff
@@ -19,6 +19,64 @@ data:
...
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    svc: logger
+roleRef:
+  kind:     Role
+  name:     psp:{{ .Release.Name }}-fission-logger-privileged
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind:      ServiceAccount
+  name:      default
```
life1347, line 77: The service account logger used is `fission-svc`, do we need to change it from `default` to `fission-svc`?

LittleFox94:
The permission to create privileged pods is required by the daemonset-controller, which uses the default sa

(cannot reply to the comment with my app)

life1347:
ok, thanks for the clarification. I will find some time to test the changes.

life1347:
@vishal-biyani I'm quite busy recently, is there any chance for you to test this PR? Thanks

vishal-biyani:
@life1347 Yes I will review and test it.

vishal-biyani:
Hey, @LittleFox94 Tested and found good for merging. The only question I have is `privileged` absolutely necessary given the other four rules are `RunAsAny`? ( I am not yet super-familiar with PSP)

LittleFox94:
Running those pods in privileged mode is another flag in the helm chart. I seem to need that on my cluster (though it was late at night, will check again), but there are definitely people who will need it (I think Openshift is strange here)

This psp is actually only a fix for the existing flag on clusters with psp enabled

vishal-biyani:
Ah, nice. I did not notice that earlier - but thanks a lot for clarifying! Keep safe in these times and I will merge this now.

LittleFox94:
Thanks, and keep safe yourself :)

Issue #576, opened 2018-03-23
Expose S3 config settings in storage service and helm chart


soamvasani:
Storage service uses a library that supports S3, but we currently only use its local volume feature.

This bug tracks:
 * [ ] exposing a flag to configure the storage system type
 * [ ] configuring S3 API endpoint
 * [ ] configuring S3 credentials
 * [ ] adding this to the helm chart

pulberg:
Still waiting for this...

vishal-biyani:
Hi, @pulberg Thanks for letting us know. We will prioritize this for an upcoming release.

rajalokan:
I'm starting implementation for this. Before I go ahead with actual implementation, wanted to get a quick review from maintainers. 

Essentially this needs three bits. 
1. Handling S3 Persistence config in helm. 
2. Modify `fission-bundle` binary to handle `runStorageSvc` with S3 storage type. 
3. Adding more generic `StorageType` interface in `stowClient.go` 

### Handling S3 Persistence config in helm: 
- Introduce a new variable `storageType`. For our case it will be `storageType: s3`. 
- Based on the value of it, additional config option will be required. One sample example is 
```
persistence:
  enabled: true
  storageType: s3
  
  s3:
    endpoint: X.amazonaws.com
    configAccessKeyId: XXXX
    configSecretAccessKey: XXX
    configRegion: XXX
```

### Modify `fission-bundle` binary to handle `runStorageSvc` with S3 storage type.
- Can be done by adding one more else if condition in `fission-bundle/main.go`. This will be hooked to a new function `runS3StorageSvc()`

### Adding more generic `StorageType` interface in `stowClient.go` 
- Presently `StorageType` is a string. Proposing to conver this to interface. 
- Introduce two new structs `LocalStorageType` and `S3StorageType`.  With this dialing logic will be handled by corresponding items. 

Please review and see if this can be a conflict or potential issues?

vishal-biyani:
Hey, @rajalokan Thanks - overall this looks really good. One thing I would like to do before we start implementation is to design the storage type struct etc. - as it is core to the configuration being passed and will affect the overall design. Can we take a stab at it here and once closed - we are good to start implementation.

pulberg:
Another thing to consider, S3 does not always mean "AWS", there are many storage arrays that provide S3 functionality.  In my case, we are using EMC ECS, and will soon be using Pure Storage arrays that have S3 functionality.

[...]

PR #1577, opened 2020-04-21
Expose S3 as storage backend

rajalokan:
- Expose S3 as storage backend.

**Update**
Converting this PR from `Open` to `Draft` as it needs below fixes before review/merge. 

- [x] Currently running storagesvc against an existing Bucket fails. Need to fix this. 
- [x] Add more test cases to ensure coverage. 

Leaving it here for comments/review regarding storage backend refactor. Current status of PR is as below : 

- [x] Addressed all comments during review.
- [x] Manually verified it using S3 config. 
- [x] Manually verified that sub directory in S3 is working as per  `persistence.s3.subDir` value in `values.yaml` 
- [x] All tests working. 
- [x] Added additional integration test for s3 backend storage. 

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1577)

### DIFF
--- a/charts/fission-core/templates/deployment.yaml
+++ b/charts/fission-core/templates/deployment.yaml
@@ -288,7 +288,7 @@ spec:
         - name: TRACE_JAEGER_COLLECTOR_ENDPOINT
           value: "{{ .Values.traceCollectorEndpoint }}"
         - name: TRACING_SAMPLING_RATE
-          value: {{ .Values.traceSamplingRate | default "0.5" | quote }}          
+          value: {{ .Values.traceSamplingRate | default "0.5" | quote }}      
         - name: ENABLE_ISTIO
           value: "{{ .Values.enableIstio }}"
         - name: FETCHER_MINCPU
@@ -398,27 +398,47 @@ spec:
         image: {{ include "fission-bundleImage" . | quote }}
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
-        args: ["--storageServicePort", "8000", "--filePath", "/fission"]
+        {{- if and (.Values.persistence.enabled) (eq (.Values.persistence.storageType | default "local") "s3") }}
+        args: ["--storageServicePort", "8000", "--storageType", "s3"]
+        {{- else }}
+        args: ["--storageServicePort", "8000", "--storageType", "local"]
+        {{- end }}
         env:
         - name: PRUNE_INTERVAL
           value: "{{.Values.pruneInterval}}"
         - name: TRACE_JAEGER_COLLECTOR_ENDPOINT
           value: "{{ .Values.traceCollectorEndpoint }}"
         - name: TRACING_SAMPLING_RATE
-          value: {{ .Values.traceSamplingRate | default "0.5" | quote }}          
+          value: {{ .Values.traceSamplingRate | default "0.5" | quote }}
+        {{- if and (.Values.persistence.enabled) (eq (.Values.persistence.storageType | default "local") "s3") }}
+        - name: STORAGE_S3_BUCKET_NAME
+          value: {{ .Values.persistence.s3.bucketName }}
+        - name: STORAGE_S3_SUB_DIR
+          value: {{ .Values.persistence.s3.subDir }}
+        - name: STORAGE_S3_ACCESS_KEY_ID
+          value: {{ .Values.persistence.s3.accessKeyId }}
+        - name: STORAGE_S3_SECRET_ACCESS_KEY
+          value: {{ .Values.persistence.s3.secretAccessKey }}
+        - name: STORAGE_S3_REGION
+          value: {{ .Values.persistence.s3.region }}
+        {{- end }}
+        {{- if ne (.Values.persistence.storageType | default "local") "s3" }}
         volumeMounts:
         - name: fission-storage
           mountPath: /fission
+        {{- end }}
         ports:
           - containerPort: 8000
             name: http
       serviceAccountName: fission-svc
+      {{- if and (.Values.persistence.enabled) (ne (.Values.persistence.storageType | default "local") "s3") }}
       volumes:
       - name: fission-storage
-      {{- if .Values.persistence.enabled }}   
         persistentVolumeClaim:
           claimName: {{ .Values.persistence.existingClaim | default "fission-storage-pvc" }}
       {{- else }}
+      volumes:
+      - name: fission-storage
         emptyDir: {}
       {{- end }}
 {{- if .Values.extraCoreComponentPodConfig }}
--- a/charts/fission-core/values.yaml
+++ b/charts/fission-core/values.yaml
@@ -110,10 +110,22 @@ router:
 
 ## Persist data to a persistent volume.
 persistence:
-  ## If true, fission will create/use a Persistent Volume Claim
+  ## If true, fission will create/use a Persistent Volume Claim unless storageType is set to s3
   ## If false, use emptyDir
   ##
   enabled: true
+
+  ## Must be set to either local or S3.
+  ## If storateType is set(other than local), one of its backend configuration must be set as below.
+  #storageType: local | s3
+
+  ## Sample configruation for AWS s3 storage backend
+  # s3:
+  #   bucketName: <awsBucketName>
+  #   subDir: <sub directory within a bucket>
+  #   accessKeyId: <awsAccessKeyId>
+  #   secretAccessKey: <awsSecretAccessKey>
+  #   region: <awsRegion>
 
   ## A manually managed Persistent Volume Claim name
   ## Requires persistence.enabled: true
--- a/cmd/fission-bundle/main.go
+++ b/cmd/fission-bundle/main.go
@@ -78,14 +78,11 @@ func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
 	}
 }
 
-func runStorageSvc(logger *zap.Logger, port int, filePath string) {
-	subdir := os.Getenv("SUBDIR")
-	if len(subdir) == 0 {
-		subdir = "fission-functions"
+func runStorageSvc(logger *zap.Logger, port int, storage storagesvc.Storage) {
+	err := storagesvc.Start(logger, storage, port)
+	if err != nil {
+		logger.Fatal("error starting storage service", zap.Error(err))
 	}
-	enableArchivePruner := true
-	storagesvc.RunStorageService(logger, storagesvc.StorageTypeLocal,
-		filePath, subdir, port, enableArchivePruner)
 }
 
 func runBuilderMgr(logger *zap.Logger, storageSvcUrl string, envBuilderNamespace string) {
@@ -200,7 +197,7 @@ Usage:
   fission-bundle --routerPort=<port> [--executorUrl=<url>]
   fission-bundle --executorPort=<port> [--namespace=<namespace>] [--fission-namespace=<namespace>]
   fission-bundle --kubewatcher [--routerUrl=<url>]
-  fission-bundle --storageServicePort=<port> --filePath=<filePath>
+  fission-bundle --storageServicePort=<port> --storageType=<storateType>
   fission-bundle --builderMgr [--storageSvcUrl=<url>] [--envbuilder-namespace=<namespace>]
   fission-bundle --timer [--routerUrl=<url>]
   fission-bundle --mqt   [--routerUrl=<url>]
@@ -295,8 +292,15 @@ Options:
 
 	if arguments["--storageServicePort"] != nil {
 		port := getPort(logger, arguments["--storageServicePort"])
-		filePath := arguments["--filePath"].(string)
-		runStorageSvc(logger, port, filePath)
+
+		var storage storagesvc.Storage
+
+		if arguments["--storageType"] != nil && arguments["--storageType"] == string(storagesvc.StorageTypeS3) {
+			storage = storagesvc.NewS3Storage()
+		} else if arguments["--storageType"] == string(storagesvc.StorageTypeLocal) {
+			storage = storagesvc.NewLocalStorage("/fission")
+		}
+		runStorageSvc(logger, port, storage)
 	}
 
 	select {}
--- a/pkg/storagesvc/storagesvc.go
+++ b/pkg/storagesvc/storagesvc.go
@@ -25,13 +25,23 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
-	_ "github.com/graymeta/stow/local"
+	"github.com/graymeta/stow"
 	"github.com/pkg/errors"
 	"go.opencensus.io/plugin/ochttp"
 	"go.uber.org/zap"
 )
 
 type (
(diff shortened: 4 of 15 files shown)


vishal-biyani:
Fixes https://github.com/fission/fission/issues/576

cmd/fission-bundle/main.go
```diff
@@ -78,14 +78,26 @@ func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
 	}
 }
 
-func runStorageSvc(logger *zap.Logger, port int, filePath string) {
```
vishal-biyani, line 81: I would still keep the name as `runStorageSvc` - weather it does use a volume or S3 - is an implementation-level detail that need not leak out/known to the caller.

rajalokan, line 81: - This is a typo. Function name will change back to `runStorageSvc` but with new signature. 
- This is a lot of logic here in `main.go`. I feel we can move this to a `Start()` method in storagesvc package. Just we are doing for others like `runTimer`, `runKubeWatcher` etc.

cmd/fission-bundle/main.go
```diff
@@ -78,14 +78,26 @@ func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
...
+	}
+	switch storageType {
+	case "s3":
+		storagesvc.RunStorageService(logger, storagesvc.ConstStorageTypeLocal,
+			filePath, subdir, port, enableArchivePruner)
+	default:
+		s3StorageType := storagesvc.NewS3Storage(storageArgs[1]...)
+		storagesvc.RunStorageService(logger, s3StorageType,
+			filePath, subdir, port, enableArchivePruner)
+	}
```
vishal-biyani, line 100: This logic should reside in `storagesvc`. The fission-bundle should simply pass a flag and rest of implementation detail should be abstracted.

pkg/storagesvc/stowClient.go
```diff
@@ -48,8 +46,9 @@ type (
 )
 
 const (
-	StorageTypeLocal StorageType = "local"
-	PaginationSize   int         = 10
+	constStorageTypeLocal string = "local"
+	constStorageTypeS3    string = "s3"
+	PaginationSize        int    = 10
```
vishal-biyani, line 51: - Constants need not pre-appended with `const` AFAIK as per idiomatic Go
- Are we making the storageTypeLocal now a package only variable?

charts/fission-core/values.yaml
```diff
@@ -110,11 +110,22 @@ router:
 
 ## Persist data to a persistent volume.
 persistence:
-  ## If true, fission will create/use a Persistent Volume Claim
+  ## If true (and storateType is not set), fission will create/use a Persistent Volume Claim
   ## If false, use emptyDir
   ##
   enabled: true
```
vishal-biyani, line 116: I personally think being explicit about storage type as shown below is clearer to read and understand

```
persistence:
  enabled: true
  storageType: local | s3 | some_future_value (Default to local for backward compatibility)

  s3:
    accessKeyId: AHAHAAKHSKD
```

charts/fission-core/templates/deployment.yaml
```diff
@@ -398,27 +398,49 @@ spec:
         image: {{ include "fission-bundleImage" . | quote }}
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
+        {{- if and .Values.persistence.storageType (.Values.persistence.enabled) }}
+            {{- if eq .Values.persistence.storageType "s3" }}
+        args: ["--storageServicePort", "8000", "--filePath", "/fission", "--storageType", "s3"]
+            {{- end }}
+        {{- else }}
```
vishal-biyani, line 405: This is related to another comment in values.yaml - once we are explicit about `storageType` - then arguments being passed to storage service will be fixed.

charts/fission-core/values.yaml
```diff
@@ -110,11 +110,22 @@ router:
...
 
+  ## Must be set only if using S3 backend. If using pvc just leave at `enabled: true`
+  ## If storateType is set, one of its backend configuration must be set as below.
+  # storageType: s3
+
+  ## Sample configruation for AWS s3 storage backend
+  # s3:
+  #   bucketName: <awsBucketName>
```
vishal-biyani, line 124: I am myself not convinced - hence I commented on the issue, but a folder within bucket should be an option?

cmd/fission-bundle/main.go
```diff
@@ -78,14 +78,11 @@ func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
 	}
 }
 
-func runStorageSvc(logger *zap.Logger, port int, filePath string) {
-	subdir := os.Getenv("SUBDIR")
-	if len(subdir) == 0 {
-		subdir = "fission-functions"
```
vishal-biyani, line 84: Removing this might break backward compatibility.

cmd/fission-bundle/main.go
```diff
@@ -78,14 +78,11 @@ func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
...
+func runStorageSvc(logger *zap.Logger, port int, filePath string, storageArgs ...string) {
+	err := storagesvc.Start(logger, filePath, port, storageArgs...)
+	if err != nil {
+		logger.Fatal("error starting storage service", zap.Error(err))
 	}
-	enableArchivePruner := true
```
vishal-biyani, line 86: Any specific reason to remove this from here - in future we could expose this via values so the user can define if pruning should be enabled etc.

cmd/fission-bundle/main.go
rajalokan: The idea is to remove all logic from `fission-bundle/main.go` to `storavesvc.go`  and make it consistent with other services in same file. Like 
```
func runMessageQueueMgr(logger *zap.Logger, routerUrl string) {
	err := mqtrigger.Start(logger, routerUrl)
	if err != nil {
		logger.Fatal("error starting message queue manager", zap.Error(err))
	}
}
```
Similarly storageSvc also has a `Start` function now which abstracts this logic. Since both statements are same ( earlier it was here and now moved to `storageSvc` so ideally this shouldn't break). Let me know if my understanding is correct?

cmd/fission-bundle/main.go
```diff
@@ -296,7 +294,16 @@ Options:
...
 		port := getPort(logger, arguments["--storageServicePort"])
 		filePath := arguments["--filePath"].(string)
-		runStorageSvc(logger, port, filePath)
+
+		if arguments["--storageType"] != nil && arguments["--storageType"] == storagesvc.StorageTypeS3 {
+			bucketName := os.Getenv("STORAGE_S3_BUCKET_NAME")
+			accessKeyId := os.Getenv("STORAGE_S3_ACCESS_KEY_ID")
+			secretAccessKey := os.Getenv("STORAGE_S3_SECRET_ACCESS_KEY")
+			region := os.Getenv("STORAGE_S3_REGION")
+			runStorageSvc(logger, port, filePath, "s3", bucketName, accessKeyId, secretAccessKey, region)
```
vishal-biyani, line 303: Can this go inside the storageSvc - as this is impl detail of Storage Service.

cmd/fission-bundle/main.go
rajalokan: Same as above. If we want to get this values from `values.yaml` we will pass this on to `Start` function and handle it there. Like it is being handled now [here](https://github.com/fission/fission/pull/1577/files?file-filters%5B%5D=.go#diff-f33eed829a16e5e68e0b2c560b158512R53)

pkg/storagesvc/storagesvc.go
```diff
@@ -43,6 +49,35 @@ type (
 	}
 )
 
+// Start function
+func Start(logger *zap.Logger, filePath string, port int, storageArgs ...string) error {
+	subdir := os.Getenv("SUBDIR")
+	if len(subdir) == 0 {
+		subdir = "fission-functions"
+	}
+	enableArchivePruner := true
```
vishal-biyani, line 58: These should be bubbled up in fission-bundle/main.go before calling storage Service. They can be made configurable in future.

cmd/fission-bundle/main.go
rajalokan: Problem with that is my little knowledge of golang. As far as I know we can't have same functions with different signatures. Therefore one single function `runStorageSvc` won't be able to handle for both `pvc/local` and `s3` cases. 

Which is why I feel we need to bisect it here and call with appropriate function signature. I will give this a thought again.

charts/fission-core/values.yaml
rajalokan: Yes. I agree. I can keep this in parallel. ONce other issues are reviewed and sorted, I will add this logic to current working setup ( using buckets).

pkg/storagesvc/stowClient.go
```diff
@@ -30,10 +31,8 @@ import (
 )
 
 type (
-	StorageType string
-
 	storageConfig struct {
-		storageType   StorageType
```
vishal-biyani, line 36: We should probably extend the following constant list and use same type?

```
const (
	StorageTypeLocal StorageType = "local"
	PaginationSize   int         = 10
)
```

vishal-biyani:
Hey @rajalokan A good start, I have left a few comments about general structure. Once those are done and we are able to test it, this looks good.

charts/fission-core/values.yaml
rajalokan: This is done now. Please verify.

charts/fission-core/templates/deployment.yaml
rajalokan: Yes. I've changed and tested for all three cases. 
- Without persistence enabled (tmpDir)
- With persistence( both implicit and explicit `storageType: local`). 
- With `storageType: s3`. 

Please verify.

charts/fission-core/values.yaml
rajalokan: I had a loot at this, but couldn't find if `stow` can create subDir within a bucket. Couldn't find any example in their repo. Also [here](https://github.com/graymeta/stow/blob/master/s3/location.go#L27) it seem they only support creating buckets right now.

cmd/fission-bundle/main.go
rajalokan: I've put this inside the `NewS3Storage` function. There we will read these env variables and create a storage stuct out of it. This is also better for case when we want to move to file based config ( from env based config). That way we will read from config file in same place. Views ?

pkg/storagesvc/storagesvc.go
rajalokan: So we have two options here. 
- Pass all relevant config params as part of `fission-bundle` cli.
- We just call fission `fission-bundle` with min arguments and rest all config comes from either ENV variables of a config file. Both of these can be done using deployment.yaml in helm chart. 

I'm doing it using second approach. In this way this logic need not be in fission-bundle/main.go . Let me know if this is reasonable ?

cmd/fission-bundle/main.go
```diff
@@ -200,7 +197,8 @@ Usage:
   fission-bundle --routerPort=<port> [--executorUrl=<url>]
   fission-bundle --executorPort=<port> [--namespace=<namespace>] [--fission-namespace=<namespace>]
   fission-bundle --kubewatcher [--routerUrl=<url>]
-  fission-bundle --storageServicePort=<port> --filePath=<filePath>
+  fission-bundle --storageServicePort=<port>
```
vishal-biyani, line 200: Would we not need always two arguments - the port and type. And inside the Storage Svc class, the type will be used to activate differently. Also, it keeps arguments for StorageService consistent irrespective of the type.

vishal-biyani:
Hi @rajalokan I created a PR for integration test with your changes here: https://github.com/fission/fission/pull/1623

I have also requested some changes in PR plus the values.yaml and template changes are not in fission-all charts yet. Would you have the bandwidth for these changes or do you want me to take it beyond this point, LMK please.

rajalokan:
No. I would love to work on this. I'm going to spend some time on this now.

charts/fission-core/templates/deployment.yaml
rajalokan: I think in current `values.yaml`  format, your comment is resolve. Can you please confirm.

vishal-biyani:
Great, thanks. A really good integration test would be to set up a single pod Minion in the cluster - point to it and verify the code indeed goes there. We may not run it every merge - but good to have checked in and for quick verification from review POV.

cmd/fission-bundle/main.go
rajalokan: This is implemented now. Please verify.

rajalokan:
I've added chart for `fission-all` and verified with `helm upgrade`. To summerize 

- [x] Addressed all comments during review.
- [x] Manually verified it using S3 config. 
- [x] Manually verified that sub directory in S3 is working as per  `persistence.s3.subDir` value in `values.yaml` 
- [x] All tests working. 
- [ ] Add any additional integration tests. Shall we do this as part of separate PR ?

rajalokan:
Sure. I will work on integration test with minio tomorrow.

cmd/fission-bundle/main.go
rajalokan: This is done. We have a generic `Start` function in `storagesvc` which takes `storage` interface. Based on this interface ( whether `local` or `s3`), it evaluates appropriate backend. 

Please mark this resolved if this is what needs to be done here ?

pkg/storagesvc/stowClient.go
rajalokan: 1. Done 
2. Yes. We have an interface level method which can be used from outside of package. It is `getStorageType() string`

Is this alright ?

rajalokan:
@vishal-biyani This is PR is ready for review. 
- All your requested changes are done. 
- I've verified manually that all the flows work. 
- Added integration tests for s3 backend. 

Please review.

vishal-biyani:
Hey @rajalokan Thanks - I finally got to test this one out. Except for one missing env variable - which I added in my Helm - it worked like a charm. Thanks for this work. 

One thing I realized is if the setup was using volume before and someone switched to S3 - there is no migration done automatically today, so we need to think of that as a separate job that we can execute (Not part of this PR - as a separate task). Also same when someone was using a S3 and then switches to a volume-based storage service.

One nitpick/small issue is in case of Minio - you are forced to specify a region (The blank value lead to error)

I will merge this soon from #1629 - I hope you ok with that?

rajalokan:
Yea. We need to handle migration. I can work on that in another PR. 

Yes you can merge if this is fine. Thanks for the patience with this.

vishal-biyani:
@rajalokan On second thought, I think simple documentation for people upgrading would be good enough. Commands to copy files from existing PVC and then insert into Minio/S3 is probably fine.

rajalokan:
Yep. We can write something immediately before next release. I can update the doc.

vishal-biyani:
Merged in https://github.com/fission/fission/pull/1629 

Thanks a ton @rajalokan for getting this done!

Issue #1585, opened 2020-04-27
Package waiting for builder pod to be ready should be better


vishal-biyani:
Right now the package watcher/builder waits for only 15 iterations for the builder pod to be ready. Some environments need a slightly higher time. This should be more of a retry backoff mechanism before failing

https://github.com/fission/fission/blob/f606c3705f81832008aba130de7a0f5845387583/pkg/buildermgr/pkgwatcher.go#L103-L105

rahulchheda:
@vishal-biyani , i think you are talking about exponential backoff mechanism? Or are you trying to implement some other backoff mechanism?

vishal-biyani:
@rahulchheda Yes that's what I was thinking as an initial simple implementation.

rahulchheda:
@vishal-biyani PTAL: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/goroutinemap/exponentialbackoff/exponential_backoff.go

vishal-biyani:
I am ok with either implementation @rahulchheda , do you want to send a PR?

rahulchheda:
Yes, @vishal-biyani, I would like to take up this issue.

vishal-biyani:
Closed by #1614

PR #1614, opened 2020-05-18
[WIP] (feat) Added Exponential BackOff for Retry in builder

rahulchheda:
Signed-off-by: Rahul M Chheda <rahul.chheda@mayadata.io>

Fix for issue: https://github.com/fission/fission/issues/1585

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1614)


pkg/buildermgr/pkgwatcher.go
```diff
@@ -100,8 +100,13 @@ func (pkgw *packageWatcher) build(buildCache *cache.Cache, srcpkg *fv1.Package)
 		return
 	}
 
+	// Create a new BackOff for health check on environment builder pod
+	healthCheckBackOff, err := utils.NewBackOff(500*time.Millisecond,time.Minute/10,5)
+	if err != nil {
+		pkgw.logger.Error("Unable to create BackOff for Health Check", zap.Error(err))
```
vishal-biyani, line 106:
```suggestion
		pkgw.logger.Error("Unable to create BackOff for Health Check", zap.Error(err), zap.String("package_name", pkg.ObjectMeta.Name))
```

pkg/buildermgr/pkgwatcher.go
```diff
@@ -100,8 +100,13 @@ func (pkgw *packageWatcher) build(buildCache *cache.Cache, srcpkg *fv1.Package)
 		return
 	}
 
+	// Create a new BackOff for health check on environment builder pod
+	healthCheckBackOff, err := utils.NewBackOff(500*time.Millisecond,time.Minute/10,5)
```
vishal-biyani, line 104: So at least for builder pod especially for platforms like JVM the `maxInterval` of 6 seconds seems small. Can we go for something like 120 seconds?

vishal-biyani:
Thanks for the PR @rahulchheda - except two minor comments, looks good. Also one nitpick - backoff seems like a single word so `backoff.go` might be a better name instead of camel case one. I will test this sometime today/tomorrow once you have made changes and then I will merge it

rahulchheda:
Thanks for the review @vishal-biyani , let me just do some more modification

rahulchheda:
Btw, @vishal-biyani , Do you want me to include some logic around maxRetries, similar to maxInterval?

vishal-biyani:
@rahulchheda maxretries would be indeed good. I don't think trying till infinity is useful.

rahulchheda:
Right now it checks till the currentBackOffDuration <= maxInterval, I can include one more variable to take in the MaxRetry, let me update the PR. I think this would be useable further down the line.

rahulchheda:
Sorry for unwanted changes, will fix in final commit, PTAL @vishal-biyani

examples/go/module-example/go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/fission/fission/examples/go/go-module-example
 
 require github.com/golang/example v0.0.0-20170904185048-46695d81d1fa
+
+go 1.13
```
vishal-biyani, line 5: Should we move this change out of this PR?

rahulchheda: Sure, got this added into git staged my mistake, will fix in next commit.

pkg/utils/backoff.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+const (
+	DefaultInitialInterval = 500 * time.Millisecond
+	DefaultMultiplier      = 1.5
+	DefaultMaxInterval     = 60 * time.Second
```
vishal-biyani, line 11: Just tested and even with 60 seconds it does not work with JVM env. I would suggest bumping up default max to 300s to cover potential other scenarios.

rahulchheda:
@vishal-biyani Made the appropriate changes, could you take a look?

pkg/utils/backoff.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"fmt"
+	"time"
+)
+
+const (
+	DefaultInitialInterval = 500 * time.Millisecond
+	DefaultMultiplier      = 1.5
+	DefaultMaxInterval     = 300 * time.Second
+	DefaultMaxCount        = 30
```
vishal-biyani, line 12: I just tested this and 30 count is happening before the max interval. Can we bump this to be inline with new max interval please?

vishal-biyani:
@rahulchheda Thanks and sorry for delay. Can you please fix the CI build issue and one minor comment - otherwise looks good!

vishal-biyani:
Thanks @rahulchheda this looks good - I will do a quick test and merge this later this week.

vishal-biyani:
Thanks a lot, @rahulchheda - I verified and this works nicely!

Issue #1594, opened 2020-05-07
Request Body Size: HTTP 413


vishal-biyani:
On Slack conversation, Ashish Jain mentioned that when using a body greater than 1MB for a function call was throwing `413 Request Entity Too Large` error. This issue is for tracking and fixing the problem.

LittleFox94:
Educated guess without looking at anything: Ingress involved and not set correctly?

vishal-biyani:
Hey, @LittleFox94 this is a problem specific to the NodeJS environment. Express/Body-parser package by default limits body size to 100KB (https://www.npmjs.com/package/body-parser#limit). 

We bumped it up to 1MB but some users need bigger limits:

https://github.com/fission/fission/blob/aa8c27fd5d9a1b92062d87e73a0e96f1c9b7ba7a/environments/nodejs/server.js#L102-L105

vishal-biyani:
Closed by #1618

PR #1618, opened 2020-05-20
Added support for setting bodyParser limit param via environment variable

therahulbhati:
#1594 User can set BODY_PARSER_LIMIT env variable via spec file

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1618)


vishal-biyani:
@therahulbhati overall the change looks good, it would be good to add a sample in examples directory. And I am assuming you have tested this as well?

therahulbhati:
@vishal-biyani I have added an example in example/spec-example. I tested this by creating an image of nodejs container after making changes in server.js

vishal-biyani:
Thanks for this change @therahulbhati merged 🎉 !

PR #1642, opened 2020-06-12
Allow user to override value of influxdb image

funkypenguin:
Hey guys!

In our environment, we enforce a policy whereby all images are stored on a local registry. As such, we needed to be able to override the default image used for influxdb. This PR updates the chart to allow the user to specify an influxdb image, as well as to disable influxdb altogether (perhaps they don't care about logs, or they already have an existing influxdb installation)

Cheers!
D

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1642)


vishal-biyani:
Thanks for the change @funkypenguin!

PR #1644, opened 2020-06-12
Permit disabling pre-upgrade checks (because.. Istio)

funkypenguin:
Hey guys,

We're using Istio in our cluster with mTLS, and as a result, every pod has an istio-proxy sidecar, managing the mTLS. This presents a problem for Jobs, since the sidecar hangs around after the job completes, resulting in an incomplete job.

Hopefully [this'll be solved in 1.19](https://github.com/kubernetes/enhancements/issues/753), but until then, I've added this PR to allow users like me to disable pre-upgrade checks (at their own peril).

Cheers!
D

### DIFF
--- a/charts/fission-all/templates/pre-upgrade-job.yaml
+++ b/charts/fission-all/templates/pre-upgrade-job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.preUpgradeChecks }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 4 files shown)


harshthakur9030:
Hey @funkypenguin ! Will something like this solve the purpose? https://github.com/redboxllc/scuttle

funkypenguin:
Awesome, thanks @sanketsudake ! I'll close this PR then in favour of #2229 :)

PR #1646, opened 2020-06-12
Add serviceaccount for nats-streaming

funkypenguin:
Hey guys!

We use PSPs to secure our cluster, and often have to create app-specific rolebindings to allow various pods to have greater-than-default privileges. 

In my case, I don't allow pods to run as root by default (which nats-streaming wants to do), so in order to make a rolebinding for nats-streaming, I needed it to use a non-default service account (else I'd be permitting any pod using the default service account to run as root).

This PR simply introduces another service account, aligned with the naming convention already in use for fission-builder, fission-fetcher, etc, so that rolebindings can be clearly defined if necessary.

Cheers!
D

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1646)


vishal-biyani:
Thanks for this PR @funkypenguin - this LGTM, I will merge it soon! Thanks a ton for other PRs too, taking a look at them.

vishal-biyani:
Thanks again for this @funkypenguin

PR #1670, opened 2020-07-21
Update values.yaml with mqt keda configuration

therahulbhati:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1670)


charts/fission-all/values.yaml
```diff
@@ -303,6 +303,6 @@ canaryDeployment:
 
 ## Message Queue Trigger Kind, KEDA: enable and configuration
 mqt_keda:
-  enabled: true
+  enabled: false
   connector_images:
-    kafka: fission/keda-kafka
+    kafka: fission/keda-kafka:1
```
vishal-biyani, line 308: Let's split image and tag into separate values

charts/fission-core/values.yaml
```diff
@@ -197,3 +197,9 @@ canaryDeployment:
 # Use these flags to enable opentracing, the variable is endpoint of Jaeger collector in the format shown below
 #traceCollectorEndpoint: "http://jaeger-collector.jaeger.svc:14268/api/traces?format=jaeger.thrift"
 #traceSamplingRate: 0.75
+
+## Message Queue Trigger Kind, KEDA: enable and configuration
+mqt_keda:
+  enabled: false
+  connector_images:
+    kafka: fission/keda-kafka:1
```
vishal-biyani, line 205: Same comment as the fission-all/values.yaml

pkg/mqtrigger/kafka/main.go
```diff
@@ -26,6 +26,13 @@ import (
 	"github.com/fission/fission/pkg/mqtrigger/util"
 )
 
+// Following code snippet is from KEDA project and adapted for Fission.
```
vishal-biyani, line 29: Let's remove this from here along with rest of Kafka code and use fission/keda-connector for that

therahulbhati: yes, add that in fission/keda-connector

charts/fission-core/values.yaml
therahulbhati: done

charts/fission-all/values.yaml
therahulbhati: Done

charts/fission-all/values.yaml
```diff
@@ -308,6 +308,11 @@ canaryDeployment:
 
 ## Message Queue Trigger Kind, KEDA: enable and configuration
 mqt_keda:
-  enabled: true
+  enabled: false
   connector_images:
-    kafka: fission/keda-kafka
+    kafka:
+      image: fission/keda-kafka
```
vishal-biyani, line 314: Probably call it fission/keda-kafka-http-connector? (LMK if you think that's too big a name though) So there is no ambiguity?

therahulbhati: Done

Issue #1485, opened 2020-01-09
Java later versions support and images


vishal-biyani:
Currently, the Java version tested and officially supported in Fission is Java8. It might be good to test and publish images for the latest versions of Java - 11, 12, 13, etc.

PR #1674, opened 2020-07-22
new jvm environments

sahil-lakhwani:
Signed-off-by: sahil-lakhwani <sahilakhwani@gmail.com>

Addresses #1485 

This PR is an effort to support versions of JVM above 8.

- After reading about the changes in class loaders in Java 9, I think the current code for loading classes is sufficient with only a minor change to load user functions correctly.
- Use JRE images instead of JDK.
- The PR modifies release scripts, adding Java versions 9, 10 and 11. Version 11 is the last LTS release. 
We sure can support later versions, but those are not recommended for production. (Need thoughts on this)

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1674)


environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -59,10 +60,8 @@
```
vishal-biyani, line 60: I think one of the goals to remove the dependency on Spring in this class too - and use stock Java Servlet or similar object.

```
ResponseEntity<String> specialize(@RequestBody FunctionLoadRequest req) {
```

environments/jvm/src/main/java/io/fission/Server.java
```diff
@@ -59,10 +60,8 @@
 
 			jarFile = new JarFile(file);
 			Enumeration<JarEntry> e = jarFile.entries();
-			URL[] urls = { new URL("jar:file:" + file + "!/") };
+			URL[] urls = { file.toURI().toURL() };
 
-			// TODO Check if the classloading can be improved for ex. use something like:
-			// Thread.currentThread().setContextClassLoader(cl);
 			if (this.getClass().getClassLoader() == null) {
 				cl = URLClassLoader.newInstance(urls);
```
vishal-biyani, line 66: I think we should look at this classloader example and see if we can do better here: https://stackoverflow.com/questions/60764/how-to-load-jar-files-dynamically-at-runtime/59743937#59743937

sahil-lakhwani: Yes, I did read on that solution. The idea there is to have a custom class loader as the system class loader, which is more suited for a complex structure where you have to control loading of multiple classes. Also reading the cons - https://github.com/update4j/update4j/wiki/Documentation#cons-2 - I think there would be more work in this than the results we get.
Any thoughts are welcome if I am missing anything.

environments/jvm/src/main/java/io/fission/Server.java
sahil-lakhwani: Yes, there is an ongoing work on that. I'll have a separate PR for that.

environments/jvm/src/main/java/io/fission/Server.java
mordechaim: Update4j contributed here, the cons listed are only relevant to the specific implementation of update4j's classloading model. Simply providing your own classloader as the system loader just works out of the box with the sole benefit of being able to cast `ClassLoader.getSystemClassLoader()` to your type.

environments/jvm/Dockerfile-11
```diff
@@ -0,0 +1,16 @@
...
+# To reuse the build cache, here we split maven dependency
+# download and package into two RUN commands to avoid cache invalidation.
+COPY pom.xml .
+RUN mvn dependency:go-offline
+
+COPY src /usr/src/myapp/src/
+RUN mvn package
+
+FROM openjdk:11-jre
+VOLUME /tmp
```
vishal-biyani, line 13: Is /tmp being used?

environments/jvm/Dockerfile-11
```diff
@@ -0,0 +1,16 @@
...
+COPY --from=BUILD /usr/src/myapp/target/env-java-0.0.1-SNAPSHOT.jar /app.jar
+ENTRYPOINT java ${JVM_OPTS} -Djava.security.egd=file:/dev/./urandom -jar /app.jar --server.port=8888
```
vishal-biyani, line 15: What does the java.security.egd do?

environments/jvm/Dockerfile-11
sahil-lakhwani: It wasn't, removed.

environments/jvm/Dockerfile-11
sahil-lakhwani: The argument `-Djava.security.egd=file:/dev/./urandom` tells the JVM to run use` /dev/urandom` instead of `/dev/random` for strong random numbers. This is done to workaround the problem of less entropy on docker containers, which reportedly can be a problem for JVM blocking or the boot speed.

sahil-lakhwani:
@vishal-biyani Change since the last review - I have restored the name of JVM `Dockerfile-8` to `Dockerfile`, and respected changes in release scripts so that the image pushed is `fission/jvm-env`, as we have today.

vishal-biyani:
@sahil-lakhwani This LGTM, please rebase once.

vishal-biyani (approved):

vishal-biyani:
Hey @sahil-lakhwani has the equivalent PR been raised on fission/environments repo? If not can you please do that, I will close it here.

sahil-lakhwani:
@vishal-biyani Not for the `jvm` environment - I'll add it.
There's one for `jvm-jersey` though

Issue #235, opened 2017-06-20
Docker containers as functions


soamvasani:
This issue tracks a feature request to run docker container images as functions.

Depends on #193.

yqf3139:
In fission, we assume that environment specific containers work in the same pattern. Two or three http endpoints are exposed to receive the function specialization request, function call requests and metrics requests in the future.

To use docker containers as functions, there are maybe two patterns. The first is that it works like the common fission environment containers but specialization is not needed. It listens function calls on a http endpoint. The  second is that it reads request from env or command line variables, writes the response to stdout or other services and then exits. So that is one request per function instance. In this case, we need to define the exchange protocols(translate triggers events into the env variables, and translate the response for the client).

erwinvaneyk:
As a step towards this goal, currently we have the binary environment which supports the execution of any Linux executable (@yqf3139's solution 2).  Although I think docker container execution should be discouraged, as it is by definition not following the operation abstraction of FaaS, I can imagine that there are situations where it is unavoidable.

I like @yqf3139 idea of having 'already' specialized environments, as it fits in the already existing idea of having more configurable environments (like environments with multiple specializations or environments). 

The easiest way to allow execution of arbitrary this is to create an image which is more or less the fetcher + binary env. An alternative would be is to provide binaries of the fetcher + binary env which users can add to their own images in order to implement the environment interface.

So to be concrete for this issue we need the following:
- [ ] Implement a way to provide configurable environments, allowing for the immediate specialization of an environment.
- [ ] Create an image or multi-platform binary that implements the environment interface.
- [ ] Way for users to use this binary/image in their own images.

soamvasani:
> Although I think docker container execution should be discouraged

FaaS is a poorly defined term, and we should evolve with use cases.

There's definitely cases where images-as-functions is useful. E.g. people who already have a process in place to output docker images, but want the resource utilization and UX improvements they can get from fission.

> fetcher + binary env

The arbitrary binary use case is very different from the arbitrary image use case. There is no need for fetcher in this case, the function is already in the image.

> So to be concrete for this issue we need the following:

See #193 for a simpler and more direct approach. There's no need to make environments more complex; we can allow the user to specify a docker image as a function and short-circuit most of the poolmgr. Fetcher will not be used at all. Such functions would have no environment.

erwinvaneyk:
Agreed, fetcher is irrelevant for this use case indeed.

panigrahis:
[...]

PR #1681, opened 2020-08-11
containers as functions

sahil-lakhwani:
Signed-off-by: sahil-lakhwani <sahilakhwani@gmail.com>

Addresses #235 

This PR adds a feature to support Docker containers as functions. User will be able to specify a Docker image which will be run as a Fission Function.
It assumes the Docker image exposes port 8888 to serve any HTTP requests.

Changes: 
CLI: A new parameter `--image` is added for `fn create` and `fn update`
Executor: A new executor `container` is added which creates the necessary Kubernetes resources.

Example: 
`fission fn create  --executortype=container --name=carts --port=80 --command="./app" --args="--port=80"   --image weaveworksdemos/catalogue:0.3.5
`

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1681)


vishal-biyani:
Hey @sahil-lakhwani A few things that would be good before we can start 

1. Can you please add a working example of a container function, which should be different from usual functions.
2. An ideal example would be a microservice with more than a few API endpoints and should be possible to access without the need for creating multiple routing endpoints
3. Also some tests - unit, as well as integration, would be really good.

pkg/executor/executortype/container/container.go
```diff
@@ -0,0 +1,481 @@
...
+	service := &apiv1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        svcName,
+			Labels:      deployLabels,
+			Annotations: deployAnnotations,
+		},
+		Spec: apiv1.ServiceSpec{
+			Ports: []apiv1.ServicePort{
+				{
+					Name:       "http-env",
```
harshthakur9030, line 324: Hey @sahil-lakhwani , are the values here intentionally hard-coded?

RealHarshThakur:
Some of the pending items which can be addressed in other PRs are:

- Client-side redirects: Supporting redirects based on user input. As of now, frontend applications may try to fetch JS and CSS files at `/` when the function is hosted at `/frontend` . The list of redirects could get exhaustive as a lot of such files can be used.

- Scaling function pods to 0 by enabling HPA based on metrics from Router. Simplest way to achieve this would be to use KEDA. If not, we might have to install other third party component like [this](https://github.com/DirectXMan12/k8s-prometheus-adapter). This also applies to NewDeploy

pkg/executor/executortype/container/common.go
```diff
@@ -0,0 +1,153 @@
...
+	k8s_err "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	fv1 "github.com/fission/fission/pkg/apis/core/v1"
+)
+
+// getResources gets the resources(CPU, memory) set for the function
+func (cn *Container) getResources(fn *fv1.Function) apiv1.ResourceRequirements {
```
vishal-biyani, line 35: nitpick - wonder if we can modify the similar method here [func (deploy *NewDeploy) getResources(env *fv1.Environment, fn *fv1.Function) apiv1.ResourceRequirements {](https://github.com/fission/fission/blob/04350fc0d7c18baa23158388fc22d989cd61aa30/pkg/executor/executortype/newdeploy/newdeploy.go#L337-L338) to reuse common logic. We can also refactor later, not going to hold this PR for just this

vishal-biyani (approved):
Largely looks good to me, one small nitpick comment - we can address in this one or next one.

pkg/executor/executortype/container/common.go
sanketsudake: Hi Vishal, we will fix it in another cycle. There are lot of places where we need to combine code across executors.

pkg/router/functionHandler.go
```diff
@@ -257,6 +258,9 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 				req.URL.Path = "/"
 			}
 
+			roundTripper.logger.Info(req.URL.Host)
```
sanketsudake, line 261: Remove these statements and unwanted code.

pkg/utils/utils.go
```diff
@@ -47,6 +47,20 @@ func UrlForFunction(name, namespace string) string {
 	return fmt.Sprintf("%v/%v", prefix, name)
 }
 
+// RemoveFuncName removes the name of the function
```
sanketsudake, line 50: Probably not required, remove after evaluation.

PR #1697, opened 2020-08-31
Non-root users access Secrets and ConfigMaps 

atsai1220:
When using custom environments with a non-root user, secrets and configmaps are inaccessible even if `fsGroup` is configured since the files are only accessible by the owner. Adding group read and execute allows `fsGroup` configuration to allow non-root users to access secrets and configmaps.

Additionally, it is not recommended to run privileged containers but it is sometimes unavoidable. Accessing function secrets and configmaps should not require root access.

Say your Dockerfile has this user
```Dockerfile
ENV USER=fission \
    UID=1000 \
    GID=1000

RUN addgroup --gid "$GID" "$USER" \
    && adduser \
    --disabled-password \
    --gecos "" \
    --home "/" \
    --ingroup "$USER" \
    --no-create-home \
    --uid "$UID" \
    "$USER"

USER 1000

ENTRYPOINT ["python3"]
CMD ["server.py"]

```

Then your Fission environment would be

```yaml
apiVersion: fission.io/v1
kind: Environment
metadata:
  name: fission-secret
  namespace: default
spec:
  builder:
    command: build
    image: fission/python-builder
  imagepullsecret: your-pull-secret
  keeparchive: false
  poolsize: 3
  resources: {}
  runtime:
    podspec:
      securityContext:
        fsGroup: 1000
    image: your-custom-image/name:1.0
  version: 2

```

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1697)


vishal-biyani:
Thanks a lot for this change @atsai1220 - very useful. I will merge this in a bit.

vishal-biyani:
Integration tests passed in https://github.com/fission/fission/pull/1710 will merge this once it it green!

PR #1700, opened 2020-09-03
Skaffold kind

vishal-biyani:
This PR changes the Skaffold and some manifests to make it work with cloud clusters as well as with Kind. This will enable a seamless way of building and deploying across both the targets.

Future improvements could be:

- Build the only fission-bundle image and skip Fetcher and preupgradecheck (Most cases fetcher and upgrade won't change a lot). This is right now not working with Skaffold patches but works by commenting lines in build and artifactOverrides sections.

- Debugging support

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1700)


vishal-biyani:
Hey @ankitjain235 Do you mind trying this out please?

Issue #854, opened 2018-08-09
Custom metric based HPA


vishal-biyani:
Currently, the HPA in new deployment executor only supports CPU (As the HPA v1 only supported CPU). The v2beta1 HPA has support for custom metrics and Fission should support that. 

Initially, the v2beta1 HPA should support single metric but eventually also add support for multiple metrics.

On a related note, the new metric server replaces heapster, details: https://github.com/fission/fission/issues/812

Issue #812, opened 2018-07-24
Heapster to Metric Server switch


vishal-biyani:
Currently, Heapster is deployed based on user input to enable HPA for Autoscaling. Heapster is [deprecated in Kubernetes 1.11 and will be removed in 1.13](https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md). Fission needs to switch from Heapster to metric server and test that setup works on a Kubernetes setup.

sanketsudake:
We need to get rid of heapster from helm chart.

sanketsudake:
We have removed Heapster via https://github.com/fission/fission/pull/2256 .  Fission relies on metric-server now.

PR #1711, opened 2020-09-10
Add custom-metrics support, upgrade to autoscaling/v2beta2 for HPA

ankitjain235:
This PR addresses https://github.com/fission/fission/issues/854 and https://github.com/fission/fission/issues/812.

Tasks to be done:
- [ ] Remove heapster resource configurations and add metrics-server configurations.
- [x] Update code to use autoscaling/v2beta2 and support existing feature targetCPU CLI flag using custom metrics.
- [x] Add custom metrics to ExecutionStrategy and add relevant functionality.
- [ ] Add test cases for custom metrics.
- [ ] Support scaling policies if possible : https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-policies.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1711)


harshthakur9030:
Hey @ankitjain235 , I noticed `v2beta2` is supported from K8s 1.18. Can we keep both autoscaling versions and execute based on k8s version?

ankitjain235:
@harshthakur9030 So actually the `v2beta2` was introduced in Kubernetes 1.12 while the `spec.behaviour` was added to HPA spec in 1.18 which allows us to provide custom scaling policies: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior.
So if we want, we can add support for the scaling policies specifically for 1.18+.

harshthakur9030:
> So if we want, we can add support for the scaling policies specifically for 1.18+.

@ankitjain235  Yes, can you please create a new issue to track it ?

charts/fission-all/values.yaml
```diff
@@ -267,8 +267,10 @@ analytics: true
 ## Internally used for generating an analytics job for non-helm installs
 analyticsNonHelmInstall: false
 
-## Enable Heapster only in clusters where heapster does not exist already
-heapster: false
+# ## Enable Heapster only in clusters where heapster does not exist already
+# heapster: false
```
harshthakur9030, line 271: From the above file(templates/deployment.yaml), I noticed heapster was removed. Does this need to be removed too?

pkg/apis/core/v1/types.go
```diff
@@ -405,6 +406,9 @@ type (
 
 		// This is the timeout setting for executor to wait for pod specialization.
 		SpecializationTimeout int
+
+		// Only for newdeploy to provide custom metrics support added in autoscaling/v2beta2
```
harshthakur9030, line 410: nit: Can you please change the comment to start with `CustomMetrics` so that golint doesn't show any warnings?

pkg/executor/executortype/newdeploy/newdeploymgr.go
```diff
@@ -573,7 +574,8 @@ func (deploy *NewDeploy) updateFunction(oldFn *fv1.Function, newFn *fv1.Function
 
 	deployChanged := false
 
-	if oldFn.Spec.InvokeStrategy != newFn.Spec.InvokeStrategy {
+	// TODO : Replace with custom equal function if required
```
harshthakur9030, line 577: I do think custom function would be better because reflection doesn't show errors during compilation and is a performance intensive package. Please let me know if this approach is worth it

pkg/apis/core/v1/validation.go
```diff
@@ -295,6 +297,21 @@ func (is InvokeStrategy) Validate() error {
 	return result.ErrorOrNil()
 }
 
+// Function to check if Target CPU utilization is added through custom metrics
+// func checkIfCPUDefinedThroughCustomMetrics(es ExecutionStrategy) bool {
+
```
harshthakur9030, line 302: Is this to check whether it's working? Or will it be uncommented later?

sanketsudake:
In 1.23 HPA graduated to GA. We need to think about adopting v2 instead of v2beta. 
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#horizontalpodautoscaler-v2-graduates-to-ga

sanketsudake:
Working on new set of changes based on existing PR. As changes in the PR are outdated.

https://github.com/fission/fission/pull/2421
https://github.com/fission/fission/pull/2423

sanketsudake:
Closing this PR , please track referenced PR for further updates.

PR #1766, opened 2020-10-01
added extra contents to  README.md

Animasingh123:
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1766)


bhavin192 (commented):
Thank you for your willingness to contribute. I don't think we need to specify this explicitly, the LICENSE text already covers this plus every major source file has a copyright header.

PR #1814, opened 2020-10-07
Add sequential flag to mqtrigger

### LINKED ISSUE #1569: Sequential function invokations when using Kafka triggers

<pre>
$ fission version
client:
  fission/core:
    BuildDate: "2020-02-03T08:40:57Z"
    GitCommit: bda974a72c9093e241c1dae6a7fc1a2d16e28b02
    Version: 1.8.0
server:
  fission/core:
    BuildDate: "2020-02-03T08:40:57Z"
    GitCommit: bda974a72c9093e241c1dae6a7fc1a2d16e28b02
    Version: 1.8.0
</pre>

**Describe the bug**

Not a bug but more a feature request.

From what I see in the [code](https://github.com/fission/fission/blob/f1e243c145239e2280a177c6acf8c23a3e3feb68/pkg/mqtrigger/messageQueue/kafka/kafka.go#L185) of `mqtrigger`, the processing of incoming messages is performed in concurrently.

For some use cases, preserving ordering of function calls can be very important from a business point of view (especially if the functions  have side-effects). Thus, it would be interesting to have an option to configure the message consumption policy at the trigger level: either concurrent or sequential.

What do you think ?

daniel-shuy:
Fixes #1569 

Adds a `sequential` flag to the `mqtrigger` function. If set to `true`, consumes messages sequentially instead of concurrently.

I wasn't sure whether to add support for all 3 MQs (Kafka, Azure Queue Storage, NATS), as they currently have different behaviors (Kafka and Azure Queue Storage MQs currently consumes messages concurrently, while NATS MQ consumes messages sequentially). Therefore I separated the implementation for each MQ into its own commit so that they can be dropped if we decide not to support them.

Note that if support for the `sequential` flag is added to the NATS MQ, it is a breaking change, as it would mean changing the default behavior from sequential to concurrent.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1814)

### DIFF
--- a/pkg/mqtrigger/messageQueue/kafka/kafka.go
+++ b/pkg/mqtrigger/messageQueue/kafka/kafka.go
@@ -185,7 +185,12 @@ func (kafka Kafka) Subscribe(trigger *fv1.MessageQueueTrigger) (messageQueue.Sub
 	go func() {
 		for msg := range consumer.Messages() {
 			kafka.logger.Debug("calling message handler", zap.String("message", string(msg.Value[:])))
-			go kafkaMsgHandler(&kafka, producer, trigger, msg, consumer)
+			msgHandler := func() { kafkaMsgHandler(&kafka, producer, trigger, msg, consumer) }
+			if trigger.Spec.Sequential {
+				msgHandler()
+			} else {
+				go msgHandler()
+			}
 		}
 	}()
 
--- a/pkg/mqtrigger/messageQueue/nats/nats.go
+++ b/pkg/mqtrigger/messageQueue/nats/nats.go
@@ -129,112 +129,120 @@ func (nats Nats) Unsubscribe(subscription messageQueue.Subscription) error {
 func msgHandler(nats *Nats, trigger *fv1.MessageQueueTrigger) func(*ns.Msg) {
 	return func(msg *ns.Msg) {
 
-		// Support other function ref types
-		if trigger.Spec.FunctionReference.Type != fv1.FunctionReferenceTypeFunctionName {
-			nats.logger.Fatal("unsupported function reference type for trigger",
-				zap.Any("function_reference_type", trigger.Spec.FunctionReference.Type),
-				zap.String("trigger", trigger.ObjectMeta.Name))
-		}
+		cb := func() {
+			// Support other function ref types
+			if trigger.Spec.FunctionReference.Type != fv1.FunctionReferenceTypeFunctionName {
+				nats.logger.Fatal("unsupported function reference type for trigger",
+					zap.Any("function_reference_type", trigger.Spec.FunctionReference.Type),
+					zap.String("trigger", trigger.ObjectMeta.Name))
+			}
 
-		// with the addition of multi-tenancy, the users can create functions in any namespace. however,
-		// the triggers can only be created in the same namespace as the function.
-		// so essentially, function namespace = trigger namespace.
-		url := nats.routerUrl + "/" + strings.TrimPrefix(utils.UrlForFunction(trigger.Spec.FunctionReference.Name, trigger.ObjectMeta.Namespace), "/")
-		nats.logger.Debug("making HTTP request", zap.String("url", url))
+			// with the addition of multi-tenancy, the users can create functions in any namespace. however,
+			// the triggers can only be created in the same namespace as the function.
+			// so essentially, function namespace = trigger namespace.
+			url := nats.routerUrl + "/" + strings.TrimPrefix(utils.UrlForFunction(trigger.Spec.FunctionReference.Name, trigger.ObjectMeta.Namespace), "/")
+			nats.logger.Debug("making HTTP request", zap.String("url", url))
 
-		headers := map[string]string{
-			"X-Fission-MQTrigger-Topic":      trigger.Spec.Topic,
-			"X-Fission-MQTrigger-RespTopic":  trigger.Spec.ResponseTopic,
-			"X-Fission-MQTrigger-ErrorTopic": trigger.Spec.ErrorTopic,
-			"Content-Type":                   trigger.Spec.ContentType,
-		}
+			headers := map[string]string{
+				"X-Fission-MQTrigger-Topic":      trigger.Spec.Topic,
+				"X-Fission-MQTrigger-RespTopic":  trigger.Spec.ResponseTopic,
+				"X-Fission-MQTrigger-ErrorTopic": trigger.Spec.ErrorTopic,
+				"Content-Type":                   trigger.Spec.ContentType,
+			}
 
-		// Create request
-		req, err := http.NewRequest("POST", url, bytes.NewReader(msg.Data))
+			// Create request
+			req, err := http.NewRequest("POST", url, bytes.NewReader(msg.Data))
 
-		if err != nil {
-			nats.logger.Error("failed to create HTTP request to invoke function",
-				zap.Error(err),
-				zap.String("function_url", url))
-			return
-		}
-
-		for k, v := range headers {
-			req.Header.Set(k, v)
-		}
-
-		var resp *http.Response
-		for attempt := 0; attempt <= trigger.Spec.MaxRetries; attempt++ {
-			// Make the request
-			resp, err = http.DefaultClient.Do(req)
 			if err != nil {
-				nats.logger.Error("sending function invocation request failed",
+				nats.logger.Error("failed to create HTTP request to invoke function",
+					zap.Error(err),
+					zap.String("function_url", url))
+				return
+			}
+
+			for k, v := range headers {
+				req.Header.Set(k, v)
+			}
+
+			var resp *http.Response
+			for attempt := 0; attempt <= trigger.Spec.MaxRetries; attempt++ {
+				// Make the request
+				resp, err = http.DefaultClient.Do(req)
+				if err != nil {
+					nats.logger.Error("sending function invocation request failed",
+						zap.Error(err),
+						zap.String("function_url", url),
+						zap.String("trigger", trigger.ObjectMeta.Name))
+					continue
+				}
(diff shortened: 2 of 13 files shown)


pkg/mqtrigger/messageQueue/nats/nats.go
```diff
@@ -129,112 +129,120 @@ func (nats Nats) Unsubscribe(subscription messageQueue.Subscription) error {
...
+						zap.String("function_url", url),
+						zap.String("trigger", trigger.ObjectMeta.Name))
+					continue
+				}
+				if resp == nil {
+					continue
+				}
+				if err == nil && resp.StatusCode == http.StatusOK {
+					// Success, quit retrying
+					break
```
harshthakur9030, line 183: Can you please change this to an appropriate `return` statement?

pkg/mqtrigger/messageQueue/nats/nats.go
```diff
@@ -129,112 +129,120 @@ func (nats Nats) Unsubscribe(subscription messageQueue.Subscription) error {
...
+			for k, v := range headers {
+				req.Header.Set(k, v)
 			}
-			if err == nil && resp.StatusCode == http.StatusOK {
-				// Success, quit retrying
-				break
+
+			var resp *http.Response
+			for attempt := 0; attempt <= trigger.Spec.MaxRetries; attempt++ {
+				// Make the request
```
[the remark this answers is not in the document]
daniel-shuy: @harshthakur9030 should this be in a separate PR? I'm actually just moving the entire code block into an anonymous function so that I can conditionally call it as a goroutine, I didn't change any of the existing logic

vishal-biyani:
Hey @therahulbhati This will also impact all Keda connectors I believe?

daniel-shuy:
@vishal-biyani I don't think so, don't Keda connectors have their own implementation (https://github.com/fission/keda-connectors)?

Looking at the implementation of the Keda Kafka HTTP Connector (https://github.com/fission/keda-connectors/tree/master/kafka-http-connector), I don't think it will have this issue as it is consuming messages sequentially, unlike the mqtrigger Kafka connector which is consuming messages concurrently.

pkg/mqtrigger/messageQueue/nats/nats.go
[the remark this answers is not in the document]
harshthakur9030: Yes, that bug existed before you touched the code base. Since you have already touched this part of the code base, I was hoping you could do it. Everything remains the same, just the for loop which sets the headers needs to go into the loop which makes the requests.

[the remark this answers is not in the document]
daniel-shuy: Ok sure, I can do it

pkg/mqtrigger/messageQueue/nats/nats.go
daniel-shuy: ? Why? If we return here the rest of the function (closing the response body, checking if the response returned an error, trigger ack message, publishing to response topic) won't execute

harshthakur9030: Oh yes! My bad. One quick question though: Shouldn't we consider for status codes between 200 and 300 to be successful rather than just 200?

pkg/mqtrigger/messageQueue/nats/nats.go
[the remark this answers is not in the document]
daniel-shuy: Done!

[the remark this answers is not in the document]
harshthakur9030: Thanks !!

pkg/mqtrigger/messageQueue/nats/nats.go
daniel-shuy: I don't think the router can return anything other than HTTP `200` for a successful request? Maybe @vishal-biyani can confirm

pkg/mqtrigger/messageQueue/kafka/kafka.go
```diff
@@ -185,7 +185,12 @@ func (kafka Kafka) Subscribe(trigger *fv1.MessageQueueTrigger) (messageQueue.Sub
 	go func() {
 		for msg := range consumer.Messages() {
 			kafka.logger.Debug("calling message handler", zap.String("message", string(msg.Value[:])))
-			go kafkaMsgHandler(&kafka, producer, trigger, msg, consumer)
+			msgHandler := func() { kafkaMsgHandler(&kafka, producer, trigger, msg, consumer) }
+			if trigger.Spec.Sequential {
+				msgHandler()
+			} else {
+				go msgHandler()
+			}
```
vishal-biyani, line 193: Won't we need a `wg.Add` & `wg.Done` similar to AzureQueue here?

pkg/mqtrigger/messageQueue/nats/nats.go
```diff
@@ -129,112 +129,120 @@ func (nats Nats) Unsubscribe(subscription messageQueue.Subscription) error {
...
+						zap.String("topic", trigger.Spec.ResponseTopic),
+						zap.String("trigger", trigger.ObjectMeta.Name))
+				}
+			}
+		}
+
+		if trigger.Spec.Sequential {
+			cb()
+		} else {
+			go cb()
```
vishal-biyani, line 245: Just checking if `wg.Add` and `wg.Done` similar to AzureQueue would be needed here?

vishal-biyani (changes requested):
Hi @daniel-shuy Thanks for this PR and looks great except for 2 really minor comments. Please let me know your thoughts and I will merge this one.

pkg/mqtrigger/messageQueue/nats/nats.go
daniel-shuy: I took a look at the AzureQueue mqtrigger implementation, and seems like it is using `sync.WaitGroup` to notify [AzureQueueSubscription.done](https://github.com/daniel-shuy/fission/blob/mqtrigger-sequential/pkg/mqtrigger/messageQueue/azurequeuestorage/asq.go#L84) (a `chan bool`) (see [asq.go#L275](https://github.com/daniel-shuy/fission/blob/mqtrigger-sequential/pkg/mqtrigger/messageQueue/azurequeuestorage/asq.go#L275)).

The other mqtrigger implementations do not have anything equivalent, so it shouldn't be needed.

vishal-biyani:
Thanks @daniel-shuy this looks good, I will merge in a day or so, again thanks for your effort on this one 🎉

vishal-biyani (approved):

chetanpdeshmukh:
Hello @daniel-shuy ,
Thanks again for the PR!

Thank you for contributing to Fission project. Could you please fill out [this form](https://docs.google.com/forms/d/e/1FAIpQLSdhxtWL5N2PU7MYC2IRRAzYw-CrG9QtsYNvkiiOPVEQj8i51w/viewform?usp=sf_link), so we can send you the well deserved awesome swags :)

Team Fission

daniel-shuy:
@chetanpdeshmukh yay, thanks! The form is restricted though

chetanpdeshmukh:
> @chetanpdeshmukh yay, thanks! The form is restricted though

Hey @daniel-shuy can you give it a try now? Should work smoothly

daniel-shuy:
@chetanpdeshmukh It works now, I've filled it up, thanks!

daniel-shuy:
@vishal-biyani just noticed that this PR hasn't been merged!

AnatoliyYakimov:
Hello, @daniel-shuy! Nice work you did! Any updates on this pull request?
We would like to use fission in our project, and this feature would help us a lot to keep consistency of our system.

daniel-shuy:
@AnatoliyYakimov Unfortunately I don't have permission to merge, I'm also wondering why the PR hasn't been merged after so long 😅

AnatoliyYakimov:
Maybe they just forgot)
@vishal-biyani please merge this feature if it's ready. We would be happy to this it in next release of Fission!)

daniel-shuy:
I've resolved the conflicts with the latest `master` branch

AnatoliyYakimov:
Also looking at the code, i saw that we continue to consume messages from topic if lambda returned an error. So if lambda returns an error, we still lose order, if'd i get it right.
For cases where ordering strictly required it would be better to have an option that allows you to stop processing entierly when facing an error.
What do you think about it?

And some personal question: im not very strong in go, and don't know Fission code that deep, so i don't understend what happens if after polling some messages weren't commited. Does sacrama (go kafka lib, if i spelled it right) return this messages from cache, commit all polled messages anyway or it will try to poll again from previous commit?

daniel-shuy:
@AnatoliyYakimov Oh wow, you're right, didn't think of that. We should break out of the loop on error and repoll Kafka again, good catch!

AnatoliyYakimov:
If error is not transient and we will poll again and again, we will DDOS our lambdas. So maybe we need some kind of error retry count and then stop trigger. I think in Kubeless you can set retry count, before sending to DLQ. Maybe spec could look like this:
onError: SendToErrorTopic | StopProcessing
retryCount: 5
SendToErrorTopic and StopProcessing are error handling strategies. They can implement abstract interface, so in this case we could easily create new strategies for our mqttriggers. Also, if we abstract away from MQ implementation, we can reuse these strategies for any availible MQ.

daniel-shuy:
@AnatoliyYakimov actually, let me get back to you on this. sarama is returning a channel, which is looped over sequentially. I'll need to look into sarama's code to see if its caching messages

sanketsudake:
Archiving this PR for now due to inactivity (see the `archive-pr` label). We revisit archived PRs periodically — please rebase and reopen if you'd like to continue.

Issue #1730, opened 2020-09-25
Insecure connection to TLS enabled Kafka using Fission MQT


abbaspw21:
**Is your feature request related to a problem? Please describe.**
We want to connect TLS enabled Kafka via fission MQT without authentication in an insecure way as we just want to transmit the data in encrypted form.

**Describe the solution you'd like**
Can we have a environment variable in Fission MQT deployment which allows connection to TLS enabled kafka without authentication 

It can be achieved via enabling ‘InsecureSkipVerify’ flag while defining tls config in : 

https://github.com/fission/fission/blob/ef65602b79f0acbd5f238ba3bfcc736fa0f59149/pkg/mqtrigger/messageQueue/kafka/kafka.go#L195 

```
        tlsConfig.InsecureSkipVerify = true
```

hd40910:
Tagging @vishal-biyani

vishal-biyani:
Hey @abbaspw21 The Kafka integration already has Kafka TLS as a flag at the Fission level, please reinstall with the flag as false. 

https://github.com/fission/fission/blob/ef65602b79f0acbd5f238ba3bfcc736fa0f59149/charts/fission-all/values.yaml#L179-L188

abbaspw21:
Hi @vishal-biyani,

We wanted to have TLS enabled but we don't want the certificates to get validated (insecure connection to TLS endpoint ).

```
kafka:
  enabled: false
  # note: below link is only for reference. 
  # Please use the brokers link for your kafka here. 
  brokers: 'broker.kafka:9092' # or your-bootstrap-server.kafka:9092/9093
  tls :
    enabled: true
  authentication:
    tls:
      enabled: false
      caCert: '' # path to certificate containing public key of CA authority
      userCert: '' # path to certificate containing public key of the user signed by CA authority
      userKey: '' # path to private key of the user
```
Th indicator tls.enabled means communication is happening using TLs protocol without any authentication.

vishal-biyani:
Got it - thanks, that makes sense @abbaspw21 Then what you pointed probably makes sense. Do you know if you folks would be switching to Keda based MQT anytime soon because that might affect how much we want this feature?

abbaspw21:
Hi @vishal-biyani ,

Currently we don't have any solid plans for Keda based MQT.
So it would much appreciated if we can get a private patch for this issue.

vishal-biyani:
Closed by https://github.com/fission/fission/pull/1829

PR #1829, opened 2020-10-13
Added flag for insecureSkipVerfiy

therahulbhati:
Addresses #1730

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1829)

### DIFF
--- a/pkg/mqtrigger/messageQueue/kafka/kafka.go
+++ b/pkg/mqtrigger/messageQueue/kafka/kafka.go
@@ -201,8 +201,9 @@ func (kafka Kafka) getTLSConfig() (*tls.Config, error) {
 
 	tlsConfig.Certificates = []tls.Certificate{cert}
 
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/mqtrigger/messageQueue/kafka/kafka.go
```diff
@@ -201,8 +201,9 @@ func (kafka Kafka) getTLSConfig() (*tls.Config, error) {
 
 	tlsConfig.Certificates = []tls.Certificate{cert}
 
-	if err != nil {
-		return nil, err
+	skipVerify, err := strconv.ParseBool(os.Getenv("INSECURE_SKIP_VERIFY"))
+	if err != nil && skipVerify {
```
vishal-biyani, line 205: Something seems odd here - so if err is not nil and assuming skipVerify is true then set the value. But also no separate handling of error per se. Should we split error handling - as in if there is an error, then return and then set flag separately?

therahulbhati: There was problem in the logic. I have fixed the issue.

Issue #1828, opened 2020-10-12
Error fetching configmap/secret to function from fission-function namespace


Magomedali:
**Fission/Kubernetes version**

<pre>
Fission version: 1.11.0
Kubectl version: v1.15.11-eks
</pre>

**Kubernetes platform (e.g. Google Kubernetes Engine)**

AWS/eks

**Describe the bug**
Error fetching configmap and secret when run function.
Secrets "scr-access-token" is forbidden: User "system:serviceaccount:fission-function:fission-fetcher" cannot get resource "secrets" in API group "" in the namespace "fission-function"

**To Reproduce**

Step 2: Create configmap or secret in the namespace "fission-function"
Step 3: Create function with that secret/configmap
Step 3: Run this function

**Expected result**

Running function with successfully fetching secrets and configmaps from namespace "fission-function"

**Actual result**

Fetching failed

**Screenshots/Dump file**

[...]

PR #1836, opened 2020-10-14
Added clusterrolebinding for read configmap/secret from fission namespaces

Magomedali:
Solution #1828

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1836)


therahulbhati:
The issue that PR addresses of "Not being able to access configmap/secret in fission-function namespace"  #1828 is not valid as the configmap/secret and fission-function were in different namespaces, due to which user was facing the issue. Hence closing the PR.

Issue #1720, opened 2020-09-19
Alert user if setting poolsize to zero


swaroopnuli:
**Fission/Kubernetes version**

<pre>
$ fission --version
1.11.0
$ kubectl version
AWS EKS 1.17.9
</pre>

**Kubernetes platform (e.g. Google Kubernetes Engine)**
AWS EKS

**Describe the bug**

Poolmanger Functions associated with Environments having poolsize set to 0 are failing.

Please provide an indication if functions cannot be created with poolsize 0.
Otherwise have flexiblity of scalingup even with poolsize 0
**To Reproduce**

**Expected result**

**Actual result**

**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

vishal-biyani:
The Poolsize for pool manager functions can not be zero because the pods are specialized from pool. You need a pool of at least 1 size for pool manager functions! I am closing this issue but please let us know if you have other questions.

swaroopnuli:
Got it. It will be helpful if there is an alert or poolsize=0 disabled to avoid it.

Divya063:
Maybe having a check for poolsize while creating or updating the env would work @vishal-biyani can I pick this up?

vishal-biyani:
@Divya063 Yes please pick it up. Yes this is a CLI change and can be added as a check

sanketsudake:
Closing issue since respective checks added via #1837. The fix would be available in upcoming release 1.13.

PR #1809, opened 2020-10-05
Added check for poolsize

Divya063:
Closes #1720

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1809)

PR #1837, opened 2020-10-14
Pool size warning

therahulbhati:
Addresses #1720

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1837)

### DIFF
--- a/pkg/fission-cli/cmd/environment/update.go
+++ b/pkg/fission-cli/cmd/environment/update.go
@@ -29,6 +29,7 @@ import (
 	fv1 "github.com/fission/fission/pkg/apis/core/v1"
 	"github.com/fission/fission/pkg/fission-cli/cliwrapper/cli"
 	"github.com/fission/fission/pkg/fission-cli/cmd"
+	"github.com/fission/fission/pkg/fission-cli/console"
 	flagkey "github.com/fission/fission/pkg/fission-cli/flag/key"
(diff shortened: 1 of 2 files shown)


RealHarshThakur (approved):

RealHarshThakur:
We can close this https://github.com/fission/fission/pull/1809 too , if this PR is merged

sanketsudake:
@vishal-biyani mentioned there is an offline discussion about current PR, so we would be holding a merge.

vishal-biyani:
So even if poolsize is zero - it should not lead to an error but a warning. For ex. for NewDeploy - users do set poolsize to zero and that is intentional.

pkg/fission-cli/cmd/environment/update.go
```diff
@@ -103,6 +103,9 @@ func updateExistingEnvironmentWithCmd(env *fv1.Environment, input cli.Input) (*f
 
 	if input.IsSet(flagkey.EnvPoolsize) {
 		env.Spec.Poolsize = input.Int(flagkey.EnvPoolsize)
+		if env.Spec.Poolsize == 0 {
```
vishal-biyani, line 106: If the poolsize is set to zero - error should happen only in case of PoolManager. For NewDeploy it actually makes sense to keep poolsize as 0!

vishal-biyani:
LGTM

RealHarshThakur (approved):

PR #1847, opened 2020-10-26
Keda integration was broken due to a typo. Fixed it, now it should work

mahendrabagul:
There was a wrong key used in values.yaml. I have corrected it, now it should work.

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1847)


mahendrabagul:
There is already a PR for this, closing this one

Issue #1845, opened 2020-10-21
Consider not using cluster-admin


coderanger:
Hi there, this is 100% a drive-by issue because I was helping someone in Kubernetes Slack with a problem but I wanted to make sure this is on your radar. In your deployment manifests, you bind the fission ServiceAccount to `cluster-admin`. That seems highly unnecessary and a big security hole. I'm sure it makes setup much simpler since it looks like the main controller deployment sets up a lot of secondary API objects after startup, but maybe at least support a more secure mode where Fission only gets the permissions it needs for normal operations and those secondary global objects are set up via the manifests? For example, it looks like you install your CustomResourceDefinition objects this way.

vishal-biyani:
Hi @coderanger - thanks for bringing this to notice, this is definitely a big no, will fix it ASAP!

PR #1867, opened 2020-11-26
Updated role and permission for fission-crd cluster role

therahulbhati:
fixes https://github.com/fission/fission/issues/1845

---
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/1867)


charts/fission-all/templates/deployment.yaml
```diff
@@ -49,9 +49,104 @@ rules:
...
+  name: fission-cr-admin
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - pods
+  - secrets
+  - services
+  - serviceaccounts
```
RealHarshThakur, line 73: Creating SA might still be a security concern. Once user has installed fission, the service accounts would exist by default. The only reason we require this permission is because the controller will create if they don't already exist, right? If so,  two scenarios where this helps is:
* Getting into a state where service accounts are deleted 
*  Incomplete installation 
Can't we just give an error to the user  through when such a thing happens?

charts/fission-all/templates/deployment.yaml
```diff
@@ -49,9 +49,104 @@ rules:
...
+  resources:
+  - scaledjobs
+  - scaledjobs/finalizers
+  - scaledjobs/status
+  - triggerauthentications
+  - triggerauthentications/status
+  verbs:
+  - '*'
+- apiGroups:
+  - keda.k8s.io
```
RealHarshThakur, line 139: We'll need to change this soon for KEDA 2.0 . Can we not add both API groups?

therahulbhati: Yes, I have added for both keda.k8s.io and keda.sh

vishal-biyani:
Hey @therahulbhati Can we also test the upgrade from a previous Fission with this PR please once?

charts/fission-all/templates/deployment.yaml
therahulbhati: https://github.com/fission/fission/issues/1928

PR #1906, opened 2021-01-07
Fix Fission cli can be more like k8s.

tosone:
Before we can delete function like this `fission fn delete --name func-name`, but now we can operate this action with command `fission fn delete func-name`, can be more like k8s.

### DIFF
--- a/pkg/fission-cli/cliwrapper/cli/cli.go
+++ b/pkg/fission-cli/cliwrapper/cli/cli.go
@@ -24,6 +24,9 @@ type (
 	Input interface {
 		//Parse(input interface{}) error
 
+		// Args get all of args from command line
+		Args(int) string
+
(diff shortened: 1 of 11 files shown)


vishal-biyani:
Hi @tosone Will this be backward compatible - as in if we pass the name even then it works? Also wanted to confirm if we are only changing the delete part of CLI right?

tosone:
Emmm, this pr not backward compatible. Can't be compatible in both ways, because the `function name` is a required parameter.

vishal-biyani:
@tosone In the interest of backward compatibility - are you ok if we don't merge this change? Sorry but want to ensure that other users don't get affected.

tosone:
It is ok to reject this pr. Made this during reading the source code.

Thank you for your reply.

tosone:
Just close this PR.

Issue #1875, opened 2020-12-03
Incorrect status code on max concurrency 


therahulbhati:
Using poolmanager strategy, we can rate-limit the number of concurrent pods. 

**Expected Result**
When the rate limit is hit, one should get a 429 HTTP status code.

**Actual Result**
502 HTTP status code is received instead.

PR #1924, opened 2021-02-01
Fix status code for rate limit

RealHarshThakur:

Fixes https://github.com/fission/fission/issues/1875
Returns the correct status code for rate limits. 
Since there's no way to fetch correct response code from RoundTrip, had to rely on parsing it from the error message. 

Also, I think we  shouldn't retry when we hit the rate limit, instead user is expected to retry later. This way, we avoid holding requests for long time and save resources.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module github.com/fission/fission
 
-go 1.14
-
 require (
 	contrib.go.opencensus.io/exporter/jaeger v0.1.0
 	github.com/Azure/azure-sdk-for-go v12.4.0-beta+incompatible
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/fission/fission
 
+go 1.14
+
 require (
 	contrib.go.opencensus.io/exporter/jaeger v0.1.0
 	github.com/Azure/azure-sdk-for-go v12.4.0-beta+incompatible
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/fission/fission
 
+go 1.14
```
therahulbhati, line 3: Can we please change version to go 1.12.17

pkg/router/functionHandler.go
```diff
@@ -191,11 +192,17 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 				// We might want a specific error code or header for fission failures as opposed to
 				// user function bugs.
 				statusCode, errMsg := ferror.GetHTTPError(err)
-				// if statusCode == http.StatusTooManyRequests {
-				// 	time.Sleep(executingTimeout)
-				// 	executingTimeout = executingTimeout * time.Duration(roundTripper.funcHandler.tsRoundTripperParams.timeoutExponent)
-				// 	continue
-				// } else {
+				if statusCode == http.StatusTooManyRequests {
+					return &http.Response{StatusCode: http.StatusTooManyRequests,
```
therahulbhati, line 196: can we return nil response and error directly like `return nil, err`

pkg/router/functionHandler.go
```diff
@@ -573,14 +579,20 @@ func (fh functionHandler) getProxyErrorHandler(start time.Time, rrt *RetryingRou
 			msg = "function not responses before the timeout"
 			fh.logger.Error(msg, zap.Any("function", fh.function), zap.Any("request_header", req.Header))
 		default:
-			status = http.StatusBadGateway
-			msg = "error sending request to function"
-			fh.logger.Error(msg, zap.Error(err), zap.Any("function", fh.function), zap.Any("request_header", req.Header))
+			code, msg := ferror.GetHTTPError(err)
```
therahulbhati, line 582: Handle the error status code detection directly in GetHTTPError

therahulbhati (approved):
LGTM

PR #1929, opened 2021-02-05
Multiarch release

RealHarshThakur:
Adding release script for multiarch. 

User is required to run ./hack/buildx.sh and can then run the release.sh file


hack/buildx.sh
```diff
@@ -0,0 +1,3 @@
+docker run --rm --privileged linuxkit/binfmt:v0.8
```
vishal-biyani, line 1: Is this file used somewhere?

hack/release.sh
```diff
@@ -70,6 +70,32 @@ push_all() {
...
+
+
+push_builder_image() {
+    TAG=$1 make multiarch-builder
+}
+
+# Push pre-upgrade-checks image
+push_pre_upgrade_checks_image() {
+    TAG=$1 make multiarch-preupgrade
+}
```
vishal-biyani, line 92: Do we need these given we have the image-multiarch which covers all images?

vishal-biyani (changes requested):
Two minor comments, looks good otherwise. Might test it once before merging.

hack/buildx.sh
RealHarshThakur: This starts the emulator , we need to run this file before we start the multiarch build.

hack/release.sh
RealHarshThakur: Might help someone who is just working to test one particular component especially if it's on ARM or something other than AMD.

hack/buildx.sh
vishal-biyani: Can you please add documentation at top of the script?

hack/release.sh
vishal-biyani: Sure, let it be around!

hack/buildx.sh
RealHarshThakur: Done

Makefile
```diff
@@ -53,7 +53,20 @@ image-multiarch:
...
+
+multiarch-fetcher:
+	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/fetcher:$(TAG) --push -f cmd/fetcher/Dockerfile.fission-fetcher .
+
+multiarch-builder:
+	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/builder:$(TAG) --push -f cmd/builder/Dockerfile.fission-builder .
+
+multiarch-preupgrade:
+	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/pre-upgrade-checks:$(TAG) --push -f cmd/preupgradechecks/Dockerfile.fission-preupgradechecks . 
+
```
sanketsudake, line 69: We need to add newly added reporter component here.

RealHarshThakur: Done

sanketsudake (approved):

Makefile
```diff
@@ -58,6 +58,21 @@ image-multiarch:
 	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/preupgradechecks:$(TAG) --push -f cmd/preupgradechecks/Dockerfile.fission-preupgradechecks .
 	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/reporter:$(TAG) --push -f cmd/reporter/Dockerfile.reporter .
 
+multiarch-bundle:
+	docker buildx build --platform=$(PLATFORMS) -t $(REPO)/fission-bundle:$(TAG) --push -f cmd/fission-bundle/Dockerfile.fission-bundle .
```
vishal-biyani, line 62: Hey @RealHarshThakur This seems a bit odd to me. So in platforms we have 3 but we are using same tag for all 3 arch images? I think what this will do is override the previous two and only last one will be available?

vishal-biyani: Never mind - I think as pointed out in our Slac conversation - this will push all three arch images!

sanketsudake:
@RealHarshThakur can you please resolve conflicts ?

sanketsudake (approved):

PR #1938, opened 2021-02-12
Maintain CRD folder to keep track of Schemas

RealHarshThakur:
For a fresh installation, there's nothing new that needs to be done. 

For users who want to upgrade from 1.11 or later(from when validations were introduced), they have to do the following:
- kubectl apply -f crds/v1/

- Install the new release


pkg/crd/crd.go
```diff
@@ -85,16 +161,16 @@ func EnsureFissionCRDs(logger *zap.Logger, clientset *apiextensionsclient.Client
...
-				Scope:   apiextensionsv1beta1.NamespaceScoped,
+				Group: crdGroupName,
+				Scope: apiextensionsv1beta1.NamespaceScoped,
 				Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
 					Kind:     "Environment",
 					Plural:   "environments",
 					Singular: "environment",
 				},
 				PreserveUnknownFields: boolPtr(false),
-				Validation:            environmentValidation,
```
vishal-biyani, line 97: Is there a reason we are removing the validation here and other places?

pkg/crd/crdvalidations.go
```diff
@@ -53,17 +53,61 @@ var (
 		},
 	}
 
+	// Function validation schema properties
+	functionSchemaPropsV2 = map[string]apiextensionsv1beta1.JSONSchemaProps{
```
vishal-biyani, line 57: Overall changes look ok but I have a few points which we can discuss. I will also try this out with some scenarios and then we can close!

cmd/preupgradechecks/main.go
```diff
@@ -65,12 +65,15 @@ Options:
...
 	}
 
-	if !crdBackedClient.IsFissionReInstall() {
+	installed, err := crdBackedClient.IsFissionReInstall()
+	if !installed {
 		logger.Info("nothing to do since CRDs are not present on the cluster")
 		return
 	}
 
+	if err != nil {
```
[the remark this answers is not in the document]
RealHarshThakur: Nope, we really don't do have to do anything in that case. This existed before too, just added it to check the version now

cmd/preupgradechecks/preupgradechecks.go
```diff
@@ -66,18 +65,30 @@ func makePreUpgradeTaskClient(logger *zap.Logger, fnPodNs, envBuilderNs string)
...
 // IsFissionReInstall checks if there is at least one fission CRD, i.e. function in this case, on this cluster.
 // We need this to find out if fission had been previously installed on this cluster
-func (client *PreUpgradeTaskClient) IsFissionReInstall() bool {
+func (client *PreUpgradeTaskClient) IsFissionReInstall() (bool, error) {
 	for i := 0; i < maxRetries; i++ {
-		_, err := client.apiExtClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(FunctionCRD, metav1.GetOptions{})
+		crd, err := client.apiExtClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(FunctionCRD, metav1.GetOptions{})
 		if err != nil && k8serrors.IsNotFound(err) {
-			return false
+			return false, nil
```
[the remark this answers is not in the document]
RealHarshThakur: Yep, missed this.

cmd/preupgradechecks/preupgradechecks.go
```diff
@@ -66,18 +65,30 @@ func makePreUpgradeTaskClient(logger *zap.Logger, fnPodNs, envBuilderNs string)
...
+				foundV2 = true
+			}
+		}
+		client.logger.Sugar().Infof("Crd %v", crd.Spec.Versions)
+		if !foundV2 {
+			return true, errors.New("Version 1 of CRD is deprecated. Please apply Version 2 of CRD")
 		}
 		if err == nil {
-			return true
+			return true, nil
```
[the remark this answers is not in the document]
RealHarshThakur: In this case, error is nil so we're good

crds/v2/environment.yaml
```diff
@@ -0,0 +1,254 @@
...
+                    The grace time for pod to perform connection draining before
+                    termination. The unit is in seconds.
+                  type: integer
+                  format: int64
+                version:
+                  description: Version is the Environment API version
+                  type: integer
+    - name: v2
+      served: true
+      storage: true
```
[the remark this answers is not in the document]
RealHarshThakur: It signifies the version stored in etcd . Only one version can be set to true. https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

pkg/crd/crd.go
RealHarshThakur: They've been moved to the upper layer in the `Schema`.   https://github.com/fission/fission/pull/1938/files#diff-1d74bd0caacd00451f3f6d75e70ac4a780cfe74a82fb5fa774e464935e68680fR80

vishal-biyani:
@RealHarshThakur Let's switch to v2alpha1 from v2 as the next version.

vishal-biyani:
@RealHarshThakur Can you please resolve conflicts on this one

crds/v2/canarconfigs.yaml
```diff
@@ -0,0 +1,30 @@
...
+    listKind: CanaryConfigList
+    plural: canaryconfigs
+    singular: canaryconfig
+  preserveUnknownFields: true
+  scope: Namespaced
+  versions:
+  - name: v1
+    served: true
+    storage: false
+  - name: v2
```
vishal-biyani, line 20: Should these also not change to v2alpha1?

vishal-biyani:
Hey @RealHarshThakur There are a bunch of references in Go names to V2 - we should change all of them to v2alpha1 for consistency and to avoid confusion later

RealHarshThakur:
@vishal-biyani I had to generate the CRDs again, I did that now . Fixed the conflict too. I think we should merge this PR first though: https://github.com/fission/fission/pull/1976

pkg/crd/crdvalidations.go
```diff
@@ -54,6 +54,7 @@ var (
 					Description: "RequestsPerPod indicates the maximum number of concurrent requests that can be served by a specialized pod.\n This is optional. If not specified default value will be taken as 1",
 				},
 			},
+			XPreserveUnknownFields: boolPtr(true),
```
vishal-biyani, line 57: I don't think this is needed - since `RequestsPerPod` has equivalent schema validation, did you see find any issues during tests if this is not given?

pkg/crd/crdvalidations.go
```diff
@@ -166,6 +170,7 @@ var (
 					Description: "LastUpdateTimestamp will store the timestamp the package was last updated metav1.Time is a wrapper around time.Time which supports correct marshaling to YAML and JSON.",
 				},
 			},
+			XPreserveUnknownFields: boolPtr(true),
```
vishal-biyani, line 173: Same comment on this like `RequestsPerPod` field

pkg/crd/crdvalidations.go
RealHarshThakur: This is so that we can add fields inside `spec` in the future and won't have to ask end users to apply the CRDs, until we figure out the schema for next version.

PR #1944, opened 2021-02-25
feat: Add image pull policy when create KEDA connector deployments

chenjr0719:
This PR adds `imagePullPolicy` when creating KEDA connector deployments. It will use the `pullPolicy` value from the helm chart by default. To modify this policy, you can modify the environment variable of `mqtrigger-keda` without changing the helm chart value.


vishal-biyani:
This looks good @chenjr0719 - thanks for the PR!

PR #1951, opened 2021-03-16
Add a SharedIndexInformer for pods to router.

irgeek:
While playing around with Fission I found I was unable to get more than about 5 req / sec out of my cluster, regardless of Fission settings. After digging deeper I found the router component was making a call into the k8s APIs on every request to check the pod's status and k8s was rate-limiting that check.

This PR adds a [SharedIndexInformer](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer) for pods to the router so it is kept up-to-date with pod changes and doesn't need to consult k8s on every request. With these changes the same cluster was able to handle over 200 req / sec, and at that rate all CPUs were 100% used.


therahulbhati (approved):
LGTM

vishal-biyani (approved):
Looks good!

vishal-biyani:
Thank you for your PR @irgeek - this is awesome!

PR #1956, opened 2021-03-18
Fix log sort order when multiple series are returned

LittleFox94:
Series are sorted internal by "order by", but the set of all series returned is not. Tried some ways to sort completely on database, found no way .. client side sort works at least

Multiple series happen when there are multiple pods involved for example


vishal-biyani:
Thanks for this change @LittleFox94 - looks good!

PR #1966, opened 2021-03-22
Add a warning if spec is applied on a dirty worktree

### LINKED ISSUE #1181: fission spec apply -- warn or error on dirty git tree
Since fission spec apply is often used for applying specs from a git repo, it would be useful for it to warn or refuse to work (by defautl) when the git tree is dirty.

hellozee:
Fixes #1181

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927 // indirect
 	github.com/containerd/continuity v0.0.0-20201208142359-180525291bb7 // indirect
+
 	github.com/dchest/uniuri v0.0.0-20160212164326-8902c56451e9
 	github.com/dnaeon/go-vcr v1.1.0 // indirect
 	github.com/docker/go-connections v0.4.0 // indirect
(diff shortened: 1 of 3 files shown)


vishal-biyani:
Hey @hellozee Thanks for this change - very useful.

Can you please paste some screen/text grabs of testing done on CLI and resolve conflict and we are good to merge!

hellozee:
Here is a screenshot, 
ছবি

Also, I am not sure about the updates to go.sum/go.mod files.

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/fission/fission
 
+go 1.15
```
vishal-biyani, line 3: We will have to switch to version that is in use right now - for now I think is 1.12.X - can you please change that?

vishal-biyani:
One last request @hellozee Can you please help with conflicts?

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -185,6 +188,27 @@ func (opts *ApplySubCommand) run(input cli.Input) error {
 	return nil
 }
 
+func warnIfDirtyWorkTree(path string) {
+	repo, err := git.PlainOpen(path)
+	if err != nil {
+		return
```
sanketsudake, line 194: The function should not ignore error silently. Can you please return an error from all places, let the caller decide how to handle the error.

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -123,6 +124,8 @@ func (opts *ApplySubCommand) run(input cli.Input) error {
 			}
 		}
 
+		warnIfDirtyWorkTree(filepath.Clean(specDir + "/.."))
```
sanketsudake, line 127: If this function returns an error we should log it here, but not fail the entire program. This helps us in making sure this function is working properly from a future maintenance perspective.

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -185,6 +188,27 @@ func (opts *ApplySubCommand) run(input cli.Input) error {
...
+		return
+	}
+
+	status, err := workTree.Status()
+	if err != nil {
+		return
+	}
+
+	if !status.IsClean() {
+		console.Warn("Worktree is not clean")
```
sanketsudake, line 208: We are giving a warning to the user, we should also provide an action to resolve them.
```
worktree is not clean, please ensure you have committed changes to git.
```

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -185,6 +188,27 @@ func (opts *ApplySubCommand) run(input cli.Input) error {
 	return nil
 }
 
+func warnIfDirtyWorkTree(path string) {
+	repo, err := git.PlainOpen(path)
```
sanketsudake, line 192: Would this work if a user is not using Git in his project or it's just a local folder and not git repo. Can we check if they have enabled Git or not in their project.

hellozee: It would return an error which I ignore and return from that point, as you pointed above

sanketsudake: As per documentation 
https://pkg.go.dev/gopkg.in/src-d/go-git.v4#PlainOpen
PlainOpen calls returns ErrRepositoryNotExists if not valid repository, we should ignore this error only. Rest we should log.

hellozee: Okay

sanketsudake (approved):

sanketsudake:
Thanks for you contribution @hellozee 🎉

sjk7524068:
> 

I met the same problem, what dose this mean, should I do something to fix ?

PR #1978, opened 2021-03-31
Serve only one request and garbage collect

therahulbhati:
Added functionality to enable one pod to serve only one request in its lifetime, to enable this set `--onceonly=true` or `--yolo=true` . It is recommended to set `--idletimeout=0` so that function pod is garbage collected as soon as possible.


pkg/apis/core/v1/types.go
```diff
@@ -357,6 +357,10 @@ type (
 		// RequestsPerPod indicates the maximum number of concurrent requests that can be served by a specialized pod
 		// This is optional. If not specified default value will be taken as 1
 		RequestsPerPod int `json:"requestsPerPod,omitempty"`
+
+		// OnceOnly specifies if specialized pod will serve exactly one request in its lifetime and would be garbage collected after serving that one request
+		// This is optional. If not specified default value will be taken as false
+		OnceOnly bool `json:"onceOnly,omitempty"`
```
RealHarshThakur, line 363: I am thinking if we should rename this to `Job` or `OneOff` . WDYT? @therahulbhati  @vishal-biyani

vishal-biyani: Job is overloaded term, also used in Kubernetes and we are not using the k8s job so that might be confusing. I think once only means same as "one off" and I don't have preference of one over other. Rahul must have some thoughts?

therahulbhati: I agree with @vishal-biyani about `Job`, `OneOff` sounds better to me than `OnceOnly`.

RealHarshThakur (approved):

PR #1982, opened 2021-04-02
Websocket reaper

RealHarshThakur:
This PR holds all the changes needed to support Websockets. Broadly classified:
* Fetcher has two new endpoints which can be called to generate Kubernetes events which Reaper will pick up to know when the websocket is active/inactive
* Reaper picking up the events and acting based on it rather than the last access time of the function 

Websocket environment related changes are in this PR : https://github.com/fission/environments/pull/49

To try this PR out:
* Role and rolebinding needs to be created in namespace as environment in order to list pods and create events. 
`kubectl apply -f fetcher-pod-sa.yaml`

Manually tested it by:
- Creating the above mentioned environment and trying the steps in it


cmd/fetcher/app/server.go
```diff
@@ -115,6 +115,8 @@ func Run(logger *zap.Logger) {
 	mux.HandleFunc("/specialize", f.SpecializeHandler)
 	mux.HandleFunc("/upload", f.UploadHandler)
 	mux.HandleFunc("/version", f.VersionHandler)
+	mux.HandleFunc("/websocket", f.WebsocketHandler)
```
vishal-biyani, line 118: I think `websocket` is a pretty generic and misleading name for this endpoint. I would use something that represents the fact that we are sending an event to indicate that the environment has started some long-running activity - and don't reap me. Something like `/start_long_connection` or something of that sort please.

cmd/preupgradechecks/Dockerfile.fission-preupgradechecks
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine as godep
+FROM golang:1.14-alpine as godep
```
vishal-biyani, line 1: Let's make 1.14 update a separate PR, please.

pkg/executor/executortype/poolmgr/gpm.go
```diff
@@ -682,3 +693,90 @@ func (gpm *GenericPoolManager) idleObjectReaper() {
...
+	)
+
+	stopper := make(chan struct{})
+	defer close(stopper)
+	informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			mObj := obj.(metav1.Object)
+			fmt.Println()
+			fmt.Println()
+			fmt.Println()
```
vishal-biyani, line 758: Clean up?

pkg/executor/executortype/poolmgr/gpm.go
```diff
@@ -682,3 +693,90 @@ func (gpm *GenericPoolManager) idleObjectReaper() {
...
+
+}
+
+// InactiveEventChecker checks if the pod has emitted an inactive event
+func (gpm *GenericPoolManager) InactiveEventChecker(kubeClient *kubernetes.Clientset) {
+
+	informer := k8scache.NewSharedInformer(
+		&k8scache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				options.FieldSelector = "involvedObject.kind=Pod,type=Normal,reason=Inactive"
```
vishal-biyani, line 739: `Inactive` is a too generic an event name and going to be used by others so easy to confuse, I would suggest we come up with a properly qualified names for events so that we can clearly identify. For ex. something like FISSION_WS_ACTIVATED/TERMINATED or something of that sort.

pkg/fetcher/fetcher.go
```diff
@@ -662,3 +688,77 @@ func (fetcher *Fetcher) SpecializePod(ctx context.Context, fetchReq FunctionFetc
 
 	return errors.Wrapf(err, "error specializing function pod after %v times", maxRetries)
 }
+
+// WebsocketHandler is used to generate websocket events in Kubernetes
+func (fetcher *Fetcher) WebsocketHandler(w http.ResponseWriter, r *http.Request) {
```
vishal-biyani, line 693: nitpick A better name for method

pkg/fetcher/fetcher.go
```diff
@@ -662,3 +688,77 @@ func (fetcher *Fetcher) SpecializePod(ctx context.Context, fetchReq FunctionFetc
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+	pods, err := fetcher.kubeClient.CoreV1().Pods(fetcher.Info.Namespace).List(metav1.ListOptions{
+		FieldSelector: "metadata.name=" + fetcher.Info.Name,
+	})
+	if err != nil {
+		fetcher.logger.Error("Failed to get the pod", zap.Error(err))
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+	for _, pod := range pods.Items {
```
vishal-biyani, line 710: Ideally we should be getting only pod right? Can we not simply retrieve the first one?

vishal-biyani (changes requested):

cmd/preupgradechecks/Dockerfile.fission-preupgradechecks
RealHarshThakur: It's actually a different PR. Just built this on top of it. We just need to merge that one first https://github.com/fission/fission/pull/1976

cmd/fetcher/app/server.go
RealHarshThakur: The event name itself needs to be in camel case. So I'm thinking the end point name should be just `longlivedConnection`  and the event name can be `InitiateLongLivedConnection`. WDYT?

pkg/executor/executortype/poolmgr/gpm.go
RealHarshThakur: For initiating, I am thinking `InitiateLongLivedConnection`  . I am not sure if we should use the term `terminate` . In the environment, we're actually checking if there are any active connections and when we find that there are no active connections, we send out an event so technically there are no connections to terminate. How about just calling it what is: `NoActiveConnections` ?

pkg/fetcher/fetcher.go
RealHarshThakur: Yeah, length of `pod.Items` will always be 1 since we've used a field selector to obtain it. I don't think there's other way here

cmd/fetcher/app/server.go
vishal-biyani: `LongLivedConnection` seems a bit too specific. How about `/eventStart` as the name of endpoint - which is to indicate that you are indicating the start of an event. The actual event could be `wsConnStartedEvent`?

cmd/fetcher/app/server.go
```diff
@@ -115,6 +115,8 @@ func Run(logger *zap.Logger) {
...
+	mux.HandleFunc("/inactive", f.InactiveHandler)
```
vishal-biyani, line 119: Similar to above, we could name this endpoint `/endEvent` and event has message `wsConnStopedEvent`. On second thought how about just having a generic `/event` API endpoint and the API internally makes a distinction about which events etc.?

pkg/executor/executortype/poolmgr/gpm.go
vishal-biyani: `NoActiveConnections` Looks good. Also check related comment earlier!

cmd/fetcher/app/server.go
RealHarshThakur: We could make this a single POST endpoint but right now it's a very simple get call being made from `Environment` to these  two different endpoints .  Since we're exposing these endpoints to developers who might want to customize their environment , it's nice to have that simplicity IMO. Otherwise, the boilerplate to make a POST request gets added in the environment code. Taking a step back , `event` makes perfect sense to have in endpoint name but I am wondering if that's something a user has to know about? They might start to question the event mechanism in place and retry sending events which probably won't be necessary. WDYT? @vishal-biyani

PR #1999, opened 2021-04-13
Add a SharedIndexInformer for services and deployments to NewDeploy executor.

irgeek:
This brings the NewDeploy executor behaviour into sync with GenericPoolManager behaviour (from #1951) by caching Kubernetes services and deployments used in per-request function validation.


pkg/executor/executortype/newdeploy/newdeploymgr.go
```diff
@@ -127,13 +131,19 @@ func MakeNewDeploy(
 		nd.envController = envController
 	}
 
+	informerFactory := k8sInformers.NewSharedInformerFactoryWithOptions(kubernetesClient, 0, k8sInformers.WithNamespace(namespace))
```
RealHarshThakur, line 134: Fission functions/environments can be in any namespace . We need to scope this down somehow at some point but until then, we'll have to go with all namespaces here. Clearly, this can be a lot, wondering if we can filter the informer with a label/annotation to uniquely identify fission resources  cc : @vishal-biyani

irgeek: I believe it is possible,  but it looks a little more complicated to get working. I don't think I'll have time to fix this in the near future, but I did find an example in the `kube-proxy` source code:

https://github.com/kubernetes/kubernetes/blob/e824c59913ffcb4e13a8d54afc97d8a3dd802076/cmd/kube-proxy/app/server.go#L712-L729

Perhaps someone else can have a crack at it.

vishal-biyani: Ok, looks like https://github.com/kubernetes/client-go/blob/v0.21.1/informers/factory.go#L96 should do it - we will need to test this ones and then we can merge changes from @irgeek followed by another small PR. @sanketsudake Can you please take this up since this is in area you are working on right now?

sanketsudake:
Moved these changes under https://github.com/fission/fission/pull/2061 and made a couple of additional changes by resolving review comments.

sanketsudake:
Closing this PR since changes moved to other PR. Please reopen in case needed.

PR #2014, opened 2021-05-07
Add preupgrade check for RPP

RealHarshThakur:

This PR adds a preupgrade check for the newer field RPP . A similar pattern can be followed whenever a new field is added

### DIFF
--- a/cmd/preupgradechecks/main.go
+++ b/cmd/preupgradechecks/main.go
@@ -71,5 +71,9 @@ Options:
 		return
 	}
 
+	err = crdBackedClient.LatestSchemaApplied()
+	if err != nil {
+		panic("New CRDs are not applied")
+	}
 	crdBackedClient.VerifyFunctionSpecReferences()
 }
--- a/cmd/preupgradechecks/main.go
+++ b/cmd/preupgradechecks/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"log"
+	"os"
 
 	"github.com/docopt/docopt-go"
 	"go.uber.org/zap"
@@ -73,7 +74,8 @@ Options:
 
 	err = crdBackedClient.LatestSchemaApplied()
 	if err != nil {
-		panic("New CRDs are not applied")
+		logger.Info("New CRDs are not applied")
+		os.Exit(1)
 	}
 	crdBackedClient.VerifyFunctionSpecReferences()
 }
--- a/cmd/preupgradechecks/main.go
+++ b/cmd/preupgradechecks/main.go
@@ -74,7 +74,7 @@ Options:
 
 	err = crdBackedClient.LatestSchemaApplied()
 	if err != nil {
-		logger.Info("New CRDs are not applied")
+		logger.Fatal("New CRDs are not applied")
 		os.Exit(1)
 	}
 	crdBackedClient.VerifyFunctionSpecReferences()
--- a/cmd/preupgradechecks/main.go
+++ b/cmd/preupgradechecks/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"log"
-	"os"
 
 	"github.com/docopt/docopt-go"
 	"go.uber.org/zap"
@@ -66,8 +65,8 @@ Options:
 			zap.Error(err))
 	}
 
-	installed := crdBackedClient.IsFissionReInstall()
-	if !installed {
+	crd := crdBackedClient.GetFunctionCRD()
+	if crd == nil {
 		logger.Info("nothing to do since CRDs are not present on the cluster")
 		return
 	}
@@ -75,7 +74,6 @@ Options:
 	err = crdBackedClient.LatestSchemaApplied()
 	if err != nil {
 		logger.Fatal("New CRDs are not applied")
-		os.Exit(1)
 	}
 	crdBackedClient.VerifyFunctionSpecReferences()
 }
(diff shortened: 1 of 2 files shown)


cmd/preupgradechecks/preupgradechecks.go
```diff
@@ -66,15 +67,33 @@ func makePreUpgradeTaskClient(logger *zap.Logger, fnPodNs, envBuilderNs string)
...
+	crd, err := client.apiExtClient.ApiextensionsV1().CustomResourceDefinitions().Get("functions.fission.io", metav1.GetOptions{})
+	if err != nil {
+		client.logger.Error("Could not get the CRD")
+		return err
+	}
+	// Any new field added in Function spec can be checked here provided the substring matches the description in CRD Validation of the field
+	if strings.Contains(crd.Spec.String(), "RequestsPerPod") {
+		return nil
+	}
+	client.logger.Error("Could not find the newer fields")
```
sanketsudake, line 95: Please add field and CRD details.
It could be like
`could not find  Spec.RequestPerPod in functions.fission.io CRD`

cmd/preupgradechecks/main.go
```diff
@@ -71,5 +71,9 @@ Options:
 		return
 	}
 
+	err = crdBackedClient.LatestSchemaApplied()
+	if err != nil {
+		panic("New CRDs are not applied")
```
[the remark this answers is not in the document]
RealHarshThakur: We need the program to fail with exit code 1 . We could use `error`, then use `os.Exit(1)`  . 

Reason is that if the preupgrade job fails(meaning exit code 1 ), only then fission will be not be upgraded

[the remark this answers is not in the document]
sanketsudake: I prefer exit codes over panic. Panic shows unexpected behaviour, and drops trace to user, errors assume known behaviour and program exits more gracefully.

cmd/preupgradechecks/preupgradechecks.go
RealHarshThakur: @sanketsudake Just thinking that we should probably keep this generic as there could be a lot of newer fields and it doesn't make sense to log about each of them . WDYT?

cmd/preupgradechecks/main.go
```diff
@@ -71,5 +72,10 @@ Options:
 		return
 	}
 
+	err = crdBackedClient.LatestSchemaApplied()
+	if err != nil {
+		logger.Info("New CRDs are not applied")
```
sanketsudake, line 77: Please use `logger.Fatal`.  It does error logging as well as exit.
Also  errors should small letter than capital letter.

```
logger.Fatal("new CRDs are not applied yet")
```

cmd/preupgradechecks/preupgradechecks.go
sanketsudake: I would probably prefer a specific error message to generic error messages.  Having a unique messages for error scenario is more preferred than generic as it makes easier. I have seen even developers going further and assigning a unique identifier to each error messages so that they easily match error logs and source code. This creates predictable behaviour.

cmd/preupgradechecks/preupgradechecks.go
```diff
@@ -66,15 +67,33 @@ func makePreUpgradeTaskClient(logger *zap.Logger, fnPodNs, envBuilderNs string)
 // IsFissionReInstall checks if there is at least one fission CRD, i.e. function in this case, on this cluster.
 // We need this to find out if fission had been previously installed on this cluster
 func (client *PreUpgradeTaskClient) IsFissionReInstall() bool {
```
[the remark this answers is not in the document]
sanketsudake:
```
		_, err := client.apiExtClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(FunctionCRD, metav1.GetOptions{})
		if err != nil && k8serrors.IsNotFound(err) {
                      continue
		}
                return true
	}
        return false
```

RealHarshThakur:
@sanketsudake Addressed the comments. PTAL

cmd/preupgradechecks/main.go
```diff
@@ -71,5 +72,10 @@ Options:
 		return
 	}
 
+	err = crdBackedClient.LatestSchemaApplied()
+	if err != nil {
+		logger.Fatal("New CRDs are not applied")
+		os.Exit(1)
```
sanketsudake, line 78: This `os.Exit` is redundant. Can we please remove it ?

cmd/preupgradechecks/preupgradechecks.go
```diff
@@ -69,14 +70,30 @@ func (client *PreUpgradeTaskClient) IsFissionReInstall() bool {
...
 	}
-
 	return false
 }
 
+// LatestSchemaApplied ensures that the end user has applied the latest CRDs generated to the cluster.
+// For future reference: whenever a new field is added, we need to check for that field's existence in this function
+func (client *PreUpgradeTaskClient) LatestSchemaApplied() error {
+	client.logger.Info("Checking if user has applied the latest CRDs")
+	crd, err := client.apiExtClient.ApiextensionsV1().CustomResourceDefinitions().Get("functions.fission.io", metav1.GetOptions{})
```
sanketsudake, line 84: This line is already getting called. Can we just rename `IsFissionReInstall` to `IsFisisionInstalled` and reuse function here ?
Also `functions.fission.io` is already defined with variable `FunctionCRD`

sanketsudake (approved):

PR #2015, opened 2021-05-09
Add Google Analytics reporting in place of existing custom reporting

sanketsudake:


charts/fission-all/templates/post-install-job.yaml
```diff
@@ -22,15 +22,20 @@ spec:
       labels:
         release: {{ .Release.Name }}
         app: {{ template "name" . }}
-      annotations:
-        {{- if .Values.enableIstio }}
-        "sidecar.istio.io/inject": "false"
-        {{- end }}
```
vishal-biyani, line 28: What is the intention of removing this one?

charts/fission-all/templates/post-upgrade-job.yaml
```diff
@@ -22,15 +22,20 @@ spec:
       labels:
         release: {{ .Release.Name }}
         app: {{ template "name" . }}
-      annotations:
-        {{- if .Values.enableIstio }}
-        "sidecar.istio.io/inject": "false"
-        {{- end }}
```
vishal-biyani, line 28: Similar to previous Istio annotation comment

charts/fission-core/templates/nonhelm-install-job.yaml
```diff
@@ -26,6 +26,16 @@ spec:
...
-          image: "fission/alpinecurl"
-          command: ["sh", "-c", "/usr/bin/curl -m 5 -H \"Content-Type: application/json\" -X POST -d '{\"type\": \"yaml-post-install\", \"chartName\": \"{{ .Chart.Name }}\", \"chartVersion\": \"{{ .Chart.Version }}\"}' https://g.fission.io/analytics || true"]
+        {{- if .Values.imageTag }}
+          image: {{ .Values.postInstallReportImage }}:{{ .Values.imageTag }}
+        {{- else }}
+          image: {{ .Values.postInstallReportImage }}
+        {{- end }}
+          imagePullPolicy: {{ .Values.pullPolicy }}
+          command: [ "/reporter" ]
+          args: ["event", "-c", "fission-use", "-a", "helm-post-install", "-l", "{{ .Chart.Name }}-{{ .Chart.Version }}"]
```
vishal-biyani, line 36: If this file is used for non helm install - then should the `helm-post-install` not be `yaml-post-install` or something similar?

hack/release-build.sh
```diff
@@ -85,6 +85,19 @@ build_builder_image() {
     docker tag $tag fission/builder:latest
 }
 
+# Build pre-upgrade-checks image
```
vishal-biyani, line 88: I think misplaced comment

test/test_utils.sh
```diff
@@ -96,6 +97,19 @@ build_and_push_pre_upgrade_check_image() {
...
+    image_tag=$1
+    cache_image=$2
+    travis_fold_start build_and_push_reporter_image $image_tag
+
+    docker build -q -t $image_tag -f $ROOT/cmd/reporter/Dockerfile.reporter --cache-from ${cache_image} --build-arg GITCOMMIT=$(getGitCommit) --build-arg BUILDDATE=$(getDate) --build-arg BUILDVERSION=$(getVersion) .
+
+    gcloud_login
+
+    gcloud docker -- push $image_tag &
+    travis_fold_end build_and_push_reporter_image
```
vishal-biyani, line 110: 1. We are not using the GCR and Travis both
2. I think we should not doing analytics part as part of test - that will skew results.

test/build.sh
```diff
@@ -24,6 +24,7 @@ setupCIBuildEnv
 build_and_push_go_mod_cache_image $REPO/go-mod-image-cache
 build_and_push_fission_bundle $REPO/$IMAGE:$TAG $REPO/go-mod-image-cache
 build_and_push_pre_upgrade_check_image $PRE_UPGRADE_CHECK_IMAGE:$TAG $REPO/go-mod-image-cache
+build_and_push_reporter_image $REPORTER_IMAGE:$TAG $REPO/go-mod-image-cache
```
vishal-biyani, line 27: I think not needed as part of build every time for test. Will increase build time too

skaffold.yaml
```diff
@@ -41,6 +44,7 @@ deploy:
         artifactOverrides:
           image: "fission"
           preUpgradeChecksImage: "preupgradechecks"
+          postInstallReportImage: "reporter"
```
vishal-biyani, line 47: Skaffold should turn analytics off.

vishal-biyani (changes requested):
Some very minor comments, LGTM otherwise

charts/fission-all/templates/post-install-job.yaml
sanketsudake: Added back, as I don't see any value of Istio injection in these post hooks.

sanketsudake:
@RealHarshThakur Could you please elaborate more on it ? Probably I don't have the context to the solution you are referring to. Any references you can point to from fission-bundle?

RealHarshThakur:
@sanketsudake  I mean we run a lot of components from Fission-Bundle itself(Router, Executor, MQT-KEDA) , so I was thinking if it made sense to keep reporter as part of it too.  https://github.com/fission/fission/blob/master/cmd/fission-bundle/main.go

PR #2018, opened 2021-05-10
Update k8s client go

mbilski:
When using the old k8s client the idle reaper object fails with `Unexpected watch close - watch lasted less than a second and no items received` and exists. After that fission stops removing specialized pods. You need to manually restart executor to make it do the cleanup.

Update to the latest version of 0.17.x k8s client-go fixes this problem.


sanketsudake:
Hi @mbilski,

Thanks for your contribution. We recently updated client-go to 1.19 and respective libraries as part of PR https://github.com/fission/fission/pull/2033

PR #2033, opened 2021-05-21
Crds update

RealHarshThakur:
- Updated code-generator from [upstream-1.19](https://github.com/fission/code-generator/tree/upstream-1.19)
- Updated client-go and respective libraries to v0.19.2
- Added manifests generated via controller-gen 
- Validations enabled via CRD manifests
- Changed Go version to 1.15
- Added kustomize for CRD installation


pkg/crd/crd.go
```diff
@@ -58,147 +61,12 @@ func ensureCRD(logger *zap.Logger, clientset *apiextensionsclient.Clientset, crd
 	return err
 }
 
-// EnsureFissionCRDs creates the CRDs
+// EnsureFissionCRDs defines all the CRDs that need to be ensured
 func EnsureFissionCRDs(logger *zap.Logger, clientset *apiextensionsclient.Clientset) error {
+	crds := []string{"canaryconfigs.fission.io", "packages.fission.io", "environments.fission.io", "functions.fission.io", "httptriggers.fission.io", "kuberneteswatchtriggers.fission.io", "messagequeuetriggers.fission.io", "timetriggers.fission.io"}
```
sanketsudake, line 66: We should have constants for these or populated from types.go directly.

pkg/crd/crd.go
```diff
@@ -18,58 +18,42 @@ package crd
...
-	maxRetries := 5
-
-	for i := 0; i < maxRetries; i++ {
-		_, err = clientset.ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), crd, metav1.GetOptions{})
-		if err == nil {
-			return nil
-		}
+// EnsureFissionCRDs checks if all Fission CRDs are present
+func EnsureFissionCRDs(logger *zap.Logger, clientset *apiextensionsclient.Clientset) error {
+	crds, err := clientset.ApiextensionsV1().CustomResourceDefinitions().List(context.TODO(), metav1.ListOptions{})
```
RealHarshThakur, line 31: Is it not simpler to just `Get()` each CRD as we have the names stored anyway in `crdsExpected`?  If the `Get()` call fails , we can append to `multierrror`

kind.yaml
```diff
@@ -8,7 +8,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.16.15
+  image: kindest/node:v1.19.11
```
RealHarshThakur, line 11: What does this mean for minimum version of Fission? 
Only one test failed on 1.16, so we could still say we support 1.16?

sanketsudake: Fission installation already says 1.19+ Kubernetes version. So we won't support any issues raised by users, user can run if they don't have an option but officially we won't have the bandwidth to support it.

image

pkg/crd/crd.go
sanketsudake: LIst call fetches all CRDs in go compared to `Get` calls. So preferred it. Since we not apply CRDs via `kubectl create` Its expected user should have applied it, if not user just need to restart pod otherwise it would be in CrashLoop.

RealHarshThakur:
LGTM! Can't approve as I've commits in the PR too

pkg/crd/crd.go
RealHarshThakur: Users could have a lot of CRDs in the cluster, in which case a `Get` call could be cheaper. Is it possible to use a field selector or label selector in the `List()` to narrow down the fission CRDs and just count if there are 8 of them?

sanketsudake (approved):

PR #2036, opened 2021-05-23
Add router cache back for performance comparison

sanketsudake:


RealHarshThakur:
@sanketsudake  just wondering how much latency this is decreasing. Have you checked that with Linkerd, etc?

vishal-biyani:
LGTM

PR #2047, opened 2021-05-31
Enable prefix based routing

RealHarshThakur:


pkg/router/functionHandler.go
```diff
@@ -220,13 +222,22 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
...
-			// doesn't do any routing.  In the future if we have
-			// multiple functions per container, we could use the
-			// function metadata here.
-			// leave the query string intact (req.URL.RawQuery)
-			req.URL.Path = "/"
-
+			klog.Info(req.URL.Path)
+			functionNamespacedURL := "/" + fnMeta.Namespace + "/" + fnMeta.Name
+			defaultNamespacedURL := "/fission-function/" + fnMeta.Name
+			klog.Info(functionNamespacedURL)
```
sanketsudake, line 228: Please use round tripper existing logger.

pkg/router/functionHandler.go
```diff
@@ -220,13 +221,18 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
...
 			req.URL.Host = roundTripper.serviceURL.Host
 
-			// To keep the function run container simple, it
-			// doesn't do any routing.  In the future if we have
-			// multiple functions per container, we could use the
-			// function metadata here.
-			// leave the query string intact (req.URL.RawQuery)
-			req.URL.Path = "/"
-
+			functionNamespacedURL := "/" + fnMeta.Namespace + "/" + fnMeta.Name
```
sanketsudake, line 224: @RealHarshThakur Can you please comments on respective blocks ? We should highlight `why` part considering conditional flow here.

sanketsudake: Wait does infinite wait, if we are stuck in package build times would increase and may keep running for a longer duration. Probably need to have wait with some timeout. 
I can try this as a separate PR where a fix would be applicable across all tests.

pkg/router/functionHandler.go
```diff
@@ -221,18 +221,19 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
...
-			defaultNamespacedURL := "/fission-function/" + fnMeta.Name
-			if strings.HasPrefix(req.URL.Path, functionNamespacedURL) || strings.HasPrefix(req.URL.Path, defaultNamespacedURL) {
-				req.URL.Path = "/"
-			} else {
-				if roundTripper.funcHandler.httpTrigger.Spec.Prefix != nil {
+			if roundTripper.funcHandler.httpTrigger != nil && roundTripper.funcHandler.httpTrigger.Spec.Prefix != nil {
+				functionNamespacedURL := "/" + fnMeta.Namespace + "/" + fnMeta.Name
+				defaultNamespacedURL := "/fission-function/" + fnMeta.Name
+				if strings.HasPrefix(req.URL.Path, functionNamespacedURL) || strings.HasPrefix(req.URL.Path, defaultNamespacedURL) {
+					req.URL.Path = "/"
```
RealHarshThakur, line 228: We need this block outside the if block as all functions might not have http triggers.

pkg/controller/httpTriggerApi.go
```diff
@@ -132,7 +132,11 @@ func (a *API) checkHTTPTriggerDuplicates(t *fv1.HTTPTrigger) error {
 			// Same resource. No need to check.
 			continue
 		}
-		if ht.Spec.RelativeURL == t.Spec.RelativeURL && ht.Spec.Method == t.Spec.Method && ht.Spec.Host == t.Spec.Host {
+		urlMatch := false
+		if ht.Spec.RelativeURL == t.Spec.RelativeURL || (ht.Spec.Prefix != nil && t.Spec.Prefix != nil && *ht.Spec.Prefix != "" && *ht.Spec.Prefix == *t.Spec.Prefix) {
```
RealHarshThakur, line 136: Do we need to check if the value is an empty string? I don't think it's possible to run into that scenario because of the `omitempty` json struct tag, it is supposed to remove the field(set it as nil) if the value is empty string

pkg/router/functionHandler.go
```diff
@@ -220,12 +221,20 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 			req.URL.Scheme = roundTripper.serviceURL.Scheme
 			req.URL.Host = roundTripper.serviceURL.Host
 
-			// To keep the function run container simple, it
-			// doesn't do any routing.  In the future if we have
-			// multiple functions per container, we could use the
-			// function metadata here.
-			// leave the query string intact (req.URL.RawQuery)
-			req.URL.Path = "/"
+			if roundTripper.funcHandler.httpTrigger != nil && roundTripper.funcHandler.httpTrigger.Spec.Prefix != nil && *roundTripper.funcHandler.httpTrigger.Spec.Prefix != "" {
```
RealHarshThakur, line 224: Same question as above regarding empty strings check

sanketsudake: Sane as above. 
Just trying to explicit here, as after adding these checks tests started passing without a timeout. Avoiding assumption and possible code failure.

pkg/fission-cli/cmd/httptrigger/create.go
```diff
@@ -95,13 +104,10 @@ func (opts *CreateSubCommand) complete(input cli.Input) error {
...
 	prefix := input.String(flagkey.HtPrefix)
-	if len(prefix) != 0 && len(triggerUrl) > 2 {
-		console.Warn("Prefix will take precedence over URL/RelativeURL")
-	}
-
-	if len(prefix) == 0 && len(triggerUrl) < 2 {
-		console.Error("You need to supply either Prefix or URL/RelativeURL")
-		os.Exit(1)
+	if prefix == "/" {
+		return errors.New("url with only root path is not allowed")
```
RealHarshThakur, line 108: What will happen if we allow URL with root path prefix?

sanketsudake: It may give entire path space to function and by default, we also have internal routes.

pkg/fission-cli/flag/flag.go
```diff
@@ -113,6 +113,7 @@ var (
 	FnConcurrency           = Flag{Type: Int, Name: flagkey.FnConcurrency, Aliases: []string{"con"}, Usage: "Maximum number of pods specialized concurrently to serve requests", DefaultValue: 500}
 	FnRequestsPerPod        = Flag{Type: Int, Name: flagkey.FnRequestsPerPod, Aliases: []string{"rpp"}, Usage: "Maximum number of concurrent requests that can be served by a specialized pod", DefaultValue: 1}
 	FnOnceOnly              = Flag{Type: Bool, Name: flagkey.FnOnceOnly, Aliases: []string{"yolo"}, Usage: "Specifies if specialized pod will serve exactly one request in its lifetime"}
+	FnSubUrl                = Flag{Type: String, Name: flagkey.FnSubUrl, Usage: "Sub URL to check if function internally support routing"}
```
vishal-biyani, line 116: I did not understand the role of this and description is not clear?

crds/v1/fission.io_httptriggers.yaml
```diff
@@ -79,12 +79,14 @@ spec:
...
                 type: string
+              prefix:
+                description: 'Prefix with which functions are exposed. NOTE: Prefix takes precedence over URL/RelativeURL. Prefix takes precendence over RelativeURL Note that it does not treat slashes specially ("/foobar/" will be matched by the prefix "/foo") so you may want to use a trailing slash here.'
+                type: string
               relativeurl:
                 description: RelativeURL is the exposed URL for external client to access a function with.
                 type: string
             required:
             - functionref
-            - relativeurl
```
vishal-biyani, line 87: I don't know if there is a way in CRD validation to say that one of the `relativeURL` or `prefix` is needed?

vishal-biyani (changes requested):
A few minor comments - let's get them in please!

pkg/fission-cli/flag/flag.go
sanketsudake: When user is caling `fission function test` he can only hit root endpoint of function but he wants to hit any other routes he can pass sub URL
eg.
```
$ fission function test --name nextjs-function

$ fission function test --name nextjs-function --suburl /api/hello
```

pkg/apis/core/v1/validation.go
```diff
@@ -417,6 +417,12 @@ func (spec HTTPTriggerSpec) Validate() error {
 		}
 	}
 
+	if spec.Prefix != nil && *spec.Prefix != "" && !strings.HasPrefix(*spec.Prefix, "/") {
+		result = multierror.Append(result, fmt.Errorf("leading / missing in HTTPTriggerSpec.Prefix"))
+	}
+	if spec.RelativeURL != "" && !strings.HasPrefix(spec.RelativeURL, "/") {
+		result = multierror.Append(result, fmt.Errorf("leading / missing in HTTPTriggerSpec.RelativeURL"))
+	}
```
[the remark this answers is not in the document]
sanketsudake: Removed this validation. I don't find the immediate necessity of. this, mux router can handle both.

pkg/fission-cli/flag/flag.go
RealHarshThakur: @sanketsudake can we rename it to sub path? Sub url seems misleading as URL implies protocol+address

pkg/fission-cli/cmd/httptrigger/create.go
RealHarshThakur: If we have a function hello which has endpoints `/home`, `/world` and user wants to expose it as `http://IP/home` / `http://IP/world`, then it would be possible if we kept the prefix as `/` . Internal route of the function would be `http://IP/Namespace/hello`

PR #2050, opened 2021-06-01
Added fission upgrade workflow

jithindevasia:


test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,96 @@
...
+      echo "Test failed"
+    fi
+}
+
+build_docker_images () {
+    echo "Building new docker images"
+    docker build -t fission-bundle -f cmd/fission-bundle/Dockerfile.fission-bundle .
+    docker build -t fetcher -f cmd/fetcher/Dockerfile.fission-fetcher .
+    docker build -t builder -f cmd/builder/Dockerfile.fission-builder .
+    docker build -t reporter -f cmd/reporter/Dockerfile.reporter .
```
RealHarshThakur, line 70: We need to build the preupgrade image too

test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,96 @@
...
+    go version
+    docker version
+    kubectl version
+    helm version
+}
+
+install_stable_release () {
+    echo "Creating namespace $ns"
+    kubectl create ns $ns
+    echo "Installing Fission $STABLE_VERSION"
```
RealHarshThakur, line 31: From 1.13 , we need to create CRDs before installing. Maybe we can have an if statement here

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,54 @@
...
+        config: kind.yaml
+        
+    - name: Setup kubectl & fetch node information
+      run: |
+        kubectl cluster-info --context kind-kind
+        kind get kubeconfig --internal >$HOME/.kube/config
+        kubectl get nodes
+        sudo apt-get install -y apache2-utils
+
+    - name: Configure fission and run upgrade test
```
RealHarshThakur, line 52: Just wondering if it's possible to do each of these in an individual test:
* Install fission
* Create fission objects(functions, environments, http triggers, mqt, etc)
* Upgrade fission 
* Test the objects

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+      with:
+        go-version: '1.15.12'
+
+    - name: Setup Helm
+      uses: Azure/setup-helm@v1
+      with:
+        version: v3.3.4
+    
+    - name: Update helm
+      run: helm repo add helm https://charts.helm.sh/stable
```
sanketsudake, line 32: This could specific to releases before 1.13. With the latest helm, the stable repo is now deprecated. If we still need this, we should move it `install_stable_release` function.

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+    - name: Setup Helm
+      uses: Azure/setup-helm@v1
+      with:
+        version: v3.3.4
+    
+    - name: Update helm
+      run: helm repo add helm https://charts.helm.sh/stable 
+
+    - name: Setup Skaffold
```
sanketsudake, line 34: We can remove `skaffold` as we are not using it in this flow.

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+jobs:
+  
+  upgrade-functionality-test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout action sources
+      uses: actions/checkout@v2.3.4
+
+    - name: Checkout fission sources
```
sanketsudake, line 15: We should not clone examples repo, as we need to single file from repo only later and it's fetched via curl later.

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,71 @@
...
+      uses: engineerd/setup-kind@v0.5.0
+      with:
+        config: kind.yaml
+        
+    - name: Setup kubectl & fetch node information
+      run: |
+        kubectl cluster-info --context kind-kind
+        kind get kubeconfig --internal >$HOME/.kube/config
+        kubectl get nodes
+        sudo apt-get install -y apache2-utils
```
sanketsudake, line 50: Probably we don't need apache2-utils in this flow.

test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+set -e
+
+ns="f-ns"
+ROOT=$(pwd)
+REPO="docker.io/library"
+STABLE_VERSION=1.12.0
+HELM_VARS="helmVars=repository=docker.io/library,image=fission-bundle,pullPolicy=IfNotPresent,imageTag=latest,fetcher.image=docker.io/library/fetcher,fetcher.imageTag=latest,postInstallReportImage=reporter,preUpgradeChecksImage=preupgradechecks"
```
sanketsudake, line 8: We should two separate vars. If only one is need we should rename this one.
```
HELM_VARS_PREV_RELEASE
HELM_VARS_LATEST_RELEASE
```

test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,111 @@
...
+
+getDate () {
+    echo $(date -u +'%Y-%m-%dT%H:%M:%SZ')
+}
+
+getGitCommit () {
+    echo $(git rev-parse HEAD)
+}
+
+getNameSpace () {
```
sanketsudake, line 22: Unused function.

test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+set -e
+
+ns="f-ns"
+ROOT=$(pwd)
+REPO="docker.io/library"
+STABLE_VERSION=1.12.0
```
sanketsudake, line 7: Please rename to `PREV_STABLE_VERSION`

test/upgrade_test/fission_objects.sh
```diff
@@ -0,0 +1,111 @@
...
+install_stable_release () {
+    echo "Creating namespace $ns"
+    kubectl create ns $ns
+    echo "Installing Fission $STABLE_VERSION"
+    helm install \
+    --namespace $ns \
+    --name-template fission \
+    https://github.com/fission/fission/releases/download/${STABLE_VERSION}/fission-all-${STABLE_VERSION}.tgz
+    mkdir temp && cd temp && curl -Lo fission https://github.com/fission/fission/releases/download/${STABLE_VERSION}/fission-cli-linux && chmod +x fission && sudo mv fission /usr/local/bin/ && cd .. && rm -rf temp
+    sleep 10
```
sanketsudake, line 43: Why do we need sleep here? If not needed we should remove it.

Please add comments/reason in code wherever we are using `sleep`.

.github/workflows/upgrade_test.yaml
jithindevasia: Removed examples repo from workflow.

.github/workflows/upgrade_test.yaml
jithindevasia: Removed skaffold installation from workflow.

.github/workflows/upgrade_test.yaml
jithindevasia: Removed apache.

test/upgrade_test/fission_objects.sh
jithindevasia: Renamed version variable as requested.

test/upgrade_test/fission_objects.sh
jithindevasia: Removed this function since I added it for testing purpose.

test/upgrade_test/fission_objects.sh
jithindevasia: We don't use any extra vars for stable installation but I have renamed the existing variable to HELM_VARS_LATEST_RELEASE.

test/upgrade_test/fission_objects.sh
jithindevasia: This sleep is required here to become all pods active. Have added comments  in script.

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,55 @@
...
+        
+    - name: Dump system info
+      run: |
+        source ./test/upgrade_test/fission_objects.sh dump_system_info
+
+    - name: Install and configure previous stable fission
+      run: |
+        source ./test/upgrade_test/fission_objects.sh install_stable_release \
+        && create_fission_objects \
+        && test_fission_objects
```
RealHarshThakur, line 44: We don't need to test fission objects here as it's the stable one.

.github/workflows/upgrade_test.yaml
jithindevasia: Removed helm upgrade and tested the workflow, nothing breaks.

.github/workflows/upgrade_test.yaml
jithindevasia: We need to make sure that the functions created in stable version works before upgrading even if it stable.

.github/workflows/upgrade_test.yaml
```diff
@@ -0,0 +1,55 @@
+name: Upgrade functionality test
+
+on:
```
sanketsudake, line 3: Please change this to

```
on:
  workflow_dispatch:
```

We would merge this workflow with master. Currently we don't want to run it over every commit so we can run it manually. Once we are sure we can enable for push to `master` branch.

Once this change is done, we are good to merge.

jithindevasia: @sanketsudake Added workflow dispatch.

sanketsudake (approved):

sanketsudake:
Thanks @jithindevasia for your PR 🎉

PR #2052, opened 2021-06-02
Wait for small time to avoid timeout in specs tests

sanketsudake:

### DIFF
--- a/test/tests/test_specs/test_spec_merge/test_spec_merge.sh
+++ b/test/tests/test_specs/test_spec_merge/test_spec_merge.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -euo pipefail
+source $(dirname $0)/../../../utils.sh
 
 env_p=nodep
 fn_p=nodehellop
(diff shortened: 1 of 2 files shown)


test/tests/test_specs/test_spec_merge/test_spec_merge.sh
```diff
@@ -19,19 +20,23 @@ pushd $(dirname $0)
 
 fission spec apply
```
RealHarshThakur, line 21: Can we use the --wait flag in this command instead of sleep later?

test/tests/test_specs/test_spec_multifile.sh
```diff
@@ -40,6 +40,9 @@ fission fn create --spec --name $fn --env $env --deploy "multifile/*" --entrypoi
 log "Applying specs"
 fission spec apply
```
RealHarshThakur, line 41: Can we use the --wait flag in this command instead of sleep later?

test/tests/test_specs/test_spec_merge/test_spec_merge.sh
sanketsudake: Wait does infinite wait, if we are stuck in package build times would increase and may keep running for a longer duration. Probably need to have wait with some timeout.
I can try this as a separate PR where a fix would be applicable across all tests.

test/tests/test_specs/test_spec_multifile.sh
sanketsudake: same as above.

RealHarshThakur (approved):

Issue #2057, opened 2021-06-06
Change rbac references from v1beta1 to v1


sanketsudake:
**Describe the bug**
Warning in executor log
```
 W0605 14:01:57.445456       1 warnings.go:67] rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
```
```
pkg/utils/rbacutils.go:27:	rbac "k8s.io/api/rbac/v1beta1"
```

**To Reproduce**
Observe the Fission executor log, the above warning, can be noticed.

** Fix
Change RBAC API references from v1beta to v1

PR #2059, opened 2021-06-07
Move RBAC v1beta1 to v1

RealHarshThakur:
Fixes https://github.com/fission/fission/issues/2057


vishal-biyani:
Once we make this change, I am assuming Fission won't work for 1.16 or lower versions of K8S (The RBAC was deprecated in 1.17). Can you please confirm which version we will be able to support?

RealHarshThakur:
@vishal-biyani  With the recent client-go upgrade PR in the master, we're supporting 1.18+

sanketsudake (approved):

PR #2060, opened 2021-06-07
Add metrics to Router and Executor

RealHarshThakur:
Adding requests recieved metric at Router and function uid label to all metrics of 

We can now find the the number of requests received at Router before any processing with fission_requests_received

 We can find out the number of pending/failed requests by querying Prometheus like this: sum(fission_requests_received{name="hello-1-15"}) - sum(fission_function_calls_total{name="hello-1-

We can differentiate between functions which were deleted and re-deployed with same name with help of function uid label

 Added a metric to measure specialization 

Added a metric to measure the time taken to get service URL from Executor

Fixed few golint warnings

To find average request duration, we can :
rate(fission_function_duration_seconds_sum [30m])/rate(fission_function_duration_seconds_count[30m])


blackfly19:
Already merged in #2398 and #2399 .

PR #2064, opened 2021-06-10
Support for multiple HTTP verbs in routes/HTTPTrigger

sanketsudake:

### DIFF
--- a/pkg/apis/core/v1/types.go
+++ b/pkg/apis/core/v1/types.go
@@ -658,9 +658,14 @@ type (
 		// +optional
 		Prefix *string `json:"prefix,omitempty"`
 
+		// Deprecated: Use Methods instead of Method.
 		// HTTP method to access a function.
 		// +optional
 		Method string `json:"method"`
+
+		// HTTP methods to access a function
+		// +optional
+		Methods []string `json:"methods"`
 
 		// FunctionReference is a reference to the target function.
 		FunctionReference FunctionReference `json:"functionref"`
--- a/pkg/apis/core/v1/types.go
+++ b/pkg/apis/core/v1/types.go
@@ -665,7 +665,7 @@ type (
 
 		// HTTP methods to access a function
 		// +optional
-		Methods []string `json:"methods"`
+		Methods []string `json:"methods,omitempty"`
 
 		// FunctionReference is a reference to the target function.
 		FunctionReference FunctionReference `json:"functionref"`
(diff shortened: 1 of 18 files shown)


pkg/apis/core/v1/types.go
```diff
@@ -650,10 +650,15 @@ type (
...
 		RelativeURL string `json:"relativeurl"`
 
+		// Deprecated: Use Methods instead of Method.
 		// HTTP method to access a function.
 		// +optional
 		Method string `json:"method"`
 
+		// HTTP methods to access a function
+		// +optional
+		Methods []string `json:"methods"`
```
RealHarshThakur, line 660: Optional fields should have omit empty tags 
```suggestion
		Methods []string `json:"methods,omitempty"`
```

sanketsudake: This is intentional, just trying to avoid nil checks as well in future this would be a compulsory or non-nil field once we remove Method.

RealHarshThakur: I think we can remove the optional tag in the next API version and then it would be okay to not have the `omitempty` tag there but in this version as we're marking it as optional , it's recommended to have the tag. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

pkg/fission-cli/flag/flag.go
```diff
@@ -116,7 +116,7 @@ var (
 	FnSubPath               = Flag{Type: String, Name: flagkey.FnSubPath, Usage: "Sub Path to check if function internally supports routing"}
 
 	HtName              = Flag{Type: String, Name: flagkey.HtName, Usage: "HTTP trigger name"}
-	HtMethod            = Flag{Type: String, Name: flagkey.HtMethod, Usage: "HTTP Method: GET|POST|PUT|DELETE|HEAD", DefaultValue: http.MethodGet}
+	HtMethod            = Flag{Type: StringSlice, Name: flagkey.HtMethod, Usage: "HTTP Methods: GET,POST,PUT,DELETE,HEAD. To mention single method: --method GET and for multiple methods --method GET --method POST.", DefaultValue: []string{http.MethodGet}}
```
RealHarshThakur, line 119: Can we have a `Methods` flag and deprecate the `Method` flag here? It would be nicer from user perspective to just do : `--methods=GET,POST`

sanketsudake: This would be a wider changes if we change name of flag. Since this is targeted towards limited users as of now, I think we should be good.
Even if we get specific request of `GET,POST` kind of syntax we can support it within existing flag. Where both `GET,POST` and `--method GET --method POST` formats can be supported. So I dont see need of overriding existing flags. It can accept existing as well as two formats discussed.

Issue #2042, opened 2021-05-27
Enable multiples kubernetes version in testing in Github CI flow


sanketsudake:
**Is your feature request related to a problem? Please describe.**
Currently, in the existing CI flow, we are able to verify only one Kubernetes version at a time, based on [kind config ](https://github.com/fission/fission/blob/master/kind.yaml). We need support for multiple Kubernetes version verification in the same flow.

**Describe the solution you'd like**
We should be able to run CI flow for multiple versions.
Eg. If we want to verify 1.19 & 1.21 in the existing CI flow we should be able to that.

- We need a way to mention a list of  Kubernetes version against which we would like to verify Fission.
- Verification of Kubernetes version and tests should run in parallel.
- List of versions would be hardcoded in CI flow and we can change it as needed, depending on deprecation or new Kubernetes releases.
- Build time should not get impacted as we are verifying multiple versions, since verification should be done in parallel.
- These fixes might require changes in test scripts and the way we access Kubernetes context currently.

**Additional context**
-

vishal-biyani:
@sanketsudake Captured here already: https://github.com/fission/fission/issues/1603

hellozee:
Feel free to copy workflows from [here](https://github.com/tetratelabs/istio)

sanketsudake:
Closing this as we are able to verify multiple Kubernetes versions with CI.

PR #2082, opened 2021-06-20
Add multi k8s version and os support in ci tests

suramya-shah:
As mentioned in [2042](https://github.com/fission/fission/issues/2042) there is only single os and k8s version support in the integration tests workflow:
image

The proposed change adds multiple k8s version tests support to the integration tests along with multi os version support:

image

### DIFF
--- a/.github/workflows/push_pr.yaml
+++ b/.github/workflows/push_pr.yaml
@@ -3,11 +3,20 @@ name: Fission CI
 on:
   push:
     branches:
-      - master   
+      - master
+    paths:
+      - '**.go'
+      - 'charts/**'
+      - 'test/**'
   pull_request:
     branches:
       - master
+    paths:
+      - '**.go'
+      - 'charts/**'
+      - 'test/**'
   workflow_dispatch:
+
 jobs:
   # Job to run change detection
   integration-test:
@@ -36,7 +45,7 @@ jobs:
         uses: Azure/setup-helm@v1
         with:
           version: v3.3.4
-          
+
       - name: Kind Clutser
         uses: engineerd/setup-kind@v0.5.0
         with:
--- a/.github/workflows/push_pr.yaml
+++ b/.github/workflows/push_pr.yaml
@@ -41,7 +41,7 @@ jobs:
         uses: engineerd/setup-kind@v0.5.0
         with:
           image:  ${{ matrix.kindimage }}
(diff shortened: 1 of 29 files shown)


.github/workflows/push_pr.yaml
```diff
@@ -3,16 +3,20 @@ name: Fission CI
...
 jobs:
   # Job to run change detection
   integration-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest , ubuntu-20.04, ubuntu-18.04] 
+        kindimage : [ 'kindest/node:v1.19.11@sha256:7664f21f9cb6ba2264437de0eb3fe99f201db7a3ac72329547ec4373ba5f5911','kindest/node:v1.20.7@sha256:e645428988191fc824529fd0bb5c94244c12401cf5f5ea3bd875eb0a787f0fe9',  'kindest/node:v1.21.1@sha256:fae9a58f17f18f06aeac9772ca8b5ac680ebbed985e266f711d936e91d113bad' ]
```
sanketsudake, line 19: Kind images should we mention in simple format?

```
kindimage : [ 'kindest/node:v1.19.11', 'kindest/node:v1.20.7', 'kindest/node:v1.21.1' ]
```

Any reason for choosing different OS ? Can we just use `ubuntu-latest` instead ?


Also I think different kubernets version wont take effect as below kind cluster step has kind version hardcoded.

```
      - name: Kind Clutser
        uses: engineerd/setup-kind@v0.5.0
        with:
          config: kind.yaml
```
Reference kind.yaml,
https://github.com/fission/fission/blob/3f99b483cb65fd0d01a6591da7d626ccf6c77bcf/kind.yaml#L11

sanketsudake (changes requested):

.github/workflows/push_pr.yaml
suramya-shah: Hi @sanketsudake, 
1- as per https://github.com/kubernetes-sigs/kind/releases it is advised to use exact images including digest.
2- thanks sanket, looks like missed the change adding now.

sanketsudake (approved):

sanketsudake:
Thanks, @suramya-shah for your PR 🎉

PR #2092, opened 2021-06-25
Shared informers

sanketsudake:
This PR replaces existing controllers with SharedIndexInformers. 
PR helps us reduce the overall memory footprint and optimize the number of controllers we create via executor.
In the executor, we got a reduction in no informers we run as we are adding different executor types as listener and informer is shared across executor types.
As we are using typed informers at most of the places helps to take advantage of the Go type system at static checks.

Also long-pending cleanup for controllers and improved code readability 😄


pkg/executor/executortype/poolmgr/funchandlers.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/fission/fission/pkg/utils"
+)
+
+func getIstioServiceLabels(fnName string) map[string]string {
+	return map[string]string{
+		"functionName": fnName,
+	}
+}
+
+func (gpm *GenericPoolManager) FunctionEventHandlers(kubernetesClient *kubernetes.Clientset, fissionfnNamespace string, istioEnabled bool) k8sCache.ResourceEventHandlerFuncs {
```
RealHarshThakur, line 40: Would it be possible to have a controller for environments too? We wouldn't need the cache then, I think.
Probably out of scope for this PR but just wondering since all fission objects are managed this way now.

sanketsudake: Yes, I am thinking of the same. Having a controller for environments would help. There would be one common informer across all executor.

RealHarshThakur (approved):
LGTM

gauravgahlot (approved):
LGTM :rocket:

PR #2098, opened 2021-07-02
Enable race detector in integration tests

sanketsudake:


sanketsudake:
Need to re-explore this integration as we are using goreleaser.

PR #2113, opened 2021-07-14
Labels and annotation support for functions and environments

sanketsudake:
This PR adds,

1. Support for adding labels and annotations to fuctions & environment via fission CLI thrrough create & update command.

2. Change ensures labels and annotations assigned to environment would reflect on pods created via executortype poolmanager and newdeploy.

Usage:

```sh

fission env create --name nodejs --image fission/node-env --poolsize 1 --labels="env=node,author=sanket" \
   --annotation "abc.com/foo=bar" --annotation "secret-inject: my-secret"

fission env update --name nodejs --labels="env=node,author=sanket,another-label=value"

```


sanketsudake:
Verified with both specs and CLI. Also verified with kubectl, labels & annotations are getting assigned to function pod.

PR #2129, opened 2021-07-27
Fetcher remove special readiness check for pod specialize on start

sanketsudake:

### DIFF
--- a/cmd/fetcher/app/server.go
+++ b/cmd/fetcher/app/server.go
@@ -24,7 +24,6 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


vishal-biyani:
@sanketsudake I think this will cause error conditions which will be difficult to track. Unless the specialisation is done - in case of newdeploy - the pod is not ready to load the code and execute. So I believe this check should stay as it is.

sanketsudake:
This PR does give the benefit of responding faster in case min scale 0 and max scale 1. But in the case of multiple new deploy pods, we face an issue where requests are routed via service to non-ready pods. Probably we need to rethink strategy multiple retries from the router and make it better based on watch. 
We cant trade off accuracy for speed so dropping off PR.

PR #2156, opened 2021-08-08
Enable user to toggle OpenTracing

gauravgahlot:


charts/fission-all/templates/router.yaml
```diff
@@ -55,8 +55,10 @@ spec:
             value: {{ .Values.router.svcAddressUpdateTimeout | default "30s" | quote }}
           - name: ROUTER_UNTAP_SERVICE_TIMEOUT
             value: {{ .Values.router.unTapServiceTimeout | default "3600s" | quote }}
+          - name: OPENTRACING_ENABLED
+            value: {{ .Values.openTracing.enabled | default false | quote }}
           - name: TRACE_JAEGER_COLLECTOR_ENDPOINT
-            value: "{{ .Values.traceCollectorEndpoint }}"
+            value: "{{ .Values.openTracing.collectorEndpoint }}"
```
sanketsudake, line 61: Trailing whitespace

cmd/fetcher/app/server.go
```diff
@@ -116,12 +118,15 @@ func Run(logger *zap.Logger) {
...
 
 	mux := http.NewServeMux()
-	mux.HandleFunc("/fetch", f.FetchHandler)
-	mux.HandleFunc("/specialize", f.SpecializeHandler)
-	mux.HandleFunc("/upload", f.UploadHandler)
-	mux.HandleFunc("/version", f.VersionHandler)
-	mux.HandleFunc("/wsevent/start", f.WsStartHandler)
-	mux.HandleFunc("/wsevent/end", f.WsEndHandler)
+
+	if openTracingEnabled {
```
sanketsudake, line 122: We are disabling these routes if `openTracingEnabled` is not set ? Probably doesn't seem right to me. We should have these routes enabled even if `openTracingEnabled` is set to false.

cmd/fetcher/app/server.go
```diff
@@ -137,11 +142,13 @@ func Run(logger *zap.Logger) {
 	})
 
 	logger.Info("fetcher ready to receive requests")
-	err = http.ListenAndServe(":8000", &ochttp.Handler{
-		Handler: mux,
-	})
-	if err != nil {
-		log.Fatal(err)
+	if openTracingEnabled {
+		err = http.ListenAndServe(":8000", &ochttp.Handler{
```
sanketsudake, line 146: Same as above, if we openTracingEnabled is set to false, fetcher wont run HTTP server.

pkg/controller/api.go
```diff
@@ -194,9 +194,19 @@ func (api *API) GetSvcName(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, service.Name+"."+podNamespace)
 }
 
-func (api *API) GetHandler() http.Handler {
+func (api *API) GetHandler(openTracingEnabled bool) http.Handler {
 	r := mux.NewRouter()
+	if openTracingEnabled {
+		api.registerHandlers(r)
```
sanketsudake, line 200: Same as above, if openTracing is disabled all handles should work.

pkg/controller/api.go
```diff
@@ -257,15 +267,14 @@ func (api *API) GetHandler() http.Handler {
...
 
-func (api *API) Serve(port int) {
+func (api *API) Serve(port int, openTracingEnabled bool) {
 	address := fmt.Sprintf(":%v", port)
 	api.logger.Info("server started", zap.Int("port", port))
-	err := http.ListenAndServe(address, api.GetHandler())
-	api.logger.Fatal("done listening", zap.Error(err))
+
+	if openTracingEnabled {
+		err := http.ListenAndServe(address, api.GetHandler(openTracingEnabled))
```
sanketsudake, line 277: This should work with openTracingEnabled is false as well.

pkg/executor/api.go
```diff
@@ -240,22 +240,29 @@ func (executor *Executor) unTapService(w http.ResponseWriter, r *http.Request) {
...
 // GetHandler returns an http.Handler.
-func (executor *Executor) GetHandler() http.Handler {
+func (executor *Executor) GetHandler(openTracingEnabled bool) http.Handler {
 	r := mux.NewRouter()
-	r.HandleFunc("/v2/getServiceForFunction", executor.getServiceForFunctionAPI).Methods("POST")
-	r.HandleFunc("/v2/tapService", executor.tapService).Methods("POST") // for backward compatibility
-	r.HandleFunc("/v2/tapServices", executor.tapServices).Methods("POST")
+
+	if openTracingEnabled {
+		r.HandleFunc("/v2/getServiceForFunction", executor.getServiceForFunctionAPI).Methods("POST")
```
sanketsudake, line 247: Same as above.

pkg/storagesvc/storagesvc.go
```diff
@@ -199,24 +199,30 @@ func MakeStorageService(logger *zap.Logger, storageClient *StowClient, port int)
...
 
-func (ss *StorageService) Start(port int) {
+func (ss *StorageService) Start(port int, openTracingEnabled bool) {
 	r := mux.NewRouter()
-	r.HandleFunc("/v1/archive", ss.uploadHandler).Methods("POST")
-	r.HandleFunc("/v1/archive", ss.downloadHandler).Methods("GET")
-	r.HandleFunc("/v1/archive", ss.deleteHandler).Methods("DELETE")
+
+	if openTracingEnabled {
+		r.HandleFunc("/v1/archive", ss.uploadHandler).Methods("POST")
```
sanketsudake, line 206: Same as above.

sanketsudake (changes requested):

cmd/fetcher/app/server.go
gauravgahlot: Fission will use OTEL by default instead of OpenTracing. So, until OpenTracing is explicitly enabled, all the routes will be registered via OTEL handler. For reference:

https://github.com/fission/fission/pull/2157/files#diff-5fb1bf8b708338b3c4389340f32523ea575d86a33b1503071cc37a66b52e91d5R190

gauravgahlot:
Closing in favor of #2157. It provides a much clear picture of what we are trying to achieve here. 
Also, the review feedback will be addressed in #2157 only.

Issue #1861, opened 2020-11-18
OpenTelemetry tracing support & context propagation


airhorns:
**Is your feature request related to a problem? Please describe.**

Currently, Fission supports outputting traces to an OpenTracing Jaeger-formatted trace collection endpoint. This is great -- it provides lots of insight into what Fission is doing and helps operators understand and maintain.

**Describe the solution you'd like**

OpenTracing has been merged with OpenCensus into a new open source org called OpenTelemetry, and they've re-specified their trace propagation spec and trace format. Specifically there's three things that Fission could to do to better support tracing:

 - move to the OpenTelemetry library system as it is maintained, getting new features, and supports the new seemingly preferred trace format which is Jaeger's GRPC trace format (on port 14250 by default, referred to as `model.proto` see [their docs](https://www.jaegertracing.io/docs/1.21/getting-started/#all-in-one)). Currently, it seems like Fission only supports the older, less efficient Jaeger thrift format.
 - using OpenTelemetry, allow trace parents to be sent to functions, so fission can be within a trace instead of always the root. Right now, Fission's tracing support doesn't support OpenTelemetry's trace context propagation headers, which would allow Fission to report the spans it generates as children of a caller. In practice, this means if you are invoking fission functions from other services, you lose your tracing context instead of seeing it as one big nice trace. It'd be awesome to see a trace that originated from some frontend HTTP call go through various services including fission all the way into the function being executed, and OpenTelemetry context propagation should give Fission that for free.
 - allow some spans to be suppressed, like those for the healthcheck endpoints of the components that have them. We're generating a bunch of extra spans that I feel like I don't need to store. I think there are ways to filter these out downstream, but it would be nice to just never produce them in the first place. Configuring the whole of the OpenTelemetry exporter stack via environment variables doesn't sound super fun, but at least some way of lowering the data volume would be grand. 

**Describe alternatives you've considered**

OpenTracing works fine right now, and I haven't tested if passing OpenTracing context propagation headers to fission works. 

**Additional context**

I am not sure there is an official word from the OpenTracing folks, but it seems like all the development effort is behind OpenTelemetry. A bunch of observability vendors have thrown their weight behind OpenTelemetry, and the trace propagation spec that otel implements is being developed at the W3C and I think is likely to last a lot longer than OpenTracing's.

vishal-biyani:
[...]

PR #2157, opened 2021-08-10
Fission meets OpenTelemetry

gauravgahlot:
Fixes: #1861 

Note: The web socket example does not work when using OTEL HTTP. Here is an issue related to that on [open-telemetry/opentelemetry-js-contrib](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/12).


cmd/fission-bundle/main.go
```diff
@@ -176,6 +191,55 @@ func registerTraceExporter(logger *zap.Logger, arguments map[string]interface{})
 	return nil
 }
 
+// Initializes an OTLP exporter, and configures the corresponding trace and metric providers.
+func initProvider(logger *zap.Logger, arguments map[string]interface{}) (func(), error) {
```
sanketsudake, line 195: Same as above, having this code common package would help.

pkg/controller/api.go
```diff
@@ -263,9 +265,21 @@ func (api *API) GetHandler() http.Handler {
...
+func (api *API) Serve(port int, openTracingEnabled bool) {
 	address := fmt.Sprintf(":%v", port)
 	api.logger.Info("server started", zap.Int("port", port))
-	err := http.ListenAndServe(address, api.GetHandler())
+
+	var handler http.Handler
+	if openTracingEnabled {
+		handler = &ochttp.Handler{Handler: api.GetHandler()}
+	} else {
+		handler = otelhttp.NewHandler(api.GetHandler(), "fission-controller",
```
sanketsudake, line 276: Can move some of this code in common package pkg/utils/telemetry ?

```go
func GetHandlerWithOtel(name, handler ) {
          handler = otelhttp.NewHandler(handler, name,
			otelhttp.WithMessageEvents(otelhttp.ReadEvents, otelhttp.WriteEvents),
			otelhttp.WithFilter(func(r *http.Request) bool {
				return !(strings.Compare(r.URL.Path, "/healthz") == 0)
			}),
		)
        return handler
}
		
```

sanketsudake (changes requested):

charts/README.md
```diff
@@ -79,6 +79,10 @@ Parameter | Description | Default
 `router.roundTrip.timeout` | HTTP transport request timeout | `50ms`
 `router.roundTrip.timeoutExponent` | The length of request timeout will multiply with timeoutExponent after each retry | `2` 
 `router.roundTrip.maxRetries` | Max retries times of a failed request | `10`
+`openTracing.enabled` | If true, OpenTracing is enabled | `false`
+`openTracing.collectorEndpoint` | Jaeger collector endpoint | ``
+`openTracing.samplingRate` | Probabilistic sampling rate | `0.5`
+`otelCollectorEndpoint` | OpenTelemetry collector endpoint | None
```
sanketsudake, line 85: Should we separate section for `openTelemetry` ?

```
openTelemetry:
         enabled: false
         collectorEndpoint: xyz
```

OR we can have `tracingMode` which can be either of `opentracing` , `opentelemetry` or `none`
then specific section for each mode.

charts/fission-all/values.yaml
```diff
@@ -340,9 +340,23 @@ prometheus:
 canaryDeployment:
   enabled: true
 
-# Use these flags to enable opentracing, the variable is endpoint of Jaeger collector in the format shown below
-#traceCollectorEndpoint: "http://jaeger-collector.jaeger.svc:14268/api/traces?format=jaeger.thrift"
-#traceSamplingRate: 0.75
+# Use these flags to enable OpenTracing.
```
sanketsudake, line 343: Use `below/following` instead of `these`

pkg/router/functionHandler.go
```diff
@@ -309,8 +313,21 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 			dumpReqFunc(newReq)
 		}
 
+		openTracingEnabled, err := strconv.ParseBool(os.Getenv("OPENTRACING_ENABLED"))
```
sanketsudake, line 316: We should avoid reading the env variable in the request path. Can we read it once and pass to all round-trippers created?

sanketsudake: Also if we get an error while reading here, it's not fatal req per say so we can avoid returning.

cmd/fetcher/app/server.go
```diff
@@ -85,11 +64,31 @@ func Run(logger *zap.Logger) {
...
+			}
+		}()
+	} else {
+		shutdown, err = otelUtils.InitProvider(logger, "Fission-Fetcher")
+		if err != nil {
+			logger.Fatal("error initializing provider for OTLP", zap.Error(err))
 		}
-	}()
+	}
+	if shutdown != nil {
```
sanketsudake, line 85: Should we move this part under else only ? we may not need the shutdown definition above then. Same for rest of shutdown definitions.

pkg/router/functionHandler.go
```diff
@@ -494,6 +508,17 @@ func (fh functionHandler) handler(responseWriter http.ResponseWriter, request *h
 		rrt.closeContext()
 	}()
 
+	// add attributes to current span
+	ctx := request.Context()
+	span := trace.SpanFromContext(ctx)
+	attributes := []attribute.KeyValue{
```
sanketsudake, line 514: can we move this otelutils ?
```sh
func GetAttributesForFunction(fn *fv1.Function) []attribute.KeyValue
```

I assume we would need to add these attributes in executor and fetcher also ? Or these are retained in next spans ?

pkg/router/functionHandler.go
```diff
@@ -310,7 +316,15 @@ func (roundTripper *RetryingRoundTripper) RoundTrip(req *http.Request) (*http.Re
 		}
 
 		// forward the request to the function service
-		resp, err := ocRoundTripper.RoundTrip(newReq)
+		var resp *http.Response
+		if roundTripper.funcHandler.openTracingEnabled || util.IsWebsocketRequest(newReq) {
```
sanketsudake, line 320: Let's add a detailed comment here about the `IsWebsocketRequest` exception. So that when we plan to remove opentracing in future this would be taken care by dev modifying code.

sanketsudake (approved):

PR #2160, opened 2021-08-12
Update alpine base image to 3.14

sanketsudake:

Change is required for vulnerability fixes also there were few DNS failures in recent alpine builds.


sanketsudake:
Upgrade tests passed https://github.com/fission/fission/runs/3308251092

PR #2161, opened 2021-08-12
Use pool pod controller with env informer 

sanketsudake:

- Use informers and listers in executors 
- Passing context properly in pool manager executor
- Environment updates in pool manager would not cause updates in the deployment
- Environment update minimizing downtime
- Use replicaset controller and environment delete triggers to cleanup specialized pods

### DIFF
--- a/charts/fission-all/templates/deployment.yaml
+++ b/charts/fission-all/templates/deployment.yaml
@@ -85,6 +85,7 @@ rules:
   resources:
   - deployments
   - deployments/scale
+  - replicasets
   verbs:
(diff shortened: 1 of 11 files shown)


RealHarshThakur (approved):

sanketsudake:
To verify minimum downtime created a function with poolmanager  and launched request load on it. `ab -n 50000 -c 10 localhost:8888/mytest`

After a few seconds of updated image for the environment, all pods were getting cleaned as part of it. Also didn't observe any loss of requests.

PR #2172, opened 2021-08-20
Add context for traces in executor

gauravgahlot:
Signed-off-by: Gaurav Gahlot <gauravgahlot0107@gmail.com>

### DIFF
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -62,6 +62,7 @@ type (
 	}
 
 	createFuncServiceRequest struct {
+		context  context.Context
 		function *fv1.Function
 		respChan chan *createFuncServiceResponse
 	}
@@ -96,7 +97,7 @@ func MakeExecutor(ctx context.Context, logger *zap.Logger, cms *cms.ConfigSecret
 		}(et)
 	}
 
-	go executor.serveCreateFuncServices(ctx)
+	go executor.serveCreateFuncServices()
 
 	return executor, nil
 }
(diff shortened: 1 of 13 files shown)


pkg/executor/executor.go
```diff
@@ -195,7 +195,7 @@ func (executor *Executor) serveCreateFuncServices() {
 				wg.Wait()
 
 				// get the function service from the cache
-				fsvc, err := executor.getFunctionServiceFromCache(req.function)
+				fsvc, err := executor.getFunctionServiceFromCache(ctx, req.function)
```
sanketsudake, line 198: context received from channel request should be passed here.

pkg/executor/executor.go
```diff
@@ -107,7 +107,7 @@ func MakeExecutor(ctx context.Context, logger *zap.Logger, cms *cms.ConfigSecret
 // get specialized. In other words, it ensures that when there's an
 // ongoing request for a certain function, all other requests wait for
 // that request to complete.
-func (executor *Executor) serveCreateFuncServices() {
+func (executor *Executor) serveCreateFuncServices(ctx context.Context) {
```
sanketsudake, line 110: We should not pass context to this function. Every request on the channel should pass its own context and we pass same context further.

```go
createFuncServiceRequest struct {
    context  context.Context
    function *fv1.Function
    respChan chan *createFuncServiceResponse
}
```

sanketsudake (approved):

sanketsudake:
Thanks, @gauravgahlot, these are very valuable changes and would help in resolving issues of executor. 🎉

PR #2176, opened 2021-08-25
Added redis keda connector

blackfly19:
Completed the RedisList keda connector and tested it.

### DIFF
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -140,7 +140,7 @@ router:
 
   ## Sample with a rate per time window (traces/second)
   traceSamplingRate: 0.5
-  
+
   ## Extend the container specs for the core fission pods.
   ## Can be used to add things like affinty/tolerations/nodeSelectors/etc.
   ## For example:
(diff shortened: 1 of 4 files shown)


charts/fission-core/values.yaml
```diff
@@ -269,3 +269,6 @@ mqt_keda:
     gcp_pub_sub:
       image: fission/keda-gcp-pubsub-http-connector
       tag: v0.2
+    redis:
+      image: blackfly19/keda-redislist-http-connector
+      tag: v5
```
sanketsudake, line 274: Since this would be first release for redis-http-connector, we should have version v0.1

charts/fission-core/values.yaml
```diff
@@ -269,3 +269,6 @@ mqt_keda:
     gcp_pub_sub:
       image: fission/keda-gcp-pubsub-http-connector
       tag: v0.2
+    redis:
+      image: blackfly19/keda-redislist-http-connector
```
sanketsudake, line 273: Please change this to `fission/keda-redis-http-connector`. Once keda connectors PR is merged, CI would build and push required image.

charts/fission-all/values.yaml
```diff
@@ -380,6 +379,9 @@ mqt_keda:
     gcp_pub_sub:
       image: fission/keda-gcp-pubsub-http-connector
       tag: v0.2
+    redis:
```
sanketsudake, line 382: Same comments as below

sanketsudake (changes requested):

sanketsudake (approved):

PR #2196, opened 2021-09-17
cleanup: Remove Opentracing support as no active users

sanketsudake:
References:
[1] https://github.com/fission/fission/discussions/2193
[2] https://fissionio.slack.com/archives/C3LUX6BBP/p1631706812069300

Signed-off-by: Sanket Sudake <sanketsudake@gmail.com>

PR #2201, opened 2021-09-21
enhancement: Add different samplers and propagators support with OpenTelemetry

sanketsudake:
* Samplers
  - always_on - Sampler that always samples spans, regardless of the parent span's sampling decision.
  - always_off - Sampler that never samples spans, regardless of the parent span's sampling decision.
  - traceidratio - Sampler that samples probabalistically based on rate.
  - parentbased_always_on - (default) Sampler that respects its parent span's sampling decision, but otherwise always samples.
  - parentbased_always_off - Sampler that respects its parent span's sampling decision, but otherwise never samples.
  - parentbased_traceidratio - Sampler that respects its parent span's sampling decision, but otherwise samples probabalistically based on rate.

* Propogators
  - tracecontext - W3C Trace Context
  - baggage - W3C Baggage
  - b3 - B3 Single
  - b3multi - B3 Multi
  - jaeger - Jaeger uber-trace-id header
  - xray - AWS X-Ray (third party)
  - ottrace - OpenTracing Trace (third party)

* Added tracing support via fission CLI

With support for jaeger propagator, ie, support for uber-trace-id header we can remove opentracing. 
https://github.com/fission/fission/pull/2196


sanketsudake:
Once https://github.com/open-telemetry/opentelemetry-go/issues/1698 is resolved we can get rid of code from Fission and use features sdk itself.

PR #2207, opened 2021-09-26
Added command to list pods owned by fission for environment/function

praddy26:
`fission env pods --name <enviornmentName> [--executortype <poolmgr | newdeploy>]` 
`fission fn pods --name <functionName> [--executortype <poolmgr | newdeploy>]`


sanketsudake (changes requested):

pkg/controller/environmentApi.go
```diff
@@ -238,3 +239,40 @@ func (a *API) EnvironmentApiDelete(w http.ResponseWriter, r *http.Request) {
...
+	if len(ens) != 0 {
+		selector[fv1.ENVIRONMENT_NAMESPACE] = ens
+	}
+
+	et := a.extractQueryParamFromRequest(r, "executortype")
+	if len(et) != 0 {
+		selector[fv1.EXECUTOR_TYPE] = et
+	}
+
+	pods, err := a.kubernetesClient.CoreV1().Pods("").List(r.Context(), metav1.ListOptions{
```
sanketsudake, line 263: can we use `metav1.NamespaceAll` here ?

pkg/fission-cli/cmd/function/pods.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2019 The Fission Authors.
```
sanketsudake, line 2: `2021`

pkg/controller/environmentApi.go
```diff
@@ -238,3 +239,40 @@ func (a *API) EnvironmentApiDelete(w http.ResponseWriter, r *http.Request) {
...
+func (a *API) EnvironmentApiPodList(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	name := vars["environment"]
+
+	// label selector
+	selector := map[string]string{
+		fv1.ENVIRONMENT_NAME: name,
+	}
+
+	ens := a.extractQueryParamFromRequest(r, "environmentnamespace")
```
sanketsudake, line 253: Please define constants for these, instead of hardcoding.
Same constant can be used in the client and API sides.
`environmentnamespace`, `executortype`
Same for other query parameters.

pkg/controller/functionApi.go
```diff
@@ -369,3 +369,35 @@ func getContainerLog(kubernetesClient *kubernetes.Clientset, w http.ResponseWrit
 
 	return nil
 }
+
+// FunctionApiPodList: Get list of pods currently used by function
+func (a *API) FunctionApiPodList(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	name := vars["function"]
```
sanketsudake, line 376: Please check if we need any error handling here of "function" key not in vars.

pkg/fission-cli/cmd/environment/pods.go
```diff
@@ -0,0 +1,72 @@
...
+func ListPods(input cli.Input) error {
+	return (&ListPodsSubCommand{}).do(input)
+}
+
+func (opts *ListPodsSubCommand) do(input cli.Input) error {
+
+	m := &metav1.ObjectMeta{
+		Name: input.String(flagkey.EnvName),
+		Labels: map[string]string{
+			"environmentNamespace": input.String(flagkey.NamespaceEnvironment),
```
sanketsudake, line 46: We should use constants for query parameters as mentioned above.
These ensure if we change client-side parameters, we are also changing server-side.

pkg/fission-cli/cmd/environment/pods.go
```diff
@@ -0,0 +1,72 @@
...
+			"executorType":         input.String(flagkey.EnvExecutorType),
+		},
+	}
+
+	pods, err := opts.Client().V1().Environment().ListPods(m)
```
sanketsudake, line 51: We should be explicit about `metav1.NamespaceAll` here

pkg/fission-cli/util/podutils.go
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+package util
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// PodReadyStatus returns the number of ready containers and total containers present in pod
+func PodContainerReadyStatus(pod *v1.Pod) (readyContainers, noOfContainers int) {
```
sanketsudake, line 24: We already have a function `IsReadyPod` in pkg/utils/utils.go 
Please see it it can be used.

sanketsudake (changes requested):

pkg/fission-cli/util/podutils.go
praddy26: `PodContainerReadyStatus` function returns the number of containers which are ready inside the pod and the total number of containers in the pod, which is used to populate the READY column like 2/2. Existing `IsReadyPod` return boolean value as whether the pod is ready or not.

sanketsudake: Got it. Could you please organize both functions in the same file? New one, which you have introduced. `podutils.go`

pkg/controller/functionApi.go
praddy26: This function is registered as event handler for the following route 
`r.HandleFunc("/v2/functions/{function}/pods", api.FunctionApiPodList).Methods("GET")`

Unless and untill we have the route like this `/v2/functions/{function}/pods`, it will not hit the event handler `FunctionApiPodList` 

Similar code has been used for other eventhandlers like `FunctionApiDelete`, `FunctionApiGet`, `FunctionApiUpdate` and `FunctionPodLogs`

pkg/controller/environmentApi.go
praddy26: Fixed

pkg/fission-cli/cmd/function/pods.go
praddy26: Fixed

pkg/controller/environmentApi.go
praddy26: Fixed

pkg/controller/client/v1/environment.go
```diff
@@ -163,3 +164,30 @@ func (c *Environment) List(ns string) ([]fv1.Environment, error) {
 
 	return envs, nil
 }
+
+func (c *Environment) ListPods(m *metav1.ObjectMeta) ([]apiv1.Pod, error) {
+	relativeUrl := fmt.Sprintf("environments/%s/pods", m.Name)
+
+	if len(m.Labels) != 0 {
+		relativeUrl = fmt.Sprintf("%s?environmentnamespace=%s&executortype=%s", relativeUrl, m.Labels[fv1.ENVIRONMENT_NAMESPACE], m.Labels[fv1.EXECUTOR_TYPE])
```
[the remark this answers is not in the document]
praddy26: Fixed

pkg/controller/client/v1/function.go
```diff
@@ -176,3 +178,30 @@ func (c *Function) List(functionNamespace string) ([]fv1.Function, error) {
 
 	return funcs, nil
 }
+
+func (c *Function) ListPods(m *metav1.ObjectMeta) ([]apiv1.Pod, error) {
+	relativeUrl := fmt.Sprintf("functions/%s/pods", m.Name)
+
+	if len(m.Labels) != 0 {
+		relativeUrl = fmt.Sprintf("%s?functionnamespace=%s", relativeUrl, m.Labels[fv1.FUNCTION_NAMESPACE])
```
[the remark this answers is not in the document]
praddy26: Fixed

pkg/controller/functionApi.go
praddy26: Fixed

pkg/fission-cli/util/podutils.go
sanketsudake: Eventually, we plan to move all content to pkg/utils. We want to get rid of any other utils folders.

1. Could you please move podutils.go to `pkg/utils/podutils.go` ? 
2. Also, exiting `IsReadyPod` function needs to be moved from `pkg/utils/utils.go` to `pkg/utils/podutils.go`

sanketsudake (changes requested):

pkg/apis/core/v1/const.go
```diff
@@ -125,6 +125,14 @@ const (
 	FUNCTION_UID              = "functionUid"
 	FUNCTION_RESOURCE_VERSION = "functionResourceVersion"
 	EXECUTOR_TYPE             = "executorType"
+	MANAGED                   = "managed"
+)
+
+// query string parameter key
+const (
+	ENVIRONMENT_NAMESPACE_QS = "environmentnamespace"
```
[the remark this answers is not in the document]
sanketsudake: Yes it should be fine

sanketsudake:
Minor comments otherwise PR looks good to me.

PR #2214, opened 2021-10-02
refactor: Replaced context.TODO() with request.Context() in API files

praddy26:


pkg/controller/api.go
```diff
@@ -131,20 +131,20 @@ func (api *API) extractQueryParamFromRequest(r *http.Request, queryParam string)
 }
 
 // check if namespace exists, if not create it.
-func (api *API) createNsIfNotExists(ns string) error {
+func (api *API) createNsIfNotExists(ns string, ctx context.Context) error {
```
sanketsudake, line 134: `ctx context.Context` should be the first parameter in all the functions.

pkg/controller/functionApi.go
```diff
@@ -147,13 +147,13 @@ func (a *API) FunctionApiCreate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check if namespace exists, if not create it.
-	err = a.createNsIfNotExists(f.ObjectMeta.Namespace)
+	err = a.createNsIfNotExists(f.ObjectMeta.Namespace, r.Context())
```
sanketsudake, line 150: Same as above. Please context as a first argument after changing function def.

pkg/controller/httpTriggerApi.go
```diff
@@ -123,8 +123,8 @@ func (a *API) HTTPTriggerApiList(w http.ResponseWriter, r *http.Request) {
 }
 
 // checkHTTPTriggerDuplicates checks whether the tuple (Method, Host, URL) is duplicate or not.
-func (a *API) checkHTTPTriggerDuplicates(t *fv1.HTTPTrigger) error {
-	triggers, err := a.fissionClient.CoreV1().HTTPTriggers(metav1.NamespaceAll).List(context.TODO(), metav1.ListOptions{})
+func (a *API) checkHTTPTriggerDuplicates(t *fv1.HTTPTrigger, ctx context.Context) error {
```
sanketsudake, line 126: Same as above. Context should be the first argument.

pkg/controller/api.go
sanketsudake: This is more by the convention in Go.

sanketsudake (changes requested):

sanketsudake (approved):

sanketsudake:
Thanks for finding it out and fixing it @praddy26 🎉

PR #2224, opened 2021-10-11
refactoring: Breakdown fission-all chart into components

sanketsudake:

Overview:

1. Break down `fission-all`  into different components.
2. By default, all extra components would be disabled in `fission-all`. To ensure `fission-all` is similar to existing `fission-core`.
3. We will get rid of the `fission-core` chart and only keep `fission-all`. 
4. User using `fission-core` chart would need to migrate to `fission-all`. 

## `fission-all` changes

We keep the following components by default disabled now, which were enabled by default earlier. 
If users needs they need to set corresponding values.

1. nats 
2. influxdb
3. prometheus 
4. canaryDeployment 

Broken Kubernetes yaml's into different components. <https://github.com/fission/fission/tree/charts-merge/charts/fission-all/templates>

### User notes

If you are using `fission-all`  chart and you are using any of the integrations mentioned below, you would need to enable them via chart values.

1. nats - Fission Nats integration
2. influxdb - Influxdb and logger component
3. prometheus - Prometheus disabled by default
4. canaryDeployment - Disabled by default

### User notes

`fission-core` chart is removed now and user should use `fission-all` chart instead. With recent changes in `fission-all`,  chart is upgrade compatible with `fission-core`. You can set chart values as per your need.

Testing done

- Installed old fission-core chart
```
kubectl create -k "github.com/fission/fission/crds/v1?ref=v1.15.0-rc1"
export FISSION_NAMESPACE="fission"
kubectl create namespace $FISSION_NAMESPACE
kubectl create -k "github.com/fission/fission/crds/v1?ref=1.14.1"
helm install fission fission-charts/fission-core --version 1.14.1 --namespace fission --set serviceType=NodePort,routerServiceType=NodePort
```
- Upgraded to `fission-all` chart in the PR

```console
# verified diff
helm diff upgrade fission ./charts/fission-all --namespace fission --set serviceType=NodePort,routerServiceType=NodePort 
# do upgrade
helm upgrade fission ./charts/fission-all --namespace fission --set serviceType=NodePort,routerServiceType=NodePort 
```

### DIFF
--- a/charts/fission-all/Chart.yaml
+++ b/charts/fission-all/Chart.yaml
@@ -1,18 +1,22 @@
 apiVersion: v2
 name: fission-all
 version: v1.15.0-rc1
+appVersion: v1.15.0-rc1
 description: Fission is a fast serverless framework for Kubernetes.
-keywords:
-- fission
-- serverless
 home: https://fission.io/
+icon: https://fission.io/images/logo-gh.svg
+sources:
+  - https://github.com/fission/fission
+  - https://github.com/fission/keda-connectors
+keywords:
+  - fission
+  - serverless
 maintainers:
   - name: Vishal Biyani
     email: vishal@infracloud.io
   - name: Sanket Sudake
     email: sanket@infracloud.io
 engine: gotpl
-appVersion: v1.15.0-rc1
 type: application
 dependencies:
   - name: prometheus


sanketsudake:
Related issue https://github.com/fission/fission-charts/issues/47

Issue #2199, opened 2021-09-17
Upgraded S3 Support


kferrone:
**Is your feature request related to a problem? Please describe.**
When using the fission chart and setting up the s3 as the storage backend, I am not able to give the credentials the way I want and the fission storagesvc tries to make a bucket. I don't really need it to make me a bucket, I have other means for creating buckets. Also I want the credentials I provide to fission to be only for that bucket without the ability to create the buckets. Currently the credentials need full create buckets privileges and I'm forced to let fission create the bucket.  

**Describe the solution you'd like**
I would like to provide credentials which don't allow creating buckets, just full access to a designated bucket. Also I would like to use [IRSA](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) since we are using EKS. 
Here is my desired policy which caused the storagesvc to be in a crashloop:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::my-fission-bucket",
            "Effect": "Allow",
            "Sid": "AllowListObjects"
        },
        {
            "Action": [
                "s3:DeleteObject",
                "s3:GetObject",
                "s3:PutObject"
            ],
            "Resource": "arn:aws:s3:::my-fission-bucket/*",
            "Effect": "Allow",
            "Sid": "AllowObjectsCRUD"
        }
    ]
}
```

**Describe alternatives you've considered**
Just gave the policy an action like: `s3:*`.

vishal-biyani:
Hi @kferrone Thanks for filing this issue with great details. The ask makes sense and we will prioritise this in one of upcoming releases.

kferrone:
How do I get this to happen. I still get AccessDenied when it tries to create the bucket when the bucket already exists.

PR #2275, opened 2021-12-09
Allow user to use existing bucket with same name via fission storage service

YachikaRalhan:
## Description
If a bucket name passed by a user already exists on the cloud, allow the user to use that one as storage and do not throw an error. So users don't really have to create a bucket always and no need to give create buckets privileges to access keys.

Fixes https://github.com/fission/fission/issues/2199

## Testing
Created multiple buckets on S3 having the same prefix, and the code picked the right bucket (with exact match) 

## Checklist:

### DIFF
--- a/pkg/storagesvc/stowClient.go
+++ b/pkg/storagesvc/stowClient.go
@@ -80,7 +80,10 @@ func getContainer(loc stow.Location, containerName string, cursor string) (stow.
 		}
 	}
 	if con == nil && !stow.IsCursorEnd(cursorNew) {
-		getContainer(loc, containerName, cursorNew)
+		_, err := getContainer(loc, containerName, cursorNew)
+		if err != nil {
+			return nil, err
+		}
 	}
 	return con, nil
 }
--- a/pkg/storagesvc/stowClient.go
+++ b/pkg/storagesvc/stowClient.go
@@ -88,6 +88,17 @@ func getContainer(loc stow.Location, containerName string, cursor string) (stow.
 	return con, nil
 }
 
+func getOrCreateContainer(loc stow.Location, containerName string, cursor string) (stow.Container, error) {
+	con, err := loc.CreateContainer(containerName)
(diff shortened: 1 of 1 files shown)


pkg/storagesvc/stowClient.go
```diff
@@ -83,22 +109,14 @@ func MakeStowClient(logger *zap.Logger, storage Storage) (*StowClient, error) {
 		return nil, err
 	}
 	stowClient.location = loc
-
 	con, err := loc.CreateContainer(config.storage.getContainerName())
```
sanketsudake, line 112: Can we define one more wrapper function here?


```
func getOrCreateContainer(loc stow.Location, containerName string, cursor string) (stow.Container, error) {
        con, err := loc.CreateContainer(config.storage.getContainerName())
        if err != nil && (os.IsExist(err) || strings.Contains(err.Error(), "BucketAlreadyOwnedByYou")) {
                conn, err = getContainer(loc, config.storage.getContainerName(), stow.CursorStart)
        }
        return conn, err
}
```
This reduces errors checks and overall code.

pkg/storagesvc/stowClient.go
```diff
@@ -62,6 +62,43 @@ var (
...
+	}
+	return con, nil
+}
+
+func getOrCreateContainer(loc stow.Location, containerName string, cursor string) (stow.Container, error) {
+	con, err := loc.CreateContainer(containerName)
+	if err != nil && (os.IsExist(err) || strings.Contains(err.Error(), "BucketAlreadyOwnedByYou")) {
+		con, err = getContainer(loc, containerName, stow.CursorStart)
+	}
+	if con == nil {
```
sanketsudake, line 96: We should have an additional condition here.

```
if con == nil && err == nil {
}
```

If err is not nil, ideally, we should wrap error, so that we don't lose previously captured error.

```
fmt.Errorf("Storage container %s not found: %w", containerName, err)
```

Also, in the below error container name is missing.

sanketsudake:
@mergifyio rebase

sanketsudake (approved):
Thanks for PR @YachikaRalhan 🎉

PR #2282, opened 2021-12-17
Simplify if logic in executor function

arbourd:
## Description

Reorders a function to facilitate early exists and reduce conditional blocks.

This change sets the order of operations from:

- Check if no errors, and runs non-error flow (updating a deployment)
- Check if the error is a non-existing deployment (creates a new deployment)
- Returns errors all other errors

To

- Checks if errors exist and error is not missing deployment, returns early
- Checks if error is deployment does not exist (creates a new deployment)
- Runs non-error glow

## Testing

Running tests on this branch.

## Checklist:


sanketsudake (approved):
LGTM

PR #2292, opened 2022-01-01
feature: Basic auth support with fission router

### LINKED ISSUE #2277: Basic auth support with fission router

Currently, Fission doesn't have any inbuilt support for authentication at the Fission router.
The fission router is the entry point for all function calls made by the Fission user.

Fission should have basic auth support which can be enabled while deploying function.
It could be getting a token with standard basic auth and using it for all requests made fission function calls.

**Describe alternatives you've considered**
In the current scenario, the user has to configure basic auth by himself at the ingress level, which does provide a layer of authentication for external calls. But if a user is using the Kubernetes service only then he cant configure authentication for all fission function calls.

**Additional context**

Probable set of changes required:
- Decide standard auth method strategy to be used 
- Generate random credentials at the time of deploying Fission
- Configure auth middleware in Fission router(changes required) so that all requests are checked if they are authenticated then forwarded to respective function handlers
- We would also need to ensure necessary changes are made in Fission CLI

praddy26:
## Description

This PR adds new feature to enable authentication for fission routers. 
This is **a optional feature** which can be enabled/disabled as per the needs of the user. 
On enabling this feature, an endpoint for authentication will be registered in the router and all function API calls with be routed via authentication middleware. 
The user has to first create auth token by passing credentials. 
The token generated has to be then passed in the authorization header for all susbequent API calls to access the function.

This feature can be turned on by setting the key `authentication.enabled` in `charts/fission-all/values.yaml ` to `true` during the time of deployment. 
Along with this there are other keys like 
- `authUrlPath`
-  `authUsername`
- `jwtSigningKey`
- `jwtExpiryTime` 
- `jwtIssuer` 

The description for these parameters are provided in the same file as comments.

On enabling this feature a secret by name `router` will be created in the `fission` namespace with `username`, randomly generated `password` and `jwtSigningKey`. 
This secret will be mounted as volume on the router pod.
Also the other values mentioned above will be added to the existing `feature-config` configmap and will be mounted as volume on the router pod.

fission function test command:
---------------------------------
-  If default `authUrlPath` is not used, then the custom path has to be exported to the environment variable `FISSION_AUTH_URI`

  ```bash
  export FISSION_USERNAME=$(kubectl get secrets/router --template={{.data.username}} -n fission | base64 -d)
  export FISSION_PASSWORD=$(kubectl get secrets/router --template={{.data.password}} -n fission | base64 -d)
  export FISSION_AUTH_TOKEN=$(fi token create --username $FISSION_USERNAME --password $FISSION_PASSWORD)
  ```

-  For testing function token is picked from the env variable `FISSION_AUTH_TOKEN`, 
  
  ```bash
  fission function --name <functionName>
  ```

-  If env variable is not set then token has to be passed in the header 

  ```bash
  fission function test --name <functionName> --header "Authorization:Bearer <token>" 
  ```

fission function API:
----------------------------------------------
-  Router port has to be port-forwarded to one of the port. ex: 3000
  
  ```bash
  export FISSION_USERNAME=$(kubectl get secrets/router --template={{.data.username}} -n fission | base64 -d)
  export FISSION_PASSWORD=$(kubectl get secrets/router --template={{.data.password}} -n fission | base64 -d)
  curl -v http://localhost:3000/auth/login -d '{"username":"'$FISSION_USERNAME'", "password":"'$FISSION_PASSWORD'"}'
  export FISSION_AUTH_TOKEN=<token>
  curl http://localhost:3000/<path> -H "Authorization: Bearer ${FISSION_AUTH_TOKEN}"
  ```

Fixes #2277

## Checklist:

### DIFF
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -31,6 +31,13 @@ canary:
   prometheusSvc: {{ .Values.prometheus.serviceEndpoint | default "" | quote }}
   {{- end }}
   {{- printf "\n" -}}
+auth:
+  enabled: {{ .Values.authentication.enabled | default false }}
+  {{- if .Values.authentication.enabled }}
+  authUriPath: {{ .Values.authentication.authUriPath | default "/auth/login" | quote}}
+  jwtExpiryTime: {{ .Values.authentication.jwtExpiryTime | default 120 }}
+  jwtIssuer: {{ .Values.authentication.jwtIssuer | default "fission" | quote }}
+  {{- end }}
 {{- end -}}
 
 {{/*
--- a/charts/fission-all/templates/router/deployment.yaml
+++ b/charts/fission-all/templates/router/deployment.yaml
@@ -35,6 +35,23 @@ spec:
         command: ["/fission-bundle"]
         args: ["--routerPort", "8888", "--executorUrl", "http://executor.{{ .Release.Namespace }}"]
         env:
+        {{- if .Values.authentication.enabled }}
+        - name: AUTH_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: router
+              key: username
+        - name: AUTH_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: router
+              key: password
+        - name: JWT_SIGNING_KEY
+          valueFrom:
+            secretKeyRef:
+              name: router
+              key: jwtSigningKey
+        {{- end }}      
         - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
@@ -80,6 +97,10 @@ spec:
             port: 8888
           initialDelaySeconds: 35
           periodSeconds: 5
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/config/config.yaml
+          subPath: config.yaml
         ports:
         - containerPort: 8080
           name: metrics
@@ -100,6 +121,10 @@ spec:
         terminationMessagePolicy: {{ .Values.terminationMessagePolicy }}
         {{- end }}
       serviceAccountName: fission-svc
+      volumes:
+      - name: config-volume
+        configMap:
+          name: feature-config
 {{- if .Values.router.priorityClassName }}
       priorityClassName: {{ .Values.router.priorityClassName }}
 {{- else if .Values.priorityClassName }}
--- a/charts/fission-all/templates/router/secret.yaml
+++ b/charts/fission-all/templates/router/secret.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.authentication.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: router
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    "helm.sh/hook": pre-install
+data:
+  username: {{ .Values.authentication.authUsername | b64enc | quote }}
+  password: {{ randAlphaNum 20 | b64enc | quote }}
+  jwtSigningKey: {{ .Values.authentication.jwtSigningKey | b64enc | quote }}
+{{- end }}
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -161,7 +161,6 @@ router:
   ## router resource utilization when under heavy workloads.
   ##
   displayAccessLog: false
-
   ## svcAnnotations is the annotations to be added to the service resource created for router.
   ##
   # svcAnnotations:
@@ -482,6 +481,37 @@ prometheus:
 ##
 canaryDeployment:
   enabled: false
+
+## Enable authentication for fission function invocation via Fission router
+##
+authentication:
+  ## set this flag to true if you need authentication
+  ## for all function invocations
+  ## default 'false'
+  ##
+  enabled: false
+  ## authUriPath defines authentication endpoint path
+  ## via router
+  ## default '/auth/login'
+  ##
+  authUriPath:
+  ## authUsername is used as a username for authentication
+  ## default 'admin'
+  ##
+  authUsername: admin
+  ## jwtSigningKey is the signing key used for
+  ## signing the JWT token
+  ##
+  jwtSigningKey: serverless
+  ## jwtExpiryTime is the JWT expiry time
+  ## in seconds
+  ## default '120'
+  ##
+  jwtExpiryTime: 
+  ## jwtIssuer is the issuer of JWT
+  ## default 'fission'
+  ##
+  jwtIssuer: fission
 
 ## Use the following flags to enable OpenTracing.
 ## Note: OpenTracing support will be removed in an upcoming release.
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/go-git/go-git/v5 v5.4.2
 	github.com/go-ini/ini v1.63.2 // indirect
 	github.com/go-openapi/spec v0.20.4
+	github.com/golang-jwt/jwt/v4 v4.2.0
 	github.com/gorilla/mux v1.8.0
 	github.com/gotestyourself/gotestyourself v2.2.0+incompatible // indirect
 	github.com/graymeta/stow v0.2.7
--- a/go.sum
+++ b/go.sum
@@ -492,7 +492,10 @@ github.com/gogo/protobuf v1.2.2-0.20190730201129-28a6bbf47e48/go.mod h1:SlYgWuQ5
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/gogo/protobuf v1.3.2 h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
 github.com/gogo/protobuf v1.3.2/go.mod h1:P1XiOD3dCwIKUDQYPy72D8LYyHL2YPYrpS2s69NZV8Q=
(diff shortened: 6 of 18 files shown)


charts/fission-all/templates/router/deployment.yaml
```diff
@@ -35,6 +35,23 @@ spec:
         command: ["/fission-bundle"]
         args: ["--routerPort", "8888", "--executorUrl", "http://executor.{{ .Release.Namespace }}"]
         env:
+        {{- if .Values.authentication.enabled }}
+        - name: ROUTER_USERNAME
```
sanketsudake, line 39: Should we prefer `AUTH_USERNAME` and `AUTH_PASSWORD` ?

charts/fission-all/templates/_helpers.tpl
```diff
@@ -31,6 +31,13 @@ canary:
   prometheusSvc: {{ .Values.prometheus.serviceEndpoint | default "" | quote }}
   {{- end }}
   {{- printf "\n" -}}
+auth:
+  enabled: {{ .Values.authentication.enabled | default false }}
+  {{- if .Values.authentication.enabled }}
+  authUrlPath: {{ .Values.authentication.authUrlPath | default "/auth/login" | quote}}
+  jwtExpiryTime: {{ .Values.authentication.jwtExpiryTime | default 120 }}
```
sanketsudake, line 38: I dont see any limitation of this config as of now. But in future if we wish to support different auth methods we need to make sure auth config is flexible for that. Please see from your perspective also we need any changes here from that perspective.

charts/fission-all/templates/router/secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.authentication.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: router
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+type: Opaque
```
sanketsudake, line 8: Should we skip type here? This way user default secret method would be used while installation ?

charts/fission-all/templates/router/secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.authentication.enabled }}
+apiVersion: v1
```
sanketsudake, line 2: We can use `pre-install` hook for this resource ?
https://helm.sh/docs/topics/charts_hooks/

charts/fission-all/values.yaml
```diff
@@ -479,6 +478,36 @@ prometheus:
...
+authentication:
+  ## set this flag to true if you need authorization
+  ## for all route calls
+  ##
+  enabled: false
+  ## authUrlPath is the router path for
+  ## authentication endpoint, defaults to /auth/login
+  ## deafult '/auth/login'
+  ##
+  authUrlPath: /auth/login
```
sanketsudake, line 491: Just wondering if tomorrow if we consider an external auth provider? Would it work with these params?

pkg/fission-cli/cmd/token/create.go
```diff
@@ -0,0 +1,97 @@
...
+	jsonValue, _ := json.Marshal(values)
+
+	kubeContext := input.String(flagkey.KubeContext)
+	// Portforward to the fission router
+	localRouterPort, err := util.SetupPortForward(util.GetFissionNamespace(), "application=fission-router", kubeContext)
+	if err != nil {
+		return err
+	}
+
+	authAuthenticatorUrl := "http://127.0.0.1:" + localRouterPort + "/auth/login"
```
sanketsudake, line 72: As URL is configurable, we can support some ENV variables. If the user sets the ENV variable, we would consider provided URL.

sanketsudake (changes requested):

charts/fission-all/templates/router/deployment.yaml
praddy26: Fixed

charts/fission-all/templates/router/secret.yaml
praddy26: Fixed

charts/fission-all/values.yaml
praddy26: The auth URI  value is looked for in the env variable `FISSION_AUTH_URI`. If it is not set or if it is a empty string then it defaults to `/auth/login`

charts/fission-all/values.yaml
```diff
@@ -483,6 +482,37 @@ prometheus:
...
+  ##
+  authUrlPath:
+  ## authUsername is used as username for authentication
+  ## of router
+  ##
+  authUsername: infracloud
+  ## jwtSigningKey is the signing key used for
+  ## signing the jwt token
+  ##
+  jwtSigningKey: converge2021
```
sanketsudake, line 504: Can we keep it `serverless` ?

charts/fission-all/values.yaml
```diff
@@ -483,6 +482,37 @@ prometheus:
...
+  enabled: false
+  ## authUrlPath is the router path for
+  ## authentication endpoint, defaults to /auth/login
+  ## default '/auth/login'
+  ##
+  authUrlPath:
+  ## authUsername is used as username for authentication
+  ## of router
+  ##
+  authUsername: infracloud
```
sanketsudake, line 500: Please change user to `admin`.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/blend/go-sdk v1.20211025.3 // indirect
 	github.com/containerd/continuity v0.2.1 // indirect
 	github.com/dchest/uniuri v0.0.0-20200228104902-7aecb25e1fe5
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
sanketsudake, line 13: This is deprecated library. Please use https://github.com/golang-jwt/jwt

go.sum
```diff
@@ -258,6 +258,7 @@ github.com/dchest/uniuri v0.0.0-20200228104902-7aecb25e1fe5 h1:RAV05c0xOkJ3dZGS0
 github.com/dchest/uniuri v0.0.0-20200228104902-7aecb25e1fe5/go.mod h1:GgB8SF9nRG+GqaDtLcwJZsQFhcogVCJ79j4EdT0c2V4=
 github.com/deepmap/oapi-codegen v1.6.0/go.mod h1:ryDa9AgbELGeB+YEXE1dR53yAjHwFvE9iAUlWl9Al3M=
 github.com/denisenkom/go-mssqldb v0.10.0/go.mod h1:xbL0rPBG9cCiLr28tMa8zpbdarY27NDyej4t/EjAShU=
+github.com/dgrijalva/jwt-go v3.2.0+incompatible h1:7qlOGliEKZXTDg6OTjfoBKDXWrumCAMpl/TFQ4/5kLM=
```
sanketsudake, line 261: This is deprecated library. Please use https://github.com/golang-jwt/jwt

sanketsudake (changes requested):

go.sum
praddy26: Fixed

go.mod
praddy26: Fixed

charts/fission-all/values.yaml
praddy26: Fixed

charts/fission-all/values.yaml
praddy26: Fixed

charts/fission-all/values.yaml
```diff
@@ -483,6 +482,37 @@ prometheus:
 canaryDeployment:
   enabled: false
 
+
```
sanketsudake, line 485: Can we add documentation for section above `authentication:` ?

Configuration options to enable authentication for fission function invocation via Fission router

pkg/fission-cli/cmd/token/command.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/spf13/cobra"
+
+	wrapper "github.com/fission/fission/pkg/fission-cli/cliwrapper/driver/cobra"
+	"github.com/fission/fission/pkg/fission-cli/flag"
+)
+
+func Commands() *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a JWT token",
```
sanketsudake, line 29: Create a JWT token for fission function invocation via Fission router

Note: Authentication should be enabled to use this feature.

sanketsudake (changes requested):

pkg/fission-cli/cmd/token/command.go
praddy26: Fixed

charts/fission-all/values.yaml
praddy26: Fixed

charts/fission-all/values.yaml
```diff
@@ -483,6 +482,38 @@ prometheus:
...
+  ## in seconds
+  ## default '120'
+  ##
+  jwtExpiryTime: 
+  ## jwtIssuer is the issuer of JWT
+  ## default 'fission'
+  ##
+  jwtIssuer: fission
+
+
```
sanketsudake, line 516: We should have a single blank line.

charts/fission-all/values.yaml
```diff
@@ -483,6 +482,38 @@ prometheus:
 canaryDeployment:
   enabled: false
 
+## Enable authentication for fission function invocation via Fission router
+##
+authentication:
```
sanketsudake, line 487: Please add relevant command to retrieve username and password if authencation is enabled. 

charts/fission-all/templates/NOTES.txt

sanketsudake (changes requested):

charts/fission-all/values.yaml
praddy26: Fixed

charts/fission-all/values.yaml
praddy26: Fixed

sanketsudake (approved):

PR #2293, opened 2022-01-03
Added ignorenotfound flag for all resources' deletion

### LINKED ISSUE #2108: Support resource deletion as successful when resource is not found
Currently while deleting any Fission CRD (like package, function) using Fission CLI, if the specified resource is not found then command exits with non zero error code.

We are looking for something similar to what we have in K8 where while deleting a resource if we pass `--ignore-not-found=true ` then in place of throwing non zero error code if resource is not found for deletion K8 skips the deletion of the resource and exits gracefully.

blackfly19:
## Description
Added ignorenotfound flag for all resources to avoid error when a resource is deleted.

Fixes #2108 

## Testing
I rebuilt the cli and tested the flag.

## Checklist:


pkg/fission-cli/flag/key/key.go
```diff
@@ -121,6 +121,7 @@ const (
 	EnvVersion         = "version"
 	EnvImagePullSecret = "imagepullsecret"
 	EnvExecutorType    = "executortype"
+	EnvIgnoreNotFound  = "ignoreNotFound"
```
sanketsudake, line 124: Lets define a single flag `IgnoreNotFound = "ignorenotfound"`
please note string should be "ignorenotfound". We won't be using camelcase in CLI options.

pkg/fission-cli/flag/flag.go
```diff
@@ -170,6 +170,7 @@ var (
 	EnvVersion                = Flag{Type: Int, Name: flagkey.EnvVersion, Usage: "Environment API version (1 means v1 interface)", DefaultValue: 1}
 	EnvImagePullSecret        = Flag{Type: String, Name: flagkey.EnvImagePullSecret, Usage: "Secret for Kubernetes to pull an image from a private registry"}
 	EnvExecutorType           = Flag{Type: String, Name: flagkey.EnvExecutorType, Usage: "Executor type of pod in environment; one of 'poolmgr', 'newdeploy', 'container'"}
+	EnvIgnoreNotFound         = Flag{Type: Bool, Name: flagkey.EnvIgnoreNotFound, Usage: "Skips the deletion of environment if it's not found", DefaultValue: false}
```
sanketsudake, line 173: This can be IgnoreNotFound.

Usage:

```
Avoid error if Fission resource not found during deletion
```

pkg/fission-cli/cmd/environment/delete.go
```diff
@@ -43,7 +43,11 @@ func (opts *DeleteSubCommand) do(input cli.Input) error {
 
 	err := opts.Client().V1().Environment().Delete(m)
 	if err != nil {
-		return errors.Wrap(err, "error deleting environment")
+		if input.Bool(flagkey.EnvIgnoreNotFound) {
+			return nil
+		} else {
+			return errors.Wrap(err, "error deleting environment")
+		}
```
sanketsudake, line 50: This would improve readability.
```
if err != nil {
         if !input.Bool(flagkey.EnvIgnoreNotFound) {
	        return errors.Wrap(err, "error deleting environment")
         }
         return nil
}
```

pkg/fission-cli/cmd/environment/delete.go
```diff
@@ -19,12 +19,12 @@ package environment
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/fission/fission/pkg/fission-cli/cliwrapper/cli"
 	"github.com/fission/fission/pkg/fission-cli/cmd"
 	flagkey "github.com/fission/fission/pkg/fission-cli/flag/key"
+	"github.com/pkg/errors"
```
sanketsudake, line 27: This should be in top imports.

sanketsudake (changes requested):

pkg/fission-cli/util/resourcedeletion.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import (
+	"strings"
+)
+
+func IsNotFound(err error) bool {
+	var result bool = strings.HasSuffix(strings.TrimSpace(err.Error()), "not found")
```
sanketsudake, line 8: Can we please just use return ?
```
return strings.HasSuffix(strings.TrimSpace(err.Error()), "not found")
```
No need of additional variable.

pkg/fission-cli/util/resourcedeletion.go
```diff
@@ -0,0 +1,10 @@
+package util
```
sanketsudake, line 1: Please add licence at top of file.
Please rename file to `errcheck.go`

pkg/fission-cli/flag/flag.go
```diff
@@ -76,6 +76,8 @@ var (
 
 	KubeContext = Flag{Type: String, Name: flagkey.KubeContext, Usage: "Kubernetes context to be used for the execution of Fission commands", DefaultValue: ""}
 
+	IgnoreNotFound = Flag{Type: Bool, Name: flagkey.IgnoreNotFound, Usage: "Avoid error if Fission resource not found during deletion", DefaultValue: false}
```
sanketsudake, line 79: Please check usage to .
```
Treat "resource not found" as a successful delete
```
Reference https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#delete

sanketsudake (changes requested):

sanketsudake (approved):

PR #2337, opened 2022-01-27
make graceperiod configurable for container functions

### LINKED ISSUE #2284: Make graceperiod configurable for container functions

Fission version above 1.14.1

**Describe the bug**
We have added support for Container functions in 1.14.1. 
Currently, the grace period for container functions is hardcoded. We should make it configurable.
https://github.com/fission/fission/blob/76f51e977de9118494b74f929c383f1392f1d8b8/pkg/executor/executortype/container/deployment.go#L170

In the case of the pool manager and new deploy, the grace period is handled at the environment level.
Whereas for container functions we need to have it function pod spec.

**Expected result**
The grace period should be configurable for container functions.

**Additional context**
We can utilize [podspec.terminationGracePeriodSeconds](https://doc.crds.dev/github.com/fission/fission/fission.io/Function/v1@v1.15.0#spec-podspec-terminationGracePeriodSeconds) to accept user value for container functions creation/update and set it accordingly during function pod creation.

Something similar to https://github.com/fission/fission/blob/76f51e977de9118494b74f929c383f1392f1d8b8/pkg/executor/executortype/poolmgr/gp_deployment.go#L55

Users should be able to pass grace period value for container functions from function create/update CLI.
The value should be accepted for container functions only.

shubham4443:
## Description

Added flag to configure grace period while creating container functions.
For negative input I've given a warning. Let me know if you want to throw an error instead. 0 is a valid input according to this - https://doc.crds.dev/github.com/fission/fission/fission.io/Function/v1@v1.15.0#spec-podspec-terminationGracePeriodSeconds. Also in the same doc it says default is 30 seconds. I believe it is a typo mistake. Shouldn't it be 360 seconds?

Fixes #2284 

## Checklist:

### DIFF
--- a/pkg/executor/executortype/container/deployment.go
+++ b/pkg/executor/executortype/container/deployment.go
@@ -172,6 +172,9 @@ func (cn *Container) getDeploymentSpec(ctx context.Context, fn *fv1.Function, ta
 	}
 
 	gracePeriodSeconds := int64(6 * 60)
+	if *fn.Spec.PodSpec.TerminationGracePeriodSeconds >= 0 {
+		gracePeriodSeconds = *fn.Spec.PodSpec.TerminationGracePeriodSeconds
+	}
 
 	podAnnotations := make(map[string]string)
 
--- a/pkg/fission-cli/cmd/function/command.go
+++ b/pkg/fission-cli/cmd/function/command.go
@@ -167,6 +167,7 @@ func Commands() *cobra.Command {
 			flag.FnCfgMap, flag.FnSecret,
 			flag.FnExecutionTimeout,
 			flag.FnIdleTimeout,
+			flag.FnTerminationGracePeriod,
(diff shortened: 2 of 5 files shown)


sanketsudake (approved):

sanketsudake:
@shubham4443  https://doc.crds.dev/github.com/fission/fission/fission.io/Function/v1@v1.15.0#spec-podspec-terminationGracePeriodSeconds. grace period mentioned here is from podspec, which we inherit from Kubernetes PodSpec. We really dont have liberty to change it here in docs. As CRD definitions are generated from CRD types.

shubham4443:
> 

Understood. I have mentioned 6 mins as the default value while giving the warning message.

sanketsudake:
Thanks @shubham4443 for the PR 🎉

PR #2341, opened 2022-02-01
CLI Check

### LINKED ISSUE #1183: [Feature] Fission CLI check
Inspired by Linkerd CLI, I propose we add a check command to Fission CLI which can act in two scenarios:

1) Before installing Fission into the cluster - the user should be able to do a "pre-check" and see what exists or does not in the cluster for Fission to work properly. Potentially it can also emit YAMLs and install from CLI itself

```
$ fission check --pre
kubernetes
--------------
√ kubernetes version is compatible

storage
------------------
√ cluster can provision a disk

```

2) After Fission is installed in the cluster - a check everything is healthy!

```
$ fission check
fission
--------------
√ all fission controllers running fine

storage
------------------
√ cluster can provision a disk

fission-version
---------------
√ can determine the latest version
√ cli is up-to-date
```

shubham4443:
## Description

Added a new command which checks whether fission can run inside the cluster properly `fission check --pre`. Additionally, it also checks whether everything is healthy after installing fission `fission check`

### Sample outputs
- `fission check --pre`
image

- `fission check`
image
I believe because I was running in dev mode, my current fission version is showing 0.0.0

- Some failed cases -
image
image

Fixes #1183 

## Checklist:

### DIFF
--- a/cmd/fission-cli/app/app.go
+++ b/cmd/fission-cli/app/app.go
@@ -10,6 +10,7 @@ import (
 	"github.com/fission/fission/pkg/fission-cli/cliwrapper/driver/cobra/helptemplate"
 	"github.com/fission/fission/pkg/fission-cli/cmd"
 	"github.com/fission/fission/pkg/fission-cli/cmd/canaryconfig"
+	"github.com/fission/fission/pkg/fission-cli/cmd/check"
 	"github.com/fission/fission/pkg/fission-cli/cmd/environment"
 	"github.com/fission/fission/pkg/fission-cli/cmd/function"
 	"github.com/fission/fission/pkg/fission-cli/cmd/httptrigger"
@@ -21,6 +22,7 @@ import (
 	"github.com/fission/fission/pkg/fission-cli/cmd/timetrigger"
 	"github.com/fission/fission/pkg/fission-cli/cmd/version"
 	"github.com/fission/fission/pkg/fission-cli/console"
+
 	"github.com/fission/fission/pkg/fission-cli/flag"
 	flagkey "github.com/fission/fission/pkg/fission-cli/flag/key"
 	"github.com/fission/fission/pkg/fission-cli/util"
@@ -81,7 +83,7 @@ func App() *cobra.Command {
 	groups = append(groups, helptemplate.CreateCmdGroup("Trigger Commands", httptrigger.Commands(), mqtrigger.Commands(), timetrigger.Commands(), kubewatch.Commands()))
 	groups = append(groups, helptemplate.CreateCmdGroup("Deploy Strategies Commands", canaryconfig.Commands()))
 	groups = append(groups, helptemplate.CreateCmdGroup("Declarative Application Commands", spec.Commands()))
-	groups = append(groups, helptemplate.CreateCmdGroup("Other Commands", support.Commands(), version.Commands()))
+	groups = append(groups, helptemplate.CreateCmdGroup("Other Commands", support.Commands(), version.Commands(), check.Commands()))
 	groups.Add(rootCmd)
 
 	flagExposer := helptemplate.ActsAsRootCommand(rootCmd, nil, groups...)
--- a/cmd/fission-cli/app/app.go
+++ b/cmd/fission-cli/app/app.go
@@ -1,3 +1,16 @@
+/*
+Copyright 2022 The Fission Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package app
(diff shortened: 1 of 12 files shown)


cmd/fission-cli/app/app.go
```diff
@@ -21,6 +22,7 @@ import (
 	"github.com/fission/fission/pkg/fission-cli/cmd/timetrigger"
 	"github.com/fission/fission/pkg/fission-cli/cmd/version"
 	"github.com/fission/fission/pkg/fission-cli/console"
+
```
sanketsudake, line 25: Please remove blank line.

cmd/fission-cli/app/app.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/fission/fission/pkg/fission-cli/cliwrapper/driver/cobra/helptemplate"
 	"github.com/fission/fission/pkg/fission-cli/cmd"
 	"github.com/fission/fission/pkg/fission-cli/cmd/canaryconfig"
+	"github.com/fission/fission/pkg/fission-cli/cmd/check"
```
sanketsudake, line 13: Looks like we are missing copyright in this file. Can you please add it ?
Reference: https://github.com/fission/fission/blob/3ac188124e348c2463a9e8b46bae9ab8967ce255/pkg/utils/gitrepo/gitrepo.go#L1

pkg/fission-cli/cmd/check/check.go
```diff
@@ -0,0 +1,37 @@
+package check
```
sanketsudake, line 1: Looks like we are missing copyright in this file. Can you please add it ?
Please do same for rest of new files.
Reference:

https://github.com/fission/fission/blob/3ac188124e348c2463a9e8b46bae9ab8967ce255/pkg/utils/gitrepo/gitrepo.go#L1

pkg/fission-cli/cmd/check/command.go
```diff
@@ -0,0 +1,21 @@
+package check
+
+import (
+	wrapper "github.com/fission/fission/pkg/fission-cli/cliwrapper/driver/cobra"
```
sanketsudake, line 4: Please fix this according to goimports standard.

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,233 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/fission/fission/pkg/controller/client"
+	"github.com/fission/fission/pkg/fission-cli/util"
```
sanketsudake, line 11: Please fix goimports

sanketsudake (changes requested):

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,247 @@
...
+
+	version, err := hc.kubeAPI.ServerVersion()
+	if err != nil {
+		return err
+	}
+
+	major, _ := strconv.Atoi(version.Major)
+	minor, _ := strconv.Atoi(version.Minor)
+	apiVersion := [3]int{major, minor, 0}
+	minAPIVersion := [3]int{1, 19, 0}
```
shubham4443, line 94: Minimum Kubernetes version required is hardcoded. I want to confirm whether this should go inside `pkg/apis/core/v1/const.go

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (hc *HealthChecker) CheckFissionVersion() error {
+	ver := util.GetVersion(hc.FissionClient)
+
+	major, _ := strconv.Atoi(strings.Split(ver.Client["fission/core"].Version, ".")[0])
+	minor, _ := strconv.Atoi(strings.Split(ver.Client["fission/core"].Version, ".")[1])
+	fix, _ := strconv.Atoi(strings.Split(ver.Client["fission/core"].Version, ".")[2])
+
+	apiVersion := [3]int{major, minor, fix}
+	minAPIVersion := [3]int{1, 15, 1}
```
shubham4443, line 131: Latest fission released version is hardcoded. I want to confirm whether this should go inside `pkg/apis/core/v1/const.go`

pkg/healthcheck/healthcheck.go
sanketsudake: I agree we should have constants for Kubernetes version. 
We can consider https://github.com/fission/fission/blob/master/pkg/fission-cli/util/constants.go

sanketsudake: This seems better place `pkg/apis/core/v1/const.go` Lets use `pkg/apis/core/v1/const.go`

pkg/healthcheck/healthcheck.go
sanketsudake: We should check client vs server version here. Ideally we should require hardcoding.
client version is embedded in build info, where as server info is fetched from fission.
Client version should match with server version.

sanketsudake: Please refere server and client versions from https://github.com/fission/fission/blob/3ac188124e348c2463a9e8b46bae9ab8967ce255/pkg/fission-cli/util/util.go#L172

sanketsudake (changes requested):

sanketsudake (approved):

PR #2360, opened 2022-02-14
Added function to sanitize strings

blackfly19:
## Description
Added a function to sanitize strings to fix security issues

## Checklist:

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -206,7 +206,7 @@ func DownloadUrl(ctx context.Context, httpClient *http.Client, url string, local
 	return nil
 }
 
-func SanitizeString(str string) string {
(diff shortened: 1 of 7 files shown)


pkg/utils/utils.go
```diff
@@ -205,3 +205,9 @@ func DownloadUrl(ctx context.Context, httpClient *http.Client, url string, local
 
 	return nil
 }
+
+func SanitizeString(str string) string {
```
sanketsudake, line 209: we can think of using replacer here https://pkg.go.dev/strings#NewReplacer
Could be common across function calls as seems to be safe for concurrent use by multiple goroutines.

sanketsudake (changes requested):

PR #2369, opened 2022-02-22
feature: Added commandline flags to set env var on the runtime and builder containers

### LINKED ISSUE #2297: fission environment create should allow adding environment variables to builder

The fission node builder allows pass through of the NODE_ENV variable, but there is no simple way of setting this variable.

`fission environment create --builder-env` cli option to allow setting static environment variables

praddy26:
## Description

We were able to set the env variables on the builder and runtime containers in the environment pod using spec file, but this was not possible when we created the environment using the following command 
`fission environment create`

This PR will add flags `--builder-env` and `--runtime-env` to the above command using which env variable can be set on the builder and runtime containers. These flags are optional and the env vars set using these flags cannot be updated.

Fixes #2297 

## Testing

The fission environment was tested using the following command
`fission env create --name nodejs --image fission/node-env:latest  --builder-env FISSION=serverless --builder-env INFRACLOUD=converge2021 --runtime-env NODE_ENV=prod --runtime-env AWS_REGION=us-east-1`

Then the environment object was inspected using the command `kubectl describe nodejs` and found that the env vars were set on the runtime and builder containers.
 
## Checklist:


pkg/fission-cli/util/util.go
```diff
@@ -372,6 +372,20 @@ func UpdateMapFromStringSlice(dataMap *map[string]string, params []string) bool
 	return updated
 }
 
+// UpdateEnvVarFromStringSlice parses key, val from "key=val" string array and updates passed []v1.EnvVar
+func UpdateEnvVarFromStringSlice(envVar []v1.EnvVar, params []string) bool {
```
sanketsudake, line 376: Should we redefine this function as 

```
func GetEnvVarFromStringS(params []string) (envVar []v1.EnvVar, error)
```

Also, please add unit tests for the function in util_test.go.

pkg/fission-cli/util/util.go
```diff
@@ -372,6 +372,20 @@ func UpdateMapFromStringSlice(dataMap *map[string]string, params []string) bool
...
+	updated := false
+	for idx, m := range params {
+		keyValue := strings.SplitN(m, "=", 2)
+		if len(keyValue) == 2 {
```
sanketsudake, line 380: Since we have mentioned  2 as an argument to `SplitN` function.
It will return 2 or <2 no of strings. We should have validation accordingly both for success/failure case.
Please add unit tests so that this gets validated.

pkg/fission-cli/util/util.go
```diff
@@ -372,6 +372,20 @@ func UpdateMapFromStringSlice(dataMap *map[string]string, params []string) bool
...
+			envVar[idx].Name = keyValue[0]
+			envVar[idx].Value = keyValue[1]
+			updated = true
```
sanketsudake, line 383: If we change function prototype this should not be true.
Essentially we should make this function as much as pure as possible. Avoid side effects.

pkg/fission-cli/flag/key/key.go
```diff
@@ -131,6 +131,8 @@ const (
 	EnvImagePullSecret = "imagepullsecret"
 	EnvExecutorType    = "executortype"
 	EnvForce           = force
+	EnvBuilder         = "builder-env"
```
sanketsudake, line 134: We mostlly have CLI arguments without `-` but I think this makes sense for the env argument.

sanketsudake (changes requested):

sanketsudake (approved):

sanketsudake:
@praddy26 could you please rebase with latest master ?

PR #2370, opened 2022-02-22
[helm chart] fission deployments resources

carlosrmendes:
Fix helm templating on `resources:` of router deployment and add `resources:` attribute to other fission deployments


sanketsudake:
Thanks @carlosrmendes for the PR. It looks neat to me. 🎉

PR #2378, opened 2022-03-08
Upgraded archiver

### LINKED ISSUE #2376: Upgrade archiver to v3.5.1
Fission is currently using [archiver@v2.1.0](https://github.com/mholt/archiver/). We need to upgrade it to v3.5.1. But there are some major changes made.

Match() function: It is no longer part of zipFormat struct. It's now part of Zip struct and it also a pointer receiver so we'll have to make changes accordingly. The argument type has changed from (filename string) to (file io.ReadSeeker).

Make() function: This function is no longer available in the library. We'll have to find a suitable replacement for it.

Open() function: It is no longer part of zipFormat struct. It's now part of Zip struct and it also a pointer receiver so we'll have to make changes accordingly. The argument type has changed from (src string, dst string) to (in io.Reader, size int64)

blackfly19:
## Description
Upgraded archiver package version to v3.5.1 and made changes accordingly.

Fixes #2376 

## Checklist:

### DIFF
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -28,7 +28,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/mholt/archiver"
+	"github.com/mholt/archiver/v3"
 	"github.com/pkg/errors"
 	uuid "github.com/satori/go.uuid"
 	"go.opencensus.io/plugin/ochttp"
@@ -352,7 +352,21 @@ func (fetcher *Fetcher) Fetch(ctx context.Context, pkg *fv1.Package, req Functio
 		}
 	}
 
-	if archiver.Zip.Match(tmpPath) && !req.KeepArchive {
+	file, err := os.Open(tmpPath)
+	if err != nil {
+		logger.Error("Error opening file",
+			zap.Error(err))
+		return http.StatusInternalServerError, err
+	}
+	defer file.Close()
+	match, err := archiver.DefaultZip.Match(file)
+	if err != nil {
+		logger.Error("Error matching file",
+			zap.Error(err))
+		return http.StatusInternalServerError, err
+	}
+
+	if match && !req.KeepArchive {
 		// unarchive tmp file to a tmp unarchive path
 		id, err := uuid.NewV4()
 		if err != nil {
@@ -377,7 +391,7 @@ func (fetcher *Fetcher) Fetch(ctx context.Context, pkg *fv1.Package, req Functio
 
 	// move tmp file to requested filename
 	renamePath := filepath.Join(fetcher.sharedVolumePath, req.Filename)
-	err := fetcher.rename(tmpPath, renamePath)
+	err = fetcher.rename(tmpPath, renamePath)
 	if err != nil {
 		logger.Error("error renaming file",
 			zap.Error(err),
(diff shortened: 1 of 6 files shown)


pkg/utils/utils.go
```diff
@@ -94,7 +94,7 @@ func MakeZipArchive(targetName string, globs ...string) (string, error) {
 	}
 
 	// zip up the file list
-	err = archiver.Zip.Make(targetName, files)
```
sanketsudake, line 97: https://github.com/mholt/archiver/blob/de0d89e255e17c8d75a40122055763e743ab0593/zip.go#L32


 ```
import archiver/zip

func MakeZip(zipPath string, filePaths []string) error {
	out, err := os.Create(zipPath)
	if err != nil {
		return fmt.Errorf("error creating %s: %v", zipPath, err)
	}
	defer out.Close()

w := zip.NewWriter(output)
	for _, fpath := range filePaths {
		if err := zipFile(w, fpath); err != nil {
			w.Close()
			return err
		}
	}

	return w.Close()
}
```

File: pkg/fetcher/fetcher.go [hunk 2]
```diff
@@ -621,14 +622,14 @@ func (fetcher *Fetcher) archive(src string, dst string) error {
...
+	return archiver.DefaultZip.Archive(files, dst)
 }
 
 // unarchive is a function that unzips a zip file to destination
 func (fetcher *Fetcher) unarchive(src string, dst string) error {
-	err := archiver.Zip.Open(src, dst)
+	err := archiver.DefaultZip.Unarchive(src, dst)
 	if err != nil {
-		return errors.Wrap(err, "failed to unzip file")
+		return fmt.Errorf("failed to unzip file: %v", err)
```
sanketsudake, line 632: `"failed to unzip file: %w"`

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -486,16 +489,18 @@ func localArchiveFromSpec(specDir string, aus *spectypes.ArchiveUploadSpec) (*fv
 	}
 
 	if len(files) > 1 || !isSingleFile {
-		// zip up the file list
-		archiveFile, err := os.CreateTemp("", fmt.Sprintf("fission-archive-%v", aus.Name))
+		// zip up the file list.Adding .zip because archive function expects that.
```
sanketsudake, line 492:
```
// Generate archive name with .zip extension and pack all files under it.
```

pkg/fission-cli/cmd/spec/apply.go
```diff
@@ -486,16 +489,18 @@ func localArchiveFromSpec(specDir string, aus *spectypes.ArchiveUploadSpec) (*fv
...
+		archiveFile, err := os.CreateTemp("", fmt.Sprintf("fission-archive-%v-*.zip", aus.Name))
 		if err != nil {
 			return nil, err
 		}
 		archiveFileName = archiveFile.Name()
-		err = archiver.Zip.Make(archiveFileName, files)
+		archiver.DefaultZip.OverwriteExisting = true
```
sanketsudake, line 498: Let's add a comment here, why we need to set it `true` and reset to `false` afterwards.

sanketsudake: I do see a issue here though, if we have a concurrent calls happening to `Archive` and we setting `archiver.DefaultZip.OverwriteExisting` to `true`

sanketsudake (changes requested):

sanketsudake (approved):

PR #2441, opened 2022-05-27
Ensure poolmanager fn address validation even if pod has active connections

### LINKED ISSUE #2438: Executor sending the IP of deleted pods to the Router

Fission version: 1.15.1
K8S version: v1.21.10-gke.2000

Google Kubernetes Engine

**Describe the bug**

Under above average load when routing requests to function pods, sometimes the executor will propose the IP for a function pod that has already been deleted.

We were seeing high load for one specific function, and while this was happening Fission was scaling up and down pods to accommodate these function executions. And while this was happening, some of these pods failed their liveness checks and were deleted by K8S.

Long after the pod was deleted, the executor would continue to serve the deleted pod's IP to it's routers. This would cause the routers to timeout and ultimately fail the upstream function call.

In this screenshot we see the bad IP in question being served, a readiness probe failure, K8S killing the pod and finally the router continuing to try and access this IP.

image

This situation continued until we restarted the executor and it cleared it's internal cache.

**To Reproduce**

We attempted to reproduce this in a small scale by deleting a pod with `--force --grace-period=0` but were not able to reproduce it. However, under sustained load all day the affected function ended up with 10 or so bad IPs in it's cache that were not cleared until we restarted the executor.

**Expected result**

We would have expected that the executor would no longer serve the IP of a dead pod after it was killed. The cached entries should have been purged when the pod was removed.

**Actual result**

The following short circuit (`active > 1 ||`) allows the router to continue sending bad IPs:

https://github.com/fission/fission/blob/90c479b23cd5e90339788cf286f721fda45e223f/pkg/executor/api.go#L80

In the case where the executor believes a pod is processing a request it avoids checking to see if the pod is still valid.

blackfly19:
## Description
Executor sending invalid ips because validation function is getting skipped as the active number of requests allow it to pass the condition. 
This error occurs when a pod is killed but the untap function is not called which influences the value of active variable.

Fixes #2438

## Checklist:

### DIFF
--- a/pkg/executor/api.go
+++ b/pkg/executor/api.go
@@ -77,7 +77,7 @@ func (executor *Executor) getServiceForFunctionAPI(w http.ResponseWriter, r *htt
 		if err == nil {
 			// if a pod is already serving request then it already exists else validated
 			logger.Debug("from cache", zap.Int("active", active))
-			if active > 1 || et.IsValid(ctx, fsvc) {
+			if et.IsValid(ctx, fsvc) {
 				// Cached, return svc address
 				logger.Debug("served from cache", zap.String("name", fsvc.Name), zap.String("address", fsvc.Address))
 				executor.writeResponse(w, fsvc.Address, fn.ObjectMeta.Name)


sanketsudake:
@mergifyio rebase

PR #2461, opened 2022-06-22
Added support to set builder and fn pod specs via helm chart

### LINKED ISSUE #1820: Feature Request: Override Environment settings from newdeploy function spec
I will  have many functions that accesses NFS volumeMount resources with permissions that vary between user/group. I would like to be able to deploy a function that can set the `securityContext` "user", "group", "supplementalGroups" values so that the function runs with the correct permissions for the data on the mounted volumes. Currently both the poolmgr and newdeploy executor type functions rely on a named reference to a shared Environment. This means the `securityContext` can only be defined at the environment and every function either has to use that named env, or they would have to deploy unique copies of the environment for every new variation of podSpec values. 
It seems unwieldy to have to maintain many Environment specs and use the "name" with the right securityContext to match the desired function. 

Because a newdeploy executor results in a discrete deployment instance for each function, it would make sense that the function would have access to override all of the values of the named environment. Or to even fully define its own environment (less of a priority, but seems to be a related feature). With this ability, I would be able to set the `securityContext` on a per-function basis. This also seems to make sense as a feature as being in the spirit of the newdeploy executor and making it entirely self-contained instead of coupled to a environment spec.

**Describe alternatives you've considered**
The only alternative I can see for working around my problem is to manage the creation of a new copy of the environment with a unique name that pairs with each function that needs a different user/group/supplementaryGroups

**Additional context**
For some additional background on the nature of this problem, I work for a Visual Effects studio with a primary reliance on NFS mounted filesystems. These filesystems enforce group level permissions that map to projects. That is, if I do not have access to Lord of the Rings, I should not be able to access files in the Lord of the Rings project group. If we are deploying functions that expect to operate on project data, we need to make sure the user in the project, deploying the function, can set their group permissions for the container.

### LINKED ISSUE #2401: fission settings to avoid running as root across fission components and functions pods

fission 1.15.1
kubernetes 1.20

k3s

**Describe the bug**

The fetcher container runs as root, and as such writes files as root, which requires functions to run as root.

**To Reproduce**

Run any sample function

**Expected result**

The function runs as a non-root user.

**Actual result**
The function runs as the root user

**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

It is not acceptable to run containers as root, generally. Fission should be able to operate without root. Minimally the files extracted by fetcher should be readable by non-root user (this is not the case). Through my environment configuration I have marked the pod as `runAsNonRoot` and also specified `runAsUser`, and `runAsGroup` with success.

blackfly19:
## Description
The users can now set the pod spec for builder and fn pods via helm chart.
Currently we have set some default securitycontext for the pods. Before there were no permissions set and the user would by default enter root when `kubectl exec` into pod. Now the permissions have been set and the user will not be able to access root directory in poolmgr and newdeploy pods.

Fixes #2401
Fixes #1820 

## Testing
Added unit tests and did some manual testing.

## Checklist:

### DIFF
--- a/charts/fission-all/templates/executor/configmap.yaml
+++ b/charts/fission-all/templates/executor/configmap.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: podspecpatch
+  namespace: {{ .Release.Namespace }}
+data:
+  spec: |
+    securityContext:
+      fsGroup: 10001
+      runAsGroup: 10001
+      runAsNonRoot: true
+      runAsUser: 10001
--- a/charts/fission-all/templates/executor/configmap.yaml
+++ b/charts/fission-all/templates/executor/configmap.yaml
@@ -1,8 +1,9 @@
+{{- if .Values.runtimePodSpec.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: runtime-podspec-patch
-  namespace: {{ .Release.Namespace }}
 data: 
   spec: |
-    {{- toYaml .Values.podSpec | nindent 4 }}
+    {{- toYaml .Values.runtimePodSpec.podSpec | nindent 4 }}
+{{- end -}}
--- a/charts/fission-all/templates/executor/configmap.yaml
+++ b/charts/fission-all/templates/executor/configmap.yaml
@@ -1,12 +1,8 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: podspecpatch
+  name: runtime-podspec-patch
   namespace: {{ .Release.Namespace }}
-data:
+data: 
   spec: |
-    securityContext:
-      fsGroup: 10001
-      runAsGroup: 10001
-      runAsNonRoot: true
-      runAsUser: 10001
+    {{- toYaml .Values.podSpec | nindent 4 }}
--- a/charts/fission-all/templates/executor/deployment.yaml
+++ b/charts/fission-all/templates/executor/deployment.yaml
@@ -59,6 +59,8 @@ spec:
           value: {{ .Values.pprof.enabled | quote }}
         - name: HELM_RELEASE_NAME
           value: {{ .Release.Name | quote }}
+        - name: FISSION_NAMESPACE
+          value: {{ .Release.Namespace }}
         {{- include "opentracing.envs" . | indent 8 }}
         {{- include "opentelemtry.envs" . | indent 8 }}
         resources:
--- a/charts/fission-all/templates/executor/deployment.yaml
+++ b/charts/fission-all/templates/executor/deployment.yaml
@@ -59,8 +59,6 @@ spec:
           value: {{ .Values.pprof.enabled | quote }}
         - name: HELM_RELEASE_NAME
           value: {{ .Release.Name | quote }}
-        - name: FISSION_NAMESPACE
-          value: {{ .Release.Namespace }}
         {{- include "opentracing.envs" . | indent 8 }}
         {{- include "opentelemtry.envs" . | indent 8 }}
         resources:
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -745,13 +745,17 @@ pprof:
 ## Enable runtimePodSpec and add spec to your poolmgr or newdeploy pods
 ##
 runtimePodSpec:
-  enabled: true
+
+  ## Setting it false by default so that integration tests pass
+  ##
+  enabled: false
+  
   ## Checkout PodSpec in https://fission.io/docs/reference/crd-reference/#runtime
   ##
-  
   podSpec:
-  ## Default podspec to improve security of the pods
-  ##
+
+    ## Default podspec to improve security of the pods
+    ##
     securityContext:
       fsGroup: 10001
       runAsGroup: 10001
@@ -761,13 +765,17 @@ runtimePodSpec:
 ## Enable builderPodSpec and add spec to your env builder pods
 ##
 builderPodSpec:
-  enabled: true
+
+  ## Setting it false by default so that integration tests pass
+  ##
+  enabled: false
+
   ## Checkout PodSpec in https://fission.io/docs/reference/crd-reference/#builder
   ##
-  
   podSpec:
-  ## Default podspec to improve security of the pods
-  ##
+
+    ## Default podspec to improve security of the pods
+    ##
     securityContext:
       fsGroup: 10001
       runAsGroup: 10001
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -743,14 +743,34 @@ pprof:
   enabled: false
 
 ## Enable runtimePodSpec and add spec to your poolmgr or newdeploy pods
+##
 runtimePodSpec:
   enabled: true
-## Refer https://doc.crds.dev/github.com/fission/fission/fission.io/Environment/v1@v1.16.0#spec-runtime-podspec
-## To write podspec
+  ## Checkout PodSpec in https://fission.io/docs/reference/crd-reference/#runtime
+  ##
+  
   podSpec:
-## Default podspec to improve security of the pods
+  ## Default podspec to improve security of the pods
+  ##
     securityContext:
       fsGroup: 10001
       runAsGroup: 10001
       runAsNonRoot: true
       runAsUser: 10001
+
+## Enable builderPodSpec and add spec to your env builder pods
+##
+builderPodSpec:
+  enabled: true
+  ## Checkout PodSpec in https://fission.io/docs/reference/crd-reference/#builder
+  ##
+  
+  podSpec:
+  ## Default podspec to improve security of the pods
+  ##
+    securityContext:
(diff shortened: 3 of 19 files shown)


charts/fission-all/templates/executor/configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: podspecpatch
```
sanketsudake, line 4: runtime-podspec-patch

charts/fission-all/templates/executor/configmap.yaml
```diff
@@ -0,0 +1,12 @@
...
+  namespace: {{ .Release.Namespace }}
+data:
+  spec: |
+    securityContext:
```
sanketsudake, line 8: This section should come from helm values

sanketsudake (changes requested):

charts/fission-all/templates/executor/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: runtime-podspec-patch
+  namespace: {{ .Release.Namespace }}
```
sanketsudake, line 5: Remove namespace from here.

charts/fission-all/templates/executor/deployment.yaml
```diff
@@ -59,6 +59,8 @@ spec:
           value: {{ .Values.pprof.enabled | quote }}
         - name: HELM_RELEASE_NAME
           value: {{ .Release.Name | quote }}
+        - name: FISSION_NAMESPACE
```
sanketsudake, line 62: Rename variable to POD_NAMESPACE.

pkg/executor/util/util.go
```diff
@@ -113,3 +115,24 @@ func ConvertConfigSecrets(ctx context.Context, fn *fv1.Function, kc kubernetes.I
 	}
 	return envFromSources, nil
 }
+
+func GetSpecConfigMap(ctx context.Context, kubeClient kubernetes.Interface, podSpec apiv1.PodSpec) (*apiv1.PodSpec, error) {
+	var additionalSpec apiv1.PodSpec
+
+	podSpecPatch, err := kubeClient.CoreV1().ConfigMaps(os.Getenv("FISSION_NAMESPACE")).Get(ctx, "runtime-podspec-patch", metav1.GetOptions{})
```
sanketsudake, line 122: Define constants for configmap name.

sanketsudake: We should not fail if configmap not found.

charts/fission-all/templates/executor/deployment.yaml
sanketsudake, line 62: Please remove this we not being used.

charts/fission-all/values.yaml
```diff
@@ -741,3 +741,12 @@ mqt_keda:
 ##
 pprof:
   enabled: false
+
+runtimePodSpec:
```
sanketsudake, line 745: Please add documentation in detail.

1. Significance of this function
2. Purpose of enable/disable
3. Reference for podSpec, provide some documentation link as we are considering specific structure..

pkg/executor/executor.go
```diff
@@ -270,6 +272,16 @@ func StartExecutor(ctx context.Context, logger *zap.Logger, functionNamespace st
 
 	executorInstanceID := strings.ToLower(uniuri.NewLen(8))
 
+	body, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
sanketsudake, line 275: define a function `GetCurrentNamespace() (string, error)`

sanketsudake: `GetPodSpecFromConfigMap(cmName, cmNs) (*podSpec, error)`

pkg/executor/util/util.go
```diff
@@ -113,3 +114,24 @@ func ConvertConfigSecrets(ctx context.Context, fn *fv1.Function, kc kubernetes.I
...
+
+func GetSpecConfigMap(ctx context.Context, kubeClient kubernetes.Interface, podSpec apiv1.PodSpec, podSpecConfigMap *apiv1.ConfigMap) (*apiv1.PodSpec, error) {
+
+	if podSpecConfigMap == nil {
+		return nil, nil
+	}
+
+	var additionalSpec apiv1.PodSpec
+
+	err := yaml.Unmarshal([]byte(podSpecConfigMap.Data["spec"]), &additionalSpec)
```
sanketsudake, line 126: This should be done at start of executor code only.

sanketsudake (changes requested):

pkg/executor/executor.go
```diff
@@ -270,6 +270,16 @@ func StartExecutor(ctx context.Context, logger *zap.Logger, functionNamespace st
 
 	executorInstanceID := strings.ToLower(uniuri.NewLen(8))
 
+	namespace, err := utils.GetCurrentNamespace()
+	if err != nil {
+		logger.Warn("Current namespace not found %v", zap.Error(err))
+	}
+
+	podSpecPatch, err := util.GetSpecFromConfigMap(ctx, kubernetesClient, fv1.RuntimePodSpecConfigmap, namespace)
```
sanketsudake, line 278: we are still executing this function even if error is returned by GetCurrentNamespace.

If GetCurrentNamespace returns error we should not execut this function.

pkg/executor/executortype/poolmgr/gp_deployment.go
```diff
@@ -150,6 +150,15 @@ func (gp *GenericPool) genDeploymentSpec(env *fv1.Environment) (*appsv1.Deployme
 		},
 	}
 
+	updatedPodSpec, err := util.CheckAndMergeSpec(pod.Spec, gp.podSpec)
```
sanketsudake, line 153: We can write this as
```
if gp.podSpec != nil {
        updatedPodSpec, err := MergePodSpec(pod.Spec, gp.podSpec)
      if err == nil {
              pod.Spec = *updatedPodSpec
      } else {
            // dump warning.
     }
}
I think we can get rid of CheckAndMergeSpec function.

**Review by sanketsudake — CHANGES_REQUESTED** [review]

File: pkg/buildermgr/envwatcher.go [hunk 11]
```diff
@@ -503,6 +506,16 @@ func (envw *environmentWatcher) createBuilderDeployment(env *fv1.Environment, ns
 		},
 	}
 
+	if envw.podSpec != nil {
```
sanketsudake, line 509: Please rename to podSpecPatch

charts/fission-all/values.yaml
```diff
@@ -741,3 +741,36 @@ mqt_keda:
 ##
 pprof:
   enabled: false
+
+## Enable runtimePodSpec and add spec to your poolmgr or newdeploy pods
+##
+runtimePodSpec:
+  enabled: true
+  ## Checkout PodSpec in https://fission.io/docs/reference/crd-reference/#runtime
```
sanketsudake, line 749: Please change indentation level as discussed and make changes to match style

pkg/executor/executor.go
```diff
@@ -270,6 +271,17 @@ func StartExecutor(ctx context.Context, logger *zap.Logger, functionNamespace st
 
 	executorInstanceID := strings.ToLower(uniuri.NewLen(8))
 
+	var podSpecPatch *apiv1.PodSpec
+	namespace, err := utils.GetCurrentNamespace()
+	if err != nil {
+		logger.Warn("Current namespace not found %v", zap.Error(err))
```
sanketsudake, line 277: Please check return type Error function, and use "%s" if required.

pkg/executor/util/util.go
```diff
@@ -113,3 +114,20 @@ func ConvertConfigSecrets(ctx context.Context, fn *fv1.Function, kc kubernetes.I
...
+func GetSpecFromConfigMap(ctx context.Context, kubeClient kubernetes.Interface, cm string, cmns string) (*apiv1.PodSpec, error) {
+
+	podSpecPatch, err := kubeClient.CoreV1().ConfigMaps(cmns).Get(ctx, cm, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	var additionalSpec apiv1.PodSpec
+
+	err = yaml.Unmarshal([]byte(podSpecPatch.Data["spec"]), &additionalSpec)
```
sanketsudake, line 127: We can direction return from here, no need of additional error check.

pkg/utils/utils.go
```diff
@@ -214,3 +215,14 @@ func IsZip(filename string) (bool, error) {
 	defer f.Close()
 	return archiver.DefaultZip.Match(f)
 }
+
+//Returns namespace where fission is deployed.
+func GetCurrentNamespace() (string, error) {
+
+	//This file contains the namespace and can be found in each container.
```
sanketsudake, line 222: add space after `// `

sanketsudake (changes requested):

sanketsudake (approved):

Issue #2460, opened 2022-06-22
Archive CLI returning invalid object URL with S3 bucket


blackfly19:
**Fission/Kubernetes version**

<pre>
$ fission --version
$ kubectl version
</pre>

**Kubernetes platform (e.g. Google Kubernetes Engine)**

**Describe the bug**
Archive CLI provides a way to get the url of an object stored using storageSvc. In case of S3 bucket I am unable to access the object with the URL returned from using the command `fission ar get-url`.

**To Reproduce**
Use Aws S3 in place of local storage. Upload files using the CLI and then get its URL. Use the URL to access the object.

**Expected result**
It should be able to retrieve the object.

**Actual result**
It returns an error of an invalid URL and the object is not retrieved.

**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

PR #2465, opened 2022-06-29
Added fix to retrieve correct s3 urls

blackfly19:
## Description
With this fix, you'll get s3 urls generated via pattern instead of getting them from the stow client. The s3 urls generated by stow are for the it's internal uses only and they can't be used to access the object externally.

It generates object links using this pattern : `https://<bucket-name>.s3.amazonaws.com/<filename>`

Fixes #

## Testing
- Did an e2e test using fission specs which is working.
- Bucket name is a must otherwise storagesvc pod crashes so the url will never be incomplete.
- The url is independent of region so even if you don't mention it neither the pod nor the url is affected.
- Tested if it returns an error if the object does not exist.

## Checklist:


blackfly19:
Temporary solution to issue #2460.

sanketsudake:
@mergifyio rebase

sanketsudake (approved):

PR #2539, opened 2022-09-11
Allow namespace configuration for different CRD resources in Fission

sanketsudake:
## Description
These changes will allow to configure CRD resources according to namespaces. Previously CRD could have only be configured in default namespace provided by user in values.yaml file under defaultNamespace field.

Helm Changes
We have added two new field in helm.
1. additionalFissionNamespaces: [] - This will be a type of array, where user can provide multiple namespaces.
    eg-
   ```
    additionalFissionNamespaces: 
     - namespace1
     - namespace2
     - namespace3
      ```
2. singleDefaultNamespace: boolean - This will be a boolean value and default value will be true. If this field sets to true then fission will always use only default namespace provided by user under defaultNamespace field. If this set to false then fission will use namespaces defined in additionalFissionNamespaces field along with defaultNamespace.

For example - if values in helm defined something like this, then fission will use default, namespace1, namespace2 and namespace3.
```
defaultNamespace: default
singleDefaultNamespace: false
additionalFissionNamespaces: 
 - namespace1
 - namespace2
 - namespace3
 ``` 

Fixes #

## Checklist:

### DIFF
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -71,3 +71,12 @@ This template generates the image name for the deployment depending on the value
 - name: OTEL_PROPAGATORS
   value: "{{ .Values.openTelemetry.propagators }}"
 {{- end }}
+
+{{- define "additional-namespace.envs" }}
+- name: FISSION_RESOURCE_NAMESPACES
+{{- if not .Values.singleDefaultNamespace }}
+  value: "{{ .Values.defaultNamespace }},{{ join "," .Values.additionalFissionNamespaces }}"
+{{- else }}
+  value: {{ .Values.defaultNamespace }}  
+{{- end }}
+{{- end }}
--- a/charts/fission-all/templates/_helpers.tpl
+++ b/charts/fission-all/templates/_helpers.tpl
@@ -72,7 +72,7 @@ This template generates the image name for the deployment depending on the value
   value: "{{ .Values.openTelemetry.propagators }}"
 {{- end }}
 
-{{- define "additional-namespace.envs" }}
+{{- define "fission-resource-namespace.envs" }}
 - name: FISSION_RESOURCE_NAMESPACES
 {{- if not .Values.singleDefaultNamespace }}
   value: "{{ .Values.defaultNamespace }},{{ join "," .Values.additionalFissionNamespaces }}"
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

File: charts/fission-all/templates/buildermgr/deployment.yaml [hunk 1]
```diff
@@ -55,6 +55,10 @@ spec:
           value: {{ .Values.pprof.enabled | quote }}
         - name: HELM_RELEASE_NAME
           value: {{ .Release.Name | quote }}
+        {{- if not .Values.singleDefaultNamespace }}
+        - name: ADDITIONAL_NAMESPACES
```
sanketsudake, line 59: `FISSION_RESOUCE_NAMESPACES`

charts/fission-all/templates/buildermgr/deployment.yaml
```diff
@@ -55,6 +55,10 @@ spec:
...
+          value: {{ join "," .Values.additionalFissionNamespaces | quote }}
```
sanketsudake, line 60: This should consider default namespace also.

charts/fission-all/templates/_helpers.tpl
```diff
@@ -71,3 +71,12 @@ This template generates the image name for the deployment depending on the value
 - name: OTEL_PROPAGATORS
   value: "{{ .Values.openTelemetry.propagators }}"
 {{- end }}
+
+{{- define "additional-namespace.envs" }}
```
sanketsudake, line 75: please name this section `fission-resource-namespace.envs`.

charts/fission-all/values.yaml
```diff
@@ -75,6 +75,15 @@ builderNamespace: fission-builder
 ##
 defaultNamespace: default
 
+## If true, fission will only watch for fission components created in the `defaultNamespace` above
```
sanketsudake, line 78: Please change documentation accordingly for default namespace as well.

charts/fission-all/values.yaml
```diff
@@ -75,6 +75,15 @@ builderNamespace: fission-builder
...
+##
+singleDefaultNamespace: true
+
+## Fission will watch the following namespaces along with the `defaultNamespace` for fission components.
+## Only works if `singleDefaultNamespace` is false.
```
sanketsudake, line 83: Please mention samples for `additionalFissionNamespaces` parameter.

PR #2543, opened 2022-09-14
Add ability to configure object reaper interval

### LINKED ISSUE #2464: New environment variable for executor container
Whenever user create a function with executor type new deploy and container by providing min scale property to 0, then a function name doIdleObjectReaper() runs after every 5 second(hard coded in the function) to check the idle timeout of function that user created and once there is idle timeout then doIdleObjectReaper() scale down the replica of deployment.

So instead of calling this function doIdleObjectReaper() after every 5 second, it will be better to make this configurable in helm chart.

Create a new environment variable for executor container named IDLE_REAPER_TIMEOUT with numeric value, value will be consider in seconds. Set initial value as 5. 

Later on we can change this with the help of helm chart. Now the doIdleObjectReaper() will be call as per the value get fetch by environment variable.

**Additional context**
These changes will only be applicable for executor type as new deploy and container.
Changes needs to me make in function idleObjectReaper() for new deploy and container both.

Tri0L:
## Description

Added properties to configure object reaper interval, global and specific to exec type.

`OBJECT_REAPER_INTERVAL` - global
`NEWDEPLOY_OBJECT_REAPER_INTERVAL` - for `newdeploy` type
`CONTAINER_OBJECT_REAPER_INTERVAL` - for `container` type
`POOLMGR_OBJECT_REAPER_INTERVAL` - for `poolmgr`

Fixes #2464

## Checklist:


pkg/executor/executortype/container/containermgr.go
```diff
@@ -781,3 +782,34 @@ func getDeploymentObj(kubeobjs []apiv1.ObjectReference) *apiv1.ObjectReference {
 	}
 	return nil
 }
+
+func getObjectReaperInterval(logger *zap.Logger, defaultReaperInterval int) int {
```
sanketsudake, line 786: Rather than defining this function at all three executor types, I am just thinking we can write this in better fashion and avoid code duplication.


```
func getObjectReaperInterval(logger *zap.Logger, executorType  ExecutorType, defaultReaperInterval int) int
```

We can pass these parameters at the time of function call. Defined in `pkg/apis/core/v1`

```
const (
	ExecutorTypePoolmgr   ExecutorType = "poolmgr"
	ExecutorTypeNewdeploy ExecutorType = "newdeploy"
	ExecutorTypeContainer ExecutorType = "container"
)
```

Variable names for particular executor types can be constructed at runtime.

https://go.dev/play/p/XDAOUeQU5kM
 So this can generalize code. 

Tomorrow if we think of defining variables at each executor level we can use the same approach.

sanketsudake (changes requested):

pkg/executor/executortype/container/containermgr.go
Tri0L: Absolutely agree. I was confused by the fact that the ISTIO settings are duplicated in each separately.
What do you think about putting my method in pkg/utils/utils.go?

sanketsudake: probably `pkg/executor/util/util.go` would be good place for this IMO. Since function is executor specific.

shubham-bansal96: @Tri0L 
You will need to make changes in helm charts as well, so user can set the values for the same while installing fission.
POOLMGR_OBJECT_REAPER_INTERVAL
NEWDEPLOY_OBJECT_REAPER_INTERVAL
CONTAINER_OBJECT_REAPER_INTERVAL
OBJECT_REAPER_INTERVAL

Tri0L: @shubham-bansal96 Yes, I will add them. I just wanted to make sure that my edits basically make sense and can be accepted.

pkg/executor/executortype/newdeploy/newdeploymgr.go
```diff
@@ -132,7 +133,7 @@ func MakeNewDeploy(
 		useIstio:               enableIstio,
 
 		defaultIdlePodReapTime:     2 * time.Minute,
-		objectReaperIntervalSecond: time.Duration(getObjectReaperInterval(logger, 5)) * time.Second,
+		objectReaperIntervalSecond: time.Duration(executorUtils.GetObjectReaperInterval(logger, fv1.ExecutorTypeContainer, 5)) * time.Second,
```
shubham-bansal96, line 136: fv1.ExecutorTypeContainer, please change it to fv1.ExecutorTypeNewdeploy

pkg/executor/executortype/poolmgr/gpm.go
```diff
@@ -156,7 +157,7 @@ func MakeGenericPoolManager(
 		enableIstio:                enableIstio,
 		poolPodC:                   poolPodC,
 		podSpecPatch:               podSpecPatch,
-		objectReaperIntervalSecond: time.Duration(getObjectReaperInterval(logger, 5)) * time.Second,
+		objectReaperIntervalSecond: time.Duration(executorUtils.GetObjectReaperInterval(logger, fv1.ExecutorTypeContainer, 5)) * time.Second,
```
shubham-bansal96, line 160: change here as well, fv1.ExecutorTypeContainer to fv1.ExecutorTypePoolmgr

pkg/executor/executortype/container/containermgr.go
shubham-bansal96: @Tri0L your changes looks good, Please go ahead with helm chart changes.

pkg/executor/executortype/newdeploy/newdeploymgr.go
Tri0L: Fixed

pkg/executor/executortype/container/containermgr.go
Tri0L: I added helm chart settings, and tests

charts/fission-all/values.yaml
```diff
@@ -155,6 +155,26 @@ executor:
     runAsUser: 10001
     runAsGroup: 10001
 
+  ## Object Reaper
+  ## objectReaperInterval represents GLOBAL interval to run process that reaps objects after certain idle time.
+  ## Also you can set different objectReaperInterval for specific executor type. See poolmgs/newdeploy/container section
+  ## Default: 5s
+  ##
+  objectReaperInterval: 5s
```
shubham-bansal96, line 163: Please change objectReaperInterval value from 5s to 5.

charts/fission-all/values.yaml
```diff
@@ -155,6 +155,26 @@ executor:
...
+
+  poolmgr: {}
+    ## objectReaperInterval specific to poolmgr executor type
+    ##
+    ## objectReaperInterval: 5s
```
shubham-bansal96, line 168: change here, newdeploy  and in container as well.

charts/fission-all/values.yaml
```diff
@@ -155,6 +155,26 @@ executor:
...
     runAsGroup: 10001
 
+  ## Object Reaper
+  ## objectReaperInterval (seconds) represents GLOBAL interval to run process that reaps objects after certain idle time.
+  ## Also you can set different objectReaperInterval for specific executor type. See poolmgs/newdeploy/container section
+  ## Default: 5
+  ##
+  objectReaperInterval: 5
+
+  poolmgr: {}
```
shubham-bansal96, line 165: instead of creating poolmgr, newdeploy and container as list type, we can create this as simple field name as poolmgrObjReaperInterval, newdeployObjReaperInterval, containerObjReaperInterval with default value empty.

poolmgrObjReaperInterval: ""
newdeployObjReaperInterval: ""
containerObjReaperInterval: ""

shubham-bansal96 (changes requested):

charts/fission-all/values.yaml
Tri0L: This is for the future, most likely executors will have more variables.
For example `defaultIdlePodReapTime`. `useIstio` also good candidate to move from code to config files.

Tri0L:
@sanketsudake @shubham-bansal96 could you please helm me with CI Lint and Unit tests. Why this step is failing?

pkg/executor/util/util.go
```diff
@@ -128,3 +133,39 @@ func GetSpecFromConfigMap(ctx context.Context, kubeClient kubernetes.Interface,
 
 	return &additionalSpec, err
 }
+
+func GetObjectReaperInterval(logger *zap.Logger, executorType fv1.ExecutorType, defaultReaperInterval int) int {
+
```
sanketsudake, line 138: Line no 141-151 and line no 155-164 seems almost duplicate code.

Probably we can refactor code here a bit.

Let's say we define function.
```
func getValueFromEnv(envVar string) (int, error)
```

Then in `func GetObjectReaperInterval(****) int`
1. We first call `getValueFromEnv(getExecutorEnvVariableName(executorType))`, if err not nil return value fetched
2. Then we can getValueFromEnv(""OBJECT_REAPER_INTERVAL"), if err not nil return value fetched.
3. Otherwise return the default value.

This would make code bit easy to read.

sanketsudake (changes requested):

shubham-bansal96:
> @sanketsudake @shubham-bansal96 could you please help me with CI Lint and Unit tests. Why this step is failing?

So this basically means that your import in go file is not sorted, please check util_test.go file. Try sorting that one and hopefully i should work.

pkg/executor/util/util.go
Tri0L: What do you think if I create `func getUIntValueFromEnv(envVar string) (uint, error)`, and check that env var exists and can be parsed to uint in this function? Looks like good util method.

Or use something like this https://pkg.go.dev/go.deanishe.net/env@v0.5.1#GetUint

sanketsudake: Yes, this is also interesting. 

In that case, it would be a very simple and easy call,

```
return GetUint(getExecutorEnvVariableName(executorType), GetUint("OBJECT_REAPER_INTERVAL", defaultValue))
```

However, if we can just write a function of the same essence for this purpose that would be great, rather than having another import.

`GetUIntValueFromEnv` is better name considering function same as `GetUint`.

Tri0L: Did it.

sanketsudake (approved):

PR #2559, opened 2022-09-28
Fix for preupgrade checks of CRDs during fission upgrade

shubham-bansal96:
Fission fails to upgrade because of preupgrade check of CRDs. Fission has a logic of pre-upgrade checks of CRD and this logic works while upgrading fission, In case of fresh install of fission, it works fine.

Fixes #

## Checklist:


shubham-bansal96:
Changes not needed. Need to make sure the process of fission upgrade.
That is, upgrade helm repo first and then apply CRDs.

PR #2560, opened 2022-09-29
Separate service accounts for each component

runitmisra:
## Description

Changes to have separate service accounts for each fission component

## Checklist:


charts/fission-all/templates/fluentbit/fluentbit.yaml
```diff
@@ -173,7 +173,6 @@ spec:
             - name: fluentbit-config
               mountPath: /fluent-bit/etc/
               readOnly: true
-      serviceAccountName: fission-svc
```
sanketsudake, line 176: This pod containers logger container, requires a pod get/list/watch permissions.

I see function log test failed.

```
2022-10-04T06:55:34.546352417Z stderr F E1004 06:55:34.546231       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.25.2/tools/cache/reflector.go:169: Failed to watch *v1.Pod: failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:fission:default" cannot list resource "pods" in API group "" at the cluster scope
```

charts/fission-all/templates/pre-upgrade-checks/clusterrole.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-fission-cr-preupgrade
+  annotations:
+    helm.sh/hook: pre-install
```
sanketsudake, line 6: Since deployment has pre-upgrade hook, we should have a pre-upgrade hook only here and across files in pre-upgrade-check folder.

charts/fission-all/templates/pre-upgrade-checks/pre-upgrade-job.yaml
```diff
@@ -41,5 +41,5 @@ spec:
         {{- if .Values.terminationMessagePolicy }}
         terminationMessagePolicy: {{ .Values.terminationMessagePolicy }}
         {{- end }}
-      serviceAccountName: fission-svc
```
sanketsudake, line 44: We should set hook-weight here so that service account is create first then deployment.
```
  helm.sh/hook-weight: "-5"
```

https://helm.sh/docs/topics/charts_hooks/#writing-a-hook

sanketsudake (changes requested):

sanketsudake (approved):
These changes look good for now. There would be further set of PRs which would strip down these permissions further.

PR #2597, opened 2022-11-03
Validation/Mutating webhook for Fission resources

neha-Gupta1:
## Description
Added- Admission webhooks for Fission Resources
Call Fission and K8s APIs from CLI and bypassing controller.

Fixes #

## Testing

CRUD for all resources should work as earlier.
If user tries to create resource in NS which is not present we throw an error instead of creating the NS. This is to replicate k8s default functionality.

## Checklist:


sanketsudake:
Closing PR as merging changes via https://github.com/fission/fission/pull/2608

PR #2612, opened 2022-11-10
Add Fission version API to router for CLI consumption

neha-Gupta1:
## Description
Exposed a version API on router to fetch server version info.
Used the API from CLI instead of using controller APIs

Fixes #

## Checklist:

### DIFF
--- a/pkg/fission-cli/util/util.go
+++ b/pkg/fission-cli/util/util.go
@@ -18,8 +18,10 @@ package util
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/fission-cli/util/util.go
```diff
@@ -149,20 +149,48 @@ func GetVersion(ctx context.Context, input cli.Input, cmdClient cmd.Client) info
...
 
-	serverInfo, err := client.V1().Misc().ServerInfo()
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		msg := fmt.Sprintf("HTTP error %v", resp.StatusCode)
+		console.Warn(msg)
+		return nil
+	}
+
+	var serverInfo info.ServerInfo
```
sanketsudake, line 172: Since we returned empty valid server info earlier, should we replace it?
This may break functions calling this function as they try to dereference nil pointer.

Should we move this definition to the top and return if any error ?

```
	var serverInfo info.ServerInfo
```

sanketsudake (changes requested):

sanketsudake (approved):

PR #2623, opened 2022-11-16
Get logs from Pods using Kubernetes API for function log command

neha-Gupta1:
## Description
Added functionality to fetch logs from pods using k8s APIs.

Fixes #

## Testing
Run command 
`fission fn log --name <fn_name>`

sample output:
log_output

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/function/log.go
+++ b/pkg/fission-cli/cmd/function/log.go
@@ -19,6 +19,7 @@ package function
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -61,7 +62,7 @@ func (opts *LogSubCommand) do(input cli.Input) error {
 	}
 
 	server, err := util.GetApplicationUrl(input.Context(), opts.Client(), "application=fission-api")
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "no available pod for port-forwarding with label selector") {
 		return err
 	}
 
--- a/pkg/fission-cli/cmd/function/log.go
+++ b/pkg/fission-cli/cmd/function/log.go
@@ -83,6 +83,7 @@ func (opts *LogSubCommand) do(input cli.Input) error {
 			case <-requestChan:
 				logFilter := logdb.LogFilter{
 					Pod:            fnPod,
+					PodNamespace:   input.String(flagkey.NamespacePod),
 					Function:       f.ObjectMeta.Name,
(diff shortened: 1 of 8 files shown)


pkg/fission-cli/cmd/function/log.go
```diff
@@ -60,13 +61,13 @@ func (opts *LogSubCommand) do(input cli.Input) error {
 		return errors.Wrap(err, "error getting function")
 	}
 
-	server, err := util.GetApplicationUrl(input.Context(), opts.Client(), "application=fission-api")
-	if err != nil {
-		return err
+	logDBOptions := logdb.LogDBOptions{
+		CTX:    input.Context(),
```
sanketsudake, line 65: Context we should not pass in struct I think. We can modify GetLogsDB so that it accepts context.
Function can use or ignore context.

sanketsudake (changes requested):

pkg/fission-cli/logdb/kubernetes_log.go
```diff
@@ -0,0 +1,142 @@
...
+	fv1 "github.com/fission/fission/pkg/apis/core/v1"
+	"github.com/fission/fission/pkg/fission-cli/cmd"
+)
+
+type LogDBOptions struct {
+	Client cmd.Client
+}
+
+type kubernetesLogs struct {
+	ctx    context.Context
```
sanketsudake, line 43: We shouldn't hold context in the structure.

sanketsudake (changes requested):

pkg/fission-cli/logdb/kubernetes_log.go
```diff
@@ -0,0 +1,142 @@
...
+	client cmd.Client
+}
+
+func (k kubernetesLogs) GetLogs(logFilter LogFilter) (podLogs *bytes.Buffer, err error) {
+	podLogs, err = GetFunctionPodLogs(k.ctx, k.client, logFilter)
+	return podLogs, err
+}
```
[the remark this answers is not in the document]
neha-Gupta1: 1. We get proper logs for pool manager based pods
2. `-f` flag help us to watch logs. And we fetch for newer logs every second. Which makes sure only new logs are printed
3. Continuously run requests against function- Works fine. New logs flow in with -f flag.
4. **Not sure about this.**  Need more testing here.
5. Since we fetch pod list before each log stream call, we stay updated about newer pods. And are able to fetch logs for the same.

PR #2631, opened 2022-11-21
Add sync triggers debounce in router

angelini:
## Description

We run a Fission deployment with about 12k functions and 12k triggers.

When the K8S Informers for HTTPTrigger and Function are instantiated their state is synchronized by calling the `AddFunc` event handler for each function and trigger. In our deployment, this means 24k calls to `syncTriggers`.

Every time `syncTriggers` is called the router's `Mux` is rebuilt from scratch as part of `getRouter`.

In our deployment, this causes our routers to max out their CPU for a couple hours after they are rebooted. The routers are still able to serve traffic while this is happening, but they time out many more requests than they should.

## Testing

We have this code running in production and the results were immediately visible. All of our router pods were rebooted without any kind of significant CPU usage.

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.19
 
 require (
 	github.com/Shopify/sarama v1.37.2
+	github.com/bep/debounce v1.2.1
 	github.com/dchest/uniuri v1.2.0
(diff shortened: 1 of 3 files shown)


sanketsudake (approved):
LGTM. Thanks for the PR 🎉

PR #2635, opened 2022-11-23
Deployments to work with specific namespaces instead of all namespaces

shubham-bansal96:
## Description
With the support of multi tenancy, fission will work with roles instead of cluster roles. And to achieve this fission needs to work with only specific namespaces instead of all namespaces.
With these changes fission will look for deployments only in specific namespaces as provided by user in helm chart.

Fixes #

## Checklist:

### DIFF
--- a/pkg/executor/reaper/reaper.go
+++ b/pkg/executor/reaper/reaper.go
@@ -363,7 +363,8 @@ func CleanupRoleBindings(ctx context.Context, logger *zap.Logger, client kuberne
 			return nil
 		}
 		for _, namespace := range GetReaperNamespace() {
-			cleanupRoleBindings(namespace) //ignore err
+			//ignore error
+			cleanupRoleBindings(namespace) //nolint errcheck
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


pkg/executor/reaper/reaper.go
```diff
@@ -192,137 +227,159 @@ func CleanupRoleBindings(ctx context.Context, logger *zap.Logger, client kuberne
...
+			return nil
 		}
+		for _, namespace := range GetReaperNamespace() {
+			//ignore error
+			cleanupRoleBindings(namespace) //nolint errcheck
+		}
+	}
+}
+
+func GetReaperNamespace() map[string]string {
```
sanketsudake, line 372: Can we avoid mutations to `DefaultNSResolver()` outside of the module?
Other go modules should utilize values in read-only mode.

Only the utils package should be able to modify it.

Please consider moving this logic under ns resolver only.

sanketsudake (changes requested):

pkg/executor/reaper/reaper.go
shubham-bansal96: done

PR #2649, opened 2022-12-01
K8s informer to work with specific namespaces for builder manager

shubham-bansal96:
## Description
With the support of multi tenancy, fission will work with roles instead of cluster roles. And to achieve this fission needs to work with only specific namespaces instead of all namespaces.
With these changes informers for builder manager will work only in specific namespaces as provided by user in helm chart.

Fixes #

## Checklist:


sanketsudake (approved):
LGTM

PR #2669, opened 2022-12-13
Create role and rolebinding for event-fetcher in multiple namespaces

shubham-bansal96:
## Description
As we have removed creating role and role bindings through backend code, So we need to make sure all role and role binding should get created through helm chart.

Currently fission is not able to create package for functions other than default namespace because of missing role and role binding of event-fetcher.
Now we have added roles and role binding for event-fetcher and with these changes fission will now able to create package for functions in multiple namespaces.

Fixes #

## Checklist:

### DIFF
--- a/charts/fission-all/templates/_function-access-role.tpl
+++ b/charts/fission-all/templates/_function-access-role.tpl
@@ -106,7 +106,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: {{ .Release.Name }}-fission-fetcher-pod-reader
+  name: {{ .Release.Name }}-fission-fetcher-websocket
   namespace: {{ .namespace }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
--- a/charts/fission-all/templates/_function-access-role.tpl
+++ b/charts/fission-all/templates/_function-access-role.tpl
@@ -45,6 +45,24 @@ rules:
   - secrets
   verbs:
   - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .namespace }}
+  name: {{ .Release.Name }}-event-fetcher
+rules:
+- apiGroups:
+  - ""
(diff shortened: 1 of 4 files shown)


charts/fission-all/templates/_function-access-role.tpl
```diff
@@ -45,6 +45,24 @@ rules:
...
+  - ""
+  resources:
+  - "events"
+  verbs:
+  - "get"
+  - "list"
+  - "watch"
+  - "create"
+  - "update"
+  - "patch"
```
sanketsudake, line 65: Please remove the following permission from fission-fetcher
```
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - list
  ```
  
  And add pod get permission in the current event-fetcher role.
  
  Please rename event-fetcher role to fission-fetcher-websocket
  and same for rolebinding.

Comment (sanketsudake) [reply charts/fission-all/templates/_function-access-role.tpl:65]: With this, we will also change fetcher code WsStartHandler and WsEndHandler

**Review by sanketsudake — CHANGES_REQUESTED** [review]

File: charts/fission-all/templates/_function-access-role.tpl (hunk 1 above)
Comment (shubham-bansal96) [reply charts/fission-all/templates/_function-access-role.tpl]: done

File: charts/fission-all/templates/_function-access-role.tpl [hunk 2]
```diff
@@ -75,5 +97,27 @@ roleRef:
...
+    {{- if and (.Values.builderNamespace) (eq .namespace "default") }}
+    namespace: {{ .Values.builderNamespace }}
+    {{- else }}
+    namespace: {{ .namespace }}
+    {{- end }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Name }}-fission-fetcher-pod-reader
```
sanketsudake, line 109: Please change to `{{ .Release.Name }}-fission-fetcher-websocket`

shubham-bansal96: done

PR #2677, opened 2022-12-24
Fix issue with updating timetrigger without --cron flag

### LINKED ISSUE #2596: Updating timetrigger without --cron flag is not updating the timetrigger
Kind

**Describe the bug**
Updating time trigger without --cron flag is actually not updating time trigger. CLI says time trigger updated but it's not actually updated.

**To Reproduce**
1. Create two fission function.
```
fission function create --name hellogo --env go --src hello.go --entrypoint Handler
fission function create --name hellogo2 --env go --src hello.go --entrypoint Handler
```
2. Create two time trigger and associate/add hellogo function with first trigger and hellogo2 function with second trigger.
```
fission timetrigger create --name trigger1 --function hellogo --cron '5 * * * * *'
fission timetrigger update --name trigger2 --function hellogo2 --cron '6 * * * * *'
```
Now check for timer pod logs, Everything looks good till now. Trigger1 is running hellogo function and trigger2 is running hellogo2 function.

Try updating trigger trigger1 and this time while updating, associate/add hellogo2 function to trigger1 without --cron flag.
```
fission timetrigger update --name trigger1 --function hellogo2
```
CLI says, time trigger updated successfully, check for timer pod logs again. After updating trigger, now trigger1 should run hellogo2 function instead of hellogo function. But it's still running hellogo function.

**Expected result**
After updating trigger, trigger1 should run hellogo2 function instead of hellogo function

**Actual result**
Trigger1 still running hellogo function.

**Additional context**
NOTE - This works fine if we update time trigger by providing --cron flag. Once we run this command then trigger1 will start running hellogo2 function.
```
fission timetrigger update --name trigger1 --function hellogo2 --cron '6 * * * * *'
```
Inside AddUpdateTimeTrigger function, fission have a check as mentioned below, this means time trigger will only get updated if there is a changes in cron. Need to change this condition.
```
if item.trigger.Spec.Cron != timeTrigger.Spec.Cron
```

shubham4443:
Signed-off-by: Shubham Nazare <shubham4443@gmail.com>

## Description

Simply removed the condition.

Fixes #2596 

## Checklist:

### DIFF
--- a/pkg/timer/timerSync.go
+++ b/pkg/timer/timerSync.go
@@ -98,7 +98,11 @@ func (ws *TimerSync) TimeTriggerEventHandlers(ctx context.Context) {
 			},
 			UpdateFunc: func(_ interface{}, obj interface{}) {
 				timeTrigger := obj.(*fv1.TimeTrigger)
-				ws.AddUpdateTimeTrigger(timeTrigger)
+				if item, ok := ws.timer.triggers[crd.CacheKeyUID(&timeTrigger.ObjectMeta)]; ok {
+					if item.trigger.ObjectMeta.ResourceVersion != timeTrigger.ObjectMeta.ResourceVersion {
+						ws.AddUpdateTimeTrigger(timeTrigger)
+					}
+				}
 			},
 			DeleteFunc: func(obj interface{}) {
 				timeTrigger := obj.(*fv1.TimeTrigger)
--- a/pkg/timer/timerSync.go
+++ b/pkg/timer/timerSync.go
@@ -61,14 +61,12 @@ func (ws *TimerSync) AddUpdateTimeTrigger(timeTrigger *fv1.TimeTrigger) {
 	ws.logger.Debug("cron event")
 
 	if item, ok := ws.timer.triggers[crd.CacheKeyUID(&timeTrigger.ObjectMeta)]; ok {
-		if item.trigger.Spec.Cron != timeTrigger.Spec.Cron {
-			if item.cron != nil {
-				item.cron.Stop()
-			}
-			item.trigger = *timeTrigger
-			item.cron = ws.timer.newCron(*timeTrigger)
-			logger.Debug("cron updated")
+		if item.cron != nil {
+			item.cron.Stop()
 		}
+		item.trigger = *timeTrigger
+		item.cron = ws.timer.newCron(*timeTrigger)
+		logger.Debug("cron updated")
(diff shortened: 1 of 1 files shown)


pkg/timer/timerSync.go
```diff
@@ -61,14 +61,12 @@ func (ws *TimerSync) AddUpdateTimeTrigger(timeTrigger *fv1.TimeTrigger) {
...
 
 	if item, ok := ws.timer.triggers[crd.CacheKeyUID(&timeTrigger.ObjectMeta)]; ok {
-		if item.trigger.Spec.Cron != timeTrigger.Spec.Cron {
-			if item.cron != nil {
-				item.cron.Stop()
-			}
-			item.trigger = *timeTrigger
-			item.cron = ws.timer.newCron(*timeTrigger)
-			logger.Debug("cron updated")
+		if item.cron != nil {
```
shubham-bansal96, line 64: We want to update cron only if there is any change in spec of time trigger. With these changes cron will get updated everytime even if there is no change in trigger spec.

You can add one more check in TimeTriggerEventHandlers function under updateFunc, something like
if oldtrigger.ObjectMeta.ResourceVersion != newtrigger.ObjectMeta.ResourceVersion {
					ws.AddUpdateTimeTrigger(timeTrigger)
				}

shubham-bansal96 (changes requested):

pkg/timer/timerSync.go
```diff
@@ -98,7 +98,11 @@ func (ws *TimerSync) TimeTriggerEventHandlers(ctx context.Context) {
 			},
 			UpdateFunc: func(_ interface{}, obj interface{}) {
```
shubham-bansal96, line 99: Instead of checking from map, You can do something like this.

UpdateFunc: func(oldObj interface{}, newObj interface{}) {
					oldTimeTrigger := oldObj.(*fv1.TimeTrigger)
					newTimeTrigger := newObj.(*fv1.TimeTrigger)
					if oldTimeTrigger.ObjectMeta.ResourceVersion != newTimeTrigger.ObjectMeta.ResourceVersion  {
						ws.AddUpdateTimeTrigger(newTimeTrigger)
					}
			}

shubham-bansal96 (changes requested):

shubham-bansal96 (approved):

sanketsudake (approved):

PR #2680, opened 2023-01-07
feat(cli): provide imagepullsecret option for container as function

### LINKED ISSUE #2569: `fission function run-container` missing --imagepullsecret option
I tried to run a `fission function run-container` to an image at a registry with authentication, and I got a `ErrImagePull` status because it was unable to authenticate.

A workaround for this problem is to [create a secret](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#create-an-imagepullsecret) and use it in a [spec file](https://doc.crds.dev/github.com/fission/fission/fission.io/Function/v1@v1.16.2) for the function at spec.podspec.imagePullSecrets[].name.

However, it would be better if we could supply a --imagepullsecret to the fission cli, so we could run it both without a spec file, or without manually editting it.

shubham4443:
Signed-off-by: Shubham Nazare <shubham4443@gmail.com>

## Description

While creating container as function, if the image is from a private registry then the user can specify the `imagepullsecret` through the cli.

Fixes #2569 

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/function/command.go
+++ b/pkg/fission-cli/cmd/function/command.go
@@ -173,6 +173,7 @@ func Commands() *cobra.Command {
 			flag.RunTimeMinCPU, flag.RunTimeMaxCPU, flag.RunTimeMinMemory,
 			flag.RunTimeMaxMemory, flag.ReplicasMin,
 			flag.ReplicasMax, flag.RunTimeTargetCPU,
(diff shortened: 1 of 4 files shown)


sanketsudake (approved):
LGTM

shubham-bansal96 (approved):

PR #2706, opened 2023-01-24
Replace existing REST APIs with gRPC

shubham4443:
## Description

This PR is the first step of replacing all the existing APIs which are written in REST architectural style with gRPC. gRPC has benefits over REST such as speed, communication protocol (gRPC uses HTTP/2).

Please note that -
1. This is still at a proof-of-concept stage.
2. We are beginning with `executor` first.
3. `generate-proto.sh` has protobuf related commands for reference.
4. `executor.proto` has a sample `UnaryEcho` service for testing purpose.
5. the executor gRPC server is running at port 50051 along with the HTTP server. This port is exposed on the pod as well on the service.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/2706)


sanketsudake:
Thanks for this exploration — it was a useful proof of concept. We recently revisited internal-transport optimization and decided not to pursue a gRPC migration:

- Since the EndpointSlice data plane (RFC-0002), the warm request path has no internal RPC left to accelerate. The router admits warm traffic from an in-memory informer-fed index and proxies straight to the pod.
- The executor RPC now fires only on cold start / saturation paths, where latency is dominated by pod specialization, not protocol overhead.
- The router→function hop must stay plain HTTP (user functions speak HTTP), so a gRPC migration could never touch the per-request path.
- gRPC over ClusterIP Services load-balances poorly (long-lived HTTP/2 pins to one backend), so doing it right needs headless Services + client-side LB or a mesh — more machinery for near-zero hot-path gain.

Instead we are landing targeted transport fixes (connection pooling on the trigger publishers and fetcher client, HMAC body-handling improvements). This PR also predates the router listener split and internal HMAC auth, so a revival would need a from-scratch design. Closing — happy to revisit if profiling ever shows protocol overhead on a specific path.

PR #2725, opened 2023-02-20
Added a common label into the fission core components that will help us checking the logs

### LINKED ISSUE #1479: [Feature] Set a common label into the fission core components that will help us checking the logs
The requirement is to be able to check the logs of all the components at once, we can achieve that using [stern](https://github.com/wercker/stern) as of now, but if we set a common label (let's say `application=fission`) to all the fission core components we can easily check the log using 
```
kubectl logs -n fission -l application=fission 
```

hiteshwani29:
## Description

This PR will add the feature to see logs of all fission components with a single label
```
kubectl logs -n fission -l part-of='fission'
```

Fixes #1479

## Checklist:

### DIFF
--- a/charts/fission-all/templates/buildermgr/deployment.yaml
+++ b/charts/fission-all/templates/buildermgr/deployment.yaml
@@ -5,15 +5,18 @@ metadata:
   labels:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     svc: buildermgr
+    component: fission-core
 spec:
   replicas: 1
   selector:
     matchLabels:
       svc: buildermgr
+      component: fission-core
   template:
     metadata:
       labels:
         svc: buildermgr
+        component: fission-core
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/path: "/metrics"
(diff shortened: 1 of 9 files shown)


charts/fission-all/templates/buildermgr/deployment.yaml
```diff
@@ -5,15 +5,18 @@ metadata:
   labels:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     svc: buildermgr
+    part-of: fission
```
viveksinghggits, line 8: I would rather prefer `application` instead of `part-of`. I will @sanketsudake review the PR. We should also see if value should be figured out from chart or release.

hiteshwani29: `application`  label is already used in the some of the fission component deployments.
We need to use some different label to achieve this

viveksinghggits: if fission is already setting the labels in the core components. Do we really need this PR?

sanketsudake: IMO if you have installed Fission with helm chart, typically there would be release name labels etc. 
OR chart label.
```
chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```

I am not ok with PR overall since it solves a specific case and doesn't help in broad perspective.

One way I can consider this is by allowing custom labels/annotations for Fission users via helm chart values, where users can use them for any use-case or purpose they are targeting.

sanketsudake:
Closing this PR as doesnt do what is intended. Overall my thoughts are fix should consider broader perspective as mentioned in above comment. 

Feel free to raise new PR if you another set of changes.

PR #2736, opened 2023-03-07
fix the invalid fn.spec.Concurrency

### LINKED ISSUE #1939: Relaxed concurrency in poolmanager strategy
Due to the current implementation of cache(pool cache) for the pool manager, sometimes number of specialized pods (function pods) are more than the `concurrency` parameter, this happens when there is burst of requests and number of incoming requests is greater than `concurrency` parameter.

Solution
This can be improved by applying locking mechanism but would degrade performance of the whole specialization.

saltbo:
## Description

use throttler to avoid simultaneous specialization

Fixes #1939

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/go-openapi/spec v0.20.7
 	github.com/golang-jwt/jwt/v4 v4.4.3
 	github.com/google/go-cmp v0.5.9
+	github.com/google/uuid v1.3.0
 	github.com/gorilla/mux v1.8.0
 	github.com/graymeta/stow v0.2.8
 	github.com/hashicorp/go-multierror v1.1.1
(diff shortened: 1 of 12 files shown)


sanketsudake:
Actually in context of poolmanager functions,

concurrency = no of specialized pods for a function
rpp = request per pod

total no of requests function can handle = concurrency * rpp

@pranoyk raised PR https://github.com/fission/fission/pull/2737
which make these constraints more strict. It also considers pods which are in specialisation, ie, virtual capacity. Capacity which would be available in future.

In future via https://github.com/fission/fission/pull/2737 we can different strategies to in allocation of requests to pods,
- Round Robin
- Aggressive pod specialisation. (serve requests in best way by utilising all available capacity)
- Compressed pod specialisation etc ( server more requests in less no of pods)

saltbo:
Thanks for your reply. That sounds great to me.  When is it possible to be merged?

I will reopen a PR for the error warp.

PR #2747, opened 2023-03-22
handle addr when port is passed in host:port format in the StartServer func

### LINKED ISSUE #2744: the PPROF doesn't work because of too many colons in address
v1.18.0

<pre>
$ fission version
$ kubectl version
</pre>

**Describe the bug**

**To Reproduce**

set PPROF_ENABLED=true
**Expected result**

pprof works well
**Actual result**
2023-03-20T07:04:50.454Z	INFO	httpserver/server.go:17	starting server	{"service": "pprof", "addr": ":localhost:6060"}
2023-03-20T07:04:50.454Z	ERROR	httpserver/server.go:21	server error	{"service": "pprof", "addr": ":localhost:6060", "error": "listen tcp: address :localhost:6060: too many colons in address"}
**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

NikhilSharmaWe:
## Description

This PR ensures to pass the correct value of addr in case when port is passed in `host:port` format in the [StartServer](https://github.com/fission/fission/blob/main/pkg/utils/httpserver/server.go#L11) func.

Fixes #2744

## Checklist:

### DIFF
--- a/pkg/utils/httpserver/server.go
+++ b/pkg/utils/httpserver/server.go
@@ -4,13 +4,17 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"go.uber.org/zap"
 )
 
 func StartServer(ctx context.Context, log *zap.Logger, svc string, port string, handler http.Handler) {
+	if !strings.Contains(port, ":") {
+		port = fmt.Sprintf(":%s", port)
+	}
 	server := http.Server{
-		Addr:    fmt.Sprintf(":%s", port),
+		Addr:    port,
 		Handler: handler,
 	}
 	l := log.With(zap.String("service", svc), zap.String("addr", server.Addr))


pkg/utils/httpserver/server.go
```diff
@@ -4,13 +4,20 @@ import (
...
 	"go.uber.org/zap"
 )
 
 func StartServer(ctx context.Context, log *zap.Logger, svc string, port string, handler http.Handler) {
+	var addr string
+	if strings.Contains(port, ":") {
+		addr = port
+	} else {
+		addr = fmt.Sprintf(":%s", port)
+	}
```
NikhilSharmaWe, line 18: This ensures that if the port is passed as `host:port` like in the case when pprof is **enabled** (like `localhost:6060`) format then do not add extra `:` in the addr. The addr value in other cases is passes like `:port` which means that the server will listen and serve for all hosts at port equal to `port`.

shubham-bansal96 (approved):

pkg/utils/httpserver/server.go
```diff
@@ -4,13 +4,20 @@ import (
...
 	"fmt"
 	"net/http"
+	"strings"
 
 	"go.uber.org/zap"
 )
 
 func StartServer(ctx context.Context, log *zap.Logger, svc string, port string, handler http.Handler) {
+	var addr string
+	if strings.Contains(port, ":") {
```
sanketsudake, line 14: if !strings.Contains(port, ":") {
      port = fmt.Sprintf(":%s", port)
)

can be done in single condition.

sanketsudake (approved):

PR #421, opened 2017-12-03
Istio integration

life1347:
This is the very first step for fission to integrate with Istio and there are still couple problems need to be solved. For those  interested in trying integrate fission with istio, following is the set up tutorial (Tutorial also post on [here](https://tachingchen.com/blog/fission-and-istio-integration/)). (https://github.com/fission/fission/issues/278, https://github.com/fission/fission/issues/344) 
## Test Environment
* Google Kubernetes Engine: 1.8.4-gke.0

### Create Kubernetes v1.8+ alpha cluster

Enable both RBAC & initializer features on kubernetes cluster.

** NOTICE ** `kubectl logs -f` is [not working correctly](https://github.com/kubernetes/kubernetes/issues/54205#issuecomment-340111607) on GKE alpha cluster right now.

```
$ gcloud container clusters create istio-demo \
	--enable-kubernetes-alpha \
    --machine-type=n1-standard-2 \
    --num-nodes=1 \
    --no-enable-legacy-authorization \
    --zone=<ZONE> \
    --cluster-version=1.8.4-gke.0
```

### Grant cluster admin permissions

Grant admin permission for `system:serviceaccount:kube-system:default` and current user.

```
# for system:serviceaccount:kube-system:default
$ kubectl create clusterrolebinding --user system:serviceaccount:kube-system:default kube-system-cluster-admin --clusterrole cluster-admin

# for current user
$ kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)
```

### Set up Istio environment

Download Istio 0.2.12

```
curl -L https://git.io/getLatestIstio | sh -
```

Apply istio related YAML files

```
$ kubectl apply -f istio-0.2.12/install/kubernetes/istio.yaml
$ kubectl apply -f istio-0.2.12/install/kubernetes/istio-initializer.yaml
```

Since helm uses jobs to install charts, we have to remove `jobs` from `initializers.rules.resources` due to a known [bug](https://github.com/istio/issues/issues/93) in istio pilot as a workaround.

```
$ kubectl edit initializerconfigurations
```

```
apiVersion: admissionregistration.k8s.io/v1alpha1
initializers:
- name: sidecar.initializer.istio.io
  rules:
  - apiGroups:
    - '*'
    apiVersions:
    - '*'
    resources:
    - deployments
    - statefulsets
    # As a workaround right now, remove jobs here 
    # due to a known bug (https://github.com/istio/issues/issues/93) 
    - jobs 
[...]

PR #2752, opened 2023-03-29
refactor: update the client retry by the pkg retryablehttp

### LINKED ISSUE #2749: Why retry so many times for the builder?

<pre>
$ fission version
$ kubectl version
</pre>

**Describe the bug**

https://github.com/fission/fission/commit/23942fdf7db917eae3052e8c4b7e356c0c189b63#diff-16303cbae6f1706622e3c61a3329d7989c89bc30dab964865c8d11852c1807ed

There is a logic of 20 retries, and there is no distinction between network errors and build errors. that's not useful for the build error, And A large number of useless retries take up a lot of memory, they crowd out the resources of normal build tasks.

**To Reproduce**

A limit memory builder with a normal function and a build error function

**Expected result**

Only retry for the network error. Or the HTTP status code > 500

**Actual result**
Any not 200 status code will be retried

**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

saltbo:
## Description

The old retry policy is too simple, we can use the pkg retryablehttp to replace it

Fixes #2749

## Checklist:

### DIFF
--- a/pkg/builder/client/client.go
+++ b/pkg/builder/client/client.go
@@ -21,10 +21,9 @@ import (
 	"context"
 	"encoding/json"
 	"io"
-	"net/http"
 	"strings"
-	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.uber.org/zap"
@@ -39,12 +38,13 @@ type (
 	Client struct {
 		logger     *zap.Logger
 		url        string
-		httpClient *http.Client
+		httpClient *retryablehttp.Client
 	}
 )
 
 func MakeClient(logger *zap.Logger, builderUrl string) *Client {
-	hc := &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+	hc := retryablehttp.NewClient()
+	hc.HTTPClient.Transport = otelhttp.NewTransport(hc.HTTPClient.Transport)
 	return &Client{
 		logger:     logger.Named("builder_client"),
 		url:        strings.TrimSuffix(builderUrl, "/"),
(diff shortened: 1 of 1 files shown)


sanketsudake (approved):

pkg/builder/client/client.go
```diff
@@ -39,12 +38,13 @@ type (
...
 		logger     *zap.Logger
 		url        string
-		httpClient *http.Client
+		httpClient *retryablehttp.Client
 	}
 )
 
 func MakeClient(logger *zap.Logger, builderUrl string) *Client {
-	hc := &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+	hc := retryablehttp.NewClient()
```
shubham-bansal96, line 46: You should define maxRetries here as well.
hc.RetryMax = 20

shubham-bansal96 (changes requested):

pkg/builder/client/client.go
sanketsudake, line 46: Default max retry time is 30 seconds where as min is 1 in library.
Max tries default is 4.
https://github.com/hashicorp/go-retryablehttp/blob/7f638666931a119b8d7a1239e44da11db0d446a4/client.go#L47

We need to see if we are honoring current max retry time.

saltbo: > However 20 retries is too much we and reduce retries and increase wait duration.

Yes, I agree. The original intention of this PR is to change the number of retries and retry strategy.

The current retry policy is from the PR Istio integration (https://github.com/fission/fission/pull/421), I think it wants to retry only for the network error, not the build error.  short time with 20 retries for the network error may be suitable, But not suitable for the build error.

shubham-bansal96: Make sense. As per current logic written in code, max retry time won't not more than 2 second.
So i think we can go ahead with these changes without specifying `hc.RetryMax = 20`

shubham-bansal96 (approved):

PR #2843, opened 2023-10-06
helm: Add support for HorizontalPodAutoscaller on router

Lunik:
## Description

This PR add new options in the Helm chart values allowing to deploy and HorizontalPodAutoscaler for the `router` component.

This is basically the implementation of this part of the doc : https://fission.io/docs/installation/advanced-setup/#create-hpa-for-router-service

By default, nothing change for the user.

User can select the Kubernetes autoscaling `apiVersion` allowing support for user with cluster version under v1.23. By default the `autoscalingApiVersion` used is `autoscaling/v2` (support v1.23+ and required in v1.26+).
The `router.autoscaling.metrics` handle the full metrics structure regardless of the `apiVersion`.

I have placed the value `autoscalingApiVersion` at the root allowing a reuse if we want to add autoscaling to other components of Fission in the futur.

Didn't find any

## Testing

- Deploying the plateform without setting any values => same as before
- Deploying the plateform with `autoscaling.enabled=true`
```
kubectl get hpa

NAME     REFERENCE           TARGETS         MINPODS   MAXPODS   REPLICAS   AGE
router   Deployment/router   <unknown>/80%   1         10        2          9m14s
```

## Checklist:

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/2843)

### DIFF
--- a/charts/fission-all/templates/router/deployment.yaml
+++ b/charts/fission-all/templates/router/deployment.yaml
@@ -11,7 +11,7 @@ metadata:
     svc: router
     application: fission-router
 spec:
-{{- if not .Values.router.deployAsDaemonSet }}
+{{- if and (not .Values.router.deployAsDaemonSet) (not .Values.router.autoscaling.enabled) }}
   replicas: {{ .Values.router.replicas | default 1 }}
 {{- end }}
   selector:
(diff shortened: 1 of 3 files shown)


sanketsudake:
Thanks for this contribution — it was the right idea. Router autoscaling has since landed in main via #3416: the chart now has an opt-in `router.autoscaling` block that renders `templates/router/hpa.yaml` (autoscaling/v2), plus a mutually-exclusive KEDA ScaledObject option. Since main covers this use case, I'm closing the PR. Sorry it sat so long, and thanks for pushing the feature forward.

PR #2867, opened 2023-11-05
Add interfaces for executor/fetcher/storagesvc clients

sanketsudake:
## Description

- Making all fission contained http clients internal and adding interfaces for them
- This would help in testing services individually and test them with mocks

Fixes #

## Checklist:

### DIFF
--- a/pkg/router/router.go
+++ b/pkg/router/router.go
@@ -52,7 +52,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/fission/fission/pkg/crd"
-	executorClient "github.com/fission/fission/pkg/executor/client"
+	eclient "github.com/fission/fission/pkg/executor/client"
 	"github.com/fission/fission/pkg/throttler"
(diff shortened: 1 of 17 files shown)


pkg/router/router.go
```diff
@@ -98,7 +98,7 @@ func serve(ctx context.Context, logger *zap.Logger, port int,
 }
 
 // Start starts a router
-func Start(ctx context.Context, clientGen crd.ClientGeneratorInterface, logger *zap.Logger, port int, executorURL string) error {
+func Start(ctx context.Context, clientGen crd.ClientGeneratorInterface, logger *zap.Logger, port int, executor eclient.ClientInterface) error {
```
pranoyk, line 101: Why did we move `executor eclient.ClientInterface` to the parameter?

sanketsudake: So that we can unit test router by mocking the executor calls.
This allows us to write tests with router and envtests and skipping rest of fission services.

PR #2899, opened 2024-01-08
Add kubernetes QPS and Burst limit configuration

sanketsudake:
## Description

Added configurable kubernetes client limits burst and QPS
Default QPS 200 and burst 500.

Fixes #

## Checklist:

### DIFF
--- a/pkg/crd/client.go
+++ b/pkg/crd/client.go
@@ -35,6 +35,11 @@ import (
 	"github.com/fission/fission/pkg/utils"
 )
 
+const (
+	EnvKubeClientQps   = "KUBE_CLIENT_QPS"
+	EnvKubeClientBurst = "KUBE_CLIENT_BURST"
+)
+
 type (
 	ClientGeneratorInterface interface {
 		GetRestConfig() (*rest.Config, error)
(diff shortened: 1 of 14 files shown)


pkg/crd/client.go
```diff
@@ -60,6 +65,20 @@ func (cg *ClientGenerator) getRestConfig() (*rest.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	qps, _ := utils.GetUIntValueFromEnv(EnvKubeClientQps)
```
pranoyk, line 69: Should we use `utils.GetIntValueFromEnv` instead of `utils.GetUIntValueFromEnv`?

pranoyk: Or rather use `utils.GetUIntValueFromEnv` only to get max QPS and max ClientBurst since these are always positive integers.

sanketsudake: actual usage is in float and int so sticked to types as suggested by linter and code scanner

```
cg.restConfig.QPS = float32(qps)
cg.restConfig.Burst = burst
```

pranoyk (approved):

PR #2979, opened 2024-07-10
Fix: Storage leak in Builder and Fetcher

### LINKED ISSUE #2821: Storage Leak in Fetcher

<pre>
$ fission version
client:
  fission/core:
    BuildDate: "2023-05-18T12:33:09Z"
    GitCommit: f955d118
    Version: v1.19.0
server:
  fission/core:
    BuildDate: "2023-05-18T12:33:09Z"
    GitCommit: f955d118
    Version: v1.19.0

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.2", GitCommit:"7f6f68fdabc4df88cfea2dcf9a19b2b830f1e647", GitTreeState:"clean", BuildDate:"2023-05-17T14:13:27Z", GoVersion:"go1.20.4", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27+", GitVersion:"v1.27.3-eks-a5565ad", GitCommit:"78c8293d1c65e8a153bf3c03802ab9358c0e1a14", GitTreeState:"clean", BuildDate:"2023-06-16T17:32:40Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
</pre>

EKS (eks.3)
**Describe the bug**

I observed that env builder pod (that has builder and fetcher container) keeps getting evicted because of over utlization of ephemeral storage. On checking the container, I found that it still has files related to the packages it built previously which are piling up (since I was building more than 100 functions together simultaneously).

**To Reproduce**
Create an environment (version 3) and build a lot of packages.

**Expected result**

Ideally once the package is built, all its files should be removed from the builder pod

**Actual result**
Builder Pod is keeping all the files of previously generated packages

**Screenshots/Dump file**

<pre>
$ fission support dump
</pre>

**Additional context**

On digging the fetcher code, it was confirmed in the `UploadHandler` that even after a successful response from the `ssClient.Upload` func, the files are not deleted.

soharab-ic:
## Description

- Builder pod contains builder and fetcher containers and they share a volume.
- Builder build the packages and fetcher uploads the packages to storage service.
- In case of build/upload failure, packages remains in the shared volume. These packages are of no use and needs to be cleaned up. 
- Added a `Clean` API in builder package. This API will be called by `buildermgr` for deleting src pkg after pkg is build and deployment is uploaded. I wanted to delete the src package from builder pod directly after failed/successful build but `buildermgr` uses `retryablehttp` client so in case of build failure `buildermgr` will retry build by sending `Build` request.
- To delete deployment pkg, fetcher `UploadHandler` will delete the deployment pkg after successful/failed upload.

Fixes #2821

## Testing

- Added a negative test for builder's `Clean` API.
- Adding a positive test involves creating a `/packages` directory for which `root` permissions are required.Therefore, did not add any positive test.
```
{"level":"info","ts":"2024-07-10T10:45:48.563Z","logger":"builder","caller":"builder/builder.go:197","msg":"builder received clean request","request":{"srcPkgFilename":"hello-test-fdebb12e-60ff-4e95-b41b-e3531444a4a4-kfezxl"}}
{"level":"info","ts":"2024-07-10T10:45:48.564Z","logger":"builder","caller":"builder/builder.go:175","msg":"clean request complete","elapsed_time":0.001311227}
```

## Checklist:

### DIFF
--- a/cmd/builder/app/server.go
+++ b/cmd/builder/app/server.go
@@ -32,6 +32,7 @@ func Run(ctx context.Context, logger *zap.Logger, mgr manager.Interface, shareVo
 	builder := builder.MakeBuilder(logger, shareVolume)
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", builder.Handler)
+	mux.HandleFunc("/clean", builder.Clean)
 	mux.HandleFunc("/version", builder.VersionHandler)
 	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
--- a/pkg/builder/client/client.go
+++ b/pkg/builder/client/client.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"net/http"
 	"strings"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -37,6 +38,7 @@ import (
 type (
 	ClientInterface interface {
 		Build(context.Context, *builder.PackageBuildRequest) (*builder.PackageBuildResponse, error)
+		Clean(context.Context, *builder.PackageCleanRequest) error
 	}
 
 	client struct {
@@ -55,6 +57,10 @@ func MakeClient(logger *zap.Logger, builderUrl string) ClientInterface {
 		url:        strings.TrimSuffix(builderUrl, "/"),
 		httpClient: hc,
 	}
+}
+
+func (c *client) getCleanUrl() string {
+	return c.url + "/clean"
 }
 
 func (c *client) Build(ctx context.Context, req *builder.PackageBuildRequest) (*builder.PackageBuildResponse, error) {
@@ -85,4 +91,26 @@ func (c *client) Build(ctx context.Context, req *builder.PackageBuildRequest) (*
 	}
 
 	return &pkgBuildResp, ferror.MakeErrorFromHTTP(resp)
+}
+
+func (c *client) Clean(ctx context.Context, req *builder.PackageCleanRequest) error {
+	logger := otelUtils.LoggerWithTraceID(ctx, c.logger)
+
+	body, err := json.Marshal(req)
+	if err != nil {
+		return errors.Wrap(err, "error marshaling json")
+	}
+
+	resp, err := ctxhttp.Post(ctx, c.httpClient.StandardClient(), c.getCleanUrl(), "application/json", bytes.NewReader(body))
+	if err != nil {
+		logger.Error("error creating delete request", zap.Error(err))
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return ferror.MakeErrorFromHTTP(resp)
+	}
+
+	return nil
 }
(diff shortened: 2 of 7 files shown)


pkg/utils/utils.go
```diff
@@ -267,3 +267,28 @@ func FindFreePort() (int, error) {
 
 	return port, nil
 }
+
+func DeleteOldPackages(pkgPath, pkgType string) error {
```
sanketsudake, line 271: Please add detailed comments for this function.

cmd/builder/app/server.go
```diff
@@ -32,6 +32,7 @@ func Run(ctx context.Context, logger *zap.Logger, mgr manager.Interface, shareVo
 	builder := builder.MakeBuilder(logger, shareVolume)
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", builder.Handler)
+	mux.HandleFunc("/clean", builder.Clean)
```
sanketsudake, line 35: Please try once commenting on this line. 
Following scenario may happen,
It may happen user upgrades to newer version of fission but builder pod is old. 
In that case, clean API would not be present and client would get 404 response.
We should be able to handle this case, from client side.

pkg/builder/client/client.go
```diff
@@ -86,3 +92,25 @@ func (c *client) Build(ctx context.Context, req *builder.PackageBuildRequest) (*
...
+	}
+
+	resp, err := ctxhttp.Post(ctx, c.httpClient.StandardClient(), c.getCleanUrl(), "application/json", bytes.NewReader(body))
+	if err != nil {
+		logger.Error("error creating delete request", zap.Error(err))
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
sanketsudake, line 111: We should be able to ignore not found error etc if API is missing due to older version builder on function side.

cmd/builder/app/server.go
soharab-ic: Even if clean API is missing, the request will be forwarded to `builder.Handler` since it accepts all request because of `/` route and we get a 500 response instead of 404.

soharab-ic: Even if i change the clean API method from `POST` to `DELETE`, still the response would 405 instead of 404.
Which approach should i take @sanketsudake ?
1. Handle 500 response in client.
2. Change clean API method from POST to DELETE.

I think approach 2 is good because in first approach builder will try to build the package but in 2nd approach it will return 405 response and avoid building the package.

soharab-ic: Also i think it would be better if we replace `/` route with `/build` route.

sanketsudake: > Change clean API method from POST to DELETE.
This one sounds better. We should change existing API behaviour for no reason.

pkg/utils/utils.go
soharab-ic: Done

cmd/builder/app/server.go
soharab-ic: Done

pkg/builder/client/client.go
soharab-ic: Done

sanketsudake (approved):

PR #3000, opened 2024-08-23
Run logger as root to allow access to log files from host

markretallack:
## Description

This change sets the user for the logger pod to be root in the helm charts. 

This allows the logger app to access the log file exposed from the host where it would normally be run as a non-root user. 

This would normally result in the following error:

> "error":"symlink /var/log/containers/poolmgr-opensearch-stats-default-46119364-685c75bc5b-fz87g_default_opensearch-stats-ffc3bf78135a5d7e5b8cacc25655bf2a81a02e130a499414fae42f0021fcebb4.log /var/log/fission/poolmgr-opensearch-stats-default-46119364-685c75bc5b-fz87g_default_opensearch-stats-ffc3bf78135a5d7e5b8cacc25655bf2a81a02e130a499414fae42f0021fcebb4.log: permission denied"

None

## Testing

Deployed to EKS and tested that the 

## Checklist:

### DIFF
--- a/charts/fission-all/templates/fluentbit/fluentbit.yaml
+++ b/charts/fission-all/templates/fluentbit/fluentbit.yaml
@@ -99,7 +99,11 @@ spec:
         - name: init
           image: {{ .Values.busyboxImage | quote }}
           imagePullPolicy: {{ .Values.pullPolicy }}
-          command: ['mkdir', '-p', '/var/log/fission']
+          command: 
+            - "/bin/sh"
+            - "-c"
+            # create the log directory and allow others read/write access
+            - "mkdir -p /var/log/fission && chmod o+wr /var/log/fission"
           volumeMounts:
             - name: container-log
               mountPath: /var/log/
@@ -130,7 +134,6 @@ spec:
 {{- if .Values.logger.enableSecurityContext }}
           securityContext:
             privileged: true
-            runAsUser: 0
 {{- end }}
         - name: fluentbit
 {{- if .Values.repository }}
--- a/charts/fission-all/templates/fluentbit/fluentbit.yaml
+++ b/charts/fission-all/templates/fluentbit/fluentbit.yaml
@@ -107,7 +107,6 @@ spec:
 {{- if .Values.logger.enableSecurityContext }}
           securityContext:
             privileged: true
-            runAsUser: 0
 {{- end }}
       containers:
         - name: logger
@@ -131,6 +130,7 @@ spec:
 {{- if .Values.logger.enableSecurityContext }}
           securityContext:
             privileged: true
+            runAsUser: 0
 {{- end }}
         - name: fluentbit
 {{- if .Values.repository }}
--- a/charts/fission-all/templates/fluentbit/fluentbit.yaml
+++ b/charts/fission-all/templates/fluentbit/fluentbit.yaml
@@ -107,6 +107,7 @@ spec:
 {{- if .Values.logger.enableSecurityContext }}
           securityContext:
             privileged: true
+            runAsUser: 0
 {{- end }}
       containers:
         - name: logger


charts/fission-all/templates/fluentbit/fluentbit.yaml
```diff
@@ -130,6 +130,7 @@ spec:
 {{- if .Values.logger.enableSecurityContext }}
           securityContext:
             privileged: true
+            runAsUser: 0
```
sanketsudake, line 133: We can you provider 

Probably we should modify main values so that user can their own options.
```
logger:
     enableSecurityContext: true/false
     securityContext: {}
     # Passing security context if enabled
     # securityContext:
     #      privilege: true
     #.     runAsUser: <privilege user>
````
 I would recommend not to use root user for security reasons.

Comment (markretallack) [reply charts/fission-all/templates/fluentbit/fluentbit.yaml]: ok, will look at an alternative method

Comment (markretallack) [reply charts/fission-all/templates/fluentbit/fluentbit.yaml]: @sanketsudake  I have changed the runAsUser to a chmod in the initContainer. This means that its not using a root user to run the container, but allows the non-root user to write into the /var/log/fission directory.  Hopefully this will be more acceptable. Ideally I would use a chown to the user who is running within the container but I am not sure how to do that.

Comment (soharab-ic) [reply charts/fission-all/templates/fluentbit/fluentbit.yaml]: @markretallack As @sanketsudake has mentioned that we should provide user with options to configure `runAsUser` value in securityContext. We have created a patch with proper explanation for logger's securityContext field. Please try the patch.

```
diff --git a/charts/fission-all/templates/fluentbit/fluentbit.yaml b/charts/fission-all/templates/fluentbit/fluentbit.yaml
index f54a400f..9718fe97 100644
--- a/charts/fission-all/templates/fluentbit/fluentbit.yaml
+++ b/charts/fission-all/templates/fluentbit/fluentbit.yaml
@@ -128,10 +128,8 @@ spec:
             - name: docker-log
               mountPath: /var/lib/docker/containers
               readOnly: true
-{{- if .Values.logger.enableSecurityContext }}
           securityContext:
-            privileged: true
-{{- end }}
+            {{- toYaml .Values.logger.securityContext | nindent 12 }}
         - name: fluentbit
 {{- if .Values.repository }}
           image: "{{ .Values.logger.fluentdImageRepository }}/{{ .Values.logger.fluentdImage }}:{{ .Values.logger.fluentdImageTag }}"
diff --git a/charts/fission-all/values.yaml b/charts/fission-all/values.yaml
index f0b668f8..8225725f 100644
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -620,11 +620,22 @@ logger:
 
   ## Fluent-bit writes/reads it’s own sqlite database to record a history of tracked
   ## files and a state of offsets, this is very useful to resume a state if the ser-
-  ## vice is restarted. For Kubernetes environment with constraints like OpenShift,
+  ## vice is restarted.
+  ##
+  ## For Kubernetes environment with constraints like OpenShift,
   ## the containers are limited to write hostPath volume. Hence, we have to enable
   ## security context and set privileged to true.
   ##
-  enableSecurityContext: false
+  ## The user ID in runAsUser should have access to the `/var/log` and
+  ## `/var/lib/docker/containers` directories on your host.
+  ## On many hosts, this user might be root ,i.e., `runAsUser: 0`.
+  ## Although it is recommended not to use root user for security reasons.
+  ##
+  ## The `/var/log` and `/var/lib/docker/containers` directories on host are mounted
+  ## to logger container with volumeType `HostPath`.
+  securityContext: {}
+    # privileged: true
+    # runAsUser: 0
 
   ## Enable PodSecurityPolicies to allow privileged container
   ## Only required in some clusters and when enableSecurityContext is true

```

soharab-ic:
@markretallack We are rushing towards a new release so we have introduced our changes on top of your changes in this PR https://github.com/fission/fission/pull/3005.

PR #3016, opened 2024-09-14
Print pod log error when response status is 404 returned by test function

### LINKED ISSUE #3014: Print pod log error when response status is 404 returned by test function

<pre>
$ fission version
client:
  fission/core:
    BuildDate: "2024-07-17T02:16:12Z"
    GitCommit: 8a351e4b
    Version: v1.20.2
server:
  fission/core:
    BuildDate: "2024-09-14T02:46:30Z"
    GitCommit: cc817d91
    Version: v1.20.2
</pre>

**Describe the bug**
When I tested the function under the default namespace, the response status returned was 404. At this point, the pod log corresponding to the function is queried, but it is always queried under the 'fission-function' namespace, which results in an error because the pod cannot be found.
image

waitstory:
## Description

When I tested the function under the default namespace, the response status returned was 404. At this point, the pod log corresponding to the function is queried, but it is always queried under the 'fission-function' namespace, which results in an error because the pod cannot be found.

Fixes #3014

## Checklist:

### DIFF
--- a/crds/v1/fission.io_timetriggers.yaml
+++ b/crds/v1/fission.io_timetriggers.yaml
@@ -73,6 +73,17 @@ spec:
                 - name
                 - type
                 type: object
+              method:
+                default: POST
+                description: 'HTTP Method for trigger, ex : GET, POST, PUT, DELETE,
+                  HEAD (default: "POST")'
+                type: string
+              subpath:
+                default: /
+                description: |-
+                  Subpath to trigger a specific route if function
+                  internally supports routing, (default: "/")
+                type: string
             required:
             - cron
             - functionref
--- a/pkg/apis/core/v1/types.go
+++ b/pkg/apis/core/v1/types.go
@@ -821,6 +821,17 @@ type (
 
 		// The reference to function
 		FunctionReference `json:"functionref"`
+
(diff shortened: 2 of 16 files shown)


pkg/fission-cli/util/util.go
```diff
@@ -535,7 +535,7 @@ func FunctionPodLogs(ctx context.Context, fnName, ns string, client cmd.Client)
 
 	if len(ns) == 0 {
 		ns = metav1.NamespaceDefault
-	} else if ns != metav1.NamespaceDefault {
+	} else {
 		podNs = ns
```
soharab-ic, line 539: @waitstory Please remove `podNs` variable from this function and use `ns` variable only. Environment and function should be in the same namespace.

soharab-ic (changes requested):

pkg/fission-cli/util/util.go
soharab-ic: If possible, please make these additional changes in this PR:
- The selector used in this function will not work for `container` functions. Update the `FunctionPodLogs` function and make it work for container functions.
- Similar changes. Make [GetFunctionPodLogs](https://github.com/fission/fission/blob/main/pkg/fission-cli/logdb/kubernetes_log.go#L57) works for container functions.

soharab-ic (approved):
Please rebase with main branch and push a single commit.

Issue #3011, opened 2024-09-10
TimeTriggers: Invoke any URL instead of root of a function


soharab-ic:
**Is your feature request related to a problem? Please describe.**
Time triggers can invoke a function based on time. Time triggers can't trigger any other URLs except a function's root URL. Suppose, a microservices app is deployed in fission as a function based on container. The container hosts multiple routes and with HTTPTrigger, we can hit those routes. Now, I want to create a time trigger for one of these routes and it is not supported by fission since time triggers will only trigger the function but not any other routes exposed by this container based function.

**Describe the solution you'd like**
- Make time triggers generic to invoke any URLs not just fission functions.

**Describe alternatives you've considered**
None

**Additional context**
https://github.com/fission/fission/discussions/3009

PR #3017, opened 2024-09-14
Add `method` and `subpath` flags to timetrigger object for triggering a function

soharab-ic:
## Description

- Timetrigger can only trigger a function by HTTP POST method and it can only trigger the root of the function. If function exposes some routes then timetrigger can not trigger those routes.
- This PR adds new flags `method` and `subpath` to timetrigger `create` and `update` commands. Now user can configure the HTTP `method` used to trigger the function and also configure a `subpath` to trigger an internal route exposed by the function.

Fixes https://github.com/fission/fission/issues/3011 https://github.com/fission/fission/discussions/3009

## Checklist:

### DIFF
--- a/pkg/apis/core/v1/types.go
+++ b/pkg/apis/core/v1/types.go
@@ -821,6 +821,13 @@ type (
 
 		// The reference to function
 		FunctionReference `json:"functionref"`
+
+		// HTTP Method for trigger
+		Method string `json:"method,omitempty"`
+
+		// Subpath to trigger a specific route if function
+		// internally supports routing.
+		Subpath string `json:"subpath,omitempty"`
 	}
 	// FailureType refers to the type of failure
 	FailureType string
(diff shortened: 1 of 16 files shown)


pkg/apis/core/v1/types.go
```diff
@@ -821,6 +821,13 @@ type (
 
 		// The reference to function
 		FunctionReference `json:"functionref"`
+
+		// HTTP Method for trigger
+		Method string `json:"method,omitempty"`
```
sanketsudake, line 826: HTTP Method for trigger function call

Please check if we can use kubebuilder default annotation to set it to POST.
Also add default being POST in the CRD doc above.

pkg/apis/core/v1/types.go
```diff
@@ -821,6 +821,13 @@ type (
...
+
+		// Subpath to trigger a specific route if function
+		// internally supports routing.
+		Subpath string `json:"subpath,omitempty"`
```
sanketsudake, line 830: Same as above. Please see if we can set default to "/".

sanketsudake (approved):

facorazza:
The subpath allows to forward parameters to the query but is there a way to send a predetermined payload when using the POST method?

PR #3027, opened 2024-09-30
Upgraded to Go latest version 1.23.1

soharab-ic:

Fixes #

## Checklist:

### DIFF
--- a/pkg/executor/executortype/poolmgr/gp.go
+++ b/pkg/executor/executortype/poolmgr/gp.go
@@ -426,7 +426,7 @@ func (gp *GenericPool) specializePod(ctx context.Context, pod *apiv1.Pod, fn *fv
 					zap.String("configmap_namespace", cm.Namespace),
 					zap.String("function_name", fn.ObjectMeta.Name),
 					zap.String("function_namespace", gp.fnNamespace))
-				return fmt.Errorf("%s", fmt.Sprintf("configmap %s must be in same namespace as function namespace", cm.Name))
+				return fmt.Errorf("configmap %s must be in same namespace as function namespace", cm.Name)
 			} else {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


pkg/executor/executortype/poolmgr/gp.go
```diff
@@ -426,7 +426,7 @@ func (gp *GenericPool) specializePod(ctx context.Context, pod *apiv1.Pod, fn *fv
 					zap.String("configmap_namespace", cm.Namespace),
 					zap.String("function_name", fn.ObjectMeta.Name),
 					zap.String("function_namespace", gp.fnNamespace))
-				return fmt.Errorf(fmt.Sprintf("configmap %s must be in same namespace as function namespace", cm.Name))
+				return fmt.Errorf("%s", fmt.Sprintf("configmap %s must be in same namespace as function namespace", cm.Name))
```
sanketsudake, line 429: I think we should just get rid of fmt.Sprintf and use fmt.Errorf. Same below
```
fmt.Errorf("configmap %s must be in same namespace as function namespace", cm.Name)
```

pkg/mqtrigger/messageQueue/kafka/consumer.go
```diff
@@ -200,7 +200,7 @@ func (ch *MqtConsumerGroupHandler) kafkaMsgHandler(msg *sarama.ConsumerMessage)
 		errorString := fmt.Sprintf("request exceed retries: %v", ch.trigger.Spec.MaxRetries)
 		errorHeaders := generateErrorHeaders(errorString)
 		errorHandler(ch.logger, ch.trigger, ch.producer, ch.fnUrl,
-			fmt.Errorf(errorString), errorHeaders)
+			fmt.Errorf("%s", errorString), errorHeaders)
```
sanketsudake, line 203: Should be errors.New if we are doing formatting. Same for rest. Lets avoid using formatting unnecessarily.

sanketsudake (approved):

PR #3041, opened 2024-10-15
Add leader election for executor

soharab-ic:
## Description

- Added leader election to executor for high availability.
- It is a Kubernetes resource that helps manage and ensure only one pod acts as the leader at any given time. If the leader pod fails, another pod can quickly take over, ensuring the application remains highly available.

Fixes #

## Testing

- Tested by increasing the number of replicas in executor deployment. Leader election is working as expected.
- The non-leader pods keep crashing and restarting because of `liveness` and `readiness` probes failed. To fix this issue, we need to restructure the starting steps of executor and start the server before the leader election loop. I am working on it.

## Checklist:

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3041)

### DIFF
--- a/cmd/fission-bundle/main.go
+++ b/cmd/fission-bundle/main.go
@@ -36,6 +36,7 @@ import (
 	eclient "github.com/fission/fission/pkg/executor/client"
 	"github.com/fission/fission/pkg/info"
 	"github.com/fission/fission/pkg/kubewatcher"
+	"github.com/fission/fission/pkg/leaderelection"
 	functionLogger "github.com/fission/fission/pkg/logger"
 	mqt "github.com/fission/fission/pkg/mqtrigger"
 	"github.com/fission/fission/pkg/router"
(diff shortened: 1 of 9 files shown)


cmd/fission-bundle/main.go
```diff
@@ -64,7 +66,16 @@ func runRouter(ctx context.Context, clientGen crd.ClientGeneratorInterface, logg
 }
 
 func runExecutor(ctx context.Context, clientGen crd.ClientGeneratorInterface, logger *zap.Logger, mgr manager.Interface, port int) error {
```
sanketsudake, line 68: Can we create a based test? 

If we have `etcd` in envtest, we can create two instances of executor and see if one of them only runs.

```
k8s/1.23.5-darwin-arm64
etcd		kube-apiserver	kubectl
```

sanketsudake:
Implemented via https://github.com/fission/fission/pull/3418

PR #3051, opened 2024-10-28
Fix download deploy package out of k8s cluster

lx1036:

when I run `fission pkg getdeploy --name hello-pkg-sp -o ./hello-pkg-sp.zip`, it raise errors. 

hello-pkg-sp package is a builded package, its yaml:

```
spec:
  buildcmd: ./build.sh
  deployment:
    checksum:
      sum: d32c0ab6e49f4043dc44f02fca6231be1e303bae779a10d0bc2682efcacd7278
      type: sha256
    type: url
    url: >-
      http://storagesvc.fission/v1/archive?id=%2Ffission%2Ffission-functions%2Fbb568aa7-f0c6-4933-97cf-72e180aad802
      
  environment:
    name: python-27
    namespace: default
  source:
    checksum: {}
    literal: >-
      UEsDBBQACAAAAK0+V1kAAAAAAAAAAAxxxxxxxxxxxxxx
    type: literal

```

error is like:
image

the code `fileUrl` is like `http://storagesvc.fission/v1/archive?id=124`, `storagesvcURL.String()` is like "localhost:52373" with port-forward out of k8s, so `strings.HasPrefix` should has `!`.
```
func DownloadStrorageURL(ctx context.Context, client cmd.Client, fileUrl string) (io.ReadCloser, error) {
	var resp *http.Response
	storagesvcURL, err := util.GetStorageURL(ctx, client)
	if err != nil {
		return nil, err
	}

	if !strings.HasPrefix(fileUrl, storagesvcURL.String()+"/v1/archive?id=") {
		url, err := url.Parse(fileUrl)
		if err != nil {
			return nil, err
		}
		id := url.Query().Get("id")

		client := storageSvcClient.MakeClient(storagesvcURL.String())
		resp, err = client.GetFile(ctx, id)
		if err != nil {
			return nil, err
		}
	} else {
		resp, err = http.Get(fileUrl)
	}

	if err != nil {
		return nil, err
	}
```

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/package/util/util.go
+++ b/pkg/fission-cli/cmd/package/util/util.go
@@ -170,7 +170,7 @@ func DownloadStrorageURL(ctx context.Context, client cmd.Client, fileUrl string)
 		return nil, err
 	}
 
-	if strings.HasPrefix(fileUrl, storagesvcURL.String()+"/v1/archive?id=") {
+	if !strings.HasPrefix(fileUrl, storagesvcURL.String()+"/v1/archive?id=") {
 		url, err := url.Parse(fileUrl)
 		if err != nil {
 			return nil, err


soharab-ic:
Created another PR https://github.com/fission/fission/pull/3051 for this.

PR #3100, opened 2024-12-13
Update create.go to handle namespace isolation

sofsms:
## Description

If the users on the cluster are namespace scoped they will not have permission to list environments from other namespaces. As you removed the Namespace variable from createEnvironmentFromCmd the m.Namespace will always be the default. This will not allow users to use fission cli to create environments

Fixes #

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/environment/create.go
+++ b/pkg/fission-cli/cmd/environment/create.go
@@ -64,16 +64,7 @@ func (opts *CreateSubCommand) complete(input cli.Input) error {
 // run write the resource to a spec file or create a fission CRD with remote fission server.
 // It also prints warning/error if necessary.
 func (opts *CreateSubCommand) run(input cli.Input) (err error) {
-	m := opts.env.ObjectMeta
-
-	envList, err := opts.Client().FissionClientSet.CoreV1().Environments(input.String(flagkey.Namespace)).List(input.Context(), metav1.ListOptions{})
-	if err != nil {
-		return err
-	} else if len(envList.Items) > 0 {
-		console.Verbose(2, "%d environment(s) are present in the %s namespace.  "+
-			"These environments are not isolated from each other; use separate namespaces if you need isolation.",
-			len(envList.Items), m.Namespace)
-	}
+	//m := opts.env.ObjectMeta
 
 	userDefinedNS, currentNS, err := opts.GetResourceNamespace(input, flagkey.NamespaceEnvironment)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sanketsudake:
@mergifyio rebase

pkg/fission-cli/cmd/environment/create.go
```diff
@@ -66,7 +66,7 @@ func (opts *CreateSubCommand) complete(input cli.Input) error {
 func (opts *CreateSubCommand) run(input cli.Input) (err error) {
 	m := opts.env.ObjectMeta
 
-	envList, err := opts.Client().FissionClientSet.CoreV1().Environments(m.Namespace).List(input.Context(), metav1.ListOptions{})
+	envList, err := opts.Client().FissionClientSet.CoreV1().Environments(input.String(flagkey.Namespace)).List(input.Context(), metav1.ListOptions{})
```
sanketsudake, line 69: This is not correct patch IMO. 

Namespace population should happen through  complete function, where we setup environment structure.
run just checks if if already exists, they tries to resolve it. 
We follow same approach in function/create.go

```
func (opts *CreateSubCommand) complete(input cli.Input) error {
	env, err := createEnvironmentFromCmd(input)
	if err != nil {
		return err
	}
	opts.env = env
	return nil
}
```

sanketsudake:
```
diff --git a/pkg/fission-cli/cmd/environment/create.go b/pkg/fission-cli/cmd/environment/create.go
index caa6a552..63b7def7 100644
--- a/pkg/fission-cli/cmd/environment/create.go
+++ b/pkg/fission-cli/cmd/environment/create.go
@@ -64,24 +64,22 @@ func (opts *CreateSubCommand) complete(input cli.Input) error {
 // run write the resource to a spec file or create a fission CRD with remote fission server.
 // It also prints warning/error if necessary.
 func (opts *CreateSubCommand) run(input cli.Input) (err error) {
-       m := opts.env.ObjectMeta
+       userDefinedNS, currentNS, err := opts.GetResourceNamespace(input, flagkey.NamespaceEnvironment)
+       if err != nil {
+               return fv1.AggregateValidationErrors("Environment", err)
+       }
+       // we use user provided NS in spec. While creating actual record we use the current context's NS.
+       opts.env.ObjectMeta.Namespace = userDefinedNS

-       envList, err := opts.Client().FissionClientSet.CoreV1().Environments(m.Namespace).List(input.Context(), metav1.ListOptions{})
+       envList, err := opts.Client().FissionClientSet.CoreV1().Environments(opts.env.ObjectMeta.Namespace).List(input.Context(), metav1.ListOptions{})
        if err != nil {
                return err
        } else if len(envList.Items) > 0 {
                console.Verbose(2, "%d environment(s) are present in the %s namespace.  "+
                        "These environments are not isolated from each other; use separate namespaces if you need isolation.",
-                       len(envList.Items), m.Namespace)
+                       len(envList.Items), opts.env.ObjectMeta.Namespace)
        }

-       userDefinedNS, currentNS, err := opts.GetResourceNamespace(input, flagkey.NamespaceEnvironment)
-       if err != nil {
-               return fv1.AggregateValidationErrors("Environment", err)
-       }
-       // we use user provided NS in spec. While creating actual record we use the current context's NS.
-       opts.env.ObjectMeta.Namespace = userDefinedNS
-
        // if we're writing a spec, don't call the API
        // save to spec file or display the spec to console
        if input.Bool(flagkey.SpecDry) {
```

Above would help probably, just reordering statements.

sofsms: Fixed :)

sanketsudake:
@sofsms Please fix formatting as lint is failing

sofsms:
> @sofsms Please fix formatting as lint is failing

Done

sanketsudake (approved):

PR #3136, opened 2025-01-10
PR-3051: Fix download deploy package out of k8s cluster

soharab-ic:
## Description

Original author for PR https://github.com/fission/fission/pull/3051 is inactive. Therefore created this duplicate PR with additional fixes.

Fixes #

## Testing

Verified these two scenarios:
- Deployment URL belongs to storagesvc.
- Deployment URL is outside k8s cluster.

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/package/util/util.go
+++ b/pkg/fission-cli/cmd/package/util/util.go
@@ -170,7 +170,7 @@ func DownloadStrorageURL(ctx context.Context, client cmd.Client, fileUrl string)
 		return nil, err
 	}
 
-	if strings.HasPrefix(fileUrl, storagesvcURL.String()+"/v1/archive?id=") {
+	if !strings.HasPrefix(fileUrl, storagesvcURL.String()+"/v1/archive?id=") {
 		url, err := url.Parse(fileUrl)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


pkg/fission-cli/cmd/package/util/util.go
```diff
@@ -165,12 +165,14 @@ func DownloadURL(fileUrl string) (io.ReadCloser, error) {
 
 func DownloadStrorageURL(ctx context.Context, client cmd.Client, fileUrl string) (io.ReadCloser, error) {
 	var resp *http.Response
-	storagesvcURL, err := util.GetStorageURL(ctx, client)
-	if err != nil {
-		return nil, err
-	}
+	var err error
+
+	if strings.Contains(fileUrl, "/v1/archive?id=") {
```
sanketsudake, line 170: Instead of `Contains` we should parse URL and check path. For better validation
Eg. Code below

```
func validateURL(urlStr string) (bool, error) {
    // Parse the URL string into a URL object
    parsedURL, err := url.Parse(urlStr)
    if err != nil {
        return false, fmt.Errorf("failed to parse URL: %v", err)
    }

    // Check if the path starts with /v1/archive
    if !strings.HasPrefix(parsedURL.Path, "/v1/archive") {
        return false, nil
    }

    // Get query parameters
    queryParams := parsedURL.Query()

    // Check if 'id' parameter exists
    if queryParams.Get("id") == "" {
        return false, nil
    }

    // URL matches all criteria
    return true, nil
}
```

sanketsudake (approved):

PR #3218, opened 2025-05-13
Create a publisher per cron timer

markretallack:
We want to be able to run multiple timers at the same time.

## Description

This pull request creates a publisher (used to trigger the function) per cron timer. This means that if a timer is triggered while an existing function is still running, the second function will be triggered when its meant to be. 

None

## Testing

Tested using two timers and two functions that runs a loop for 1.5 minutes.  Checked that both timers start correctly and both functions are run at the same time. 

## Checklist:


markretallack:
This is now ready for review, the CI is passing the tests etc...

sanketsudake (approved):

PR #3366, opened 2026-05-08
Drop fission-fetcher SA token from function user containers

sanketsudake:
## Description

Closes [GHSA-85g2-pmrx-r49q](https://github.com/fission/fission/security/advisories/GHSA-85g2-pmrx-r49q).

Function runtime pods (poolmgr + newdeploy backends) previously ran the user-code container under the `fission-fetcher` ServiceAccount, with the SA token auto-mounted by kubelet at `/var/run/secrets/kubernetes.io/serviceaccount`. Because that SA is granted namespace-wide `get` on `secrets` and `configmaps` (so the fetcher container can materialize `FunctionSpec.Secrets` / `FunctionSpec.ConfigMaps`), arbitrary user function code could read every secret in the function's namespace via the Kubernetes API — bypassing the `Spec.Secrets` allowlist.

This PR sets pod-level `AutomountServiceAccountToken=false` on every function pod and re-mounts the SA token via a private projected volume on the **fetcher container only**. The user container ends up with no SA token at all.

The clamp is re-applied after every `MergePodSpec` call so neither `gp.podSpecPatch`/`deploy.podSpecPatch` nor `env.Spec.Runtime.PodSpec` can re-enable the auto-mount or inject a competing mount at the SA token path.

## Behavioral change

Function code that legitimately called the Kubernetes API from inside a function pod will now fail with no token. Affected functions must declare a dedicated ServiceAccount via `Spec.PodSpec` and use it explicitly.

The container backend (`pkg/executor/executortype/container/`) is unaffected — it runs arbitrary user images without injecting `fission-fetcher`.

GHSA-85g2-pmrx-r49q (private security advisory).

## Testing

- New unit tests in both backends assert: (a) user container has no mount at the SA token path, (b) fetcher container has exactly one mount backed by the projected volume, (c) `env.Spec.Runtime.PodSpec.AutomountServiceAccountToken=true` cannot re-enable the mount, (d) a user-supplied fetcher `volumeMounts` entry at the SA token path is stripped before the projected mount is added (no kubelet duplicate-mount-path rejection).
- `make code-checks` clean, `make test-run` green locally and across all CI k8s versions (1.28, 1.32, 1.34).

## Checklist:

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3366)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Security
-
-- Function runtime pods no longer auto-mount the `fission-fetcher` ServiceAccount token in the user container.
-  User function code can no longer read namespace-wide secrets via the Kubernetes API from inside a function pod.
-  Fixes GHSA-85g2-pmrx-r49q.
-  Functions that legitimately call the Kubernetes API must declare a dedicated ServiceAccount via a `PodSpec` patch.
-
 ## [v1.17.0-rc1](https://github.com/fission/fission/tree/v1.17.0-rc1) (2022-07-06)
 
 [Full Changelog](https://github.com/fission/fission/compare/v1.16.0...v1.17.0-rc1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Security
+
+- Function runtime pods no longer auto-mount the `fission-fetcher` ServiceAccount token in the user container.
+  User function code can no longer read namespace-wide secrets via the Kubernetes API from inside a function pod.
+  Fixes GHSA-85g2-pmrx-r49q.
+  Functions that legitimately call the Kubernetes API must declare a dedicated ServiceAccount via a `PodSpec` patch.
+
 ## [v1.17.0-rc1](https://github.com/fission/fission/tree/v1.17.0-rc1) (2022-07-06)
 
 [Full Changelog](https://github.com/fission/fission/compare/v1.16.0...v1.17.0-rc1)
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -951,20 +951,7 @@ networkPolicy:
   enabled: false
 
 
-## Function-runtime pod security posture.
-##
-## The user-code container of every Fission function pod (poolmgr +
-## newdeploy backends) is started with automountServiceAccountToken=false.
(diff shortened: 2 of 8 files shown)


pkg/executor/executortype/newdeploy/newdeploy.go
```diff
@@ -237,6 +248,11 @@ func (deploy *NewDeploy) getDeploymentSpec(ctx context.Context, fn *fv1.Function
 		} else {
 			deploy.logger.Error(err, "Failed to merge the specs")
 		}
+		// Re-clamp after the merge: MergePodSpec propagates a non-nil
+		// AutomountServiceAccountToken from the patch, which would otherwise
+		// re-enable the kubelet auto-mount on the user container. See
+		// GHSA-85g2-pmrx-r49q.
+		pod.Spec.AutomountServiceAccountToken = ptr.To(false)
```
sanketsudake: use new(false) instead of ptr.To(false) everywhere.
So we wont need  explicit use of "k8s.io/utils/ptr"

charts/fission-all/values.yaml
```diff
@@ -949,3 +949,9 @@ grafana:
 ## Egress rules are intentionally not added in this round.
 networkPolicy:
   enabled: false
+
+
+# Function runtime user containers always run with
```
sanketsudake: Values is not used anywhere, we can revert any changes in values.yaml

pkg/executor/executortype/poolmgr/gp_deployment.go
```diff
@@ -168,6 +179,11 @@ func (gp *GenericPool) genDeploymentSpec(env *fv1.Environment) (*appsv1.Deployme
 		} else {
 			gp.logger.Error(err, "Failed to merge the specs")
 		}
+		// Re-clamp after the merge: MergePodSpec propagates a non-nil
+		// AutomountServiceAccountToken from the patch, which would otherwise
+		// re-enable the kubelet auto-mount on the user container. See
+		// GHSA-85g2-pmrx-r49q.
+		pod.Spec.AutomountServiceAccountToken = ptr.To(false)
```
sanketsudake: use new(false) instead of ptr.To(false). This is added in go 1.26 onwards.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
```
sanketsudake: Revert any changes in CHANGELOG for now, we will add them later.

PR #3422, opened 2026-05-29
feat: RFC-0005 WS3 — Reconciler scaffold + timer reconciler

sanketsudake:
## Summary

First PR of the RFC-0005 **WS3 Reconciler migration**: introduce the shared scaffold that lets Fission's hand-rolled, informer-driven controllers become controller-runtime native Reconcilers, and convert the **timer** subsystem as the first user (and the template for the rest).

This is the foundation the parallel stream PRs (kubewatcher, canary, mqtrigger, buildermgr, router, executor) build on.

## What's in `pkg/controller`

- **`Register`** — wires a Reconciler onto a Manager with the standard Fission event filter (`predicate.GenerationChangedPredicate`), so a controller doesn't re-trigger on its own status writes while Adds and Deletes still flow (deletion teardown relies on the delete event). Controllers inherit the Manager's leader-election scope automatically.
- **`SetConditions`** — a generic, best-effort status-condition writer that collapses the `IsAt` fast-path → `Get` → re-check → `Set` → `UpdateStatus` dance every subsystem hand-rolled (~30 lines each). It's typed by the generated clients through new `GetConditions()` accessors on the CRD types (`pkg/apis/core/v1/status_accessors.go`) and the `StatusClient` interface those clients already satisfy (compile-time asserted in the test).
- **`doc.go`** — documents the canonical `Reconcile` shape (Get → NotFound⇒teardown by NamespacedName → apply → write condition), and why deletion uses the not-found path rather than finalizers (teardown is in-memory only).

## Namespace-scoped cache

`crmanager.FissionCacheOptions` scopes a leader-elected Manager's shared cache to exactly the namespaces Fission watches (`DefaultNSResolver().FissionResourceNS`). Reconcilers watch through this cache, so this reproduces the per-namespace informers the subsystems used before the migration and keeps the RBAC footprint unchanged. The cache is inert until a controller registers an informer, so the subsystems still on `LeaderRunnable` are unaffected.

## Timer conversion (the template)

- `TimeTriggerReconciler` replaces `TimerSync` (informer + event handlers).
- The cron lifecycle moves onto `Timer` (`addUpdate`/`remove`), rekeyed by `NamespacedName` so a delete (no UID) tears the entry down via the NotFound path.
- Status goes through `controller.SetConditions`, removing `markTimeTriggerScheduled` and its `context.Background()` hack.
- Leader election preserved (same `fission-timer` lease; reconciler runs leader-only).

## Testing

- `go build ./...`, `go vet ./...`, `golangci-lint` on changed packages — clean.
- Unit tests: `pkg/controller` (SetConditions fast-path / write / generation-stamp + compile-time interface assertions) and `pkg/timer` (reconcile add → cron + conditions, idempotent re-reconcile, delete → teardown), via fake clientset.
- e2e: `./test/e2e/fetcher` and `./test/e2e/cli` pass (the in-process harness runs the timer reconciler and the namespace-scoped Managers together).

## Notes

- No CRD or API changes here. The Package/CanaryConfig `/status` subresource additions travel with the buildermgr and canary stream PRs that need them, where they're tested alongside the reconciler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3422)


pkg/utils/crmanager/crmanager.go
```diff
@@ -37,13 +39,30 @@ func NewLeaderElected(restConfig *rest.Config, lockName string, logger logr.Logg
...
+// before the controller-runtime migration and keeps RBAC requirements
+// unchanged. The cache is inert until a controller registers an informer, so
+// this is harmless for Managers that only run non-reconciler runnables.
+func FissionCacheOptions() cache.Options {
+	nsConfig := map[string]cache.Config{}
+	for _, ns := range utils.DefaultNSResolver().FissionResourceNS {
+		nsConfig[ns] = cache.Config{}
+	}
+	return cache.Options{DefaultNamespaces: nsConfig}
+}
```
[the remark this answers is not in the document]
sanketsudake: Added TestFissionCacheOptions in 61a9d370 asserting the resolved namespace set maps into the Manager cache's DefaultNamespaces. The env-driven resolution itself (default / FISSION_DEFAULT_NAMESPACE / FISSION_RESOURCE_NAMESPACES) is already covered by GetNamespaces in pkg/utils/namespace_test.go.

pkg/timer/reconciler.go
```diff
@@ -0,0 +1,59 @@
...
+// (applied in controller.Register) drops the status-only updates the old
+// UpdateFunc filtered by hand.
+type TimeTriggerReconciler struct {
+	logger        logr.Logger
+	fissionClient versioned.Interface
+	timer         *Timer
+}
+
+func (r *TimeTriggerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	tt, err := r.fissionClient.CoreV1().TimeTriggers(req.Namespace).Get(ctx, req.Name, metav1.GetOptions{})
```
sanketsudake: Reconciler is using fissionClient. It can use client with cache from controller ? to optimize reads from Kubernetes ?
Does our reconciler gets latest updates from kubernetes API server from informer and we are using those ?

sanketsudake: Good catch — you're right, and it matters since this is the template for every reconciler.

Today the controller's **watch** uses the Manager's informer cache (the `builder.For(&TimeTrigger{})` registration starts a cached informer), but the `Get` inside Reconcile went through the typed fissionClient — a direct API-server read that bypassed that warm cache. So we were paying an API round-trip per reconcile despite having the data cached.

Fixing it to the idiomatic pattern: the reconciler now reads through `mgr.GetClient()` (a cache-backed `client.Client` — Get/List served from the same informer cache the watch populates, no second cache), and status writes go through `client.Status().Update()` (writes always go direct to the API server's /status subresource). `SetConditions` is refactored to take `client.Client` accordingly. This removes the per-reconcile API read and is what we want for the high-frequency reconcilers this scaffold templates (router mux rebuilds, executor cold-start). Pushing the change now.

sanketsudake: Done in 55192142. Reconciler now reads via `mgr.GetClient()` (cache-backed) and writes status via `client.Status().Update`; `SetConditions` takes a `client.Client`. I'm propagating the same pattern to the stacked kubewatcher PR (#3423) and all subsequent stream reconcilers.

PR #3449, opened 2026-06-01
feat(crd): validate CRDs in the API server with CEL + server-side-apply markers

sanketsudake:
## What

Adds Kubernetes-native CEL validation (`x-kubernetes-validations`) and server-side-apply markers to the Fission CRDs, so the API server validates resources directly — the rules apply even if the admission webhook is not running, and GitOps tools (Argo CD, Flux) and `kubectl explain` can see them.

### Field rules now enforced by the API server
- Executor type/strategy enums and autoscale bounds (min/max scale, target CPU %).
- Archive / checksum / build-status enums.
- Environment version range; pool-size and termination-grace bounds.
- HTTPTrigger HTTP-method enums (the deprecated singular `method` may still be empty), host / CORS / ingress-path rules.
- FunctionReference name (DNS-1123) validation.
- Immutability of `Environment.spec.version`.

## Admission webhook

The admission webhook is **retained** for the checks CEL cannot express: cross-namespace references and pod-spec security (the API server does not expose `metadata.namespace` to CRD validation rules, and validating an embedded `PodSpec` exceeds the CEL cost budget), plus the package archive literal-size limit. No behavior change for those checks.

## Tests

`test/cel` exercises the new rules against a real API server (envtest): valid objects are admitted; invalid enums/ranges and immutable-field updates are rejected by the API server itself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3449)

### DIFF
--- a/crds/v1/fission.io_environments.yaml
+++ b/crds/v1/fission.io_environments.yaml
@@ -53,6 +53,9 @@ spec:
                   Available value:
                   - single
                   - infinite
+                enum:
+                - single
+                - infinite
                 type: string
               builder:
                 description: |-
@@ -10078,6 +10081,7 @@ spec:
                 type: boolean
               poolsize:
                 description: The initial pool size for environment
+                minimum: 0
                 type: integer
               resources:
                 description: |-
@@ -20162,6 +20166,7 @@ spec:
                   The grace time for pod to perform connection draining before termination. The unit is in seconds.
                   (Optional) defaults to 360 seconds
                 format: int64
+                minimum: 0
                 type: integer
               version:
                 description: |-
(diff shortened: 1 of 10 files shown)


sanketsudake:
Thanks for the review — addressed all points:

- **`test/cel` cleanup ignored `env.Stop()` error**: envtest is now started once per package via `TestMain`, and `env.Stop()` failures are written to stderr so leaked control-plane processes surface in CI logs.
- **`startEnv()` started a control plane per test**: replaced with a single `TestMain` control plane reused across cases (each test object already uses a unique name, so isolation is preserved). This also drops the package runtime noticeably.
- **Checksum CEL precedence / message**: added explicit parentheses — `((type empty) && (sum empty)) || type == 'sha256'` — and reworded the message to "checksum must be empty, or its type must be 'sha256'".
- **`CanaryConfig.weightincrement` Min/Max**: removed. The field has no `omitempty`, so a client leaving it `0` would have been newly rejected by `minimum: 1` — a behavior regression. It was a net-new constraint, so dropping it keeps the change non-breaking.

PR #3452, opened 2026-06-01
refactor(webhook): drop the HTTPTrigger, TimeTrigger and CanaryConfig admission webhooks

sanketsudake:
## What

These three CRDs no longer need an admission webhook now that the API server validates their field rules via CEL (`x-kubernetes-validations`):

- **HTTPTrigger / TimeTrigger** — their remaining parser-based rules (which CEL cannot express) move into the **router** and **timer** reconcilers and surface as status **Conditions** (`RouteAdmitted` / `Scheduled = False`) instead of admission rejections: invalid cron schedules, CORS origin/max-age, and ingress path regex.
- **CanaryConfig** — its webhook was a no-op.

The webhooks for **Function, Environment, Package, MessageQueueTrigger, KubernetesWatchTrigger** are **retained unchanged** — they enforce the checks CEL cannot express (cross-namespace references, pod-spec/container security, the environment runtime-image/name invariant, message-queue type/topic validity, reference-name DNS checks).

## Behavior change
A raw `kubectl apply` / GitOps write of an invalid cron, CORS origin, or ingress path is now **admitted and flagged with a `…=False` condition** rather than rejected at creation. The `fission` CLI still rejects these client-side, so the common path is unchanged.

## Upgrade note
The webhook no longer intercepts HTTPTrigger, TimeTrigger, or CanaryConfig. Operators relying on admission rejection of a malformed cron/CORS/ingress should check the new resource conditions (or rely on the CLI).

## Tests
- New unit tests: timer reconciler invalid-cron condition; router CORS/ingress config validation.
- Updated the CORS integration test (invalid origin is now admitted + `RouteAdmitted=False`).
- Helm webhook config and the e2e webhook manifest updated; full unit + envtest (cel/e2e) suites pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3452)


sanketsudake:
Thanks — both findings addressed:

- **Environment runtime image/name invariant**: restored. The environment webhook now re-checks that a runtime-podspec container reusing `spec.runtime.image` without an explicit command is named after the environment — this is a non-CEL, non-security correctness invariant that previously lived in `Environment.Validate()`.
- **MessageQueueTrigger stale subscription on invalid update**: resolved by keeping queue type/topic validity in the (retained) MQT webhook rather than the reconciler. Since the MQT webhook stays for the podspec allowlist anyway, moving type/topic to a reconciler condition added the early-return-without-unsubscribe drift you flagged (and depended on the validator registry being populated in the reconciler) without enabling any webhook deletion. Invalid type/topic is now rejected at admission as before, so the reconciler only ever sees valid triggers. The cron and CORS/ingress checks (whose webhooks *were* deleted) still move to reconciler conditions.

sanketsudake:
Reworked per review: rather than slimming the retained webhooks (which dropped non-CEL field checks like the package/function reference-name DNS validation you flagged), I reverted those 5 handlers to their original form and scoped this PR to **deleting the 3 webhooks that are fully covered by CEL + reconciler conditions** (httptrigger/timetrigger/canaryconfig). The retained webhooks keep `Validate()` (cross-ns + podspec + reference-name + env image/name + MQ type/topic), so no admission validation is lost. The stale service.go comment is fixed.

PR #3454, opened 2026-06-02
refactor(validation): split CRD validation into CEL-covered and webhook-only halves

sanketsudake:
## What

The Kubernetes API server now enforces the field-level CRD rules via CEL (`x-kubernetes-validations`), so the admission webhooks no longer need to re-validate them. This splits validation into two halves:

- Each webhook-relevant CRD type gains a `ValidateForAdmission()` method holding **only** the checks CEL cannot express. The 5 retained webhooks (function, environment, package, messagequeuetrigger, kuberneteswatchtrigger) call it instead of the full `Validate()`.
- The **fission CLI keeps calling the full `Validate()`** for client-side validation — `fission spec validate` / `--spec-save` are unchanged.

### What the webhooks still enforce (non-CEL)
- pod-spec / container security (function, environment) — the GHSA hardening
- the environment runtime-image/name invariant
- message-queue type/topic validity (validator registry)
- kubernetes-watch label-selector qualified key/value
- (inline, unchanged) cross-namespace references and the package archive literal-size limit

A CR's own `metadata.name` is left to the API server's built-in object-name validation.

### Moved to the API server (CEL)
The reference-name DNS-1123 checks (environment / secret / configmap / package names) are now enforced by the API server via `Pattern` markers instead of the webhook. Names that can legitimately be absent — the environment of a **container** function, and the optional **package** reference — are `omitempty` so an unset reference is skipped.

## Guarantees
- **No admission check is dropped** — every webhook security guarantee is preserved (verified: `pkg/webhook` + `pkg/apis/core/v1` tests green), and the CLI still validates everything client-side.
- Field-level rules are still enforced server-side (by CEL) and client-side (by the CLI); only the redundant webhook re-check is removed.

## Upgrade note
Minor, non-security loosenings for raw `kubectl`/GitOps writes (the `fission` CLI still rejects these): a package with an empty environment name, or a package reference with a namespace but an empty name, are no longer rejected at admission (they were nonsensical and fail later with a clear controller error).

## Tests
- New envtest fixtures (`test/cel`) prove the API server now rejects bad environment/secret/configmap/package names (and accepts an absent package/container-function environment), plus the relocated kubernetes-watch namespace cases.
- Webhook unit tests updated (the empty-namespace case moved to envtest); full `go test ./...` is green.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3454)

### DIFF
--- a/crds/v1/fission.io_functions.yaml
+++ b/crds/v1/fission.io_functions.yaml
@@ -719,6 +719,8 @@ spec:
                   description: ConfigMapReference is a reference to a kubernetes configmap.
                   properties:
                     name:
+                      maxLength: 63
+                      pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                       type: string
                     namespace:
                       type: string
@@ -738,11 +740,18 @@ spec:
                   function cannot be invoked.
                 properties:
                   name:
+                    description: |-
+                      Name of the referenced environment. Optional + omitempty: an unset
+                      reference is omitted and its Pattern skipped (a container function has
+                      no environment; a Package with an unset environment is admitted and
+                      fails later with a clear builder error — the fission CLI still rejects
+                      it). When set, it must be a DNS-1123 label.
+                    maxLength: 63
+                    pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                     type: string
                   namespace:
                     type: string
                 required:
-                - name
                 - namespace
                 type: object
               functionTimeout:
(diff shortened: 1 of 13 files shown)


sanketsudake:
Fixed the comment: `EnvironmentReference` is shared by FunctionSpec and PackageSpec, so the doc now says an *unset* reference is skipped (covering both the container-function case and the Package case, where an empty environment is admitted but fails later with a clear builder error and is still rejected by the CLI).

PR #3475, opened 2026-06-05
Concurrent Builder Pool with Scale-to-Zero + Newdeploy Build-Aware Provisioning

Stradivario:
## Summary

This PR ports the fork's concurrent builder pool and idle scale-to-zero onto upstream's controller-runtime reconcilers, and adds a newdeploy build-aware provisioning gate preventing the CrashLoopBackOff spiral when a function references a package that is still building.

---

### Phase 1 — `feat(types): add Builder.IdleTimeout and PoolSize to Environment spec`
Adds two optional fields to `Builder` in the Environment spec:
- `IdleTimeout *int64` — seconds of idle before scaling the builder pool to zero (default 600s; 0 = never).
- `PoolSize *int32` — max on-demand builder pods per environment (default 1).

Regenerated deepcopy and CRD YAML.

### Phase 2 — `feat(builder): bound concurrent builds per pod with a semaphore`
Protects the builder pod from concurrent build overload by adding a buffered-channel semaphore (cap = `MAX_PARALLEL_BUILDS`, default 1). The handler acquires/releases a slot around each build, logging queue depth. An unbuffered semaphore would deadlock every build, so `MakeBuilder` clamps values < 1 to 1.

### Phase 3–4 — `feat(buildermgr): concurrent builder pool + scale-to-zero on reconcilers`
Replaces the old environmentWatcher cache with a shared in-memory `BuilderPoolManager` injected into both the Environment and Package reconcilers, and a leader-only idle reaper:

- **`builderpool.go`** — `BuilderPoolManager`: per-env in-flight builds tracked as an idempotent set keyed by package (so requeues don't double-count), claimed builder pod IPs, and idle timing.
- **`idle_reaper.go`** — leader-only `Runnable` that scales builders with no in-flight builds to zero once their idle window elapses. Never scales down a building env.
- **`package_reconciler.go`** — `build()` now records demand, scales the builder UP toward demand (capped at `spec.builder.poolsize`), claims a Ready pod by IP, and routes the build to that pod. Waits via `RequeueAfter` (non-blocking) so it composes with bounded reconcile workers. A missing builder deployment requeues (not terminal). `RemoveBuild` releases demand if a package is deleted mid-wait.
- **`common.go`** — `buildPackage` takes a `builderPodIP`; pins fetch+build+upload to one pod (required for >1 replica since source lives on the pod's local volume), falling back to the builder Service DNS when the IP is empty.
- **`environment_reconciler.go`** — registers/forgets envs in the pool.
- **`scale.go`** — injectable `deploymentScaler` (the client-go fake lacks the scale subresource), shared by the reconciler and reaper.

### Phase 5 — `feat(newdeploy): wait for package build before provisioning`
Without this, a newdeploy function referencing a still-building package provisions a deployment whose fetcher sidecar CrashLoopBackOffs fetching a deploy archive that does not yet exist.

`waitForBuild` polls the referenced package's `BuildStatus` before `fnCreate` and `updateFuncDeployment` provision:
- `succeeded` / `none` → proceed.
- `failed` → error, skip provisioning (pod would crash-loop anyway).
- `pending` / `running` / empty → poll until it settles or `NEWDEPLOY_BUILD_WAIT_TIMEOUT` (default 600s) elapses, then provision anyway so a slow build never permanently blocks the function.

The poll returns `ctx.Err()` on shutdown rather than holding a reconcile worker for the full window.

### Phase 6 — `feat(cli): add --builder-idletimeout and --builder-poolsize env flags`
Exposes the Phase 1 fields through the CLI on `environment create` and `environment update`:
- `--builder-idletimeout` (`Int64` flag, seconds)
- `--builder-poolsize` (`Int32` flag, pod count)

Both flags are pointers — unset values fall back to the buildermgr defaults (600s idle, 1 pod).

---

## Testing

- `go build ./...` passes.
- `go test -race ./pkg/buildermgr/...` — BuilderPoolManager (demand idempotency, claim/release, reap eligibility, generation reset, leaked-demand cleanup), idle reaper (scales idle, protects busy), requeue-on-missing-deployment regression test.
- `go test -race` (full suite via `hack/runtests.sh`).
- Integration tests: HPA scale-up assertion under load, serial suite (executor restart + adopt pass).

@sanketsudake

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3475)


Stradivario:
@sanketsudake closing this one as stale.

It is a neat feature that hope at some moment you may decide it is worth looking at.

Cheers

PR #3476, opened 2026-06-05
Watch All Namespaces + Cross-Namespace Invocation Isolation

Stradivario:
## Summary

This PR ports two closely related features onto the controller-runtime codebase:

1. **Watch All Namespaces** — the fork's namesake feature, enabling Fission to manage functions and builders in any Kubernetes namespace, not just the install namespace.
2. **Cross-Namespace Invocation Isolation** — a router guard that prevents a function pod in one tenant namespace from invoking a private function in another namespace on the internal listener (`router-internal:8889`).

Both features are grounded in a Helm-chart RBAC restructure (cluster-scoped Roles when watch-all is on) and a shared namespace-propagation mechanism for the internal HMAC secret.

---

### `feat(namespaces): watch all namespaces (the fork's namesake feature)`

**Spike:** [docs/spike-watch-all-namespaces.md](docs/spike-watch-all-namespaces.md)

Ports the fork's watch-all-namespaces onto the controller-runtime architecture. The Go enabler is small: the new Manager cache already supports cluster-wide watching — `controller-runtime` treats the `""` (AllNamespaces) key in `cache.Options.DefaultNamespaces` as cluster-wide, and `FissionCacheOptions` builds that map from `NamespaceResolver.FissionResourceNS`. So `GetNamespaces()` just returns the `metav1.NamespaceAll` sentinel when `FISSION_WATCH_ALL_NAMESPACES` is set.

**Go changes:**
- `pkg/utils/namespace.go` — `FISSION_WATCH_ALL_NAMESPACES` env; `GetNamespaces()` returns the single `metav1.NamespaceAll` sentinel when set, taking all four managers (buildermgr, executor, router, mqtrigger) cluster-wide.
- `pkg/utils/serviceaccount.go` — `EnsureFetcherSA` / `EnsureBuilderSA` (logr); `runSACheck` skips the `NamespaceAll` sentinel and empty resolved namespaces.
- On-demand SA creation for dynamically-discovered namespaces (idempotent, harmless when not watch-all): executor (`poolmgr gp.setup`, `newdeploy fnCreate`) ensures the fetcher SA; buildermgr (`EnvironmentReconciler.ensureBuilder`) ensures the builder SA.

**Helm changes (cluster-wide cache requires cluster-scoped RBAC):**
- `values.yaml`: `watchAllNamespaces: true` (fork default — it is the watch-all distro).
- `_helpers.tpl`: sets `FISSION_WATCH_ALL_NAMESPACES=true` + `FISSION_RESOURCE_NAMESPACES=""` on every component when enabled.
- Role generators: emit `ClusterRole`/`ClusterRoleBinding` (one, gated to the default namespace) instead of per-namespace `Role`/`RoleBinding` when enabled.
- buildermgr/executor gain `serviceaccounts` + `roles/rolebindings` + `localsubjectaccessreviews create` so on-demand SA provisioning is not RBAC-forbidden.

Verified: `go build ./...`, `go test ./pkg/utils/... ./pkg/buildermgr/... ./pkg/executor/executortype/...`, helm lint, `helm template` renders ClusterRoles + watch-all env at `watchAllNamespaces=true` and per-namespace Roles at `=false`.

---

### `fix(namespaces): propagate fission-internal-auth Secret to dynamic namespaces`

Under watch-all-namespaces, function/builder pods land in namespaces the chart never knew about, so the `fission-internal-auth` Secret (the shared HMAC secret) is absent there. The fetcher sidecar's `FISSION_INTERNAL_AUTH_SECRET` secretKeyRef is `Optional`, so it resolves to empty, the fetcher signs nothing, and storagesvc rejects every `/v1/archive` request with HTTP 401 — functions in tenant namespaces fail to specialize and builder uploads fail. HMAC auth is an upstream addition the fork's v1.22 base never had, so the fork's watch-all never hit this.

**Fix:** `utils.EnsureInternalAuthSecret(ns)` — create-or-update the `fission-internal-auth` Secret in `ns`, mirroring the caller's own `FISSION_INTERNAL_AUTH_SECRET[_OLD]` env. No-op when internal auth is disabled or `ns` is empty; idempotent. Wired alongside `EnsureFetcherSA`/`EnsureBuilderSA` in `poolmgr gp.setup`, `newdeploy fnCreate`, and `buildermgr ensureBuilder`.

**RBAC:** buildermgr/executor gain `secrets create/update` (had only get/list/watch), so the on-demand secret write is not forbidden.

Tests: create-from-env, disabled no-op, idempotent/drift update, empty ns no-op. `go build ./...` and helm lint pass.

---

### `chore(chart): default internalAuth (HMAC) off on the fork`

**Feature doc:** [docs/features/internal-auth.md](docs/features/internal-auth.md)

Upstream defaults `internalAuth.enabled=true`, which makes the HMAC verifier on `router-internal` (and storagesvc) reject unsigned requests. The upstream KEDA HTTP connectors and the GraphQL federation gateway both send **unsigned** requests, so with HMAC on, MessageQueueTrigger delivery (RabbitMQ/Kafka/…) and federated subgraph calls fail with 401. This fork ran its whole v1.22.x life with no HMAC, so default it off to preserve that working behaviour (NetworkPolicies + namespace isolation still apply). Opt back in with `--set internalAuth.enabled=true` (then you must use signing-aware connector images).

- `values.yaml`: `internalAuth.enabled false` (+ rationale comment)
- `docs/features/internal-auth.md`: why, how to enable, and the all-or-nothing restart caveat (toggling needs all 8 secret-bearing components restarted, or verifiers/signers end up split-state → fetch failures / specialization timeouts)

Verified: helm lint; default render has no `FISSION_INTERNAL_AUTH_SECRET` env and no `fission-internal-auth` Secret; `--set internalAuth.enabled=true` restores them.

---

### `fix(namespaces): self-heal internal-auth Secret when HMAC is toggled off`

`EnsureInternalAuthSecret` now reconciles the tenant-namespace `fission-internal-auth` copy to the current internal-auth state instead of only creating it: when internal auth is disabled (empty `FISSION_INTERNAL_AUTH_SECRET` env) it **deletes** an existing copy. A leftover copy makes the function-pod fetcher sidecar (secretKeyRef always mounted) keep enforcing HMAC while the control plane no longer signs, so specialization fails with "error creating service for function: 401 Unauthorized". This is what bit a cluster that had internalAuth enabled and was then switched off: the per-namespace copies lingered and broke poolmgr/newdeploy specialization.

- delete-on-disable, scoped by a `Get` so the default-off case is a no-op
- RBAC: buildermgr/executor get `secrets delete` restricted via `resourceNames` to just `fission-internal-auth`
- test: disabled-deletes-a-leftover-secret; docs note updated

Production safety: a fresh deploy with `internalAuth.enabled=false` (the fork default) never creates these Secrets — verified `helm template` at the default has 0 `FISSION_INTERNAL_AUTH_SECRET` envs and 0 `fission-internal-auth` Secret objects, and the fetcher's secretKeyRef is Optional (missing → empty → pass-through). So no 401 regression on a clean install; this fix only matters when toggling a live one.

`go build ./...; go test ./pkg/utils/...; helm lint all pass.`

---

### `feat(router): same-namespace invocation guard for the internal listener`

**Feature doc:** [docs/features/internal-invocation-isolation.md](docs/features/internal-invocation-isolation.md)

Closes the cross-namespace invocation gap on `router-internal:8889`. The internal listener serves `/fission-function/<ns>/<name>` for **any** namespace, and with `internalAuth` disabled it is unauthenticated — so a function pod in one tenant namespace can invoke a private function in another. NetworkPolicy/HMAC gate **who** reaches the port, not **which** namespace they can target.

Opt-in via `router.enforceSameNamespaceInvocation` (default off). When enabled, a `/fission-function/<ns>/<name>` request is rejected (403) unless:
- the caller's pod is in the **same namespace** as the target function, **or**
- the caller is an **internal Fission component** — a pod in the install namespace (executor/kubewatcher/timer/mqtrigger), which legitimately invoke across namespaces.

This fits the real patterns: a federation gateway invoking its own-namespace subgraphs and a KEDA connector (deployed in the function's namespace) both pass; A→B cross-tenant invocation is blocked.

- Caller attributed by source pod IP (TCP peer; `X-Forwarded-For` ignored as spoofable) via a cluster-wide podIP→namespace informer cache; unresolved callers fail closed. Enforced per-function handler so the target namespace is unambiguous.
- Chart: `router.enforceSameNamespaceInvocation` value, `ROUTER_ENFORCE_SAME_NAMESPACE_INVOCATION` env, and a dedicated cluster-scoped ClusterRole granting the router pods get/list/watch (created only when enabled; works regardless of watchAllNamespaces).
- Reuses the router's existing `POD_NAMESPACE` downward-API env for the install ns.

Tests: same-ns allowed, internal-component cross-ns allowed, cross-ns 403, unresolved-IP 403; clientIP parsing (incl. IPv6); podIP cache incl. recycled-IP delete safety. `go build ./...`, `go test -race ./pkg/router/`, helm lint all pass; `helm template` gates env + ClusterRole on the value.

---

### `fix(router): resolve caller IP via API on same-namespace-guard cache miss`

The informer-backed podIP cache races a freshly created caller: the pod makes its first request before the watch has observed the pod's IP, so lookup misses and the guard fail-closes — wrongly rejecting legitimate same-namespace traffic (incl. a just-started federation gateway). Live verification showed same-namespace and internal-component callers getting 403 with "caller namespace unresolved".

**Fix:** on a cache miss, resolve the IP directly from the API server (server-side filtered by `status.podIP`, re-checked against the IP so it stays correct even against a client that ignores the field selector) and warm the cache. A truly unknown IP still returns not-found, so an unattributable source still fails closed. The warm informer cache remains the fast path for steady-state traffic.

The router's `fission-router-pod-reader` ClusterRole already grants pods list, so no RBAC change.

Test: `TestPodIPCacheAPIFallback` (cache miss resolves via fake clientset + warms).

---

### `docs(router): note same-namespace-guard API fallback + cold-start window`

Document that the guard resolves the caller IP via the informer cache with a direct-API fallback on miss, and that a brand-new pod's very first request (before its `status.podIP` is observable) fails closed and succeeds on retry — long-running callers (federation, functions) are unaffected, and the federation→subgraph cold-start path is safe (the caller is the long-running federation).

Verified live: cross-namespace 403, same-namespace 200, internal-component 200, federation passing.

---

## Feature Docs

| Document | Description |
|---|---|
| [docs/features/internal-auth.md](docs/features/internal-auth.md) | Why HMAC is off by default, what breaks if turned on without signing-aware callers, all-or-nothing operational caveat, tenant-namespace secret propagation under watch-all |
| [docs/features/internal-invocation-isolation.md](docs/features/internal-invocation-isolation.md) | How the same-namespace guard works, how it attributes callers, cold-start window, relationship to NetworkPolicy and HMAC controls, key files |

---

## Relationship Between the Controls

| Control | Restricts |
|---|---|
| `internalAuth` (HMAC) | *who* may call router-internal (signed callers only) |
| `networkPolicy` (router-internal ingress) | *which pods* may reach port 8889 |
| **`enforceSameNamespaceInvocation`** | *which namespace's functions* a caller may invoke |

---

## Testing

- `go build ./...`
- `go test ./pkg/utils/... ./pkg/buildermgr/... ./pkg/executor/executortype/...`
- `go test -race ./pkg/router/` (same-ns allowed, cross-ns 403, unresolved-IP 403, podIP cache API fallback, IPv6, recycled-IP delete safety)
- `helm lint`; `helm template --set watchAllNamespaces=true` asserts ClusterRole + `FISSION_WATCH_ALL_NAMESPACES=true` on all components; default render unchanged (per-namespace Roles)

@sanketsudake

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3476)

### DIFF
--- a/charts/fission-all/templates/_fission-component-roles.tpl
+++ b/charts/fission-all/templates/_fission-component-roles.tpl
@@ -27,6 +27,35 @@ rules:
   - get
   - update
   - patch
+# The buildermgr provisions the fission-builder ServiceAccount (and its Role +
+# RoleBinding) on demand in each builder namespace as environments are
+# discovered (EnsureBuilderSA -> setupSAAndRoleBindings). Required under
+# watch-all-namespaces, where namespaces are not known at install time and the
+# chart cannot pre-create the SA. Under watch-all these rules render as a
+# ClusterRole, granting the create power cluster-wide.
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - get
+  - delete
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - rolebindings
+  - roles
+  verbs:
+  - create
+  - get
+  - delete
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - localsubjectaccessreviews
+  verbs:
+  - create
 {{- end }}
 {{- define "executor-rules" }}
 rules:
@@ -54,6 +83,35 @@ rules:
   - get
   - update
   - patch
+# The executor provisions the fission-fetcher ServiceAccount (and its Role +
+# RoleBinding) on demand in each function namespace as functions are specialized
+# (EnsureFetcherSA -> setupSAAndRoleBindings). Required under watch-all-namespaces,
+# where namespaces are not known at install time and the chart cannot pre-create
+# the SA. Under watch-all these rules render as a ClusterRole, granting the create
+# power cluster-wide.
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - get
+  - delete
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - rolebindings
+  - roles
+  verbs:
+  - create
+  - get
+  - delete
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - localsubjectaccessreviews
+  verbs:
+  - create
 {{- end }}
 {{- define "kubewatcher-rules" }}
 rules:
--- a/charts/fission-all/templates/_fission-kubernetes-roles.tpl
+++ b/charts/fission-all/templates/_fission-kubernetes-roles.tpl
@@ -59,6 +59,17 @@ rules:
   - list
   - watch
 - apiGroups:
+  - ""
+  resources:
+  # The buildermgr provisions the fission-internal-auth Secret (the shared HMAC
+  # secret) on demand in each builder namespace under watch-all-namespaces, so the
+  # builder's fetcher can sign storagesvc requests. Needs create/update in addition
+  # to the get/list/watch above.
+  - secrets
+  verbs:
+  - create
+  - update
+- apiGroups:
   - apps
   resources:
   - deployments
@@ -151,7 +162,18 @@ rules:
   - get
   - list
   - watch
-{{- if .Values.executor.serviceAccountCheck.enabled }}  
+- apiGroups:
+  - ""
+  resources:
+  # The executor provisions the fission-internal-auth Secret (the shared HMAC
+  # secret) on demand in each function namespace under watch-all-namespaces, so the
+  # fetcher sidecar can sign storagesvc requests. Needs create/update in addition
+  # to the get/list/watch above.
+  - secrets
+  verbs:
+  - create
+  - update
+{{- if .Values.executor.serviceAccountCheck.enabled }}
 - apiGroups:
   - ""
   resources:
--- a/charts/fission-all/templates/_fission-kubernetes-roles.tpl
+++ b/charts/fission-all/templates/_fission-kubernetes-roles.tpl
@@ -70,6 +70,17 @@ rules:
   - create
   - update
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  # delete is scoped by name: the buildermgr removes the tenant fission-internal-auth
+  # copy when internal auth is turned off, so leftover copies don't make fetchers
+  # enforce HMAC against an unsigned control plane.
+  resourceNames:
+  - fission-internal-auth
+  verbs:
+  - delete
+- apiGroups:
   - apps
   resources:
   - deployments
@@ -173,6 +184,17 @@ rules:
   verbs:
   - create
   - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  # delete is scoped by name: the executor removes the tenant fission-internal-auth
+  # copy when internal auth is turned off, so leftover copies don't make fetchers
+  # enforce HMAC against an unsigned control plane.
+  resourceNames:
+  - fission-internal-auth
+  verbs:
+  - delete
 {{- if .Values.executor.serviceAccountCheck.enabled }}
 - apiGroups:
   - ""
--- a/charts/fission-all/templates/_fission-kuberntes-role-generator.tpl
+++ b/charts/fission-all/templates/_fission-kuberntes-role-generator.tpl
@@ -1,7 +1,11 @@
 {{- define "kubernetes-role-generator" }}
+{{- /* Under watch-all-namespaces emit a single cluster-scoped role (gated to the
+       default namespace so it is not duplicated per namespace); otherwise a
+       per-namespace Role as before. */ -}}
+{{- if or (not .Values.watchAllNamespaces) (eq .namespace .Values.defaultNamespace) -}}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: {{ if .Values.watchAllNamespaces }}ClusterRole{{ else }}Role{{ end }}
 metadata:
 {{- if eq "preupgrade" .component }}
   annotations:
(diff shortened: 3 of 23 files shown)


Stradivario:
This is like the 4th PR which i am making trying to help the community and yet still i can see nobody gives attention at all so this will be also my last PR doing anything here :) 

Months and months of waiting for anything. It looks like fission team are just too busy to accomplish some other people goals. Just make the repo private and develop it. 

Just to point out this PR fixes a security vulnerability which exposes every namespace and every person using fission at risk to access a lambda from different namespace since the internal fission router does not stop you from executing a function even if it is outside the current namespace. But do not look your PR's they are useless!
Cheers

@sanketsudake

sanketsudake:
@Stradivario Released multitenancy in Fission native way as part of 1.27 release. Please feel free to share your feedback if you see any shortcomings.  I wanted to isolate all important permissions to a new component so that current users are not hampered in any way if they want to stick with current threat model.
https://fission.io/docs/usage/multi-namespace-tenancy/


If you still see any shortcomings please feel free to file issue/PRs

Stradivario:
@sanketsudake Thank you very much man!

I really appreciate your work!

I have merged all changes from "main" branch to my fork and it appears that after my extensive testing i did not find anything buggy.

I found suitable "dynamic" strategy and tenant controller with labeling the fission resources and marking them as fission managable.

Thanks again for the great work i am going to publish a few more PR's which i find really neat for the platform but first i need to test them out. 

Closing this one since it diverged from the main by a lot and it is no longer mergable easily.

PR #3484, opened 2026-06-10
feat: OCI-native package delivery

sanketsudake:
## RFC-0001: OCI-Native Package Delivery

Function code can now be delivered as an OCI image (opt-in `Archive.OCI`) instead of a tarball, for both the poolmgr and newdeploy executors.
Cold starts skip the storagesvc GET + unzip, packages gain registry-native dedup/signing/scanning, and standard push tooling works out of the box.

### Path A — fetcher pull (works on the K8s 1.32 floor)
- New `pkg/oci`: pulls via go-containerregistry and extracts the flattened image under `os.Root` confinement (rejects traversal/symlink/hardlink entries, masks modes, 2 GiB decompression cap, optional sha256 digest pin).
- Credentials chain the fetcher SA's imagePullSecrets + per-package pull secrets via `pkg/authn/kubernetes` — no cloud SDKs; the fetcher binary grows only ~1.1 MiB.
- Fetcher: one branch in `Fetch()` — extract to a tmp dir, atomic rename to the store path, same idempotency contract as tarballs.
- Plain-HTTP registries require an explicit host allowlist (`fetcher.allowInsecureRegistries`, default empty = TLS required).
- RBAC delta: `serviceaccounts: get` on the namespaced fetcher Role (the keychain reads the SA's imagePullSecrets).

### Path B — kubelet image volumes (KEP-4639, opt-in via `executor.enableOCIImageVolume`, needs K8s >= 1.33)
- Poolmgr: per-image pools keyed `(env UID, image hash)` — byte-identical keys for non-OCI pools; pods mount the code read-only with **no fetcher container**, and specialization is a direct load call.
  Functions with Secrets/ConfigMaps or v1 envs automatically fall back to Path A.
- Newdeploy: the image volume mounts at the fetcher's store path, so the fetcher's exists-early-exit makes specialization load-only with zero fetcher change (the fetcher stays for secrets + load).
- The capability gate (`ENABLE_OCI_IMAGE_VOLUME` + server >= 1.33) is evaluated once per executor; below 1.33 OCI packages stay on Path A.

### API / CLI
- `Archive` gains an optional `oci` block (image, imagePullSecrets, subPath, digest); CEL + webhook enforce literal/url/oci mutual exclusivity.
  (CEL deliberately never references `literal`: any CEL access to a byte-format field makes the apiserver reject standard-base64 payloads — covered by the webhook instead.)
- `fission package|fn create/update --oci registry/repo:tag[@digest]`.
- Image-volume `subPath` must be a directory (kubelet rejects file subpaths); documented on the field.

### Tests
- Unit: extraction security negatives, keychain resolution, fetcher OCI fetch against an in-memory registry, pool keying/eligibility parity guards, deployment-spec shape for both executors (including post-MergePodSpec invariant re-clamps).
- Integration: registry-free CR reconciliation + CEL rejection; with the CI test registry — poolmgr/newdeploy Path A end-to-end, digest-mismatch negative, package-update rollout, Path B pod-shape proofs, and the secrets-fallback case.
- CI runs a `registry:2` Deployment on the 1.32 leg (proving the floor claim) and the 1.36 leg (image volumes).
  Verified empirically on kind v1.36.1 that no feature gates or containerd config are needed: the kubelet pulls the registry's NodePort at `localhost:30500` via containerd's built-in localhost exception.

### Using it — building a code image

The code image carries **only your code** (the runtime still comes from the environment image), laid out exactly like an extracted deployment archive:

```dockerfile
FROM scratch
COPY hello.py /
```

```bash
$ docker push registry.example.com/myteam/hello-code:v1
$ fission package create --name hello --env python --oci registry.example.com/myteam/hello-code:v1
$ fission fn create --name hello --pkg hello --entrypoint hello.main
```

**Base image recommendation: `FROM scratch`.** The image is never executed — Fission only reads its filesystem — so no shell/libc/OS layer is needed; a scratch code image is a few KB with zero CVE surface. Larger bases work too (the merged filesystem is extracted), ideally combined with `subPath` to point at the code directory. Spec-file fields beyond the CLI flag: `digest` (pin, recommended in production — tag re-pushes are not auto-detected), `subPath` (directory containing the code), `imagePullSecrets`.

A full user guide (image authoring, private/insecure registries, image-volume mode and its caveats) is prepared for fission.io, to ship with the v1.26.0 release docs.

### Backward compatibility
- Tarball path untouched and default; all new fields/flags/Helm values optional and off by default.
- Non-OCI pool keys and pod specs are locked byte-for-byte by parity tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/fission/fission/3484)


sanketsudake:
Two additions since the last green run:

**Compiled-language e2e coverage** — `TestOCIPackageGoCompiled` builds a Go plugin on-cluster with the env's builder, downloads the built `.so` via `package getdeploy`, repackages it as an OCI image, and serves it: binary artifacts survive the push/pull/extract path intact.

**Multi-agent review round** (`143707cc`) — five specialized review passes over the full diff; fixes:
- Pool identity now hashes all pod-spec-affecting archive fields (reference+digest, subPath, pull secrets) — keying by image alone could serve the wrong code root for same-image/different-subPath packages.
- Per-image (Path B) pool deployments are adopted on executor restart, so the post-adopt reaper no longer tears down warm Path B pools.
- Package-read errors during eligibility no longer silently downgrade the delivery mode (only NotFound falls back; transient errors fail the reconcile/cold start and retry).
- `package getdeploy` errors cleanly on OCI packages (was a nil-reader panic); `package update --oci` resets a stale failed/pending build status.
- New admission validation: digest-in-both-places conflicts, absolute/unclean `subPath`, OCI on source archives. Plus hardening of the image-volume helper, the Path B gate, and the load-only specialize retry loop.

Documented follow-ups (out of scope here): per-image stale-pool reaper after `package update --oci` on Path B, a serial executor-restart test with live Path B pools, and a spec save/apply round-trip test.

PR #3485, opened 2026-06-10
feat: RFC-0002 EndpointSlice-native data plane (phases 0-3)

sanketsudake:
## Summary

Implements phases 0–3 of RFC-0002: the router discovers function addresses from Kubernetes Services + EndpointSlices instead of HTTP-calling the executor on the warm path.
The design thesis is *"router admits, executor provisions"* — the executor narrows to provisioning (specialize, scale, reap), the router serves warm traffic from a slice-fed endpoint index.
Everything ships behind feature gates that default **off**; with gates at defaults this PR is behavior-neutral.

The cold-start path is byte-identical to today: the synchronous `getServiceForFunction` RPC returning a pod IP is untouched, preserving the ~100ms poolmgr cold-start budget.
The two new pod labels ride existing patches (generation in the `choosePod` relabel, `fission.io/served` in the post-specialize patch), so the cold path gains zero API writes.

## Why

Today every poolmgr request pays two executor RPCs (lookup + untap) and funnels through a single PoolCache goroutine serializing all poolmgr traffic cluster-wide — and the executor is a single point of failure for the entire poolmgr data plane.
Newdeploy scale-from-zero requests climb a dial-fail exponential backoff ladder (often 1–10s) before the executor is asked to scale up.
With the gates on, warm traffic flows with zero executor RPCs and survives executor downtime/upgrades; scale-from-zero is detected from slice emptiness and goes straight to the executor.

## How

**Executor (`executor.functionServices.enabled` → `ENABLE_FUNCTION_SERVICES`, default off):**
- Each invoked poolmgr function gets an async headless selector Service (`fn-<name>-<uid8>`); the built-in EndpointSlice controller publishes its specialized pods — no custom slice writer, no new executor RBAC.
- The selector gates on `fission.io/served` (set only after specialization succeeds) and `fission.io/function-generation` (stale-generation pods drop out on function update).
- Services are lifecycle-managed like other executor objects: owner ref when same-namespace, reconciler delete hook, instanceID adoption across restarts, orphan reaping.
- The request multiplexer is now `dispatch.Dispatcher`: per-function dedup with context-aware waiters (newdeploy/container) and independent runs (poolmgr), optionally bounded via `EXECUTOR_SPECIALIZATION_CONCURRENCY` (default 0 = today's unbounded).
  This fixes a real bug: concurrent newdeploy waiters previously re-read the cache and unconditionally wrapped a nil error, so they always received an error.
- New `POST /v2/ensureCapacity`: the router's saturation path; enforces the function concurrency cap via PoolCache (still the capacity authority) or specializes one more pod.
- Idle reaping becomes drain-aware when the gate is on: unlabel `served` (pod leaves slices, routers stop admitting within watch latency) → drain grace `clamp(fnTimeout, 30s, 5m)` → delete.

**Router (`router.endpointSliceCache.mode` → `ROUTER_ENDPOINTSLICE_CACHE_MODE`, `off|shadow|on`, default off):**
- One label-filtered EndpointSlice informer feeds a sharded copy-on-write endpoint index; every replica watches independently (no leader election, by design).
- `shadow`: every executor answer is compared against the index and classified into `fission_router_endpointcache_shadow_total{result=match|miss|lag}` — zero routing change; the steady-state non-match rate is the promotion criterion to `on`.
- `on`: poolmgr warm requests are admitted from the index (least-outstanding below `requestsPerPod`, per-pod atomic counters, CAS admission) with a release closure replacing the per-request UnTap RPC; saturation goes to `ensureCapacity` (a 404 from an older executor degrades to the legacy RPC); dial-failed endpoints are quarantined until the next slice event.
- Enforcement is per-router-replica by design (worst case ~1 extra in-flight request per pod per extra replica, degrading to brief queueing); `fission.io/concurrency-enforcement: strict` on a Function keeps today's exact RPC+UnTap path.
- Newdeploy/container keep dialing the Service DNS; zero ready endpoints bypasses the address cache and RPCs the executor proactively (replacing the backoff ladder).
- Structural track: `functionHandler.go` (988 lines, six responsibilities, a circular transport back-pointer) is decomposed along injected `AddressResolver`/`Tapper` seams — the retry transport is unit-testable against fakes for the first time.

**Compat:** streaming (RFC-0008 heartbeat/untap-on-drain), canary weighted routing, MCP, and Gateway API are address-source-only changes and keep working; upgrade order is safe in any direction (either side old → RPC fallback).
RBAC delta is read-only (`discovery.k8s.io/endpointslices` + core `services` get/list/watch on the router); NetworkPolicy is unchanged (the router already dials pod IPs).

## Verification

- Unit (all `-race`): dispatcher dedup/cancellation/bounding, index event storms with 50 concurrent readers, shadow classification table, transport retry/invalidation matrix against fake resolvers, fallback decision table (warm/cold/strict/saturated/429/degrade/quarantine), Service-ensure idempotency, drain-then-delete.
- Integration: `TestPoolmgrFunctionEndpointSlice` (Service+slice lifecycle), `TestWarmInvokeWithExecutorDown` (the headline: five consecutive warm invokes succeed with the executor scaled to zero), `TestMixedGateUpgradeOrder` (upgrade-order safety).
- CI: kind-ci runs with both gates **on** across all three k8s versions; the v1.34 leg pins the gates off post-deploy so the legacy RPC path stays covered.
- `make test-run` (full repo, race detector) green locally; `make code-checks` + `make license-check` clean; helm renders verified for both gate settings.

## Performance (measured in CI, gates off vs on — same commit)

Full data in the two analysis comments: [latency + traffic mix](https://github.com/fission/fission/pull/3485#issuecomment-4677020101) and [pprof + runtime resources](https://github.com/fission/fission/pull/3485#issuecomment-4677147638).

- Router overhead (`fission_function_overhead_seconds`) over the functional suite: median **407ms → 7.5ms**, mean 2.33s → 1.62s.
- Warm-path traffic mix with gates on: 521 index hits / 82 misses / ~29 fallbacks — **~82% of requests complete with zero executor RPCs**.
- Resources: the router gains exactly one informer (~14 goroutines, sub-MB, constant per replica, verified frame-by-frame in pprof); executor goroutine/heap/RSS/CPU per pod are unchanged within noise; no leak signature across load, rollouts, or idle (Prometheus trend returns to a flat baseline matching the pprof snapshots).
- Cold start is structurally unchanged (same RPC, same bytes, zero added API writes); the dedicated cold-start microbenchmark + k6 burst runbook remain the pre-phase-4 gate.

**Not in this PR (phase 4, next minor release):** defaults flip, optional newdeploy endpoint-level LB, and deletion of the then-dead legacy paths.


sanketsudake:
## Verification summary (merge-readiness)

**CI: all substantive checks green** across k8s 1.32 / 1.34 / 1.36 — three integration legs (gates **on** for 1.32/1.36, legacy-pinned `mode=off` for 1.34), upgrade test, lint, CodeQL, codecov.
`License Compliance` (FOSSA) is the documented repo-wide expected failure (red on `main` too) and not a merge blocker.

**Hardening that came out of driving CI green over 9 rounds** (each verified by a subsequent green run):
- RBAC for the slice watch is granted where the slices actually live (`router-dataplane` Role per function namespace), with a fail-soft preflight (degrade to legacy + loud error, never crash-loop) and controller-runtime logs finally visible in router/executor.
- Poolmgr tap liveness fixed (`TouchByAddress` falls back to the PoolCache): taps were silently 404ing since RFC-0008, and `mode=on` would have idle-reaped serving pods.
- `ensureCapacity` is an atomic reserve-or-429 inside the PoolCache actor (no TOCTOU past the concurrency cap, no accounting corruption, no nil-map executor crash).
- Streaming admission slots release on re-resolve; dial-failed endpoints quarantine immediately; admission refusals are now named (`all_busy|all_quarantined|...`) in logs and the fallback metric.
- The function Service ensure self-heals under traffic (debounced re-ensure on the warm RPC path) — a lost one-shot ensure no longer strands a function without slices.
- Specialization keeps `main`'s caller-cancellation semantics for poolmgr (load shedding); the cancellation-detach applies only where dedup bounds it (newdeploy/container), fixing coalesced-waiter poisoning.
- CI robustness: self-healing port-forwards (a router liveness blip cost 40 tests), runner disk freed before the image preload, and the legacy data plane stays covered on the v1.34 leg.

**Performance signal** (Prometheus dumps from one run, same suite; v1.34 legacy vs v1.36 EndpointSlice path):

| metric (whole run) | legacy | EndpointSlice |
|---|---|---|
| router overhead median | ~407 ms | ~7.5 ms |
| router overhead mean | ~2.33 s | ~1.62 s |
| index hits / misses / fallbacks | — | 521 / 82 / ~29 (≈82% of requests with zero executor RPCs) |

Caveat: cross-leg comparison on a functional suite; the rigorous pre-phase-4 measurement remains the runbook k6 off-vs-on burst test and the cold-start microbenchmark (cold starts ride the identical synchronous RPC by design).

Review process: two comprehensive multi-angle review passes were run on the branch; all confirmed findings are fixed in-branch with regression tests.

sanketsudake:
## Resource usage: pprof + Prometheus runtime-metrics cross-check

Companion to the latency comparison above — this verifies the RFC's resource expectations ("the informer adds a constant handful of goroutines, not per-function; executor unchanged") from two independent data sources on the all-green run [27321703131](https://github.com/fission/fission/actions/runs/27321703131): the post-run pprof artifacts and the per-leg Prometheus TSDB dumps.
The v1.34.8 leg pins the gates **off** (legacy data plane); the v1.36.1 leg runs gates **on**.

### Goroutines — pprof snapshots (post-run idle)

| Profile | Router | Executor |
|---|---|---|
| `main` (v1.34 / v1.36 legs) | 75 / 72 | — / 228 |
| PR, gates off (v1.34 leg) | 72 | — |
| PR, gates on (v1.36 leg) | 86 | 233 |

The router's +14 with gates on is attributed line-by-line: every client-go informer frame (`Reflector.startResync`, `handleAnyWatch`, `sharedProcessor.run`, `processorListener.run`/`pop`, `controller.RunWithContext`, `Informers.Start`) goes from 2 to 3 — exactly one new informer (the EndpointSlice watch) plus its HTTP/2 watch stream and context plumbing.
The fission app-frame fingerprint is **identical** between gates-off and gates-on: admission (`Index.Admit`, release closures, quarantine) runs on atomic counters inside the request goroutine and leaves zero resident goroutines.
The executor's frame fingerprint is byte-identical to `main` (the ±5 sits in shared runtime frames that vary run to run); the dispatcher leaves no residue and no drain-timer buildup appears.

### Runtime metrics over the whole run — Prometheus, per pod

| Metric (per pod) | Router, gates off | Router, gates on | Executor, gates off | Executor, gates on |
|---|---|---|---|---|
| Goroutines avg / max | 79 / 114 | 96 / 119 | 243 / 320 | 244 / 278 |
| Heap inuse avg / max (MB) | 13.6 / 17.4 | 13.2 / 17.0 | 15.2 / 21.7 | 13.5 / 22.5 |
| RSS max (MB) | 73.4 | 74.8 | 82.5 | 81.6 |
| CPU over run | 4.5s / 19min | 5.5s / 22min | 8.4s | 9.0s |

Leak check: both router legs rise under load and return to a flat baseline — gates-off settles at exactly 72 (matching its pprof snapshot), gates-on at ~86–92 (matching pprof's 86); the last-third average is *lower* than the first-third in both legs, so there is no monotonic growth.
Router CPU normalizes to ~0.24 vs ~0.25 s/min — indistinguishable, and the gates-on leg runs *more* tests (the slice suite doesn't skip there).
The executor absorbed twice the rollout churn on the gates-on leg (serial gate/scale-to-zero tests) without moving any per-pod number.

### Heap — pprof

All totals sit in a 5–12 MB band; the deltas are gzip `/metrics` scrape buffers caught mid-flight (the known noise pattern from the RFC-0005 baseline work), and same-code legs vary as much as cross-code legs.
The new code itself — endpoint index, resolvers, dispatcher, ensure-dedup map — appears **nowhere** in any heap profile (below the 512 KB sampling threshold), consistent with ~1 KB per function entry at this scale (max index size 10 in this run).

### Conclusion

The EndpointSlice data plane costs one informer of fixed overhead per router replica (~14 goroutines, sub-MB) and nothing measurable on the executor, with no leak signature under load, across rollouts, or at idle.
Caveat: the kind cluster is tiny, so scale-dependent index cost (RFC projects <50 MB at 10k functions from ~1 KB/entry cardinality) is verified by reasoning, not by these profiles — the pre-phase-4 k6 runbook remains the rigorous gate.

PR #3491, opened 2026-06-11
perf(router): RFC-0014 — shared transport (fix never-working keep-alive), allocation diet, resolver-cache removal

sanketsudake:
## Summary

Implements RFC-0014 (router hot-path efficiency), headlined by a genuine bug: **the proxy built a fresh `http.Transport` per request, so HTTP keep-alive has never worked** — every proxied request paid a TCP dial to the function pod, and the configured `MaxIdleConns: 100` was dead config.
Four bisectable phases: baseline harness → shared transport → handler hoists → resolver-cache removal, plus review follow-ups.

## Measured impact (unit benchmarks, this branch)

| Metric | Before | After | Delta |
|---|---|---|---|
| Warm RoundTrip wall time | 253,358 ns/op | 89,007 ns/op | **−65%** |
| Allocations | 45,708 B/op / 267 allocs | 19,518 B/op / 165 allocs | **−57% / −38%** |
| TCP dials for 20 serialized requests | 20 | ≤2 | pooling restored |

(End-to-end k6 numbers from a kind before/after run are posted as a PR comment.)

## What changed

**Phase 1 — shared transport (the fix):** one `http.Transport` per router process, lazily built on the shared `tsRoundTripperParams` (`sync.Once`); the `otelhttp` wrapper built once instead of per attempt.
The per-attempt dial timeout — which is the backoff-scaled **fast-retry ladder for cold pods**, not just a timeout — moves into a context value read by the shared `DialContext`; a deadline-exceeded dial classifies identically (`*net.OpError{Op:dial}, Timeout()==true`), pinned by `TestDialLadderTimeoutClassification`, and the ladder's per-attempt **growth** is pinned by `TestDialLadderGrowsPerAttempt`.
Sizing: `MaxIdleConnsPerHost` 32 (each poolmgr pod is its own host; tunable via `ROUTER_ROUND_TRIP_MAX_IDLE_CONNS_PER_HOST`), `IdleConnTimeout` 30s (under the poolmgr reap window).

**Chart bug fixed along the way:** `ROUTER_ROUND_TRIP_DISABLE_KEEP_ALIVE` used `| default true`, and sprig's `default` treats `false` as empty — the env rendered `"true"` unconditionally, making keep-alive impossible to enable even explicitly.
Previously masked (the per-request transport made the setting moot); now it is the meaningful escape hatch back to per-request connections.

**Phase 2 — handler hoists:** the round-tripper logger and per-backend proxy policies are computed once in `buildMuxes` (per backend UID, covering canary) instead of per request.
Honest numbers: marginal (217→215 allocs/op on the full handler path) — phase 1 was the lever; this is hygiene with a parity test.

**Phase 3 — resolver cache removal:** the `functionReferenceResolver`'s trigger-RV-keyed result cache was a cache of the Manager informer cache, used only during mux rebuilds (never on the request path).
Removed: −2 goroutines per router, no more full-map `Copy()` walk on every Function reconcile, and the trigger-RV staleness class is structurally gone (resolution always reads current state — pinned by test).

**Review follow-ups** (two-agent review; code-reviewer verdict: clean): distinct Info log for the pooled-conn failure class ("pooled connection failed mid-request; pod may have been reaped" — deliberately not quarantined: the failure indicts the connection, not the pod); startup log of effective transport settings; defensive 30s dial cap when the ladder value is absent; config soft-fail tests.

## The one behavioral shift, documented

A stale pooled conn to a reaped pod fails at write/read instead of dial.
Graceful reaps FIN-evict the conn from the pool; Go auto-retries replayable requests; a non-replayable POST (and the rare raced GET that may have executed) surfaces the error with the distinct log line — pinned by `TestStaleConnectionPOSTSurfacesErrorWithoutQuarantine` and the race-tolerant recovery test.
The quarantine path is untouched (it keys on dial errors, correctly).
Escape hatch: `router.roundTrip.disableKeepAlive: true`.

**Release-notes material:** every install effectively had keep-alive disabled (the chart coercion); after upgrade, pooling is ON by default with the settings logged at router startup.

## Verification

- New tests: connection-reuse regression (20 req ≤2 conns), ladder classification + growth pins, stale-conn POST/GET behavior, policy-hoist parity, fresh-resolution-after-fn-update, config soft-fail table; benchmarks committed.
- The frozen matrices — transport retry, settle dispatch, streaming release, canary, GHSA route pins — pass **unchanged**.
- `make code-checks` 0 issues; `make license-check` clean; helm renders verified for default/`true`/explicit-`false`.

### DIFF
--- a/charts/fission-all/templates/router/deployment.yaml
+++ b/charts/fission-all/templates/router/deployment.yaml
@@ -75,8 +75,14 @@ spec:
           value: {{ .Values.router.roundTrip.timeoutExponent | default 2 | quote }}
         - name: ROUTER_ROUND_TRIP_KEEP_ALIVE_TIME
           value: {{ .Values.router.roundTrip.keepAliveTime | default "30s" | quote }}
+        # NOTE: not `| default true`: sprig's default treats false as empty, so
+        # an explicit `disableKeepAlive: false` would silently coerce back to
+        # "true" and make keep-alive impossible to enable (the values.yaml
+        # default is false; nil also renders "false").
         - name: ROUTER_ROUND_TRIP_DISABLE_KEEP_ALIVE
-          value: {{ .Values.router.roundTrip.disableKeepAlive | default true | quote }}
+          value: {{ if .Values.router.roundTrip.disableKeepAlive }}"true"{{ else }}"false"{{ end }}
+        - name: ROUTER_ROUND_TRIP_MAX_IDLE_CONNS_PER_HOST
+          value: {{ .Values.router.roundTrip.maxIdleConnsPerHost | default 32 | quote }}
         - name: ROUTER_ROUND_TRIP_MAX_RETRIES
           value: {{ .Values.router.roundTrip.maxRetries | default 10 | quote }}
         - name: ROUTER_SVC_ADDRESS_MAX_RETRIES
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -426,8 +426,16 @@ router:
     ## so that kubernetes will be able to reap old function pod quickly.
     ##
     ## For details, see https://github.com/fission/fission/issues/723
+    ## Since RFC-0014 the router pools connections to function pods; this is
+    ## the escape hatch back to a fresh connection per request.
     ##
     disableKeepAlive: false
+
+    ## maxIdleConnsPerHost bounds the router's pooled idle connections per
+    ## function address (each poolmgr pod is its own address). 0 uses the
+    ## built-in default (32).
+    ##
+    maxIdleConnsPerHost: 32
 
     ## keepAliveTime is period for an active network connection to function pod.
     ##
--- a/charts/fission-all/values.yaml
+++ b/charts/fission-all/values.yaml
@@ -432,8 +432,7 @@ router:
     disableKeepAlive: false
 
     ## maxIdleConnsPerHost bounds the router's pooled idle connections per
-    ## function address (each poolmgr pod is its own address). 0 uses the
-    ## built-in default (32).
+    ## function address (each poolmgr pod is its own address).
     ##
     maxIdleConnsPerHost: 32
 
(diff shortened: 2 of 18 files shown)


sanketsudake:
## End-to-end performance evidence (local kind, main vs this branch, same cluster/hardware)

Method: kind v1.36.1, `SKAFFOLD_PROFILE=kind make skaffold-deploy` of **main** → warm-path k6 (`test/benchmark/tests/warm-path`, 20 VUs × 60s, one pre-warmed poolmgr function, `requestsPerPod=200`) → redeploy **this branch** → identical run.

| Metric | main | RFC-0014 | Delta |
|---|---|---|---|
| median | 4.17ms | 3.70ms | **−11.3%** |
| p95 | 6.97ms | 6.16ms | −11.6% |
| p99 | 9.34ms | 8.33ms | **−10.8%** |
| **max** | **185.33ms** | **28.73ms** | **−84%** (the dial tail, gone) |
| throughput (closed-loop) | 266,536 req (4,442 rps) | 302,913 req (**5,048 rps**) | **+13.6%** |
| failures | 0.00% | 0.00% | — |

The max-latency collapse is the fix's signature: with a fresh transport per request, an unlucky dial (SYN loss, scheduler hiccup) hit the long tail on any request; with pooling, steady-state requests never dial.

Unit-level (committed benchmarks): warm RoundTrip 253µs → 89µs (−65%), 45.7KB → 19.5KB, 267 → 165 allocs/op; 20 serialized requests now ride ≤2 connections (was 20 dials).

Goroutines at true idle: 87 (main) → 86 (branch) — connection pooling adds nothing at idle (pooled-conn read/write loops decay with the 30s `IdleConnTimeout`; transient +~40 directly after load is expected and observed), and the resolver-cache removal's −2 goroutines land as predicted.

The new startup log verifies effective settings on the deployed chart:
```
router proxy transport configured  disableKeepAlive=false maxIdleConnsPerHost=32 idleConnTimeout=30
```
(`disableKeepAlive=false` is itself part of the fix — the chart's `| default true` previously coerced every install to `"true"`.)

sanketsudake:
Side effect worth noting: the resolver-cache removal (phase 3) structurally eliminates the trigger-RV-keyed staleness that was root-caused as the mechanism behind the long-standing `TestGoEnv` flake (a function-only `fn update --pkg` could serve a stale function snapshot from the cache because the trigger's ResourceVersion never changed). The `currentFunction` re-read already mitigated the specialization path; with the cache gone, the staleness class no longer exists at all.

PR #3493, opened 2026-06-12
perf(router): RFC-0013 — incremental route updates (route table + handler indirection, precedence + RouteConflict conditions)

sanketsudake:
Implements [RFC-0013](docs/rfc/0013-incremental-router-routes.md) phases 0–2: the router's trigger/function reconcilers now feed per-event diffs into a route table instead of rebuilding both muxes from full LISTs on every event.
Route **shape** (path/prefix/methods/host) is split from **handler** (function snapshots, canary weights): handlers live behind a stable `HandlerRef` (one `atomic.Pointer`), so the steady-churn class — canary weight ticks, function updates — is a single pointer swap with **zero mux rebuild**, while shape changes (trigger create/delete, path/method/host edits) signal a debounced materializer that rebuilds gorilla muxes from a precedence-ordered table snapshot.

## Measured

Apple M2, fake-cache resolver, `-benchtime=10x -count=3`:

| Event | Legacy (full rebuild) | Incremental | |
|---|---|---|---|
| Canary weight tick, 10k triggers in table | **213 ms / 297 MB / 3.56 M allocs** | **21 µs / 9.7 KB / 96 allocs** | ~10,000× |
| same, 1k triggers | 21.5 ms / 30 MB / 348 k allocs | (same ~21 µs — O(1) in table size) | |

`BenchmarkMuxMatch` at 10k routes (phase-3 gate evidence): first-route 0.7 µs, last-route ~0.7 ms, miss ~0.7 ms — **under the RFC's >1 ms build trigger for the native matcher, so phase 3 stays unbuilt** (gate stays armed; the benchmark is in-tree).

## What changed

- **Phase 0** — golden tests pinning today's route-shape derivation (dual-registration boundary, empty-methods dead route, host, CORS OPTIONS append, GKE `/` fallback, internal namespace folding) + `BenchmarkBuildMuxes{100,1k,10k}` / `BenchmarkMuxMatch` baselines + `test/benchmark/tests/route-churn/` generator.
- **Refactor** — `buildMuxes`' inline derivation/registration moved into shared helpers (`routeshape.go`); the legacy path and the materializer register routes through literally the same code, pinned by the golden tests plus a dispatch-parity corpus across both modes.
- **Phase 1** — `pkg/router/routetable` (HandlerRef, RouteSpec, fn index, unresolved index); reconcilers become per-event diff sources; debounced materializer; 60s cache-fed resync as the drift guard; `ROUTER_INCREMENTAL_ROUTES` escape hatch (chart `router.incrementalRoutes`, rendered if/else so an explicit `false` survives sprig).
- **Phase 2** — specified precedence (hosted > host-less, exact > prefix, longest prefix, creationTimestamp tiebreak; methods filter rather than rank) and `RouteAdmitted=False/RouteConflict` conditions naming the winner — overlapping triggers are observable for the first time; the loser starts serving the moment the winner is deleted.
- **Review follow-ups** (two-agent review; each fix has a regression test):
  - conflict losers survive handler swaps and resync passes (the apply path consults the shadow set before marking `True`);
  - a failed materialize is sticky — conditions re-queued, dirty flag re-armed by the resync loop, `mux_materialize_failures_total` + `route_resync_failures_total` counters;
  - change detection keys on **Generation** (matching `GenerationChangedPredicate`), so status-only writes don't poison the drift counter or rebuild handlers;
  - unresolved-trigger index: a trigger applied before its function is re-admitted by the function's create event, not the next resync;
  - resync's stale-entry deletion re-reads with UID comparison (mid-pass-create race) without leaking delete+recreate entries.

## Bugs found along the way

- gorilla's `Route.Methods()` uppercases the slice it is handed **in place** — the legacy path has been silently mutating informer-owned trigger objects, and reusing the slice raced the still-serving mux's matcher (caught by the new no-flap test under `-race`). Both paths now clone per registration.
- An unresolvable trigger used to 404 while its condition claimed the route was admitted; it now reports `RouteAdmitted=False/FunctionNotFound`.

## Observability

`fission_router_route_table_applies_total{result}`, `fission_router_mux_rebuilds_total{listener,reason}` (must NOT move under canary/function churn in incremental mode), `fission_router_routes{listener}`, `fission_router_route_resync_drift_total` (CI bar: zero), plus the two failure counters above.

## Compatibility / rollout

- Behavior-identical for non-overlapping route sets; phase 2 deliberately replaces today's nondeterministic overlap winner with the specified precedence (release-note item), with `RouteConflict` conditions surfacing every affected trigger.
- The GHSA public/internal split, CORS deny-by-default, encoded-path handling, and rewrite semantics are pinned by the pre-existing suite, unchanged.
- The v1.34 CI leg pins `ROUTER_INCREMENTAL_ROUTES=false` (same step as the RFC-0002 legacy pin) so the escape hatch keeps end-to-end coverage; it is scheduled for removal one release after this ships.
- Also removes the stray `rfc/0014-router-hot-path-efficiency.md` that accidentally rode #3491 to the repo root (the canonical copy is `docs/rfc/0014`).

Verification: `go test -race ./pkg/router/... ./pkg/apis/core/v1/` green; e2e harness suites (`test/e2e/...`, real `Start` wiring with incremental ON) green; helm lint + template render checks for default/true/false; lint clean.


sanketsudake:
## Cluster-level acceptance evidence (kind, this branch)

Deployed via `SKAFFOLD_PROFILE=kind make skaffold-deploy`, then drove `test/benchmark/tests/route-churn/generate.sh` against the live router metrics.

**Startup** (zero objects):
```
router route-update mode configured  incrementalRoutes=true
fission_router_mux_rebuilds_total{listener="public",reason="shape_change"} 1
fission_router_route_resync_drift_total 0
```

**After `create 1000`** (1000 functions + 1000 canary triggers applied in ~3.5s):
```
fission_router_routes{listener="public"} 1000
fission_router_routes{listener="internal"} 1000
fission_router_mux_rebuilds_total{...,reason="shape_change"} 287   # 2000 shape changes coalesced by the 20ms debouncer
fission_router_route_table_applies_total{result="rejected"} 498    # triggers applied before their functions —
fission_router_route_table_applies_total{result="handler_swapped"} 1933  # — all re-admitted by the create-event cascade (unresolved index), not the resync
fission_router_route_resync_drift_total 0
```
The 498 `rejected` applies are the trigger-before-function apply ordering exercising the unresolved-trigger index in anger: every one re-admitted immediately when its function landed (1000/1000 routes in the table, zero drift).

**After `churn 200`** (200 random canary-weight rewrites — the steady-churn class):
```
fission_router_route_table_applies_total{result="handler_swapped"} 2132   # +199
fission_router_mux_rebuilds_total{...,reason="shape_change"} 287          # UNCHANGED — zero mux rebuilds
fission_router_route_resync_drift_total 0
```
This is the RFC's headline acceptance bar, metric-asserted on a live cluster: **canary weight churn causes zero mux rebuilds**. The `no_change` counter also grew by exactly 1000 between scrapes — one resync pass over all triggers, fully converged — confirming the Generation-keyed change detection (the RV-keyed variant would have logged ~1000 false drift corrections here, since the router had just written conditions on every trigger).

**Escape hatch, flipped live** (`kubectl set env deploy/router ROUTER_INCREMENTAL_ROUTES=false`):
```
router route-update mode configured  incrementalRoutes=false
fission_router_mux_rebuilds_total{listener="public",reason="legacy"} 4    # startup replay of 2000 objects, debounce-coalesced
```
The router restarts into the legacy full-rebuild loop and serves the same 1000 triggers.

PR #3494, opened 2026-06-12
feat: RFC-0012 — OCI-native package delivery as the default (producer, B-fetcher variant, pool reaper, Path B default)

sanketsudake:
Implements [RFC-0012](docs/rfc/0012-oci-default-package-delivery.md): OCI-native package delivery becomes the default cold-start path, in four independently shippable phases (per-phase commits).

## What changed

**Phase 1 — per-image idle pool reaper** (`pkg/executor/.../poolreaper.go`): an empty per-image (OCI) warm pool whose activity clock is older than `executor.ociPoolIdleReapTime` (default 5m, floored at the pool's pod-ready timeout + 1m) and which serves no specialized pods is destroyed; the next cold start recreates it against the kubelet's image cache. Generic env pools are NEVER reaped. Runs on the pool actor (single-writer discipline preserved); destroys are bounded (30s); pod-list uncertainty fails safe. `fission_executor_oci_pools_reaped_total` (+ `_reap_failures_total`) is the Gate C signal.

**Phase 2 — B-fetcher variant**: functions with Secrets/ConfigMaps — previously excluded from image volumes — ride Path B with the fetcher sidecar retained (newdeploy's shipped pattern): the image volume mounts at the fetcher's store path on BOTH containers, the exists-early-exit makes the fetch a no-op, and the fetcher still materializes secrets and drives the load. The variant bit joins `ociPoolHash` (no pool aliasing); KeepArchive envs are excluded consumer-side (they expect an archive FILE; OCI artifacts are directories).

**Phase 3 — consumer default flip**: `executor.enableOCIImageVolume: true`. The chart flips the default; the runtime gate degrades on k8s < 1.33 (the v1.32 CI leg proves the auto-degrade; the v1.34 legacy-pin leg strips the flag for explicit flag-off coverage).

**Phase 4 — the producer** (the real unlock): with `packageRegistry` configured, every successful build publishes its deployment directory as a digest-pinned single-layer OCI image (`pkg/oci.PushDirectory`: deterministic layer — sorted walk, zeroed timestamps — OCI media types, digest-tagged) and the Package is rewritten to `Archive{Type: oci}` with an `OCIPublished` condition. `fission fn create --code` users get image-volume delivery with zero per-package action; no registry = the tarball pipeline, byte-identical. Push failure: `fallbackToStorage` (default) rides the tarball with an `OCIPublishDegraded` condition + buildermgr log + `fission_buildermgr_oci_publish_total{result}`; strict mode fails the build with the REAL error (a pre-existing retry-after-artifact-delete interaction that masked upload errors as "no such file" is fixed). `publishedPrefix` decouples the push endpoint from the recorded consumption reference (the registry split-brain: kubelet resolves via the node, the builder via cluster DNS).

## Eligibility / delivery matrix

| Package | Cluster | Function | Delivery |
|---|---|---|---|
| tarball | any | any | fetcher download (unchanged) |
| oci | <1.33 or flag off | any | Path A: fetcher OCI pull |
| oci | ≥1.33 + flag | no secrets, v2+, non-Infinite, non-KeepArchive | **Path B-direct** (no fetcher) |
| oci | ≥1.33 + flag | secrets/configmaps | **Path B-fetcher** (sidecar for secrets only) |
| oci | ≥1.33 + flag | env v1 / Infinite / KeepArchive | Path A |

## Testing

- Unit (`-race`, all green): reap decision table + fail-safe + activity-clock contract; variant hash no-aliasing; B-fetcher/B-direct pod-spec invariants (incl. the GHSA-85g2 SA-token re-clamps); `PushDirectory` round-trip **through the actual Path A extractor** (consumption-compatibility by test, not by claim), determinism, symlink rejection; upload-handler mode table (success short-circuit / strict / fallback against a stub storagesvc); registry config policy; publish-condition transitions.
- Integration: `TestOCIProducerBuild` drives source build → push → digest-pinned OCI Package → condition → function serves via image volumes, on the v1.36 leg — which also enables the producer for the ENTIRE builder test corpus (every build test on that leg rides producer-published packages end-to-end).
- Full `./pkg/... ./test/e2e/...` sweep green locally; helm lint + render checks for all knob combinations.
- Two-agent review pre-push; all findings fixed with regression tests (reaper-vs-cold-start window, lying reap counter, strict-mode error masking, degraded-publish observability, Path A insecure-allowlist union, config-policy consistency).

## Compatibility

Additive throughout: no registry → no behavior change; flag off → old default; tarball packages untouched; existing OCI packages gain Path B after the flip. Migration of an existing built package = `fission pkg rebuild` after configuring the registry. Per-package escape hatch: `fission.io/package-delivery: tarball` annotation.


sanketsudake:
## Cold-start / warm-path comparison (kind v1.36.1, this branch)

Same `hello.py` function, 20 sequential cold starts per delivery variant (fresh function + route each iteration; first non-404 response = the specializing request), then 100 sequential warm requests against the last specialized pod. Package padded to ~1 MB so the tarball baseline really exercises the storagesvc download (small archives go literal). Apple M2 / single-node kind / local NodePort registry.

| Delivery | Cold p50 | Cold p95 | min | max | Warm p50 | Warm p95 |
|---|---|---|---|---|---|---|
| Tarball (storagesvc URL) — today's default | 122.7 ms | 197.2 ms | 98.4 | 219.1 | 4.7 ms | 7.6 ms |
| OCI **Path A** (fetcher registry pull) | 136.2 ms | 211.2 ms | 107.0 | 235.6 | 4.2 ms | 5.6 ms |
| OCI **Path B** (kubelet image volume) | **54.0 ms** | 308.5 ms | 36.2 | 1279.4 | 4.4 ms | 7.9 ms |

**Reading:**

- **Path B halves the median cold start: 122.7 → 54.0 ms (−56%).** The fetcher RPC, the per-pod archive download, and the unzip are gone from the critical path — the RFC's Gate A/B claim, measured.
- **Path B's tail (p95 309 ms, max 1.28 s) is the per-image poolsize=1 trade-off under this benchmark's worst case**: 20 back-to-back cold starts of the *same package* each consume the pool's single warm pod, so some iterations catch the deployment still replacing it (kubelet-cached pull, so refill is fast but not instant). Steady-state traffic doesn't hammer serial cold starts on one package; the generic-pool tarball baseline has 3 warm pods to draw from. This is exactly the warm-depth vs bounded-economics dial (`executor.ociPoolIdleReapTime` + poolsize lever) the RFC documents.
- **Path A ≈ tarball (+11% median)**: a registry pull costs about the same as a storagesvc download+unzip — OCI packages on pre-1.33 clusters (or with image volumes off) lose nothing meaningful, so the producer is safe to enable fleet-wide regardless of cluster version.
- **Warm path is delivery-agnostic: 4.2–4.7 ms p50 across all three** — once specialized, the serving pod is identical; delivery only moves bytes at specialization time. No warm regression from any of this PR's changes.

Raw CSVs: `rfc0012-results/` (local), reproducible via `test/benchmark/tests/cold-start/` pattern with `--deploy`/`--oci` package variants.

sanketsudake:
**Addendum: literal archives** (code embedded in the Package CR, the `--code`/<256 KB path — no storage, no tar at all):

| Delivery | Cold p50 | Cold p95 | max | Warm p50 |
|---|---|---|---|---|
| **Literal** (embedded in CR) | 111.6 ms | **130.1 ms** | 137.5 | 4.5 ms |
| Tarball (storagesvc, ~1 MB) | 122.7 ms | 197.2 ms | 219.1 | 4.7 ms |
| OCI Path A (fetcher pull) | 136.2 ms | 211.2 ms | 235.6 | 4.2 ms |
| OCI Path B (image volume) | **54.0 ms** | 308.5 ms | 1279.4 | 4.4 ms |

Two takeaways:

- **Literal saves bytes, not choreography.** It removes the download/unzip entirely, yet its median is only ~11 ms better than a 1 MB tarball — and still **2.1× slower than Path B**. The dominant cold-start cost for small code is not byte transfer but the per-pod RPC choreography (executor → fetcher `/specialize`, fetcher write, fetcher → env load); Path B is faster because it deletes the fetcher hop itself, not because it moves bytes better. Literal does have the *tightest tail* (p95 130 ms — zero network variance), the flip side of Path B's poolsize-1 refill tail under serial same-package cold starts.
- **RFC-0012 leaves literal packages completely untouched**: the producer only rewrites *built* archives (a `--code` function never enters buildermgr), and a literal package has no OCI field, so it rides the plain pool exactly as before — pinned by the unchanged non-OCI parity tests. The 256 KB CR-embedding cap (etcd cost) remains literal's structural boundary; code that outgrows it moves to the tarball or — with a registry — the producer path, where Path B then more than halves the median.

PR #3497, opened 2026-06-16
Multi-namespace tenancy: PRD + phased implementation (#3298)

sanketsudake:
## What this is

A security-first design + phased implementation for **multi-namespace tenancy** in Fission, addressing the cluster-wide-restart problem in #3298.

This PR opens with the full design under [`docs/multiple-namespace/`](https://github.com/fission/fission/tree/feat/multi-namespace-tenancy/docs/multiple-namespace) and will grow phase-by-phase. It is a **draft** while the phased implementation lands.

## The problem (#3298)

Adding a namespace via `additionalFissionNamespaces` + `helm upgrade` restarts the **entire control plane** and re-specializes every function. Root cause: `pkg/utils/namespace.go` reads the namespace set from env vars **once at `init()`** into a package global, so changing the set mutates every control-plane pod-template → rolling restart. It is structurally inevitable today, not a tuning problem.

There is also a latent isolation gap: the chart copies the **master** `fission-internal-auth` secret into every tenant namespace, and HMAC keys are derived per-*service* not per-*namespace* — so one compromised tenant namespace can forge requests as any service for any namespace.

## The approach — three pillars

1. **Zero-disruption dynamic onboarding** — a tenant-lifecycle controller + dynamic per-namespace watching (no env, no restart). Only Fission's own CRDs go cluster-wide; all core/workload resources stay per-namespace dynamic (least privilege preserved).
2. **Per-namespace derived HMAC keys** — the master never leaves the control plane; a compromised tenant can act only as itself. Version-aware signing keeps rolling upgrades 401-free.
3. **Declarative lifecycle** — a cluster-scoped `FissionTenant` CRD + `fission.io/enabled=true` label + `fission tenant` CLI, with finalizer-drained offboarding.

## Phased delivery

| Phase | Deliverable |
|---|---|
| 0 | Thread-safe `NamespaceResolver` (enabler, no behavior change) |
| 1 | `FissionTenant` CRD + `--tenantController` + CLI |
| 2 | Helm migration Job — **fixes #3298 (zero restart)** |
| 3 | Tier-A cluster-wide CRD cache + membership predicate |
| 4 | Tier-B dynamic per-namespace cache + RBAC provisioning |
| 5 | Per-namespace derived HMAC keys |
| 6 | Opt-in `tenancy.mode: cluster` (watch-all) |
| 7 | Archive ns-prefix (storagesvc content isolation) — follow-up |

Phases 0–2 close #3298 with zero security regression; 3–5 deliver true isolation. Each phase is independently reviewable.

## Relation to #3476

#3476 identifies the same operational pain and is appreciated. This design takes a different posture on the security-sensitive parts (keeps least-privilege per-namespace RBAC and internal auth on by default, with watch-all as an explicit opt-in; uses cryptographic per-namespace keys instead of a podIP heuristic). The docs' `prd.md` §9 has a concrete, side-by-side comparison.

## Docs in this PR

- `prd.md` — the PRD (design, phases, critical files, #3476 comparison)
- `testing-plan.md` — per-phase unit + integration test matrix
- `performance-benchmarking.md` — metrics, scaling sweeps, regression gates
- `backward-compatibility.md` — upgrade/rolling-restart review + safe-upgrade runbook

Closes #3298 (on completion of phases 0–5).


sanketsudake:
## Performance: benchmarks & observations — multi-namespace tenancy

This adds reproducible Go benchmarks for the hot paths the tenancy work touches, plus a regression read from the CI performance artifacts.
Net: the dynamic model's per-request cost is negligible, and there is no control-plane regression.

### Hot-path benchmarks

Committed as regression guards (`pkg/auth/hmac/bench_test.go`, `pkg/utils/namespace_bench_test.go`).
Reproduce with `go test -run '^$' -bench=. -benchmem ./pkg/auth/hmac/ ./pkg/utils/`.
Numbers below are Apple M2 (arm64); the x86 CI runners are the same order of magnitude.

| Path | When it runs | Cost | allocs/op | Note |
|---|---|---|---|---|
| `NamespaceResolver.IsTenant` | every CRD watch event + every cold-start specialization | **13 ns/op** | **0** | constant regardless of tenant count |
| `DeriveServiceKey` (master-scoped HKDF) | cold-start specialization / build | 705 ns/op | 18 | pre-existing |
| `DeriveServiceKeyNS` (namespace-scoped HKDF) | cold-start specialization in a tenant ns | 714 ns/op | 19 | **+1.3% vs master-scoped** |
| storagesvc dual-accept verify (full per-request) | each `/v1/archive` request | ~2.6 µs/op | 64 | 2 real HKDFs (steady state) |
| `FissionResourceNamespaces` (copy getter, 50 tenants) | startup / periodic reaper only | 1251 ns/op | 4 | **never per-request** |

Takeaways:

- The per-event / per-request gate the dynamic model adds — `IsTenant` — is **~13 ns and zero-allocation**, and does not grow with tenant count.
Dynamic tenancy stays off the request-latency budget.
- Namespace-scoping the HMAC derivation is **within ~1%** of the existing master-scoped derivation (it only appends `:<ns>` to the HKDF info string), and it runs only on cold-start specialization / builds, never per request.
- The full storagesvc per-request verify is **single-digit microseconds**, trivial next to the archive-body I/O it gates.
(Benchmarked with `masterOld=nil`, the steady state; during a key-rotation overlap it derives two extra candidate keys — still single-digit µs.)
- The copy-returning resolver getter is the one path that scales with tenant count, and it is deliberately confined to startup/periodic loops.
The hot paths use `IsTenant` instead — the side-by-side benchmark documents exactly that contrast.

### Control-plane footprint (CI artifact analysis)

Read from the integration-leg pprof + Prometheus TSDB artifacts, comparing the two dynamic-tenancy legs (k8s 1.32.11 / 1.36.1) against the all-legacy leg (1.34.8).

- **No goroutine leak.** Every executor/router goroutine series is flat across the run (last-third ≈ first-third), and the end-of-run baselines match the pprof snapshots.
- Dynamic mode adds a **constant** goroutine offset (~+17–20 executor, steady state): the cluster-wide Tier-A CRD cache + the FissionTenant watch + tenant re-enqueue + resolver-sync.
It does **not** grow per onboarded namespace — the CRD cache is one cluster-wide informer, not one per tenant — so it is a fixed cost, not a leak.
(The router offset is larger but partly reflects RFC-0002/0013 being *off* on the legacy-pin leg, not pure tenancy.)
- **Heap is flat** (~15–18 MB executor/router) across all three legs; dynamic mode shows no heap regression.
- Control-plane health: **0 reconcile panics, 0 mux-materialize failures, 0 resync failures**, and the dynamic legs had *fewer* executor reconcile errors than the legacy leg.

### Cleanup (`/simplify`) pass

A four-angle review (reuse / simplification / efficiency / altitude) found the code already clean on the hot paths.
The de-duplications applied (commit `0d47c620`, behavior-preserving — build + HMAC byte-compat tests + `golangci-lint` all green):

- one `deriveKey(master, info)` behind `DeriveServiceKey` / `DeriveServiceKeyNS` (guarantees the two stay byte-compatible, the property the "they never collide" guarantee relies on);
- `VerifierFromKeyOrMaster` centralizing the fetcher/builder "prefer the mounted derived key, else derive from master" decision (one home for the security choice both sidecars make);
- `fv1.MANAGED_BY_LABEL` instead of the re-spelled `fission.io/managed-by` key (one site was a bare string literal);
- `waitForCRD(label, listFn)` behind the two near-identical CRD-wait poll loops;
- reciprocal `LOCKSTEP` comments pinning the Go vs Helm copies of the fetcher/builder Role rules.

Deliberately **skipped** (noted, not changed on a green security-sensitive PR):

- removing the always-True `ServiceAccountsReady` status condition (it is observable CRD API surface);
- making the storagesvc dual-accept verifier lazy (single-digit-µs on an I/O-bound path, and cross-request key caching is *forbidden* for tenant isolation);
- the structural fix for the Go-vs-Helm Role duplication (bind shared-partial ClusterRoles by name, as the workload roles already do) — a tracked follow-up, too invasive for a cleanup pass.

PR #3508, opened 2026-06-18
refactor(router): confine gorilla/mux to pkg/router (executor + storagesvc + metrics → stdlib ServeMux)

sanketsudake:
## Context

Part of the dependency cleanup. `gorilla/mux` was archived in Dec 2022 (re-maintained since 2023). The **router** depends on it for user-facing path-template routing (`/{id}`, `/{sort:(asc|desc)}` → `X-Fission-Params-*`) and drives its own route-precedence engine, so it stays there. Every **other** consumer can use the Go 1.22 `net/http.ServeMux` method+path patterns.

After this PR, `grep -rl gorilla/mux pkg` returns **only `pkg/router/`**. (gorilla/mux remains a direct `go.mod` dep because the router still uses it — the goal is source confinement, since `fission-bundle` is one binary.)

## What changed

**`pkg/utils/metrics` — mux-agnostic middleware**
The HTTP metrics middleware no longer imports gorilla. The shared core takes a path-label function:
- `InstrumentHandler(path, next)` — static label, for stdlib `ServeMux` consumers.
- `InstrumentHandlerFunc(pathFn, next)` — per-request label, for the router.
The gorilla `CurrentRoute().GetPathTemplate()` lookup moves into `pkg/router/metrics_middleware.go` (the only package that still uses gorilla), so the router's metric labels are unchanged.

**`pkg/executor/api.go`, `pkg/storagesvc/storagesvc.go` → stdlib `ServeMux`**
- Routes become `"POST /v2/getServiceForFunction"` etc.; per-route metrics via `InstrumentHandler`.
- The HMAC verifier wraps the **whole mux** as the outermost layer, preserving the documented *verifier-outer / metrics-inner* ordering (401s handled at the auth layer; metrics see only verified requests).
- storagesvc's two `GET /v1/archive` routes (gorilla `.Queries("id","{id}")` split) merge into **`getOrListHandler`**, which dispatches on the `id` query param — stdlib `ServeMux` can't match on query. `HEAD /v1/archive` stays a distinct, more-specific pattern so `infoHandler` still wins HEAD.

**Tests** drop gorilla in executor/storagesvc/metrics/httpserver. The httpserver test uses `"/{$}"` to keep exact-`/` matching (stdlib bare `/` is a catch-all subtree).

## Behavioral parity verified

- 405-on-method-mismatch and 404-on-unknown-path: stdlib `ServeMux` matches gorilla here.
- `GET /v1/archive` + `HEAD /v1/archive`: stdlib serves the explicit HEAD pattern for HEAD (more specific than GET's implicit HEAD) — no panic, `infoHandler` wins.
- The metrics core is byte-for-byte equivalent (websocket bypass, in-flight gauge inc/dec on the same series, duration, status-code counter).

## Security observations

- HMAC verifier wiring is unchanged in effect: it remains the outermost layer with the same per-service key derivation and `/healthz`(+`/readyz`) bypass; unsigned requests to non-bypass paths still 401. Covered by the executor/storagesvc auth-wiring tests (now stdlib).
- No new external surface; `InsecureSkipVerify`/CORS posture untouched.

## Performance observations

- No per-request data-plane change. The router's hot path (gorilla) is untouched; executor/storagesvc are control-plane RPC endpoints.
- Per-route `InstrumentHandler` resolves the path label from a constant (cheaper than the old `CurrentRoute`+`GetPathTemplate` reflection for these static routes).

## Verification

- `make code-checks` → 0 issues; `make license-check` → clean
- `go test -race ./pkg/executor/... ./pkg/storagesvc/... ./pkg/utils/metrics/... ./pkg/utils/httpserver/... ./pkg/router/...` → pass
- `grep -rl gorilla/mux pkg` → only `pkg/router/*`

> Note: this branch and #3505 both touch `pkg/utils/httpserver/server_test.go` (adjacent import lines). If #3505 merges first, this branch will be rebased — trivial resolution.


sanketsudake:
Superseded by #3510, which does the executor/storagesvc migration properly via the new pkg/utils/httpmux package (and fixes the metrics-package leakage this PR introduced). Closing in favor of the httpmux phased approach.

PR #3510, opened 2026-06-19
feat(httpmux): internal HTTP router (static + {var}/{var:regexp}) + pure-recorder metrics; migrate executor/storagesvc

sanketsudake:
## Context

Brainstormed redesign (supersedes #3508). The dependency cleanup confined `gorilla/mux` to `pkg/router` by moving executor/storagesvc onto the stdlib `ServeMux`, but it left two warts: routing/registration concerns leaked into `pkg/utils/metrics` (`RegisterHandler`/`InstrumentHandler*`), and gorilla still backs the router. The agreed end-state is **one internal routing package with full gorilla removal**, done in phases. This is **Phase 1**.

## What changed

**New `pkg/utils/httpmux`** — a small SOLID layer over `net/http`:
- Fluent, method-separate API: `m.Handle("/x", h).Methods("POST").Host(...)`, `HandlePrefix`, `New(WithMiddleware(...), WithMetrics(...), WithEncodedPath())`, `Handler()`.
- Matcher: first-match in **registration order** (caller controls precedence), `405` vs `404` parity, host/method/exact/prefix. `Handle("/")` is **exact** (not a stdlib catch-all).
- `Vars(r)` / `Pattern(r)` context accessors (replace `mux.Vars` / `CurrentRoute().GetPathTemplate()`).
- `Instrument(rec, patternOf, next)` — per-route metrics with a websocket bypass; exported so the still-gorilla router can reuse it during migration.

**`pkg/utils/metrics` reverted to a pure recorder:**
- `HTTPMetricMiddleware` + `ResponseWriterWrapper` deleted; the package now exposes only **`HTTPRecorder`** (over the existing metric vecs), which `httpmux` drives. `metrics` no longer imports `gorilla/mux` or `pkg/router/util`. (It also fixes a latent bug: the old websocket-bypass missed `Connection: keep-alive, Upgrade`; httpmux's check is RFC-correct.)

**Consumers:**
- `pkg/executor`, `pkg/storagesvc` build via `httpmux.New(WithMiddleware(verifier), WithMetrics(...))`; verifier stays outermost (401 before metrics). storagesvc's two `GET /v1/archive` routes merge into `getOrListHandler` (can't match on `?id=`).
- `pkg/router/metrics_middleware.go`: interim bridge recording via `httpmux.Instrument` + a gorilla `CurrentRoute` label fn — **deleted in the router-migration phase**.

**Tests** migrated to httpmux; `server_test` switches the hardcoded `:8999` to `freePort` (kills the flake); `HTTPRecorder` gets a focused metrics test; new `httpmux` package has thorough unit tests (matching, 405/404, host, prefix, first-match, encoded path, metrics, websocket bypass, flush passthrough).

## Outcome
`grep -rl gorilla/mux pkg` → only `pkg/router`. Phases 2 (`{var}`/`{var:regex}` template matcher + parity tests) and 3 (router cutover, drop `gorilla/mux` from `go.mod`) follow.

## Verification
- `go test -race ./pkg/utils/httpmux/... ./pkg/utils/metrics/... ./pkg/utils/httpserver/... ./pkg/executor/... ./pkg/storagesvc/... ./pkg/router/...` → pass
- `make code-checks` → 0 issues; `make license-check` → clean; integration build compiles.

> Supersedes #3508 — that PR can be closed.

### DIFF
--- a/pkg/executor/api.go
+++ b/pkg/executor/api.go
@@ -15,7 +15,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/gorilla/mux"
 	"golang.org/x/sync/errgroup"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -25,6 +24,7 @@ import (
 	"github.com/fission/fission/pkg/executor/client"
 	"github.com/fission/fission/pkg/executor/executortype"
 	"github.com/fission/fission/pkg/executor/fscache"
+	"github.com/fission/fission/pkg/utils/httpmux"
 	"github.com/fission/fission/pkg/utils/httpsecurity"
 	"github.com/fission/fission/pkg/utils/httpserver"
 	"github.com/fission/fission/pkg/utils/metrics"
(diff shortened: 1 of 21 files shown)


sanketsudake:
**Updated: this PR now combines Phase 1 + Phase 2** (stacked PRs against a non-main base skip the integration/lint/CodeQL suite, so #3511 is folded in here).

Added since the initial review:
- **`{var}`/`{var:regexp}` path-template matching + `CompilePattern` validation** (`pkg/utils/httpmux/template.go`) — handles the real Fission patterns incl. the slash-spanning `/bank/{html:[a-zA-Z0-9./]+}`, returns errors instead of panicking on malformed templates, and accepts the `(asc|desc)` capturing-group pattern gorilla rejects. Not yet wired to the router (that's Phase 3).
- **404/405 now recorded under fixed labels** (`<not found>`/`<method not allowed>`) — restores the unmatched-request signal the old gorilla global metrics middleware had, without the raw-path label cardinality.

Two intentional behavior notes for reviewers: (1) the websocket-bypass check is now RFC-correct (the old `util.IsWebsocketRequest` missed `Connection: keep-alive, Upgrade` — a latent router-metrics bug); (2) 404/405 on executor/storagesvc are now counted under the fixed labels above rather than per-raw-path.

sanketsudake:
## Performance & security observations

Per the review focus on edge cases / security / performance, here are concrete measurements and the security posture of the matcher.

### Performance

Matcher microbenchmarks (`go test -bench=. -benchmem ./pkg/utils/httpmux/`, Apple M-series, Go 1.26):

| Benchmark | shape | ns/op | allocs/op |
|---|---|---:|---:|
| `BenchmarkExactMatch` | executor/storagesvc (4 exact, method-gated; hits last) | 103 | 3 |
| `BenchmarkTemplateMatch` | router `/accounts/{id}` (bare `{var}`) | 320 | 7 |
| `BenchmarkTemplateMatchRegex` | `/bank/{html:[a-zA-Z0-9\./]+}` (spans `/`) | 434 | 7 |
| `BenchmarkNoMatch404` | scan-everything-then-404 (probes/scanners) | 189 | 3 |

The executor/storagesvc paths this PR migrates are **exact** matches (~103 ns, 3 allocs) — 3-4 orders of magnitude below the request's network/proxy budget, so matcher cost is negligible at this scope. The template numbers are the baseline for the **Phase 3** router cutover, which will benchmark the hot path directly against gorilla/mux via `httpTriggers_bench_test.go`. A `bench_test.go` is included as a regression guard.

### Security

- **ReDoS-safe by construction.** Templates compile to Go `regexp` (RE2) — linear-time, no catastrophic backtracking on attacker-supplied paths.
- **Fail-loud on malformed templates.** `Handler()` now panics on a compile error instead of silently leaving a dead, never-matching route (review fix); callers handling user-supplied patterns (the router, Phase 3) validate up front via `CompilePattern`, which returns an error rather than gorilla's panic.
- **Bounded metric cardinality.** Unmatched requests are recorded under fixed `<not found>` / `<method not allowed>` labels, not the raw path — a path-scanning probe can't blow up Prometheus series.
- **Auth ordering preserved.** The HMAC `ServiceVerifier` is wired outermost (`WithMiddleware`), so it runs before dispatch — unauthenticated requests never reach a handler. Bypass lists unchanged (executor `/healthz` + `/readyz`; storagesvc `/healthz`, and only when `authSecret` is set).
- **Literal path matching is explicit.** httpmux does **not** clean/redirect `..` / `//` the way gorilla does by default. Safe for the exact, method-gated, verifier-fronted routes here (a non-canonical path simply 404s) — documented in the package doc, and flagged as a Phase 3 decision for the router's raw-request-URI-signed internal listener (with `..`/`//`/encoded-slash tests to add at that boundary).
- **No new dependencies.** Pure stdlib + the already-present `x/net/http/httpguts`; the net effect once Phase 3 lands is **one fewer** third-party dep (gorilla/mux dropped from `go.mod`).

### Edge cases covered by tests

405→200 fall-through to a later route, 405-across-routes vs host-mismatch 404, default-200 status capture, multi-var extraction, `{var:regex}` spanning `/`, capturing-group templates gorilla rejects, encoded-path matching, first-match registration order, and panic-on-invalid-template.

PR #3512, opened 2026-06-19
feat(router): migrate router onto httpmux; drop gorilla/mux (Phase 3)

sanketsudake:
## Summary

Final phase of the gorilla/mux removal.
`pkg/router` now routes through the internal `pkg/utils/httpmux` package (shipped in #3510) instead of gorilla/mux, and `go mod tidy` drops `gorilla/mux` from `go.mod` entirely.

```
$ grep -rl '"github.com/gorilla/mux"' pkg cmd
(empty)
```

This is the licensing/deprecation goal that motivated the whole effort: the archived-then-revived `gorilla/mux` is gone (the remaining `gorilla/websocket` is an unrelated indirect transitive dep).

## Router cutover

- `mutablemux.go` stores the built `http.Handler` (`atomic.Pointer[http.Handler]`) instead of `*mux.Router`; an atomic swap is one pointer write.
- `routeshape.go` / `httpTriggers.go` / `incremental.go` build `*httpmux.Mux` through the shared registration helpers and call `.Handler()` at the swap point.
- `buildMuxes` and the extracted `buildIncrementalMuxes` return `*httpmux.Mux`, so the golden shape + legacy/incremental parity tests introspect via `httpmux.Match` without driving the proxy handlers.
- `validateRouteTemplate` uses `httpmux.CompilePattern` (returns an error, no `recover`) instead of gorilla's panic-on-bad-template.
- `requesthHeader.go` uses `httpmux.Vars`; `auth.go`'s middleware is a plain `func(http.Handler) http.Handler`; `metrics_middleware.go` is deleted (its job is now httpmux's `WithMetrics` + per-route pattern label).

## httpmux additions (additive; executor/storagesvc behaviour unchanged)

- `Mux.Match(req)` reports which route would handle a request — for tests/diagnostics; the dispatcher's match loop is extracted and shared with `ServeHTTP`.
- `Route.Methods()` with an **empty** set now matches no method (a dead route), distinct from never calling `Methods()` (any method).
This aligns httpmux with gorilla and fixes the empty-method-shape contract the router relies on — caught by the legacy-vs-incremental parity test.

## Behavioral changes from gorilla (all documented + tested)

- **Auth runs pre-match**, so an unauthenticated request to an unknown path returns 401 (not 404) — it no longer reveals which paths exist.
- **404/405 recorded under constant metric labels** instead of the raw path, bounding Prometheus cardinality against path-scanning probes.
- **Capturing-group templates** (`{sort:(asc|desc)}`) are now admitted (httpmux wraps each var in a named group) rather than rejected; only genuinely malformed templates (unbalanced braces, uncompilable class) are rejected.
- **The HMAC-signed internal listener matches paths literally** (no `.`/`..`/`//` cleaning): correct here because the signature covers the raw request-URI and publishers always build clean paths via `utils.UrlForFunction`.
Locked in by `TestInternalListenerLiteralPathMatching`.

## Verification

- `go test -race ./pkg/router/... ./pkg/utils/httpmux/...` — all pass.
- `golangci-lint run` — 0 issues; `make license-check` clean.
- GHSA-3g33-6vg6-27m8 split preserved (`TestPublicMuxDoesNotRegisterInternalFunctionRoute`), legacy↔incremental dispatch parity preserved (`TestIncrementalLegacyParity`).


sanketsudake:
## Review, performance & security observations

Ran a 3-agent plugin review (general code review, silent-failure hunter, type-design) plus benchmarks and a security pass.

### Review outcome

**No Critical issues.** One **Important** finding fixed, plus doc/test hardening (commit `5609d35d`):

- **Operator `AuthUriPath` could crash-loop the router.** It was registered without validation, so a malformed value (e.g. `/auth/{`) would make `httpmux.Handler()` panic when the mux is (re)built — in the rebuild goroutine, *outside* `panicRecoveryMiddleware`. gorilla merely left a silently-dead route. Now validated via `httpmux.CompilePattern`; a bad value skips the login route (logged) instead of downing the router. Locked by `TestRouterOwnedRoutesSkipInvalidAuthPath`.
- Documented the cross-package coupling the `methodsSet` bool relies on (`registerRouteShape` calls `.Methods()` unconditionally so an empty set stays a dead route), the `updateRouter(nil)` guard, and `Mux.Match`'s test-only/recompiles-per-call cost.

The type-design pass rated the new surface highly: `Mux` deliberately has no `ServeHTTP`, so "forgot `.Handler()`" is a **compile error**, not a runtime bug; `atomic.Pointer[http.Handler]` is the correct choice over `atomic.Value` (avoids its panic-on-differing-concrete-types hazard, which is live here since the stored handler is sometimes a bare dispatcher and sometimes middleware-wrapped).

### Performance

Router-scale matcher benchmarks (`go test -bench` on `pkg/router`, Apple M-series, Go 1.26):

| Benchmark | scenario | ns/op | allocs/op |
|---|---|---:|---:|
| `BenchmarkMuxMatch/first-route` | first of 10k routes | 115 | 3 |
| `BenchmarkMuxMatch/last-route` | scan all 10k (worst case) | 38,715 | 3 |
| `BenchmarkMuxMatch/miss` | scan all 10k → 404 | 27,650 | 3 |
| `BenchmarkBuildMuxes/triggers=10000` | full rebuild + compile | 85.9 ms | — |

Matching is **O(routes) linear scan — exact parity with gorilla** (gorilla also scanned registered routes in order). At realistic route counts (tens–hundreds) match cost is sub-µs to low-µs; even the 10k-route worst case (≈39 µs) is far under the ~1 ms p99 bar and negligible against the proxy round-trip. The 86 ms full rebuild at 10k triggers is only paid on **shape** changes — RFC-0013 incremental routing keeps canary weight ticks / function updates O(1) handler swaps. A future exact-route fast-path map could cut the worst-case scan, but it was deliberately **deferred**: an earlier-registered template can shadow a literal exact path, so a naive map would break first-match precedence — correctness over a micro-opt in the hot path.

### Security

- **HMAC-signed internal listener matches paths literally** — no `.`/`..`/`//` cleaning. This is *more* correct than gorilla here: gorilla's clean-and-301-redirect would produce a path the raw-request-URI signature no longer covers. Publishers always build clean paths via `utils.UrlForFunction`. Locked by `TestInternalListenerLiteralPathMatching` (+ a public-listener mirror).
- **Auth runs pre-match**, so an unauthenticated request to an unknown path returns 401, not 404 — no path-existence disclosure.
- **404/405 recorded under constant labels**, bounding Prometheus cardinality against path-scanning probes (gorilla's fallback recorded the raw path).
- **ReDoS-safe** templates (RE2, no backtracking), carried from httpmux.
- GHSA-3g33-6vg6-27m8 listener split preserved (`TestPublicMuxDoesNotRegisterInternalFunctionRoute`).

### Documented behavioral deltas (acceptable, called out for reviewers)

- **401 responses are no longer counted** in `fission_router` HTTP metrics (auth now rejects before the per-route metric instrumentation; OTEL spans still capture them). Net: bounded cardinality, minus 401 visibility.
- **Public-listener non-canonical paths now 404** instead of gorilla's 301-redirect-to-clean. Real clients send canonical paths; this removes a surprise-redirect surface.

sanketsudake:
## Second review + simplify pass

Ran `/pr-review-toolkit:review-pr` (5 agents: code, tests, comments, silent-failure, type-design) and `/simplify` (4 agents: reuse, simplification, efficiency, altitude) again. **No Critical code issues**; the migration was rated correct and well-designed. Applied the worthwhile findings (commit `8a80b7fc`):

### Performance win (efficiency agent, confirmed)
`httpmux.Pattern` + its `patternKey` context value had **zero production callers** (metrics label by the route's closure-captured constant, not a context read), yet `withMatch` set it via `context.WithValue` on **every matched request**. Removed it; the dispatcher now copies the request context only when a templated route actually captured vars.

| Benchmark | before | after |
|---|---|---|
| `BenchmarkExactMatch` (internal `/fission-function/...` hot path) | 103 ns/op, 3 allocs | **29 ns/op, 0 allocs** |

Every function invocation through the internal listener (all static, no path vars) now matches with **zero allocations**.

### Simplification
- Extracted `registerInternalRoute` (mirrors `registerRouteShape`) so `buildMuxes` and `buildIncrementalMuxes` register the internal route pair through one shared helper — completing the "both paths go through shared helpers" symmetry.
- Hoisted a doubled `panicRecoveryMiddleware` construction; fixed a stale `internalRouteShapes` doc-comment name.

### Review fixes
- **Comment rot:** corrected a stale gorilla comment on `triggerConfigError` that contradicted the code (capturing groups now admit, no panic), and two `buildMuxes` references that are really `newListenerMuxes` on the default incremental path.
- The `AuthUriPath` skip log now names the operational consequence.
- Added `TestAuthPreMatch401OnUnknownPath` (pre-match auth → 401 not 404) and `TestRouteShapeMethodNotAllowedVsNotFound` (405 vs 404 through the served handler).

### Skipped (noted as follow-ups, out of this PR's scope)
- Deduping `router/util.IsWebsocketRequest` against httpmux's `isWebSocketUpgrade` — would change data-plane websocket detection (originated in #3510).
- A shared `atomicHandler` primitive for `mutableRouter` + `routetable.HandlerRef` — defensible duplication; a judgment call touching RFC-0013 code.
- Moving `AuthUriPath` validation to config-load time — cleaner altitude but touches `featureconfig`; current guard is acceptable.

All `pkg/router` + `pkg/utils/httpmux` tests pass with `-race`; golangci-lint 0 issues; license clean. (The earlier `lint` job red was a `TestPoolPodControllerPodCleanup` flake in a package this PR doesn't touch — passes 3/3 locally with `-race`; cleared on re-run.)

PR #3513, opened 2026-06-19
perf(httpmux/router): exact-route fast-path index + metrics alloc + websocket-detection fix

sanketsudake:
## Summary

Follow-up **B** from the gorilla→httpmux migration's performance analysis (#3512). The httpmux dispatcher linear-scans routes in registration order; for the internal listener — every `/fission-function/<ns>/<name>` invocation — that is O(routes), which at thousands of functions is tens of microseconds per request.

This adds a precedence-safe exact-route fast-path index: a `path → []route` map for literal (non-template) exact routes, consulted before the scan.

## Correctness — why it's precedence-safe

A literal exact route is indexed **only when no template or prefix route can match its path**. Shadowing depends solely on the path, so:
- when a request's path is a map key, that bucket holds **every** route that can match it → host / method / 405 resolution within the bucket is byte-identical to the scan;
- every other request (templates, prefixes, shadowed exacts, and misses) falls through to the **unchanged** linear scan.

Excluding a route from the index can never change behaviour, only speed. `fastpath_test.go` pins the tricky cases: a template **or** prefix registered *before* a literal exact still wins (the literal is shadowed out of the index, so the scan decides), plus host/method/405-vs-404, encoded paths, and white-box eligibility.

## Build cost kept near-linear

Naively, "does any matcher match path P" is O(routes) per path → O(routes²) to build (248 ms at 10k). Instead: a prefix `Q` shadows `P` iff `P[:len(Q)] == Q`, so prefixes go in a set probed by length (O(distinct prefix lengths) per path); only templates stay pairwise (typically few).

## Measured (Go 1.26)

| | before | after |
|---|---|---:|
| match — last of 10k exact routes (worst-case scan) | 38,715 ns | **~15 ns, 0 allocs** |
| match — any position, 100 / 1k / 10k routes | O(routes) | **~25 ns (flat)** |
| index build at 10k routes (shape-change path only) | — | 2.1 ms (naive pairwise was 248 ms) |

Misses still scan (a route may match via a prefix/template) — that's the rare path, not function invocation.

## Verification
- All existing `pkg/utils/httpmux` + `pkg/router` tests pass unchanged with `-race` — the fast path is transparent.
- `golangci-lint` 0 issues; license clean.


sanketsudake:
## Added two more hot-path items (per discussion)

Folded in the two genuinely-outstanding items from the router perf review (the rest were diminishing returns or deferred-with-data):

### Metrics: per-request map allocations removed
`HTTPRecorder` built a `prometheus.Labels` **map** per call (`InFlightInc` + `InFlightDec` + `Observe` ≈ 3 maps per instrumented public request). Switched to `WithLabelValues` (positional, hits the vec's child cache, no caller-side map).

`BenchmarkHTTPRecorder` (inc + observe + dec per request):

| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| before — `With(Labels{})` | 848.9 | 1011 | 7 |
| after — `WithLabelValues` | 207.9 | 3 | 1 |

The lone remaining alloc is the unavoidable `strconv.Itoa` of the status code. Behaviour identical (same vecs/labels). Internal invocations are unaffected (the internal listener has no metrics).

### Websocket detection corrected (latent bug)
`router/util.IsWebsocketRequest` used naive string equality — `Connection == "Upgrade"` — which **misses `Connection: keep-alive, Upgrade`** (the comma-token form browsers send) and case variants, so a real websocket upgrade on the proxy path could be mis-detected and buffered/instrumented instead of streamed. httpmux already had the RFC-6455-correct detector (httpguts token-list parse) for its own bypass; exported it as `httpmux.IsWebSocketUpgrade` and had `router/util` delegate, so the data plane and the mux share one correct implementation. The previously-untested detection now has coverage (`TestIsWebSocketUpgrade`).

All affected suites (`metrics`, `httpmux`, `router/util`, `router`, `executor`, `storagesvc`) pass with `-race`; lint 0 issues; license clean.

sanketsudake:
## /simplify + /pr-review-toolkit:review-pr passes

Ran both on the full PR (now also includes the `IsWebsocketRequest` wrapper removal). **No Critical or Important code issues** — the review proved the precedence-safe shadowing correct by adversarial inspection. Applied the worthwhile findings:

### /simplify (4 agents)
- Moved `IsWebSocketUpgrade` out of `recorder.go` (a metrics-Recorder file) into its own `websocket.go` + test — it's now an exported, data-plane-consumed predicate.
- Dropped the `buildExactIndex` eligibility memo — it only helped a degenerate repeat-path case while adding a map+writes on the common unique-path path. Net simpler **and** a faster build (10k-route index build **2.1 ms → 1.54 ms**).
- Collapsed three precedence tests into one table-driven test; trimmed a thrice-stated invariant doc to one source of truth.

### /pr-review-toolkit:review-pr (4 agents)
- **code-reviewer:** adversarially probed the shadowing along every axis (prefix-not-ending-`/`, prefix == path, prefix longer than path, empty prefix, `{var:regexp}` templates, host conservatism, encoded path, bucket-miss early-return, 405-vs-404) — **could not construct any divergence** between fast path and scan. Metrics label order + websocket fix confirmed correct.
- **comment-analyzer:** docs accurate, not over-trimmed, single-source-of-truth structure praised.
- **type-design-analyzer:** the derived index can't drift today (unexported, rebuilt from `routes`, two co-located construction sites); flagged that the derive-from-same-slice invariant is a convention.
- **pr-test-analyzer / type-design convergence:** the one durable gap was a **differential test**. Added it:
  - `TestExactFastPathMatchesScanDifferential` — a corpus of edge-case route sets × methods × paths × hosts (~1,200 comparisons) asserting the indexed dispatcher and a forced linear scan return identical status+body. Catches any future shadowing regression mechanically.
  - `TestExactIndexWellFormed` (buckets are literal exacts from `routes`, in order) + `TestExactIndexEngagedAtScale` (index actually populates, so a silently-disabled optimization fails a test, not just a benchmark).
- Plus the one optional micro-opt: `slices.Sort(prefixLens)` for deterministic shadow-check order.

All `pkg/utils/httpmux` + `pkg/router` + `pkg/utils/metrics` tests pass with `-race`; lint 0 issues; license clean.

PR #3527, opened 2026-06-22
refactor(cache): replace single-goroutine channel-actor caches with locks

sanketsudake:
## What

Replaces the three stacked single-goroutine "channel-actor" caches in the control plane with mutex-guarded direct access.

`pkg/cache.Cache[K,V]`, `fscache.PoolCache`, and `fscache.FunctionServiceCache` each ran an unbuffered `requestChannel` drained by one `service()` goroutine, so every operation — even pure reads — serialized behind that goroutine, and the layers stacked (one executor tap = 3 serialized channel hand-offs: `fsc` → `byAddress.Get` → `byFunction.Get`).

## Why

The CI executor goroutine profile (run 27950964210) showed 22 concurrent `/tapService` requests parked sending on `FunctionServiceCache`'s channel — head-of-line blocking on the RFC-0002 tap/accounting hot path.

Removing the actors eliminates that serialization point, deletes a dedicated goroutine per cache instance, and removes ~660 lines of channel/request/response plumbing (net −157 in production code; all three files shrink).

Public APIs are byte-identical, so there is no caller churn.

## Changes

- CP1 — `pkg/cache.Cache` → `sync.RWMutex` (Copy uses RLock; Get takes the write lock since it updates atime / evicts).
- CP2 — `fscache.PoolCache` → `sync.Mutex`; the `svcChannel` specialization hand-off is preserved exactly (the waiter blocks after the lock is released, so SetSvcValue's wake-up send can acquire the lock).
- CP3 — `fscache.FunctionServiceCache` → an fsc-level `sync.Mutex` over the four operations the actor mutually excluded (TouchByAddress / ListOld / ListOldForPool / Log).
- Cleanups: result-container capacity hints; boundary docs for the deliberate `fsc.lock` scope and the under-lock channel send.

## Behavior note (one deliberate deviation)

This is a behavior-preserving refactor with **one** intentional exception: `ListOld`'s error path. On `main`, when `byFunction.Get` missed during the scan (a TOCTOU gap under concurrent delete) the actor did a bare `return` out of its `service()` loop — killing the goroutine, never sending a response, and permanently wedging every later cache call. It now logs and returns the partial list. Covered by `TestListOldPartialReturnOnDanglingIndex`.

A pre-existing lock-free `fsvc.Atime` race in `GetByFunction`/`GetFuncSvc`/`GetByFunctionUID`/`AddFunc` (which the actor never covered either) is intentionally left out of scope and documented with a boundary comment; closing it is a follow-up.

## Test plan

- [x] `go build ./...`, `go vet`, `make license-check`, `golangci-lint` — all clean.
- [x] `-race` green across `pkg/cache`, `pkg/executor/fscache`, `poolmgr`, `executor`, and the full `pkg/router`.
- [x] Added race-validated concurrency tests for `pkg/cache` and `FunctionServiceCache`; `TestPoolCacheRequests` hammered x10 under `-race`.
- [x] All background cache goroutines removed — `fscache` goleak now needs zero ignores.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -6,20 +6,10 @@ package cache
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	ferror "github.com/fission/fission/pkg/error"
-)
-
-type requestType int
-
-const (
-	GET requestType = iota
-	SET
-	DELETE
-	EXPIRE
-	COPY
-	UPSERT
 )
 
 type (
@@ -29,24 +19,11 @@ type (
 		value V
 	}
 	Cache[K comparable, V any] struct {
+		lock                sync.RWMutex
 		cache               map[K]*Value[V]
 		ctimeExpiry         time.Duration
 		atimeExpiry         time.Duration
 		expiryCheckInterval time.Duration
-		requestChannel      chan *request[K, V]
-	}
-
-	request[K comparable, V any] struct {
-		requestType
-		key             K
-		value           V
-		responseChannel chan *response[K, V]
-	}
-	response[K comparable, V any] struct {
-		error
-		existingValue V
-		mapCopy       map[K]V
-		value         V
 	}
 )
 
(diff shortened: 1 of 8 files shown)


sanketsudake:
## Reviewer guide

All three caches share one before/after shape — an unbuffered `requestChannel` + a single `service()` goroutine becomes a mutex + direct methods — so once you've seen one, the others read fast.

**Suggested reading order (easiest → most delicate):**

1. `pkg/cache/cache.go` — the pattern in its simplest form (generic K/V cache → `sync.RWMutex`). Start here.
2. `pkg/executor/fscache/functionServiceCache.go` — a coordinator `sync.Mutex` over the four ops the actor mutually excluded; note the documented `fsc.lock` boundary (the `Get*`/`AddFunc` lock-free `Atime` writes are a pre-existing, out-of-scope race).
3. `pkg/executor/fscache/poolcache.go` — **the delicate one.** The only subtle bit is the blocking `svcChannel` hand-off: `getSvcValue` returns the waiter under the lock, and `GetSvcValue` waits on the channel *after* the lock is released, so `SetSvcValue`'s wake-up send (under the lock) can't deadlock. Comments at the send site and on the split explain why.

**Core vs. test files:**
- Core: `pkg/cache/cache.go`, `pkg/executor/fscache/{poolcache,functionServiceCache}.go`
- Tests: `*_test.go` (added race-validated concurrency tests for `cache` and `FunctionServiceCache`, plus a regression test for the `ListOld` wedge); both `main_test.go` drop now-obsolete goleak ignores.

**What proves it's behavior-preserving:**
- Public APIs are byte-identical → zero caller churn (`go build ./...` clean).
- `-race` green across `cache`, `fscache`, `poolmgr`, `executor`, full `router`.
- Both `main_test.go` files end with **zero goleak ignores** — the actor goroutines are genuinely gone, and goleak proves nothing leaked.

**One deliberate behavior change** (see the "Behavior note" in the PR description): `ListOld` now returns a partial list instead of wedging the cache on a TOCTOU index miss.

**Commits** are in dependency order (CP1 cache → CP2 PoolCache → CP3 FunctionServiceCache, then two small cleanup commits); each is individually buildable and `-race`-green.

sanketsudake:
## Before/after profile (CI pprof)

Measured from the integration job's heap + goroutine pprof.
Comparison is **loaded-vs-loaded**, same matrix leg (v1.34.8, post-test): baseline `main` (`1e1401cd`) vs this branch (`b8a0600c` — runtime-identical to tip; the later commits only add a test, comments, and slice capacity hints).

### Attributable to the refactor — load-independent

Per-cache actor `service()` goroutines exist regardless of traffic, so this is the clean signal:

| dedicated actor goroutine | before | after |
|---|---:|---:|
| executor `cache.Cache.service` | 10 | **0** |
| executor `PoolCache.service` | 3 | **0** |
| executor `FunctionServiceCache.service` | 2 | **0** |
| router `cache.Cache.service` (functionServiceMap) | 1 | **0** |
| **background goroutines removed** | **16** | **0** |

This is also why `fscache`'s goleak ignore set drops to zero.

### Head-of-line blocking on the tap path — eliminated

| executor, under load (v1.34.8 post-test) | before | after |
|---|---:|---:|
| taps parked **sending on the shared `requestChannel`** (`TouchByAddress`) | 22 | **0** |
| `tapServices` handlers serialized behind one goroutine | 22 | **0** |

Before, 22 concurrent `/tapService` requests queued on one unbuffered channel drained by a single goroutine.
That channel + goroutine no longer exist; a tap now takes a per-cache mutex and returns.

### Capture-timing variance — NOT claimed as wins

For honesty, these deltas are dominated by capture timing, not the refactor:

| | before | after | note |
|---|---:|---:|---|
| executor total goroutines | 280 | 212 | ~16 = the removed actors; remainder is in-flight-request variance |
| router total goroutines | 136 | 69 | mostly outbound keep-alive pool variance (router only lost 1 actor) |
| executor heap inuse | 11.9 MB | 9.0 MB | within per-`/metrics`-gzip + init-baseline noise |
| router heap inuse | 13.4 MB | 8.7 MB | structurally flat |

Heap is unchanged in substance — the refactor removes channels/goroutines (KB-scale), not the MB-scale scrape buffers that dominate the heap.

**Net:** 16 background goroutines and the per-tap single-channel serialization deleted; heap flat. No regression.

PR #3530, opened 2026-06-22
refactor(router): remove ROUTER_INCREMENTAL_ROUTES escape hatch (RFC-0013)

sanketsudake:
## What

Removes the RFC-0013 `ROUTER_INCREMENTAL_ROUTES` escape hatch so the per-event **incremental route materializer is the only production route-update path**. Deletes the legacy full-rebuild loop. Net **−149 lines** (14 files).

## Why

RFC-0013 incremental routing shipped as the default in 1.27, with the legacy full-rebuild loop kept behind `ROUTER_INCREMENTAL_ROUTES=false` as a deliberately one-release escape hatch (the config comment said "for one release"). 1.27 has shipped, so the window has elapsed — this completes the RFC-0013 rollout.

## How to review

Three commits, split by concern:

1. **`refactor(router)` — production source** (the meat). `config.go` drops the `incrementalRoutes` field + env parsing; `httpTriggers.go` removes the `ts.incremental` field and **deletes** the legacy `updateRouter()` loop + its `syncTriggers()` debounce; `router.go` + `reconciler.go` drop the `if ts.incremental {…} else {…legacy…}` branches so the incremental path is unconditional. `enableIncrementalRoutes()` → `initIncrementalRoutes()` (it builds the route table + featureConfigFn, so it's real init).
2. **`test(router)`** — delete `BenchmarkBuildMuxes`; rename the legacy-parity test to `TestBuildMuxesIncrementalParity`; reconciler tests now drive the incremental path (which resolves functions at reconcile time) so they get real `FunctionReference`s + backing Functions.
3. **`chore`** — drop the gate from the CI legacy-pin step (keeping the RFC-0002 pins), the Helm value/env, and the docs/CLAUDE.md wording.

**Kept on purpose:** `buildMuxes()` survives as the one-shot mux builder used by the nil-client test scaffold and the shape/parity unit tests — it is *not* the legacy loop (`updateRouter` was). All shared `routeshape.go` / `mutablemux.go` helpers and the GHSA-3g33-6vg6-27m8 public/internal listener split are untouched.

## Behavior change

`ROUTER_INCREMENTAL_ROUTES` (env) and `router.incrementalRoutes` (Helm value) no longer exist; setting them is silently ignored. There is **no data-plane behavior change** — incremental routing was already the default in 1.27; this only removes the ability to opt back into the legacy rebuild.

## Testing

- `go build ./...`, `go vet ./pkg/router/...`, `gofmt -l` clean
- `golangci-lint run ./pkg/router/...` — 0 issues
- Full `go test -race ./pkg/router/...` (with envtest) — all green
- `helm lint charts/fission-all` — 0 failed; rendered template has no `INCREMENTAL_ROUTES`


sanketsudake:
Updated this PR:

- **Rebased** onto current `main` (now includes the merged #3528 and #3529) — clean, no conflicts.
- **Folded in review-battery fixes.** A multi-dimension adversarial review (correctness / GHSA-security / quality / deslop) surfaced **no correctness or security defects** — the public/internal listener split and incremental behavior are intact. It found only stale "legacy mode" vestiges the escape-hatch removal had missed, now swept:
  - the `muxRebuilds` Prometheus Help text + comment in `metrics.go` (the `legacy` reason label is no longer emitted — only `shape_change`);
  - `materialize()` / `resyncLoop` / reconciler doc comments that still referenced the deleted full-rebuild path;
  - `docs/rfc/0013` prose (dropped a cite-then-retract benchmark note; one-sentence-per-line).

Re-verified locally: `go build ./...`, `golangci-lint ./pkg/router/...` (0 issues), and the full `go test -race ./pkg/router/...` suite — all green.

PR #3547, opened 2026-06-27
fix(fission-bundle): set the controller-runtime logger for all subsystems

sanketsudake:
## What

Set controller-runtime's process-global logger exactly once per process, at the entrypoint, instead of in each subsystem's `Start`.

- `cmd/fission-bundle/main.go` sets it before dispatching to any subsystem.
- `test/e2e/framework/services/services.go` (`StartServices`) sets it for the in-process e2e harness, which starts webhook + executor + router + the others in one process.
- The per-`Start` calls in `router`, `executor`, and `webhook` are removed (they were re-setting the same global).

## Why (found while scanning CI kind-logs)

Six control-plane subsystems each emit this at startup:

```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
Detected at:
  >  goroutine N [running]
  >  runtime/debug.Stack() ...
```

Seen in **kubewatcher, timer, mqtrigger, mqt_keda, canaryconfig, buildermgr, mcp**.
Each builds a controller-runtime manager in its `Start` but never sets the logger, so controller-runtime's own logs — **manager cache-sync, reconcile errors, leader-election** — are silently dropped, and a one-off stack trace is printed.
Only `router`/`executor`/`webhook` set it before, each in its own `Start`.

`ctrl.SetLogger` targets controller-runtime's **global** logger, so it only needs to be set once per process.
The `mgr` threaded through `main.go` is an `errgroup.Group`, not a controller-runtime manager — each subsystem makes its own — so there is no single manager to attach a logger to; the global is the right seam.
`fission-bundle` runs exactly one subsystem per process, so a single call in `main()` before the dispatch covers whichever subsystem runs.
The e2e harness is the one place that runs several managers in a single process, so it gets its own entrypoint call.

## Effect

- Restores controller-runtime diagnostics (cache-sync/reconcile/leader-election) for the six subsystems that were dropping them — a real observability gain (these were invisible in production logs).
- Removes the noisy stack-trace warning.
- Unifies the controller-runtime logger name to `controller-runtime` across subsystems (webhook previously named it `webhook`) — a harmless naming change.

No behavior change beyond logging.
`go build ./...`, `vet`, `golangci-lint`, `license-check` clean.

## Also: e2e harness teardown robustness (same file)

While verifying the `services.go` change I traced a pre-existing flake on the `lint` job's `test/e2e/cli` leg to this harness.
Each subsystem's `Start` ran in the errgroup and `os.Exit(1)`'d on any error — including at teardown, where a controller-runtime manager whose runnables exceed the 30s graceful-shutdown grace period (poolmgr/newdeploy cleanup draining slowly under CI load) makes `mgr.Start` return `"failed waiting for all runnables to end within grace period of 30s"`, aborting the whole package with no `--- FAIL` line.
All subsystem Starts now go through a `runService` helper that only `os.Exit`s when `ctx` hasn't been cancelled (a genuine startup failure), not on teardown shutdown noise.
The helper also collapses seven near-identical goroutine blocks and removes a data race on the shared `err` the webhook goroutine wrote.

### DIFF
--- a/cmd/fission-bundle/main.go
+++ b/cmd/fission-bundle/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"golang.org/x/sync/errgroup"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 	cnwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 
@@ -131,6 +132,15 @@ func main() {
 
 	// Initialize logger
 	logger := loggerfactory.GetLogger()
+
+	// Route controller-runtime's own logs (manager cache-sync, reconcile,
+	// leader-election) through our logger for EVERY subsystem. Each subsystem
+	// builds its own controller-runtime manager in its Start; without a global
+	// logger the ones that don't set it themselves (kubewatcher, timer,
+	// mqtrigger, mqt_keda, canaryconfig, buildermgr, mcp) silently drop those
+	// logs and print a one-off "log.SetLogger was never called" stack trace.
+	// Setting it once here, before any Start, covers them all.
+	ctrl.SetLogger(logger.WithName("controller-runtime"))
 
 	// Set up signal handling for graceful shutdown
 	ctx := signals.SetupSignalHandler()
--- a/cmd/fission-bundle/main.go
+++ b/cmd/fission-bundle/main.go
@@ -133,13 +133,11 @@ func main() {
 	// Initialize logger
 	logger := loggerfactory.GetLogger()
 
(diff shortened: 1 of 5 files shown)


sanketsudake:
**Verified on this PR's CI run** (kind-logs, v1.34.8 integration):

- `[controller-runtime] log.SetLogger(...) was never called` warning + stack trace: **gone from all six subsystems** (6 pods → 0).
- controller-runtime diagnostics are now **captured** where they occur (buildermgr / executor / mqtrigger-keda / router / webhook) — all INFO level, with **no previously-hidden warnings or errors** surfacing (nothing bad was being suppressed; the change is pure observability + noise removal).
- No container restarts, no OOMKills.

No behavior change beyond logging; `go build ./...` / `vet` / `golangci-lint` / `license-check` clean.

PR #3553, opened 2026-07-02
fix(cli): spec apply fails to retrigger build after source change (#3427 regression)

markretallack:
## Description

`fission spec apply` breaks source-build packages on the second and subsequent deploys when source changes.

When the CLI updates a source-build package, it overwrites `spec.Deployment` with `{checksum: {}}` from the spec file, wiping the deploy archive URL that the buildermgr wrote on the previous successful build. Prior to #3427 (Package /status subresource), the old code set `BuildStatus = pending` in the same `Update` call, which immediately triggered a rebuild. After #3427, the `/status` subresource means the API server silently strips status fields from a main-resource Update, so `BuildStatus` stays at `succeeded` and the buildermgr never rebuilds.

The result is pods crash-looping with `"failed to download url from archive: empty URL"` from the fetcher.

**The fix has two parts:**

1. **Update path**: Always retrigger (set `BuildStatus=pending` via `UpdateStatus`) for source-build packages (non-empty source + build command), not only when `BuildStatus==failed`.

2. **Equality check**: Treat a source-build package with an empty `spec.Deployment` as not-ready even if `BuildStatus==succeeded`. This handles packages left in a stuck state by a previous broken deploy — the CLI will re-apply and retrigger the build.

Fixes regression introduced in #3427

## Testing

- Deployed a multi-function Fission application (6 source-build packages) using `fission spec apply --wait --delete=true` against a v1.27.0 cluster.
- Without the fix: all packages show `buildstatus: succeeded` with empty `spec.deployment.url` after source changes. Pods fail readiness probes (fetcher: "empty URL"). Verified via `kubectl get packages -o jsonpath` and `kubectl logs` on the fetcher container.
- With the fix: all packages rebuild after `spec apply`, deploy archive URLs are populated, pods start 2/2 healthy.
- Verified idempotency: re-running `spec apply` with no source changes does not trigger unnecessary rebuilds (the `equal()` function correctly identifies packages as up-to-date when both source matches and deploy URL is present).
- Tested the stuck-state recovery: packages previously left with `buildstatus: succeeded` but no deploy URL are correctly detected as not-ready and rebuilt on the next `spec apply`.

## Checklist:

### DIFF
--- a/pkg/fission-cli/cmd/spec/apply.go
+++ b/pkg/fission-cli/cmd/spec/apply.go
@@ -668,6 +668,14 @@ func applyPackages(ctx context.Context, fclient cmd.Client, fr *FissionResources
 			// would re-apply unchanged deploy packages on every run.
 			ready := existing.Status.BuildStatus == fv1.BuildStatusSucceeded ||
 				existing.Status.BuildStatus == fv1.BuildStatusNone
+			// A source-build package that has "succeeded" but no deployment
+			// archive is in a broken state (the deploy URL was wiped by a
+			// previous spec apply). Force a re-apply so the update path can
+			// retrigger the build.
+			if ready && existing.Spec.BuildCommand != "" &&
+				!existing.Spec.Source.IsEmpty() && existing.Spec.Deployment.IsEmpty() {
+				ready = false
+			}
 			return specMatches &&
 				isObjectMetaEqual(existing.ObjectMeta, desired.ObjectMeta) &&
 				ready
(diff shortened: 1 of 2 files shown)


markretallack:
Hi, 

I have done an internal review, and these are the items:

1. Redundant rebuilds on metadata-only changes — if you only change an annotation on a source-build package (not the source), it'll trigger a rebuild unnecessarily. This is the conservative/safe approach and acceptable for a bug-fix. A more targeted condition could check current.Spec.Deployment.IsEmpty() but introduces a race risk with stale builds.

2. No unit test coverage — the reviewer noted no tests exercise the new paths. If you want the PR accepted upstream, adding table-driven tests for the equal function with cases like (succeeded + empty deploy, succeeded + non-empty deploy) would strengthen it.

Can you except as-is or would you prefer these done first?

markretallack:
Ok added tests, Note, I may of done the wrong thing, or fixed the issue in the wrong place.

sanketsudake (approved):

PR #3581, opened 2026-07-14
RFC-0026 PR1: Provisioned Concurrency & Scheduled Warming

### LINKED ISSUE #3572: RFC-0026: Provisioned concurrency & scheduled warming
Part of #3566. Design: [`docs/rfc/0026-provisioned-concurrency-scheduled-warming.md`](https://github.com/fission/fission/blob/docs/rfc-0021-0026-stateful-orchestration-wave/docs/rfc/0026-provisioned-concurrency-scheduled-warming.md) (#3565).

Let a function declare a floor of always-warm capacity. `FunctionSpec.ProvisionedConcurrency` keeps N pods **pre-specialized** (package loaded, ready to serve) at all times, with optional cron-scheduled windows for known traffic patterns. The executor specializes eagerly from the generic pool, exempts these pods from the idle reaper, and publishes them via the existing `fission.io/served` label + headless-Service path (RFC-0002) so the router sees them as instantly warm. This is Lambda provisioned concurrency — the cheapest, broadest lever in the wave: it removes cold starts entirely for opted-in functions. No statestore dependency.

### Implementation phases
- [ ] CRD field + webhook validation (cron, caps, poolmgr-only) + provisioner with base `Target` (eager specialize, reaper exemption, status, metrics)
- [ ] Scheduled windows (cron transitions, overlapping-window max, restart re-derivation)
- [ ] Generation make-before-break on updates; warm-rollback interaction with #3571
- [ ] Bench scenario + docs (generic-pool sizing, cost guidance)

### Verification
Effective-target evaluation as `rapid` properties (overlap → max, `Target:0` windows, DST crons), reconcile loop against a fake pool inside `testing/synctest` bubbles (virtual clock drives cron + reaper deterministically), no-starvation CI guardrail (warm-up burst must not regress another function's cold-start p95).

### Related (linked not closed)
- #528 cold start benchmarking (open), #205 Cold start optimization: prefetching in fetcher (open) — complementary cold-start work; provisioned concurrency is the declarative "never cold" floor alongside these heuristics.
- #253 piggyback specialization, #204 optimistic pod selection, #1008 minscale=0/HPA conflict (closed) — prior cold-start/scaling threads for context.

rijojohn85:
## Summary

Implements RFC-0026 PR1: provisioned concurrency base for poolmgr functions. Keeps N warm specialized pods per opted-in function, exempt from the idle reaper.

Closes #3572.

### What changed

**API types** (`pkg/apis/core/v1/`):
- `ProvisionedConcurrencyConfig` + `ProvisionedWindow` types added to `FunctionSpec`
- `ProvisionedReady` + `ProvisionedTarget` status fields added to `FunctionStatus`
- `FunctionConditionProvisioned` condition + reasons (`ProvisionedSatisfied`, `ProvisionedWarming`, `ProvisionedDisabled`)
- `PROVISIONED_LABEL`/`PROVISIONED_VALUE` constants (`fission.io/provisioned=true`)
- Webhook validation: poolmgr-only (CEL), target bounds (CEL), Windows-emptiness PR1 gate (Go admission webhook)

**Provisioner** (`pkg/executor/executortype/poolmgr/provisioner.go`):
- New component: periodic reconcile loop + per-event `ReconcileFunction`
- Eagerly specializes pool pods, patches `PROVISIONED_LABEL` on, publishes to function's headless Service
- Self-heal: re-warms after pod delete via periodic tick
- Generation refresh: CLI target change bumps Generation → `refreshFuncPods` recycles old-gen pods → provisioner re-warms fresh
- Disable: `--provisioned-concurrency 0` → `StopProvisioning` + status reset + condition False/`ProvisionedDisabled`

**Idle reaper** (`pkg/executor/reaper/idle/idle.go`):
- Function-level exemption: skips reaping pods for functions with non-nil `ProvisionedConcurrency` (per-pod label exemption deferred to PR2)

**CLI** (`pkg/fission-cli/cmd/function/`):
- `--provisioned-concurrency` flag on `create` + `update`

**Helm chart** (`charts/fission-all/`):
- `executor.provisionedConcurrency.{enabled,maxPerFunction,maxInflightPerFunction,reconcileInterval}` values
- Env vars: `EXECUTOR_PROVISIONED_CONCURRENCY_ENABLED`, `EXECUTOR_PROVISIONED_MAX_PER_FUNCTION`, `EXECUTOR_PROVISIONED_MAX_INFLIGHT_PER_FUNCTION`, `EXECUTOR_PROVISIONED_RECONCILE_INTERVAL`

**Integration tests** (`test/integration/`):
- 5 new helpers in `framework/provisioned.go`: `ReadyProvisionedPods`, `WaitForProvisionedPodsAtLeast`, `WaitForNoProvisionedPods`, `WaitForProvisionedStatus`, `RunningFunctionPodCount`
- `framework/function.go`: `ProvisionedConcurrency` + `IdleTimeout` fields on `FunctionOptions`
- `framework/cli.go`: `CLIExpectError` method
- `framework/executor_lifecycle.go`: `ExecutorEnvEnabled` generic gate
- 4 common-suite tests + 1 serial-suite test (see below)

**CI** (`skaffold.yaml`):
- `kind` + `kind-ci` profiles: `executor.provisionedConcurrency.enabled=true`, `reconcileInterval=10s`

### Why these choices

- **Function-level reaper exemption (not per-pod label)**: PR1 scope. Per-pod exemption deferred to PR2 — needs label lifecycle work.
- **Two reconcile drivers**: Periodic `Provisioner.Run` (10s in CI) AND per-event `ReconcileFunction`. Self-heal after pod delete is periodic-tick-only (no event fires on delete). CLI target drops converge via generation refresh, not idle reaper.
- **Generation refresh for target drops**: CLI `--provisioned-concurrency` change bumps Generation → `refreshFuncPods` recycles all old-gen pods → provisioner re-warms fresh. This is how target drops converge, not via idle reaper.
- **Conflict retry in windows test**: Provisioner status writes bump resourceVersion between Get and Update, causing stale-version conflicts that mask webhook errors. `EventuallyWithT` retries on conflict.

### Common suite (`test/integration/suites/common/provisioned_concurrency_test.go`)

1. **`TestProvisionedConcurrencyCLIRoundTrip`** (0.16s) — PASS
   - Creates fn with PC=3, asserts CR field, updates to 5, asserts. No provisioner gate needed.

2. **`TestProvisionedConcurrencyRejectsNewdeploy`** (0.06s) — PASS
   - Uses `CLIExpectError`, asserts CEL rejection message contains "provisionedConcurrency is only supported with executortype".

3. **`TestProvisionedConcurrencyRejectsWindows`** (0.11s) — PASS
   - Creates valid poolmgr fn with PC=2, then updates via typed clientset adding `Windows` field. Retries on conflict. Asserts webhook rejects with "scheduled windows are not yet supported".

4. **`TestProvisionedConcurrencyLifecycle`** (61.71s) — PASS
   - 5-phase end-to-end test:
     - **Phase 1**: Floor@0 — provisioner warms 2 pods without traffic, status converges to 2,2, condition True, smoke test serves.
     - **Phase 2**: Reaper liveness — control fn (no PC) reaped after 30s idle, provisioned fn floor holds (2,2). Proves reaper exemption.
     - **Phase 3**: Self-heal — delete one provisioned pod, provisioner replaces it via periodic tick. New pod name differs from victim.
     - **Phase 4**: Target drop 2→1 via CLI — generation refresh. Status converges to 1,1, total pod count 1, pod generation differs.
     - **Phase 5**: Disable — `--provisioned-concurrency 0`. Status 0,0, no provisioned pods, condition False/`ProvisionedDisabled`, all pods gone.

### Serial suite (`test/integration/suites/serial/provisioned_concurrency_restart_test.go`)

5. **`TestProvisionedConcurrencyExecutorRestart`** (32.20s) — PASS
   - Establishes floor (2 pods), deletes BOTH provisioned pods, restarts executor via `SetExecutorEnv` nonce + `WaitForExecutorRollout`. Fresh executor re-derives floor via periodic reconcile tick.

### Test logs

```
=== RUN   TestProvisionedConcurrencyCLIRoundTrip
--- PASS: TestProvisionedConcurrencyCLIRoundTrip (0.16s)

=== RUN   TestProvisionedConcurrencyRejectsNewdeploy
--- PASS: TestProvisionedConcurrencyRejectsNewdeploy (0.06s)

=== RUN   TestProvisionedConcurrencyRejectsWindows
--- PASS: TestProvisionedConcurrencyRejectsWindows (0.11s)

=== RUN   TestProvisionedConcurrencyLifecycle
    provisioned_concurrency_test.go:217: Phase 1: floor@0 — provisioner warms 2 pods without traffic
    provisioned_concurrency_test.go:225: Phase 2: reaper liveness — control reaped, provisioned floor holds
    provisioned_concurrency_test.go:241: Phase 3: self-heal after pod delete
    provisioned_concurrency_test.go:272: Phase 4: target drop 2→1 via CLI (generation refresh)
    provisioned_concurrency_test.go:300: Phase 5: disable — provisioned-concurrency 0
--- PASS: TestProvisionedConcurrencyLifecycle (61.71s)

=== RUN   TestProvisionedConcurrencyExecutorRestart
    provisioned_concurrency_restart_test.go:70: establishing provisioned floor before restart
    provisioned_concurrency_restart_test.go:77: deleting both provisioned pods
    provisioned_concurrency_restart_test.go:88: restarting executor
    provisioned_concurrency_restart_test.go:94: waiting for fresh executor to re-derive provisioned floor
--- PASS: TestProvisionedConcurrencyExecutorRestart (32.20s)
```

All 5 tests run against a live kind cluster with `EXECUTOR_PROVISIONED_CONCURRENCY_ENABLED=true` + `EXECUTOR_PROVISIONED_RECONCILE_INTERVAL=10s`.

### High-risk areas (review carefully)

1. **`pkg/executor/executortype/poolmgr/provisioner.go`** — core logic. Key functions:
   - `Run` — periodic reconcile loop (10s tick)
   - `ReconcileFunction` — per-event handler
   - `eagerSpecialize` — specializes pool pod + patches provisioned label
   - `UpdateFunctionStatus` — writes `ProvisionedReady`/`ProvisionedTarget` + condition
   - `StopProvisioning` — disable path (target=0/nil)

2. **`pkg/executor/reaper/idle/idle.go`** — reaper exemption. Function-level check (not per-pod label). PR2 will move to per-pod.

3. **`pkg/apis/core/v1/validation.go`** — `ProvisionedConcurrencyConfig.Validate` rejects Windows in PR1.

4. **`pkg/executor/executortype/poolmgr/reconciler.go`** — wires provisioner into `ReconcileFunction` + `refreshFuncPods` generation refresh.

### Low-risk areas (skim)

- CLI flags, Helm chart values, skaffold patches — additive, gated.
- Generated code (`pkg/generated/`, `zz_generated.deepcopy.go`, CRD YAML) — regenerated via `make codegen` + `make generate-crds`.
- Integration test framework helpers — read-only helpers, no production impact.

### Gates passed

- `make license-check` — pass
- `make code-checks` (golangci-lint) — 0 issues
- `go vet -tags=integration` — pass
- `go build ./...` — pass
- All 5 integration tests — pass against live kind cluster

#### Test plan

- [ ] CI integration tests pass (common + serial suites)
- [ ] `make code-checks` passes
- [ ] `make license-check` passes
- [ ] Review provisioner logic: floor establishment, self-heal, generation refresh, disable
- [ ] Review reaper exemption: function-level check correct
- [ ] Review webhook validation: Windows rejection message matches test assertion


.github/code-review-graph.instruction.md
```diff
@@ -0,0 +1,46 @@
+---
```
sanketsudake: Please remove changes under .github and .claude unless intentional.

CLAUDE.md
```diff
@@ -141,3 +141,42 @@ The CLI (`cmd/fission-cli` + `pkg/fission-cli/`) talks to Kubernetes directly th
   Always pass a clone (`registerRouteShape` does); the race only surfaces under `-race` with concurrent build+serve.
 - `EXECUTOR_SPECIALIZATION_CONCURRENCY` bounds **one semaphore shared by all executor types**: a small bound lets newdeploy's minutes-long `waitForDeployment` holds starve poolmgr's ~100ms specializations (head-of-line blocking; surfaced as mass cold-start timeouts in CI).
   Leave it 0 (unbounded) unless poolmgr-only pressure is the proven problem.
+
+<!-- code-review-graph MCP tools -->
```
sanketsudake: We can remove these changes unless required

.mcp.json
```diff
@@ -0,0 +1,13 @@
+{
```
sanketsudake: Please remove

CLAUDE.md
```diff
@@ -141,3 +141,42 @@ The CLI (`cmd/fission-cli` + `pkg/fission-cli/`) talks to Kubernetes directly th
...
+## MCP Tools: code-review-graph
```
sanketsudake: Please remove

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+		p.clearExcessProvisionedLabels(ctx, fn, excess)
+		// returning here cause the pods will take time to provision, status will be updated on the next tick
+		return
+	}
+
+	// if the code reaches here, ready = target, update Target
+	err = p.updateFunctionStatus(ctx, fn, ready, target)
+	if err != nil {
+		p.logger.Error(err, "Unable to update status of the function", "function", fn.Name, "namespace", fn.Namespace, "ready", ready, "target", target)
+	}
```
sanketsudake, line 265: **[correctness / rfc-adherence — medium]** Status is written **only** on the exact `ready == target` tick. Both the `ready < target` (warming) and `ready > target` (draining) branches early-return without calling `updateFunctionStatus`, so:

- `Status.ProvisionedReady` never surfaces during warm-up — it jumps straight from stale to satisfied, defeating the RFC-0026 "honest status: report how many provisioned pods are actually ready" goal.
- The `ProvisionedWarming` reason branch in `statusSet` is **unreachable dead code** (the only caller with `ready != target` is the `target == 0` path, which reports `Disabled`).

Publish status on every reconcile pass. This also flattens the `else if`-after-`return` into plain branches:

```suggestion
	if ready < target {
		delta := target - ready
		p.fireEagerSpecializations(ctx, fn, delta)
	} else if ready > target {
		excess := ready - target
		p.clearExcessProvisionedLabels(ctx, fn, excess)
	}

	// Publish observed status on every pass so ProvisionedReady and the
	// Provisioned=Warming condition surface during warm-up/drain, not only
	// once ready == target.
	if err := p.updateFunctionStatus(ctx, fn, ready, target); err != nil {
		p.logger.Error(err, "Unable to update status of the function", "function", fn.Name, "namespace", fn.Namespace, "ready", ready, "target", target)
	}
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		return err
+	}
+
+	statusSet(&latestFunc, ready, target)
+	_, err = p.fissionClient.CoreV1().Functions(fn.Namespace).UpdateStatus(ctx, &latestFunc, metav1.UpdateOptions{})
+
+	if err != nil && apierrors.IsConflict(err) {
+		// retry one with fresh version
+		if err2 := p.crClient.Get(ctx, client.ObjectKey{Name: fn.Name, Namespace: fn.Namespace}, &latestFunc); err2 != nil {
```
sanketsudake: **[correctness — medium]** `updateFunctionStatus` reads `latestFunc` from `p.crClient` (the informer cache) and writes via the direct `fissionClient`. On `IsConflict` the retry re-`Get`s from the **same cached client**, which frequently still returns the stale ResourceVersion — so the single retry conflicts again. This is the flakiness the tests worked around with `EventuallyWithT`.

Do the read-modify-write against an uncached reader, e.g. `p.fissionClient.CoreV1().Functions(fn.Namespace).Get(ctx, fn.Name, metav1.GetOptions{})`, for both the initial read and the retry. (A `RetryOnConflict` loop would be cleaner than the hand-rolled single retry.)

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+
+func (p *Provisioner) clearExcessProvisionedLabels(ctx context.Context, fn *fv1.Function, excess int) {
+	podList, err := p.listPods(ctx, fn)
+	if err != nil {
+		p.logger.Error(err, "Unable to list pods", "function", fn.Name, "namespace", fn.Namespace)
+		return
+	}
+	if len(podList.Items) < excess {
+		return
+	}
```
sanketsudake: **[correctness — low]** This guard bails out clearing **nothing** when there are fewer provisioned pods than the computed `excess`. The loop below is already dual-bounded (`i < excess && i < len(podList.Items)`), so the guard is dead in practice and, if ever hit, wrongly leaves stale `fission.io/provisioned` labels in place. Remove it. (Note: `provisioner_test.go` asserts this no-op behavior, so the test currently locks in the bug.)

```suggestion
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+		case <-ticker.C:
+			p.reconcileAll(ctx)
+		}
+	}
+}
+
+func filterOptedFunctions(fnlist *fv1.FunctionList) []fv1.Function {
+	var opted []fv1.Function
+	for _, fn := range fnlist.Items {
+		if fn.Spec.ProvisionedConcurrency != nil && fn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType == fv1.ExecutorTypePoolmgr {
```
sanketsudake: **[correctness — medium]** This matches only `ExecutorType == fv1.ExecutorTypePoolmgr` (the literal `"poolmgr"`). But `ExecutorType` **defaults to poolmgr when empty**, and the CEL admission rule explicitly admits `provisionedConcurrency` with an empty executor type (`... == '' || ... == 'poolmgr'`). A raw `kubectl`/GitOps write that sets `provisionedConcurrency` but leaves `executorType` unset therefore passes admission but is **silently ignored** by the provisioner — exactly the raw-write path RFC-0026 calls out. CLI-created functions are unaffected (the CLI defaults the flag). Treat empty as poolmgr here, e.g. `et := fn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType; (et == "" || et == fv1.ExecutorTypePoolmgr)`.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+	cfg := ProvisionerConfig{
+		MaxPerFunction:         util.AtoiOr("EXECUTOR_PROVISIONED_MAX_PER_FUNCTION", 20),
+		MaxInflightPerFunction: util.AtoiOr("EXECUTOR_PROVISIONED_MAX_INFLIGHT_PER_FUNCTION", 4),
+		ReconcileInterval:      util.DurOr("EXECUTOR_PROVISIONED_RECONCILE_INTERVAL", 30*time.Second),
+	}
+	return cfg, true
+}
+
+// NewProvisioner creates a new Provisioner. If any fields in config is missing/zero, default values
+// are applied.
```
sanketsudake: **[correctness — medium]** The doc comment says "If any fields in config is missing/zero, default values are applied", but `NewProvisioner` no longer applies any defaults (they moved to `ProvisionerConfigFromEnv`). A caller constructing a `Provisioner` with a zero `ReconcileInterval` (a future call site or a unit test) will hit `time.NewTicker(0)` in `Run`, which **panics** ("non-positive interval for NewTicker"). Either restore the zero-defaulting here (matching the doc) or clamp `ReconcileInterval` to a sane floor in `Run`, and fix the comment.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+		fissionClient:    fissionClient,
+		kubernetesClient: kubernetesClient,
+		crClient:         crClient,
+		config:           config,
+		inflight:         sync.Map{},
+	}
+}
+
+func (p *Provisioner) Run(ctx context.Context) {
+	ticker := time.NewTicker(p.config.ReconcileInterval)
```
sanketsudake: **[rfc-adherence — high]** RFC-0026 phase 1 scope is "...eager specialize, reaper exemption, status, **metrics**", and the Observability section names four meters: `fission_provisioned_target`, `fission_provisioned_ready`, `fission_provisioned_eager_specializations_total{outcome}`, `fission_provisioned_window_transitions_total`. None are implemented — there is no meter/counter/gauge anywhere in the provisioner. Without `eager_specializations_total{outcome}` in particular, warm-up failures (the `eagerSpecialize` error path) are invisible except in logs. Please add the RFC-0019-style meters (the window-transitions counter can wait for PR2 since windows are rejected here).

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+			return err
+		}
+		statusSet(&latestFunc, ready, target)
+		_, err = p.fissionClient.CoreV1().Functions(fn.Namespace).UpdateStatus(ctx, &latestFunc, metav1.UpdateOptions{})
+	}
+
+	return err
+}
+
+func (p *Provisioner) clearAllProvisionedLabels(ctx context.Context, fn *fv1.Function) {
```
sanketsudake: **[simplification — medium]** `clearAllProvisionedLabels` and `clearExcessProvisionedLabels` are the same list-then-patch loop — "clear all" is just "clear excess" with `excess >= len(pods)`. Both re-implement `listPods` + iterate + `clearProvisionedLabel` + identical error logging, so any future change (patch retry, a metric, the label-clear semantics) has to be made in two places. Collapse to one helper, e.g. `clearProvisionedLabels(ctx, fn, max int)` where `max < 0` means all.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	// filter only functions which have ProvisionedConcurrency and ExecutorType=poolmgr
+	opted := filterOptedFunctions(fnList)
+
+	// if no functions have opted in, return immediately.
+	if len(opted) == 0 {
+		return
+	}
+
+	const maxConcurrentReconciles = 10 // TODO: make this configurable
```
sanketsudake: **[cleanup — low]** `// TODO: make this configurable` ships an unresolved TODO. The provisioner already reads `EXECUTOR_PROVISIONED_*` env vars — either wire this bound to one of them or drop the TODO and document the fixed value.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+
+func (p *Provisioner) release(fnUID types.UID) {
+	v, ok := p.inflight.Load(fnUID)
+	if !ok {
+		return
+	}
+	v.(*atomic.Int32).Add(-1)
+}
+
+func (p *Provisioner) fireEagerSpecializations(ctx context.Context, fn *fv1.Function, delta int) {
```
sanketsudake, line 336: **[test-coverage — high]** The heart of the feature — `fireEagerSpecializations` → `eagerSpecialize` (`gpm.GetFuncSvc` + provisioned-label patch) and the `ready < target` warming branch of `reconcileFunction` — has **no unit coverage**. `newTestProvisioner` wires `gpm = nil`, so these paths are exercised only by the gated, skippable integration suite. The `tryAcquire` rollback (invariant P3) is likewise tested only single-threaded. Please add unit tests (a fake/stub `gpm`, and a concurrent/`-race` test for the pacing counter) so the actuation and pacing logic is covered without a live cluster.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+}
+
+// computes the effectiveTarget: minimum of provisioned concurreny target
+// and max per function. The schedule will be added in PR2
+func (p *Provisioner) effectiveTarget(fn *fv1.Function) int {
+	return min(fn.Spec.ProvisionedConcurrency.Target, p.config.MaxPerFunction)
+}
+
+func (p *Provisioner) StopProvisioning(ctx context.Context, fn *fv1.Function) {
```
sanketsudake: **[simplification — low]** There are three overlapping "provisioning off" entry points that must stay in lockstep: `reconcileFunction`'s `target == 0` branch (inlines `clearAllProvisionedLabels` + `updateFunctionStatus(0,0)`), `StopProvisioning`, and `UpdateFunctionStatusZero`. `reconciler.go` uses `StopProvisioning` + `UpdateFunctionStatusZero`; the `target == 0` branch does it a third way. Collapse to a single disable helper so the transition can't drift.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,417 @@
...
+	// function UID. Used by tryAcquire/release to pace warm-up and prevent
+	// head-of-line blocking (invariant P5).
+	inflight sync.Map // map[types.UID]*atomic.Int32
+}
+
+// ProvisionerConfigFromEnv build ProvisionerConfig from
+// EXECUTOR_PROVISIONED_* env vars.
+// Returns zero ProvisionerConfig when EXECUTOR_PROVISIONED_CONCURRENCY_ENABLED is unset/false
+func ProvisionerConfigFromEnv() (ProvisionerConfig, bool) {
+	enabled, err := strconv.ParseBool(os.Getenv("EXECUTOR_PROVISIONED_CONCURRENCY_ENABLED"))
```
sanketsudake: **[robustness — low]** Two silent-disable footguns here: (1) any `ParseBool` error on `EXECUTOR_PROVISIONED_CONCURRENCY_ENABLED` (e.g. `yes`/`on`/`1`-with-typo) is swallowed and the feature is silently off — worth a log line; (2) if `MaxInflightPerFunction` or `MaxPerFunction` is explicitly set to `0`, `tryAcquire` always rejects / `effectiveTarget` is always 0, silently disabling provisioning with no diagnostic. Consider validating/logging non-positive values.

pkg/executor/executortype/poolmgr/reconciler.go
```diff
@@ -142,13 +142,32 @@ var poolmgrWarmPodPredicate = predicate.NewPredicateFuncs(func(obj client.Object
...
 // stable function name/uid, so it only needs create/delete.
 func (gpm *GenericPoolManager) ReconcileFunction(ctx context.Context, old, fn *fv1.Function) error {
-	return reconcilePoolmgrFunc(ctx, gpm, gpm.enableIstio, old, fn)
+	err := reconcilePoolmgrFunc(ctx, gpm, gpm.enableIstio, old, fn)
+	if err != nil {
+		return err
+	}
+	if gpm.provisioner != nil {
+		if fn.Spec.ProvisionedConcurrency != nil {
+			gpm.provisioner.reconcileFunction(ctx, fn)
```
sanketsudake, line 151: **[correctness — medium]** `reconcileFunction` is now driven by **two owners in the same (leader) process**: the periodic `Provisioner.Run` tick and this per-CR-event call. They share only the in-flight pacing counter, and the count→fire decision isn't atomic, so a tick and an event can both observe `ready < target` and each fire `delta`, transiently over-provisioning (bounded by `MaxInflightPerFunction`) and racing two status writes (a source of the update conflicts noted elsewhere). It self-corrects, but consider a per-function lock or making CR events merely mark the function dirty for the single periodic owner.

pkg/executor/reaper/idle/idle.go
```diff
@@ -238,6 +238,21 @@ func (s *PoolDeleteStrategy) Reap(ctx context.Context, fsvc *fscache.FuncSvc) er
...
+	// function opts into provisioned concurrency, all its specialized
+	// pods are provisioner-managed. The narrow race between
+	// GetFuncSvc returning and the provisioner's label patch completing
+	// is accepted (design §5j) — the provisioner re-specializes on the
+	// next tick.
+	if fn, ok := s.fnByUID[fsvc.Function.UID]; ok {
+		if fn.Spec.ProvisionedConcurrency != nil {
+			return nil
+		}
+	}
```
sanketsudake, line 265: **[correctness — medium]** This exempts **every** specialized pod of an opted-in function, not just the provisioned floor. An opted-in function (`target = 2`) that bursts to N on-demand pods (up to `Concurrency`, default 500) keeps **all** of them forever while provisioned concurrency is enabled — the reaper never retires the overflow, and unlike the label-patch race this does **not** self-heal (there is no per-pod signal distinguishing floor from overflow). PR2's per-pod label fixes this, but until then please bound the exemption to `target` pods or document the overflow-retention cost. (Also: this function-level exemption has no unit test.)

pkg/apis/core/v1/validation.go
```diff
@@ -369,6 +368,10 @@ func (spec FunctionSpec) Validate() error {
 		errs = errors.Join(errs, spec.Tool.Validate())
 	}
 
+	if spec.ProvisionedConcurrency != nil {
+		errs = errors.Join(errs, spec.ProvisionedConcurrency.Validate())
+	}
```
sanketsudake: **[correctness — low]** `ProvisionedConcurrency.Validate()` runs **twice** on the full-validate path: here in `Validate()` and again inside `validateForAdmission()` (which `Validate()` calls a few lines down). The `Target`/`Windows` errors are therefore joined twice. The pod-spec and `Invocation` checks live only in `validateForAdmission`; do the same here and drop this duplicate call.

```suggestion
```

pkg/apis/core/v1/validation.go
```diff
@@ -594,6 +602,25 @@ func (tc *ToolConfig) Validate() error {
 	return errs
 }
 
+// Validate checks the provisioned concurrency config. In PR 1 only the base
+// Target is meaningful — Windows is accepted in the CRD schema (for PR 2
```
sanketsudake, line 650: **[rfc-adherence — medium]** RFC-0026 says `Target` is bounded by the namespace cap at the webhook. Validation only enforces `Target >= 1` (CEL `minimum: 1`); the cap is applied solely as a runtime clamp in `effectiveTarget` (`min(Target, MaxPerFunction)`). Consequence: a user can set `target: 100000`, it's admitted and stored, but `Status.ProvisionedTarget` reports the clamped `20` — a confusing spec-vs-status divergence with no admission error or condition explaining it. Either enforce a CEL upper bound (a static max) or surface the clamp in status/a condition. (Acknowledged as a design open-question in the RFC — worth a decision before this ships.)

pkg/crd/key.go
```diff
@@ -21,13 +21,12 @@ func (ck CacheKeyUR) String() string {
 }
 
 type CacheKeyURG struct {
-	UID             types.UID
-	ResourceVersion string
-	Generation      int64
+	UID        types.UID
```
sanketsudake, line 24: **[scope / risk — high]** Dropping `ResourceVersion` from `CacheKeyURG` re-keys the entire poolmgr↔router data plane (concurrency accounting, `UnTapService`) for **all** functions — not just provisioned ones. This looks like a legitimate fix for the `activeRequests` leak the new comment describes, but it's unrelated to RFC-0026, ships with **no regression test** for the UnTap/`activeRequests` invariant, and its blast radius touches the "two disjoint accounting modes must never mix" contract. Please split it into its own PR/commit with a focused test so it's reviewable and revertable independently of the feature.

pkg/executor/fscache/poolcache.go
```diff
@@ -273,7 +273,17 @@ func (c *PoolCache) SetSvcValue(ctx context.Context, function crd.CacheKeyURG, a
 	if _, ok := c.cache[function]; !ok {
 		c.cache[function] = NewFuncSvcGroup()
 	}
-	if _, ok := c.cache[function].svcs[address]; !ok {
+	if existing, ok := c.cache[function].svcs[address]; ok {
+		// Same {UID, Gen, address} — this is a re-specialization or
+		// concurrent specialization of the same pod. Log it so we can
+		// verify the RV race fix is merging entries correctly.
+		if c.logger.V(1).Enabled() {
+			otelUtils.LoggerWithTraceID(ctx, c.logger).V(1).Info("SetSvcValue: updating existing entry",
```
sanketsudake: **[cleanup — medium]** This `else` branch exists only to log "so we can verify the RV race fix is merging entries correctly" — investigation scaffolding from the `CacheKeyURG` fix, not durable observability. Same for the `MarkAvailable` "function miss"/"address miss" V(1) blocks below (which build a `knownAddresses` slice just to log it). The `MarkAvailable` early-return refactor itself is a good change worth keeping — just drop the debug logging that chased the cache race.

sanketsudake (commented):
## Review battery — RFC-0026 PR1 (Provisioned Concurrency)

Ran a consolidated review battery (code-review, security-review, /simplify, /deslop, thermo-nuclear quality, and a dedicated test-coverage + RFC-0026 adherence pass) and deduped the results. Overall this is a solid, well-tested-at-the-integration-level PR1, but there are a few things worth addressing before merge. Inline comments below; the highlights:

**Correctness**
- **Warming status is never published** — status only updates at `ready == target`, so `ProvisionedReady` never surfaces during warm-up and `statusSet`'s `ProvisionedWarming` branch is dead code. Directly contradicts the RFC's "honest status" goal. (suggestion inline)
- **`filterOptedFunctions` ignores the default (empty) executor type** — a raw kubectl/GitOps write with `provisionedConcurrency` and an unset `executorType` is admitted by CEL but silently not provisioned.
- **Status RMW mixes a cached read with a direct write** — the conflict retry re-reads the same lagging cache, so it's largely ineffective (the source of the test flakiness worked around with `EventuallyWithT`).
- **Function-level reaper exemption makes on-demand overflow pods immortal** (up to `Concurrency`, default 500) while PC is enabled — does not self-heal.

**RFC-0026 adherence**
- **Metrics are missing.** Phase-1 scope explicitly lists metrics (`fission_provisioned_target/_ready/_eager_specializations_total{outcome}/_window_transitions_total`); none are implemented. Status is delivered, metrics are not.
- **`fission fn get` does not surface `provisionedReady/provisionedTarget`** — the RFC's Status section says it should; today the numbers are only visible via `kubectl`. (No diff line to anchor — flagging here.)
- **Namespace cap is a runtime clamp, not an admission bound** — `target` over the cap is accepted and stored but silently clamped, diverging spec from status.

**Scope / bundling** — three cross-cutting changes ride along with the feature and would be safer as their own PRs: the global `CacheKeyURG` ResourceVersion drop (data-plane cache key for *all* functions), the envreconciler stale-cache guard, and a batch of RV-race investigation debug logging (`poolcache.go`/`gpm.go`). None are wrong, but they widen the blast radius and lack dedicated tests.

**Tests** — the integration lifecycle + serial-restart suites are thorough for the happy paths, but the core eager-specialize/pacing logic has no *unit* coverage (`gpm = nil` in the unit tests), and the RFC's prescribed `testing/synctest` bubble tests for P2/P3/P4 are absent. (`pgregory.net/rapid` property tests are reasonably deferred to PR2 with the schedule windows.)

**Kept as-is (good):** the `MarkAvailable` early-return refactor, the `AtoiOr`/`DurOr` helpers, and the CLI/Helm wiring. Security review found no exploitable issues.

_Posted via an automated multi-lens review pass._

pkg/executor/executortype/poolmgr/gpm.go
```diff
@@ -608,6 +635,11 @@ func (gpm *GenericPoolManager) adoptSpecializedPods(ctx context.Context, wg *syn
 		for i := range podList.Items {
 			pod := &podList.Items[i]
 			if !utils.IsReadyPod(pod) {
+				if len(pod.Status.ContainerStatuses) == 0 {
+					gpm.logger.Info("adopt: skipping pod, containerStatuses empty",
```
sanketsudake: **[cleanup — low]** `"adopt: skipping pod, containerStatuses empty"` logs at plain `Info` (not `V(1)`), so it fires on every adopt pass at startup for each not-yet-ready pod. Together with the V(1) debug blocks added to `UnTapService` (dups the OTel span), `IsValid` (dups the readiness check two lines down), and the redundant `uid`/`generation` fields on `MarkSpecializationFailure` (already encoded in `key.String()`), these are leftovers from the RV-race investigation. Please prune to the lines that carry lasting signal.

pkg/crd/key.go
```diff
@@ -57,15 +56,16 @@ func CacheKeyURFromObject(obj metav1.Object) CacheKeyUR {
...
-// UIDs are unique, uid+resourceVersion identifies the
-// content.
-// Generation is also included to identify latest generation of the object.
+// CacheKeyURGFromMeta creates a cache key that uniquely identifies the
+// function's content version. UID is stable for the function's
+// lifetime; Generation increments on spec changes. ResourceVersion is
+// intentionally excluded: it changes on status updates (not just spec
+// changes), and the router's informer cache may lag the executor's
+// view, causing UnTapService to miss the cache entry and leak
+// activeRequests. See ci-29472717703 v1.36.1 analysis.
```
sanketsudake: **[cleanup — nit]** Several otherwise load-bearing comments end with references that mean nothing to a future reader of the repo: CI-run-id trailers (`See ci-29472717703 v1.32.11 analysis.` here and in `podutils.go`, `gp_pod.go`; `ci-29487828565 ... PCL Phase 3` in `envreconciler.go`) and internal design-doc pointers (`(design §5j)` in `provisioner.go` and `idle.go`). Keep the reasoning, drop the ephemeral run-id / section trailers.

pkg/executor/envreconciler/reconciler.go
```diff
@@ -33,16 +33,32 @@ import (
...
+	apiReader client.Reader // uncached reader for IsNotFound verification
+	handlers  []executortype.EnvReconciler
+	lastSeen  sync.Map // client.ObjectKey -> *fv1.Environment
 }
 
 func (r *environmentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	env := &fv1.Environment{}
 	if err := r.client.Get(ctx, req.NamespacedName, env); err != nil {
 		if apierrors.IsNotFound(err) {
+			// The informer cache reported NotFound. Before destructive cleanup
```
sanketsudake, line 47: **[scope — medium]** This stale-cache `IsNotFound` verify (uncached API read + 30s requeue before pool teardown) is a substantive env-reconciler behavior change, unrelated to RFC-0026. It was needed to stabilize the integration test, but adding a live API read on every environment `NotFound` belongs in its own PR so the data-plane change is reviewed and revertable on its own merits. (The added unit coverage for it is good.)

pkg/utils/podutils.go
```diff
@@ -57,3 +66,17 @@ func PodContainerReadyStatus(pod *v1.Pod) (readyContainers, noOfContainers int)
...
 	return
 }
+
+func IsPodRunning(pod *v1.Pod) bool {
+	return pod.Status.Phase == v1.PodRunning
+}
+
+func ReadyAndRunningPodsFilter(podList *v1.PodList) []v1.Pod {
+	output := []v1.Pod{}
+	for _, pod := range podList.Items {
```
sanketsudake: **[simplification — low]** `ReadyAndRunningPodsFilter` ranges `for _, pod := range podList.Items` then takes `&pod`, copying the whole (large) `corev1.Pod` struct each iteration just to take its address. Same pattern in `clearAllProvisionedLabels` (provisioner.go). Use index iteration: `for i := range podList.Items { pod := &podList.Items[i]; ... }`.

pkg/executor/executortype/poolmgr/reconciler.go
```diff
@@ -142,13 +142,33 @@ var poolmgrWarmPodPredicate = predicate.NewPredicateFuncs(func(obj client.Object
...
+		} else {
+			// Provisioned concurrency was removed (e.g. --provisioned-concurrency 0):
+			// clear labels on any previously-provisioned pods and reset status.
+			gpm.provisioner.disableProvisioning(ctx, fn)
```
sanketsudake, line 155: **[should-fix — fleet-wide side effects on startup]** This else-branch runs for **every** poolmgr function whose spec has nil `ProvisionedConcurrency` — not just ones that had it removed. On executor startup, `funcreconciler.lastReconciled` is empty, so every function reconciles with `old == nil` and lands here. On a cluster with N non-opted-in functions that means, per restart: N `UpdateStatus` calls stamping a spurious `Provisioned=False/ProvisionedDisabled` condition on functions that never used the feature, 2N per-function gauge series (`fission_provisioned_target/ready{...}=0`) — exactly the cardinality RFC-0019 discipline avoids — and N permanent `reconcileLocks` entries via `lockFor`.

`old` is right there; gate on the actual transition:

```suggestion
		} else if old != nil && old.Spec.ProvisionedConcurrency != nil {
			// Provisioned concurrency was removed (e.g. --provisioned-concurrency 0):
			// clear labels on any previously-provisioned pods and reset status.
			gpm.provisioner.disableProvisioning(ctx, fn)
		}
```

(The `old == nil` create case can't need a disable — a just-created function has no provisioned pods or status to clear. Executor-restart cleanup of a function whose PC was removed *while the executor was down* is already covered by the label-based `clearProvisionedLabels`, which no-ops when no labeled pods exist — and that edge doesn't justify stamping conditions on the whole fleet.)

pkg/executor/executortype/poolmgr/reconciler.go
```diff
@@ -142,13 +142,33 @@ var poolmgrWarmPodPredicate = predicate.NewPredicateFuncs(func(obj client.Object
...
 // DeleteFunction satisfies executortype.FuncReconciler: it marks the function's
 // fsCache entries deleted (so the reaper recycles its pods) and removes its istio
 // Service (when enabled) or its headless function Service (RFC-0002, when enabled).
 func (gpm *GenericPoolManager) DeleteFunction(ctx context.Context, fn *fv1.Function) error {
+	if gpm.provisioner != nil {
+		// Function is being deleted: clear provisioned labels so the reaper
+		// can recycle the pods. No status write — the Function object is
+		// being removed, so a status update would race with the delete and
+		// is pointless.
+		gpm.provisioner.clearProvisionedLabels(ctx, fn, -1)
```
sanketsudake, line 171: **[should-fix — slow leak, flagged by all three review lenses]** `DeleteFunction` clears the labels but never frees the per-function map entries: `reconcileLocks` has **no** deletion path anywhere, and `inflight` is only deleted in `disableProvisioningLocked` (target→0), which is not on the delete path. Under function churn (CI, ephemeral tenants) both maps grow monotonically for the executor's lifetime — and with the unconditional else-branch at :155, `reconcileLocks` tracks the *total* function population, not just PC users.

Fix here after clearing labels:

```go
gpm.provisioner.clearProvisionedLabels(ctx, fn, -1)
gpm.provisioner.reconcileLocks.Delete(fn.ObjectMeta.UID)
gpm.provisioner.inflight.Delete(fn.ObjectMeta.UID)
```

(or a `forget(fnUID)` helper on Provisioner). Taking the per-function lock first also closes the minor race with an in-flight `reconcileFunction` re-labeling pods during teardown.

pkg/apis/core/v1/validation.go
```diff
@@ -376,6 +375,10 @@ func (spec FunctionSpec) Validate() error {
 		}
 	}
 
+	if spec.ProvisionedConcurrency != nil {
```
sanketsudake: **[prior finding — still present]** `ProvisionedConcurrency.Validate()` runs **twice** on the full-validate path: once here, and again inside `validateForAdmission()` (called at :383, which has its own `spec.ProvisionedConcurrency != nil` block at :406). A rejected spec gets the same error string `errors.Join`ed twice in CLI/webhook output. Drop this direct call — `validateForAdmission()` already covers it on every path that reaches here.

pkg/executor/reaper/idle/idle_test.go
```diff
@@ -155,6 +155,46 @@ func TestPoolDeleteStrategy_Skips(t *testing.T) {
 		f.Atime = time.Now() // fresh
 		require.NoError(t, s.Reap(t.Context(), f))
 	})
+
+	t.Run("provisioned-concurrency function is skipped (PR1 function-level exemption)", func(t *testing.T) {
```
sanketsudake, line 159: **[test-effectiveness]** Neither subtest can fail if the exemption at `idle.go:262-266` is deleted: `Reap` returns nil on the regression path too (fake-client deletes don't error), so `require.NoError` passes either way — and the "reaped normally" subtest is explicitly `_ = s.Reap(...)` asserting nothing. This exemption is the only thing keeping the warm floor alive between provisioner ticks; if it silently regresses, provisioned pods get reaped every idleTimeout and unit CI never notices.

Cheap fix with the same harness: seed the strategy's `fscache.FunctionServiceCache` with the function's fsvc, then assert the cache entry **survives** `Reap` when PC is set and is **removed** when it isn't — both subtests become discriminating.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -0,0 +1,489 @@
...
+// cache, request channel, etc.). *GenericPoolManager satisfies this
+// interface implicitly.
+type funcSvcGetter interface {
+	GetFuncSvc(ctx context.Context, fn *fv1.Function) (*fscache.FuncSvc, error)
+}
+
+// ProvisionerConfig configures the RFC-0026 provisioner.
+type ProvisionerConfig struct {
+	// MaxPerFunction is the namespace-wide cap on ProvisionedConcurrencyConfig.Target.
+	// The webhook (or CEL, if we add a namespace-scoped cap later) enforces it;
```
sanketsudake: **[nit — stale comment]** The webhook can't enforce this: `MaxPerFunction` comes from an executor env var (`EXECUTOR_PROVISIONED_MAX_PER_FUNCTION`) the webhook never sees, and `ProvisionedConcurrencyConfig.Validate()` only checks `Target >= 1` + the Windows gate. The provisioner clamp below is the *only* enforcement — which is a coherent design (raw `ProvisionedSpecTarget` + `ProvisionedClamped` condition surface it well), but the comment claims defense-in-depth that doesn't exist. Reword to say the clamp here is the sole enforcement point.

test/integration/suites/serial/provisioned_concurrency_restart_test.go
```diff
@@ -0,0 +1,97 @@
...
+	t.Log("establishing provisioned floor before restart")
+	ns.WaitForProvisionedPodsAtLeast(t, ctx, fnName, 2, 5*time.Minute)
+	ns.WaitForProvisionedStatus(t, ctx, fnName, 2, 2, 5*time.Minute)
+
+	// Delete BOTH provisioned pods. Restart alone doesn't touch them
+	// (they're independent of the executor rollout), so deletion is what
+	// forces the fresh executor to re-derive.
+	t.Log("deleting both provisioned pods")
+	pods := ns.WaitForProvisionedPodsAtLeast(t, ctx, fnName, 2, 5*time.Minute)
+	require.Len(t, pods, 2, "expected exactly 2 provisioned pods to delete")
```
sanketsudake: **[test-robustness]** Two issues here:

1. `require.Len(t, pods, 2)` is an exact-count assert fed by an *at-least* waiter — a transient third provisioned pod (double-fire overshoot: the per-function lock releases before the async `GetFuncSvc` calls finish, so a second owner can over-fire before the next tick's excess-clear) hard-fails the test. Delete **all** returned pods instead of requiring exactly 2.

2. The test can pass without proving what it claims: pods are deleted *before* the restart, but the OLD executor's reconcile tick is 10s in CI, and the gap until SIGTERM lands is seconds — the old executor can re-warm the floor before dying, making the post-rollout waits pass even if a fresh executor never re-derives anything. A discriminating variant: scale the executor to 0, delete the provisioned pods while nothing is running, restore, then assert re-warm. Fine as a follow-up, but worth a comment noting the current version is best-effort.

test/integration/suites/common/provisioned_concurrency_test.go
```diff
@@ -0,0 +1,345 @@
...
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		count, err := ns.RunningFunctionPodCount(ctx, fnName)
+		if !assert.NoErrorf(c, err, "count specialized pods for %q", fnName) {
+			return
+		}
+		assert.Equalf(c, 1, count, "function %q: want 1 specialized pod (new-gen), got %d", fnName, count)
+	}, 90*time.Second, 5*time.Second)
+
+	onePod := ns.WaitForProvisionedPodsAtLeast(t, ctx, fnName, 1, 5*time.Minute)
+	require.Len(t, onePod, 1, "expected exactly 1 provisioned pod after target drop")
```
sanketsudake: **[flake risk]** `WaitForProvisionedPodsAtLeast(..., 1, ...)` followed by `require.Len(t, onePod, 1)` — an at-least waiter feeding an exactly-1 assert. A transient second provisioned pod (double-fire overshoot before the excess-clear tick) returns 2 and hard-fails Phase 4. The status convergence at :289 already proves 1/1 — either drop this exact assert or poll for exact convergence inside an `EventuallyWithT`. Also: the 90s window for old-gen recycle + re-warm at :292-298 is the tightest in the file; Phase 2's own comment argues for multi-minute windows under `-parallel 6` CI load — same reasoning applies here.

sanketsudake (commented):
## Re-review after fixes — verification + fresh pass

Re-ran the review battery on head `953040f9` (fix-verification, fresh bug hunt on the fix commits, test-quality), deduplicated, and verified locally (build ✓, `go test -race` on all touched packages ✓, `pkg/crd` diff vs main now **empty** after the #3596 rebase, codegen/CRD/chart/CLI plumbing coherent, CEL rules match the capitalized JSON tags and treat empty `ExecutorType` as poolmgr).

**Of the 17 substantive findings from the first review: 14 verified fully fixed, 1 not fixed, 2 partial.** The fixes are high quality — status now written on every pass, conflict retry re-reads via the direct client, metrics genuinely instrumented (not just declared), the disable paths collapsed, the reaper-exemption cost honestly documented as a PR1 LIMITATION with tests, and `provisioner_test.go` is a model harness (label effects observed through fakes, CAS-tracked peak-concurrency gauge for the pacing invariant).

### Not fixed / new — should address before merge
1. **Startup status-write storm on every non-opted-in function** (inline @ reconciler.go:155) — the biggest new finding, introduced by the disable-path fix. `ReconcileFunction`'s else-branch calls `disableProvisioning` for **every** function with nil `ProvisionedConcurrency`. On executor start, `lastReconciled` is empty so every poolmgr function reconciles with `old == nil` → every function in the cluster gets a spurious `Provisioned=False/ProvisionedDisabled` condition stamped, two per-function OTel gauge series, and a permanent `reconcileLocks` entry. Gate on the actual transition.
2. **`reconcileLocks`/`inflight` map entries never freed on function delete** (inline @ reconciler.go:170) — flagged independently by all three review lenses. Slow unbounded growth under function churn; compounded by #1 (an entry per every function ever seen, not just PC users).
3. **Prior finding not fixed: `ProvisionedConcurrency.Validate()` still runs twice** (inline @ validation.go:378) on the full-validate path — the direct call plus `validateForAdmission()`'s call, so a rejected spec shows the duplicated error string.
4. **Both new reaper-exemption tests pass even with the exemption deleted** (inline @ idle_test.go:159) — the floor's only protection between ticks is unguarded by unit CI.

### Consider (test robustness)
5. **Serial restart test can pass without proving re-derivation** (inline @ restart test:79) — the old executor's 10s tick can re-warm the floor before SIGTERM lands.
6. **At-least waiter feeding an exact-count assert** (inline @ lifecycle test:301) — flake under transient double-fire overshoot (the per-function lock releases before async `GetFuncSvc` calls finish, so a second owner can over-fire; bounded by MaxInflight and healed next tick, but it makes exact-count asserts race-prone). Same pattern at restart test:79.
7. **No composed clamp test through `reconcileFunction`**: `TestProvisioner_effectiveTarget` covers the clamp and the status subtests cover the condition, but nothing drives target > cap end-to-end — swapping `target`/`specTarget` at the `fireEagerSpecializations` call site (over-provisioning 25 pods) would keep every existing test green. One `fakeGPM` test closes it.
8. **`ProvisionedConcurrencyConfig.Validate()` has zero unit tests** — Target<1 and the Windows gate are CEL/integration-only; a small table test guards the CLI-side check.
9. **No regression test for the per-function lock** (2a4b0731) — the two-owner race fix is correct (verified: both paths take it, panic-safe, no unlock mismatch) but unpinned; deleting `lockFor` leaves the suite green under `-race`.

### Nits
10. Stale comment (inline @ provisioner.go:47): claims the webhook enforces `MaxPerFunction` — it can't see the executor env var; the provisioner clamp is the *only* enforcement (which is fine and well surfaced via `ProvisionedSpecTarget` + `ProvisionedClamped`).
11. `reconciler.go:261` still says "defaults are applied in NewProvisioner" (they live in `ProvisionerConfigFromEnv`), plus leftover "Step 9" design-doc references.
12. `AtoiOr`/`DurOr` silently fall back to defaults on unparseable `EXECUTOR_PROVISIONED_MAX_*`/`_INTERVAL` values — the same footgun class the `ParseBool` fix addressed, one `console`/log line each.

Overall: the feature core is solid and converging; #1/#2 are the only things I'd hold the merge for (both are small, localized fixes), #3/#4 are cheap to fix alongside.

sanketsudake (approved):

PR #3589, opened 2026-07-17
fix(cli): fission fn test uses router internal listener + HMAC signing (#3588)

### LINKED ISSUE #3588: fission function test is broken after router public/internal listener split

<pre>
client:
  fission/core:
    BuildDate: "2026-06-22T12:02:31Z"
    GitCommit: 1e1401cd
    Version: v1.27.0
server:
  fission/core:
    BuildDate: "2026-06-22T12:02:31Z"
    GitCommit: 1e1401cd
    Version: **v1.27.0**
</pre>

OpenShift/Kubernetes.

Fission is installed in the `fission` namespace. Functions are deployed in the `fission-functions` namespace.

**Describe the bug**

`fission function test` consistently returns `404 page not found` on Fission v1.27.0, including for a minimal valid Python function.

The function itself works correctly when invoked through an HTTPTrigger on the public router. The failure is isolated to the direct invocation path used by `fission function test`.

Verbose output shows that the CLI port-forwards to the public router listener on port `8888`, then calls:

<pre> /fission-function/&lt;namespace&gt;/&lt;function&gt; </pre>

That direct invocation endpoint is no longer served by the public listener. The router therefore returns `404` before the request reaches the executor.

The installed router services appear to be configured as expected:

<pre> $ oc -n fission get svc router router-internal \ -o custom-columns='NAME:.metadata.name,PORT:.spec.ports[*].port,TARGET:.spec.ports[*].targetPort' NAME PORT TARGET router 80 8888 router-internal 8889 8889 </pre>

**To Reproduce**

Create a minimal Python function:

<pre> $ cat &gt; hello_min.py &lt;&lt;'PY' def main(): return "Hello from Fission!" PY </pre>

Create the function:

<pre> $ fission function create \ --name hello-min \ --env python \ --code hello_min.py \ --namespace fission-functions Package 'hello-min-...' created function 'hello-min' created </pre>

Test it using `fission function test`:

<pre> $ fission function test \ --name hello-min \ --namespace fission-functions \ --verbosity 2 </pre>

The verbose output shows:

<pre> Setting up port forward to application=fission-router in namespace fission Starting port forward from local port 42817 Connecting to port 8888 on pod fission/router-c7695678c-ln6qh Forwarding from 127.0.0.1:42817 -&gt; 8888 Function test url: http://127.0.0.1:42817/fission-function/fission-functions/hello-min GET /fission-function/fission-functions/hello-min HTTP/1.1 </pre>

The response is:

<pre> HTTP/1.1 404 Not Found Content-Type: text/plain; charset=utf-8 404 page not found </pre>

To confirm that the function itself works, create an HTTPTrigger:

<pre> $ fission route create \ --name hello-min-route \ --function hello-min \ --url /hello-min \ --method GET \ --namespace fission-functions </pre>

Port-forward the public router:

<pre> $ oc -n fission port-forward svc/router 8888:80 </pre>

Invoke the function through the HTTPTrigger:

<pre> $ curl http://127.0.0.1:8888/hello-min Hello from Fission! </pre>

This confirms that:

- The Python environment works.
- The package is valid.
- Function specialization works.
- The executor works.
- The public router works.
- The HTTPTrigger works.
- The failure is specific to `fission function test`.

**Expected result**

The following command should invoke the function successfully:

<pre> $ fission function test \ --name hello-min \ --namespace fission-functions </pre>

Expected output:

<pre> Hello from Fission! </pre>

The CLI should use a supported invocation mechanism, for example:

1. Use the internal router listener on `router-internal:8889` and apply the required internal authentication/signature.
2. Create and clean up a temporary HTTPTrigger.
3. Use another supported direct invocation mechanism.

If direct invocation is no longer intended to be supported, the command should be deprecated or removed, and the replacement procedure should be clearly documented.

**Actual result**

`fission function test` port-forwards to the public router listener on port `8888`:

<pre> Connecting to port 8888 on pod fission/router-c7695678c-ln6qh </pre>

It then calls the direct invocation endpoint:

<pre> /fission-function/fission-functions/hello-min </pre>

The public router responds with:

<pre> HTTP/1.1 404 Not Found 404 page not found </pre>

The request does not reach the executor, and the CLI subsequently reports:

<pre> Error: getting function logs: no active pods found for function in namespace fission-functions Error: error getting function response </pre>

**Screenshots/Dump file**

Relevant verbose CLI output is included above.

**Additional context**

The router service layout appears correct:

<pre> router 80 -&gt; 8888 router-internal 8889 -&gt; 8889 </pre>

The problem appears to be caused by the CLI selecting the public router while constructing an internal direct-invocation URL.

The Helm chart assigns the shared label:

<pre> application: fission-router </pre>

to both `router` and `router-internal`.

The port-forward implementation appears to select the first service returned for that label. In this installation, the selected service is `router`, so the CLI forwards to port `8888`.

The resulting combination is:

<pre> Internal endpoint: /fission-function/&lt;namespace&gt;/&lt;function&gt; Public listener: 8888 Result: 404 </pre>

The router public/internal listener split was introduced intentionally so that direct function invocation would no longer be exposed through the public listener.

It also appears that upgrade or integration coverage which previously used `fission function test` was changed to use an HTTPTrigger and `curl` after the public direct-invocation path began returning `404`.

That change is understandable for validating the public router, but it may also have removed coverage for the documented `fission function test` command without fixing, deprecating, or documenting the broken behavior.

Thanks.

rijojohn85:
## Summary

Fixes #3588.

`fission fn test` (sync path) returned `404 page not found` on v1.27.0 because it port-forwarded to the router **public** listener (port 8888) and requested `/fission-function/<ns>/<name>` — a path that, after the GHSA-3g33-6vg6-27m8 listener split, is served **only** by the internal listener (port 8889, `svc/router-internal`). The `--async` path already used the internal listener + HMAC signing and worked; the sync path was left behind.

### Root cause

`pkg/fission-cli/cmd/function/test.go` `do()` called `util.GetRouterURL` (public listener, 8888), then constructed a `/fission-function/<ns>/<name>` URL. Post-split, the public mux has no such route → `404` before the executor is touched. The `--async` path already did the right thing: `util.GetRouterInternalURL` (8889) + HMAC signing via `hmacauth.NewServiceSigningTransport(..., ServiceRouterInternal, ..., "/fission-function/")`.

### The fix

Both `fission fn test` (sync + `--async`) and `fission fn run` (local invoke) now share a single **`combinedHTTPRequest`** helper (`pkg/fission-cli/cmd/function/test.go`), which replaced the original `DoHTTPRequest` and an intermediate `buildInternalRequest` from earlier in this PR's history:

1. `do()` resolves the router **internal** listener URL once (`util.GetRouterInternalURL` → port-forward to 8889, or `FISSION_ROUTER_INTERNAL_URL` override) — under `input.Context()`, not the per-invocation `--timeout` context (see review follow-up below).
2. Builds `/fission-function/<ns>/<name>` via `utils.UrlForFunction` (folds default ns), appends `--subpath`, applies `--query`.
3. `combinedHTTPRequest` builds and sends the request: inits the OTel tracer provider + starts a span, owns the whole span/shutdown lifecycle internally (one `defer`, no caller cleanup contract), applies user `--header`s, then sets `X-Fission-Invoke-Mode` **LAST** when set, so `--async` stays authoritative over a user `-H "X-Fission-Invoke-Mode: sync"`.
4. Wraps the transport with HMAC signing (`ServiceRouterInternal` key) when `SignWithHMAC` is requested **and** `FISSION_INTERNAL_AUTH_SECRET` is set; empty secret = pass-through (matches `internalAuth.enabled=false`). `fn test` requests signing; `fn run`'s local invoke does not (it talks to a locally-specialized pod, not the router, so it attaches the bearer token instead).
5. Dumps request/response at `--verbosity 2` (`-vv`), matching the pre-refactor sync path.

An `invokeOptions` struct (`Method`, `URL`, `Body`, `Headers`, `SignWithHMAC`, `AttachBearerToken`, `InvokeModeHeader`) is the single contract both callers configure. `fn run`'s `invokeLocal` now calls `combinedHTTPRequest` directly instead of the old `DoHTTPRequest`.

### Review follow-up (this update)

Addressed the review round from @sanketsudake on top of the original fix:

- **401/403 now give an actionable hint on the sync path** — previously fell through to `renderInvocationFailure` → `util.FunctionPodLogs` → the same misleading `no active pods found` from the original bug report. Both status codes now return one merged message: `router rejected the request (...); set FISSION_INTERNAL_AUTH_SECRET when authentication is enabled` (the HMAC verifier only ever returns 401 for any auth failure — 403 only ever comes from unrelated CORS preflight rejection, so a single message covers both correctly).
- **`-vv`/`--verbosity 2` request/response dump restored** on the sync path (regressed when the sync and async invoke paths briefly forked).
- **Bearer token (`FISSION_AUTH_TOKEN`) forwarding is now explicit and intentional** via `invokeOptions.AttachBearerToken` — `fn test` doesn't set it (the internal listener authenticates via HMAC, not the JWT gateway); `fn run`'s local invoke does.
- **Duplication eliminated**: the two near-parallel request builders (`DoHTTPRequest`, `buildInternalRequest`) that had drifted apart — causing the bearer-token and verbose-dump regressions above — are now one `combinedHTTPRequest`, used by both `fn test` and `fn run`.
- **Router-internal URL resolution no longer bounded by the request timeout.** `util.GetRouterInternalURL` (and the port-forward setup it can trigger — K8s API calls + SPDY tunnel) now runs under `input.Context()`, not the `--timeout`/`FunctionTimeout`-bounded context, matching pre-refactor `main` behavior. Previously a short `--timeout` also had to cover cold port-forward setup latency, which had nothing to do with the function itself.
- **A bad `OTEL_*` env no longer blocks `fn test`.** `otelUtils.InitProvider` failure now warns and proceeds (tracing degrades to the OTel SDK's no-op provider) instead of hard-failing before the request is ever sent.
- **Fixed `%d` on a `time.Duration`** in the timeout-exceeded error message (was printing raw nanoseconds); now `%s` via `Duration.String()`.
- **Stale doc comments fixed**: `combinedHTTPRequest` now has an accurate doc comment (it's genuinely shared by `fn test` and `fn run`'s local invoke); the "CLI subprocess" comments in the integration test file were wrong — the CLI runs in-process (`cliMu.Lock()` + real `os.Setenv` + an `os.Stdout` pipe swap), never as a subprocess.
- **Test coverage gaps closed**: `do()`'s generic non-2xx failure path (500 → `renderInvocationFailure` + pod-log fallback), both `--subpath` leading-slash branches, `--timeout` correctly overriding a larger function-spec timeout, and a lock-in that `GetRouterInternalURL` never falls back to the public `FISSION_ROUTER_URL` var.

### Behaviour change for users

- `fission fn test` now honours `FISSION_ROUTER_INTERNAL_URL` (not `FISSION_ROUTER_URL`) for the direct-invoke override — same as `--async` already did.
- When internal auth is enabled (`internalAuth.enabled=true`, chart default), operators must export `FISSION_INTERNAL_AUTH_SECRET` before running `fission fn test`, or the router returns `401`/`403` with an actionable error naming the env var:
  ```bash
  export FISSION_INTERNAL_AUTH_SECRET="$(kubectl get secret fission-internal-auth -n fission -o jsonpath='{.data.secret}' | base64 -d)"
  fission fn test --name my-fn
  ```
  Note: the secret key is `secret`, not `master` (per `charts/fission-all/templates/_helpers.tpl` `internalAuth.envs`).
- When internal auth is disabled (pass-through mode), no env var is needed.

## Files changed

| File | Change |
|------|--------|
| `pkg/fission-cli/cmd/function/test.go` | `combinedHTTPRequest` (replaces `DoHTTPRequest`/`buildInternalRequest`); sync `do()` and `handleAsyncResponse` both use it; 401/403 hint; timeout-scoping fix; OTel warn-and-proceed; `%d`→`%s` duration fix; doc comment |
| `pkg/fission-cli/cmd/function/run.go` | `invokeLocal` migrated onto `combinedHTTPRequest` |
| `pkg/fission-cli/cmd/function/run_local.go` | comment updated (`DoHTTPRequest` → `combinedHTTPRequest`) |
| `pkg/fission-cli/cmd/function/run_test.go` | comment updated |
| `pkg/fission-cli/cmd/function/test_async_test.go` | rewritten for `combinedHTTPRequest`/`handleAsyncResponse` (request building, HMAC pass-through regression guard, bearer token, malformed headers, invalid method, async response handling) |
| `pkg/fission-cli/cmd/function/test_test.go` | new: 401/403 hint, async-dispatch regression guard, generic failure path, subpath handling, `--timeout` selection |
| `pkg/fission-cli/util/util_more_test.go` | new: `GetRouterInternalURL` never falls back to `FISSION_ROUTER_URL` |
| `test/integration/framework/cli.go` | `CLICaptureStdoutWithEnv` — combines `os.Stdout` capture with env-var overrides |
| `test/integration/suites/common/function_test_test.go` | `TestFunctionTestCLI` (sync) + `TestFunctionTestCLIAsync` (async); stale "CLI subprocess" comments fixed |
| `docs/internal-auth/00-design.md` | New "CLI direct invocation" section + CLI added to router-internal callers row |
| `docs/rfc/0017-function-developer-debugging-toolkit.md` | Invoke-path note in Phase 2 enrichment section |
| `docs/test-migration/01-migration-status.md` | `TestFunctionTestCLI` coverage note |

## Why this approach (not the alternatives)

The issue suggested three options:
1. **Use the internal listener + HMAC signing** (this PR) — matches the post-split contract, no cluster changes, `--async` already proved the pattern works.
2. **Create + clean up a temp HTTPTrigger** — heavier, mutates cluster state, flaky.
3. **Deprecate `function test`** — bad UX for a documented command.

Option 1 is the minimal, correct fix. The shared `combinedHTTPRequest` helper also eliminates the duplication that caused both the original regression and a second round of drift (bearer token, verbose dump) during review.

### Unit tests (no cluster)

```
$ go test -race -v ./pkg/fission-cli/cmd/function/... ./pkg/fission-cli/util/...

--- PASS: TestCombinedHTTPRequestForwardsRequest (0.00s)
--- PASS: TestCombinedHTTPRequestInvokeModeSetLast (0.00s)
--- PASS: TestCombinedHTTPRequestHMACSigning (0.00s)
    --- PASS: TestCombinedHTTPRequestHMACSigning/signed_when_requested_and_secret_set
    --- PASS: TestCombinedHTTPRequestHMACSigning/pass-through_when_requested_but_no_secret
    --- PASS: TestCombinedHTTPRequestHMACSigning/never_signed_when_not_requested
--- PASS: TestCombinedHTTPRequestAttachesBearerToken (0.00s)
    --- PASS: TestCombinedHTTPRequestAttachesBearerToken/attached_when_requested
    --- PASS: TestCombinedHTTPRequestAttachesBearerToken/omitted_when_not_requested
--- PASS: TestCombinedHTTPRequestMalformedHeaderErrors (0.00s)
--- PASS: TestCombinedHTTPRequestInvalidMethod (0.00s)
--- PASS: TestHandleAsyncResponseAccepted (0.00s)
    --- PASS: TestHandleAsyncResponseAccepted/invocation_id_from_header
    --- PASS: TestHandleAsyncResponseAccepted/invocation_id_falls_back_to_JSON_body
    --- PASS: TestHandleAsyncResponseAccepted/no_id_anywhere_still_succeeds_with_a_warning
--- PASS: TestHandleAsyncResponseErrorStatuses (0.00s)
    --- PASS: TestHandleAsyncResponseErrorStatuses/unauthorized_401
    --- PASS: TestHandleAsyncResponseErrorStatuses/forbidden_403
    --- PASS: TestHandleAsyncResponseErrorStatuses/server_500
    --- PASS: TestHandleAsyncResponseErrorStatuses/disabled_501
--- PASS: TestDoSyncUnauthorized (0.00s)
--- PASS: TestDoAsyncDispatch (0.00s)
--- PASS: TestDoSyncGenericFailure (0.00s)
--- PASS: TestDoSubPathHandling (0.00s)
    --- PASS: TestDoSubPathHandling/without_leading_slash
    --- PASS: TestDoSubPathHandling/with_leading_slash
--- PASS: TestDoSyncSmallerTestTimeoutGoverns (0.00s)
--- PASS: TestGetRouterInternalURLIgnoresPublicRouterVar (0.54s)
...
ok      github.com/fission/fission/pkg/fission-cli/cmd/function  5.419s
ok      github.com/fission/fission/pkg/fission-cli/util          3.147s
```
(Trimmed to the tests relevant to this PR; the full package suites — 100+ tests — pass with `-race`.)

New/changed tests of note:
- `TestCombinedHTTPRequestHMACSigning` — table-driven: signed when requested+secret set, **pass-through (not an error) when requested but no secret** (regression guard — this path briefly hard-failed during the refactor), never signed when not requested.
- `TestCombinedHTTPRequestInvokeModeSetLast` — a user `-H "X-Fission-Invoke-Mode: sync"` is overwritten by `--async` (guards the "set LAST" rule).
- `TestDoAsyncDispatch` — end-to-end regression guard for a bug caught mid-refactor: `do()`'s `--async` dispatch was briefly dropped, which would have made async invocations silently print the raw 202 body instead of the invocation id. Existing unit tests calling the invoke helper directly didn't catch this; only a test driving `do()` itself does.
- `TestDoSyncGenericFailure` / `TestDoSubPathHandling` / `TestDoSyncSmallerTestTimeoutGoverns` / `TestGetRouterInternalURLIgnoresPublicRouterVar` — close the coverage gaps called out in review (see "Review follow-up" above). The timeout test uses a 1ns deadline instead of a real sleep so it can't flake under CI load.

### Integration tests (kind cluster, internal auth enabled)

```
$ NODE_RUNTIME_IMAGE=ghcr.io/fission/node-env \
  FISSION_INTERNAL_AUTH_SECRET=$(kubectl get secret fission-internal-auth -n fission -o jsonpath='{.data.secret}' | base64 -d) \
  go test -tags=integration -run TestFunctionTestCLI -v ./test/integration/suites/common/...

=== RUN   TestFunctionTestCLI
=== PAUSE TestFunctionTestCLI
=== RUN   TestFunctionTestCLIAsync
=== PAUSE TestFunctionTestCLIAsync
=== CONT  TestFunctionTestCLI
=== CONT  TestFunctionTestCLIAsync
...
Request ID: bfc4665b-829b-44aa-a99d-513532b8a6be
--- PASS: TestFunctionTestCLI (17.78s)
--- PASS: TestFunctionTestCLIAsync (17.91s)
PASS
ok      github.com/fission/fission/test/integration/suites/common       19.285s
```

- `TestFunctionTest` — existing; hits `/fission-function/<fn>` directly via framework's signed HTTP client (route check).
- `TestFunctionTestCLI` — drives real `fission fn test` CLI in-process, forwards `FISSION_INTERNAL_AUTH_SECRET` from framework to CLI env, asserts `Hello, Fission` in output.
- `TestFunctionTestCLIAsync` — drives real `fission fn test --async` CLI in-process, asserts `Accepted (202)` + `invocationId:` in output. Skips when `ASYNC_INVOCATION_ENABLED != true`.

### Manual end-to-end verification (real kind cluster + Docker, using the built `fission` binary from this branch)

- `fission fn test --name <fn>` with `FISSION_INTERNAL_AUTH_SECRET` unset (internal auth on): returns the actionable `set FISSION_INTERNAL_AUTH_SECRET` error instead of the original misleading `no active pods found`.
- `fission fn test --name <fn> -v 2`: full request/response dump printed, confirming the `-vv` regression (review finding #2) is fixed.
- `fission fn test --name <fn> --async`: `Accepted (202)` + `invocationId: ...` printed, confirming the async dispatch path (and the mid-refactor regression it guards against) works end-to-end.
- `fission fn run-local --name hello-local --env python --code /tmp/hello.py`: pulled `ghcr.io/fission/python-env`, specialized, and returned `Hello, World!` — confirms the `invokeLocal` migration onto `combinedHTTPRequest` works over Docker with no router/HMAC involved.

### Lint / build

- `go build ./...` — clean
- `go vet ./pkg/fission-cli/...` — clean
- `gofmt -l` — clean
- `go test -race ./pkg/fission-cli/...` — all packages pass

#### Test plan

- [x] Unit tests pass (`go test -race ./pkg/fission-cli/...`)
- [x] Integration tests pass on kind with internal auth on (`TestFunctionTest`, `TestFunctionTestCLI`, `TestFunctionTestCLIAsync`)
- [x] Manual verification on a real kind cluster: missing-secret error, `-v 2` dump, `--async` dispatch
- [x] Manual verification of `fission fn run-local` over Docker
- [x] `go vet` / `gofmt` clean
- [ ] CI on this PR (all legs, including "Pin legacy data plane")


pkg/fission-cli/cmd/function/test.go
```diff
@@ -135,13 +99,23 @@ func (opts *TestSubCommand) do(input cli.Input) error {
...
+	// HMAC signing is applied when FISSION_INTERNAL_AUTH_SECRET is set; empty
+	// secret = pass-through (matches the chart's internalAuth.enabled=false).
+	req, hc, cleanup, err := opts.buildInternalRequest(ctx, input, m, method, "")
 	if err != nil {
 		return err
 	}
+	defer cleanup()
+	console.Verbose(2, "Function test url: %v", req.URL.String())
+
+	resp, err := hc.Do(req)
```
sanketsudake: **[should-fix — this is what decides whether #3588 is fully fixed]** When internal auth is enabled (chart default) and the operator hasn't exported `FISSION_INTERNAL_AUTH_SECRET`, this `hc.Do` gets 401/403 → the code below falls into `renderInvocationFailure` + `util.FunctionPodLogs`, which prints **`no active pods found`** — the exact misleading output from #3588. `invokeAsync` already handles this with an actionable message; the sync path should mirror it *before* the generic failure rendering:

```go
if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
	return fmt.Errorf("router rejected the request (%s); set FISSION_INTERNAL_AUTH_SECRET "+
		"when internal auth is enabled", resp.Status)
}
```

Otherwise a user on the default chart still hits a confusing failure and has to guess. (Nice-to-have beyond this PR: the CLI already has cluster access and could read the `fission-internal-auth` secret itself so nothing needs exporting — RBAC permitting.)

pkg/fission-cli/cmd/function/test.go
```diff
@@ -135,13 +99,23 @@ func (opts *TestSubCommand) do(input cli.Input) error {
...
+	// svc/router-internal) for /fission-function/<ns>/<name> — the public
+	// listener (8888) no longer serves that path after GHSA-3g33-6vg6-27m8.
+	// HMAC signing is applied when FISSION_INTERNAL_AUTH_SECRET is set; empty
+	// secret = pass-through (matches the chart's internalAuth.enabled=false).
+	req, hc, cleanup, err := opts.buildInternalRequest(ctx, input, m, method, "")
 	if err != nil {
 		return err
 	}
+	defer cleanup()
+	console.Verbose(2, "Function test url: %v", req.URL.String())
```
sanketsudake: **[regression]** `fission fn test -vv` used to dump the full request **and** response here. The old sync path went through `DoHTTPRequest`, which at `console.Verbosity >= 2` calls `httputil.DumpRequestOut(req, false)` before the call and `httputil.DumpResponse(resp, false)` after. This path logs only the URL, so the verbose dump is gone — while `fission fn run` still has it. Since this is a debugging command (RFC-0017) and #3588 was diagnosed *with* `--verbosity 2`, please restore it: dump the request here and the response after `hc.Do`, both guarded by `console.Verbosity >= 2`. (`net/http/httputil` is already imported.)

pkg/fission-cli/cmd/function/test.go
```diff
@@ -203,24 +190,62 @@ func (opts *TestSubCommand) invokeAsync(ctx context.Context, input cli.Input, m
 		fnURL.RawQuery = q.Encode()
 	}
 
-	req, err := http.NewRequestWithContext(ctx, method, fnURL.String(), strings.NewReader(input.String(flagkey.FnTestBody)))
+	// OTel: init provider + start a span before building the request so the
+	// span context is injected into the outgoing request via otelhttp.
+	shutdown, err := otelUtils.InitProvider(ctx, logr.Logger{}, "fission-cli")
```
sanketsudake: Two things on this block:

1. **Tracing setup hard-gates the invoke.** A failing `InitProvider` (e.g. a malformed `OTEL_*` endpoint) aborts `fission fn test` before the request is even attempted, and the error is returned unwrapped (`return nil, nil, nil, err`) so the user sees an opaque tracing error with no hint of its origin. Tracing is auxiliary — prefer warn-and-continue (span against the global no-op tracer), or at minimum wrap: `fmt.Errorf("initializing tracing: %w", err)`.

2. **Timeout scoping.** `ctx` here is the request-timeout context, so both this init and `GetRouterInternalURL` (line 175) now run under the function timeout. `main` resolved the URL under `input.Context()`. With a short `--timeout` the port-forward setup can now time out during resolution. Consider resolving/initing under `input.Context()` and applying the deadline only to the request.

pkg/fission-cli/cmd/function/test.go
```diff
@@ -203,24 +190,62 @@ func (opts *TestSubCommand) invokeAsync(ctx context.Context, input cli.Input, m
...
+	req, err := http.NewRequestWithContext(spanCtx, method, fnURL.String(), strings.NewReader(input.String(flagkey.FnTestBody)))
+	if err != nil {
+		span.End()
+		if shutdown != nil {
+			shutdown(ctx)
+		}
+		return nil, nil, nil, err
 	}
+
 	for _, h := range input.StringSlice(flagkey.FnTestHeader) {
```
sanketsudake: **[nit — silent skip]** `strings.Cut` silently drops a malformed `--header` (one with no `:`), whereas `DoHTTPRequest` *errors* on it (`len(headerKeyValue) != 2`). So `fission fn test -H "Authorization"` (typo, missing value) is swallowed here but rejected on the `run` path — same flag, two contracts. Consider erroring or `console.Warn`-ing on a header with no `:` so a typo isn't silently ignored.

pkg/fission-cli/cmd/function/test.go
```diff
@@ -203,24 +190,62 @@ func (opts *TestSubCommand) invokeAsync(ctx context.Context, input cli.Input, m
...
+// durable invocation id, which is printed; the response body is not awaited.
+func (opts *TestSubCommand) invokeAsync(ctx context.Context, input cli.Input, m *metav1.ObjectMeta, method string) error {
+	req, hc, cleanup, err := opts.buildInternalRequest(ctx, input, m, method, asyncinvoke.InvokeModeAsync)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+	resp, err := hc.Do(req)
 	if err != nil {
 		return err
```
sanketsudake: **[nit — parity]** The sync path wraps this same error as `fmt.Errorf("error executing HTTP request: %w", err)`; wrap here too so an async dial/HMAC failure isn't an unlabelled bare error.

```suggestion
		return fmt.Errorf("error executing HTTP request: %w", err)
```

test/integration/suites/common/function_test_test.go
```diff
@@ -53,3 +55,122 @@ func TestFunctionTest(t *testing.T) {
...
+	validFn := "fnt-cli-" + ns.ID
+
+	ns.CreateEnv(t, ctx, framework.EnvOptions{Name: envName, Image: runtime})
+
+	validCode := framework.WriteTestData(t, "nodejs/fn_test/hello.js")
+	ns.CreateFunction(t, ctx, framework.FunctionOptions{
+		Name: validFn, Env: envName, Code: validCode,
+	})
+
+	// Forward the internal auth secret to the CLI subprocess env so the
```
sanketsudake: **[nit — inaccurate comment]** The CLI runs **in-process** (`app.App(...)` under `cliMu.Lock()`, which is exactly why `os.Stdout` has to be captured) — there is no subprocess. Same wording appears at lines 67 and 154.

```suggestion
	// Forward the internal auth secret to the in-process CLI's env so the
```

sanketsudake (commented):
## Review — full battery (code-review · simplify · thermo-nuclear · deslop · silent-failure · test-coverage)

Ran the diff through several review lenses and verified locally (`go test -race ./pkg/fission-cli/cmd/function/...` passes). **The core fix is correct**: both sync `do()` and `invokeAsync()` now route `/fission-function/<ns>/<name>` through the internal listener (8889) with HMAC signing via the shared `buildInternalRequest` helper — exactly option 1 from #3588 — and `TestFunctionTestCLI` is a genuine end-to-end regression guard (it would 404 on the pre-fix public-listener path). `utils.UrlForFunction`, `%w` wrapping and the "invoke-mode header set LAST" invariant are all preserved and pinned by tests. Nice work.

Findings below, **deduplicated** across all lenses. Nothing is a hard blocker, but **#1 decides whether #3588 is *fully* fixed for the reporter's setup** (internal auth is on by chart default).

### Should-fix
1. **Sync path gives no `FISSION_INTERNAL_AUTH_SECRET` hint on 401/403 — it reproduces the original confusing failure.** (inline) The reporter's cluster has internal auth enabled (chart default). With this fix a user who hasn't exported the secret now gets a 401/403 that falls through to `renderInvocationFailure` → `util.FunctionPodLogs` → **`no active pods found`** — the *same* misleading output from the bug report. `invokeAsync` already special-cases 401/403 with the actionable message; the sync path should mirror it. (Stronger option: the CLI already has cluster access and could read the `fission-internal-auth` secret itself so nothing needs exporting — RBAC permitting.)
2. **`fission fn test -vv` no longer dumps the request/response.** (inline @ :113) The old sync path went through `DoHTTPRequest`, which dumped both at verbosity ≥ 2; the new path logs only the URL. This is a debugging-toolkit regression (RFC-0017), diverges from `fission fn run` (which still dumps), and #3588 itself was diagnosed *with* `--verbosity 2`.

### Consider
3. **Two near-parallel invoke builders now coexist** — `DoHTTPRequest` and `buildInternalRequest` share ~60% scaffolding (OTel init/span/transport/header parsing). That fork is what silently dropped the bearer token (#5) and the verbose dumps (#2) from the sync path, and it gives `--header` two different contracts (see inline @ :211). A single options-driven helper (URL, optional HMAC, optional invoke-mode, optional bearer, verbose) would dissolve #2/#4/#5 and the header divergence at once. Optional for a bugfix — worth a follow-up.
4. **`buildInternalRequest` returns a 4-tuple with a "callers MUST defer cleanup()" prose contract** and a name that undersells what it does (URL resolution + global OTel init + span + client + teardown). A callback form (`invokeInternal(ctx, …, func(*http.Response) error)`) would own the whole span lifecycle in one `defer`. Optional.
5. **Bearer `FISSION_AUTH_TOKEN` is no longer forwarded on the sync path.** This is *intentional* (the internal listener authenticates via HMAC, not the JWT gateway), but it's silent — worth a one-line code note + a mention in the PR/help for anyone whose function reads `Authorization`.

### Nits
6. **Timeout scoping change** (inline @ :195): `buildInternalRequest` resolves the internal URL + inits OTel under the request-timeout `ctx`; `main` used `input.Context()`. A short `--timeout` now also bounds port-forward setup.
7. **`InitProvider` failure hard-gates the whole test** and is returned unwrapped (inline @ :195) — a bad `OTEL_*` env blocks `fission fn test` before the request. Warn + proceed, or at least wrap.
8. **Pre-existing `%d` on a `time.Duration`** at `test.go:85` (`"function request timeout (%d)s exceeded"`) prints nanoseconds — and this very PR fixed the analogous `%d` in the warning 10 lines up. Use `%s` and drop the trailing literal `s`.
9. **Stale doc comment** on `DoHTTPRequest` (~:320): "Shared by `function test` and `function run` … byte-identical" — after this PR only `function run` uses it.
10. **"CLI subprocess" comments** (inline) — the CLI runs in-process under `cliMu.Lock()` (that's why `os.Stdout` is captured); there's no subprocess.
11. **Error-wrapping parity** (inline @ :250): `invokeAsync`'s `hc.Do` error is bare while sync wraps it.

### Test coverage (solid, with gaps)
- The unit tests exercise `buildInternalRequest` directly, so they *characterize the helper* but don't guard the actual regression site (`do()`'s sync dispatch) — a revert of that one line would pass every unit test; only the cluster-gated `TestFunctionTestCLI` catches it. Cheap lock-in: assert `buildInternalRequest` **errors** when only `FISSION_ROUTER_URL` is set (proves sync never falls back to the public var).
- **Sync error/failure path is untested** (non-2xx → `renderInvocationFailure` + log fallback), including the 401/403 case from #1 — add a sync test with an httptest server returning 401.
- **Subpath-with-leading-slash** branch and the `--timeout` selection are untested (one table row each).

sanketsudake (approved):

PR #3593, opened 2026-07-20
feat: RFC-0023 stateful functions — keyed state API + sticky routing

sanketsudake:
## RFC-0023: Stateful functions — keyed state API + sticky routing

Implements [RFC-0023](docs/rfc/0023-stateful-functions-keyed-state-sticky-routing.md) end to end: functions opt into durable, scoped key/value state via `FunctionSpec.State` — no embedded DB clients, no hand-wired credentials — plus opt-in sticky routing so one key's requests land on one pod. This is Cloudflare Durable Objects / Dapr-state territory, built on the RFC-0021 statestore.

## How to review this PR (89 files, ~5k lines; only 6 generated)

**Suggested reading order** — the diff is layered; follow the data flow:

1. **The API shape** — `pkg/apis/core/v1/types.go` (`StateConfig`/`StickyConfig`) + `validation.go`. Start here; everything downstream keys off these fields.
2. **The quota fix (most subtle logic)** — `pkg/statestore/interfaces.go` (`CountedKV`), `scoped.go`, `memory/kv.go`, `sqlstore/kv.go`. This replaces a racy check-then-act soft cap (deleted) with an atomic counted write. The `sqlstore` `state_quota` row-lock transaction is the load-bearing part; `pkg/statestore/statestoretest/conformance.go` `CountedQuotaConcurrent` proves it on every driver.
3. **The head** — `pkg/statesvc/{statesvc,auth,handler,index,reconciler}.go` + the `stateapi` leaf wire-contract package. `auth.go` is the security boundary (bearer + signed-admin, scope only from a verified token); `reconciler.go` is the keyspace-purge finalizer.
4. **Token injection (the trickiest constraint)** — `pkg/fetcher/statetoken.go` + `config/config.go`, `pkg/executor/util/stateenv.go`. Why the token is a fetcher-written FILE not an env var is the poolmgr-generic-pod constraint; the header comments explain it.
5. **Sticky routing** — `pkg/router/endpointcache/index.go` (`Admit` HRW branch), `transport.go` (`stickyKeyFromRequest`), `resolver*.go` threading.
6. **CLI** — `pkg/fission-cli/cmd/function/state.go`.

**Skip on first pass** (generated / mechanical): `zz_generated.deepcopy.go`, `pkg/generated/**`, `crds/v1/fission.io_functions.yaml`, `applyconfiguration/**`.

**Helm** (`charts/fission-all/templates/statesvc/*`, `tenant-controller/dynamic-cluster-roles.yaml`, `webhook-server/rbac.yaml`) mirrors the mcp/workflow head patterns; `test/helm/portdrift_test.go` pins the port/RBAC/metrics wiring.

**Highest-risk areas to scrutinize:** (a) the `sqlstore` counted-quota transaction under Postgres READ COMMITTED; (b) `auth.go` token verification (constant-time, admin scope in the signed query string not headers); (c) the `Admit` HRW branch preserving the fused pick+admit atomicity and the `Release`-closure accounting seam.

**Commits** are a story to follow in order: feature phases (crd → quota → hmac → statesvc → helm → cli → injection → sticky), then the review-battery fixes, then CI-surfaced fixes (each `fix(...)` commit message names the exact defect and why). Reviewing per-commit is viable.

### What ships

- **CRD** — `FunctionSpec.State` (`StateConfig`: keyspace, TTL, quotas, backend, `StickyConfig`), presence = opt-in like `Streaming`/`Tool`. Container executor and `allowedFunctionsPerContainer: infinite` envs are rejected (no way to isolate a per-function token).
- **`statesvc` head** (`--stateApiPort`, ClusterIP `svc/statesvc`, `functionState.enabled`) — a **scoped** function-facing surface over the statestore (`GET/PUT/DELETE` + `POST /cas` + `GET` list; If-Match version CAS; TTL headers; paged listing). Scope comes **only** from a verified per-keyspace HKDF token or the signed admin channel — never from the request. Keyspace-lifecycle finalizer purges on delete (unless retained or shared).
- **Atomic quota (S3)** — replaced `scopedKV`'s racy check-then-act soft cap with a driver-native `CountedKV` capability: the live-key count and the write are one step (memory: under mutex; sqlstore: a `state_quota` row-lock transaction). TTL-exact, no external counter to drift.
- **Token injection** — `FISSION_STATE_URL` is a plain env var; the per-function token is derived **by the fetcher** (which already holds the master secret) and written to `/userfunc/.fission-state-token` as `{namespace, keyspace, token}` JSON — no secret ever rides the pod-visible specialize arg, and no env-image contract change (an improvement over the RFC's runtime-writes-it sketch).
- **Sticky routing (S4/S5)** — a rendezvous-hash branch **inside** `endpointcache.Admit` (not a strategy object), so pick+admit stays one fused atomic op and the accounting seam is untouched. Saturated owner overflows to next-ranked (best-effort).
- **CLI** — `fission fn state get|set|delete|list` (admin channel, fails closed without the secret) and `fn create/update --state*` flags.

### Verification

Invariants S1–S6 each have a targeted test: **S1** HKDF-scope fuzzer (144k execs) + forgery matrix + in-cluster forge fixture; **S2** porcupine linearizability over the real HTTP surface + zero-lost-increment counter (unit + integration); **S3** `CountedKV` conformance run on every driver incl. a concurrent-overshoot subtest + `rapid`/`-race` on memory; **S4/S5** `rapid` determinism/reshuffle + χ²-style balance; TTL/rotation in `synctest`; finalizer matrix incl. purge-failure guard. Full local unit suite green; the RFC-0023 integration suite (`test/integration/.../function_state_test.go`) passes all 5 subtests against a kind cluster (counter/no-lost-updates, scope isolation, CLI+quota, sticky residency, finalizer purge).

Every commit is one reviewable step (CRD → quota → hmac → statesvc → helm → CLI → injection → sticky → review fixes). Reviewed with the full battery (code-review, security, simplify, thermo-nuclear, deslop, test-coverage); findings folded in as their own commits.

### Deviations & deferrals (also in the RFC status note)

- S3 landed as the "counted transaction" option, **not** counter-key CAS (a counter needs a TTL-drift heal that races in-flight reservations — caught by the mixed-churn race test).
- Admin scope claims ride the **signed query string**, not headers (replay-retarget hardening).
- Sticky sources: header/queryparam (PathParam deferred). Sticky metrics shipped as `fission_router_sticky_requests_total` / `_key_missing_total` (per-key reshuffle counting dropped — needs per-key router memory).
- **Deferred**: env-repo node/python SDK helpers, RFC-0018 local-loop wiring, RFC-0020 bench scenario, `MaxNamespaceBytes` namespace budget, per-function `Backend` driver selection (field validated, semantics later).

---

**Tenancy note:** the state token derives from the master internal-auth secret, which by design never reaches function pods under dynamic/cluster multi-namespace tenancy (those get per-namespace keys). So the feature is fully supported under **static tenancy** (the default); statesvc's cluster RBAC is wired for dynamic/cluster, but per-namespace token derivation (via the tenant controller) is a documented follow-up, and the integration tests gate to static.

---

**Docs & examples:**

- User-facing docs (practical "Function State" usage guide): [fission/fission.io#314](https://github.com/fission/fission.io/pull/314)
- Example functions (Node + Python stateful counter / session): [fission/examples#95](https://github.com/fission/examples/pull/95)


charts/fission-all/templates/webhook-server/rbac.yaml
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# The validating webhook reads Environments (get only) to enforce env-dependent
+# Function invariants CEL cannot express — currently that RFC-0023 state opt-in
+# is refused on an allowedFunctionsPerContainer: infinite environment (one
+# shared mount cannot deliver a per-function scoped state token). Functions may
+# reference an environment in any Fission-watched namespace, so the read is
+# cluster-scoped; it is get-only (uncached API reader), the minimum for a
+# point lookup.
+apiVersion: rbac.authorization.k8s.io/v1
```
sanketsudake, line 60: We need to see if these settings can be adjustable as per static, dynamic tenancy modes etc.

sanketsudake: Good call — done in 6b4cbfa2. The env read now follows the same tenancy-scoping as every other Fission-CRD read:

- **static** — a namespaced `Role` (get environments) in `defaultNamespace` + each `additionalFissionNamespaces`, bound to `fission-webhook`. Functions can only reference environments in those namespaces there, so a cluster-wide grant would be over-privileged.
- **dynamic / cluster** — a cluster-wide `ClusterRole` is added (gated on `fission.tenancyControllerEnabled`), since namespaces onboard at runtime and a function may reference an environment in any of them.

Verb stays `get`-only (uncached point lookup). `test/helm/portdrift_test.go::TestWebhookEnvReaderRBACScopesByTenancy` pins both postures.

PR #3596, opened 2026-07-21
fix(crd): drop ResourceVersion from CacheKeyURG

rijojohn85:
## Summary

- Drops `ResourceVersion` from `CacheKeyURG`, relying on `UID + Generation` only
- Fixes an `activeRequests` leak: when the executor writes a function status update (bumping RV) and the router later calls `UnTapService` with the stale RV from its informer cache, the key doesn't match the entry created by `SetSvcValue` — `UnTapService` misses the cache entry, `activeRequests` is never decremented, and the idle reaper refuses to delete the pod (`activeRequests > 0`)
- Status-only updates no longer re-key the cache; `SetSvcValue`/`UnTapService` agree on the key regardless of informer cache lag

### Blast radius

This re-keys the entire poolmgr↔router data plane (concurrency accounting, `UnTapService`) for **all** functions, not just provisioned ones. No caller accesses `.ResourceVersion` on a `CacheKeyURG` — the struct is used solely as a map key. Split out from PR #3581 (RFC-0026 provisioned concurrency) so it's reviewable and revertable independently of the feature.

#### Test plan

- [x] `TestCacheKeyURGFromMeta_StableAcrossStatusUpdates` — key is identical across RV-only changes (same UID+Generation)
- [x] `TestCacheKeyURGFromMeta_ChangesOnSpecUpdate` — key changes on Generation increment
- [x] `TestCacheKeyURGFromMeta_ChangesOnUIDChange` — key differs for different UIDs
- [x] `TestCacheKeyURG_String` — format is `UID_Generation`
- [x] `go build ./...` passes
- [x] `go test -race -count=1 ./pkg/crd/...` passes

### DIFF
--- a/pkg/crd/key.go
+++ b/pkg/crd/key.go
@@ -20,12 +20,12 @@ func (ck CacheKeyUR) String() string {
 	return fmt.Sprintf("%v_%v", ck.UID, ck.ResourceVersion)
 }
 
-type CacheKeyURG struct {
+type CacheKeyUG struct {
 	UID        types.UID
 	Generation int64
 }
 
-func (ck CacheKeyURG) String() string {
+func (ck CacheKeyUG) String() string {
 	return fmt.Sprintf("%v_%v", ck.UID, ck.Generation)
 }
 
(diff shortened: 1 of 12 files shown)


pkg/crd/key.go
```diff
@@ -21,13 +21,12 @@ func (ck CacheKeyUR) String() string {
 }
 
 type CacheKeyURG struct {
```
sanketsudake: Can we drop R from all struct and methods ? CacheKeyUG

sanketsudake (approved):
We didn't have status in the resource enabled, but recently we have enabled status, so generation should suffice, I guess. 
There were some situations, I think, where generation was not getting updated and function images were updated via env etc. Hopefully we don't see those issues again after this fix.

PR #3606, opened 2026-07-24
RFC-0026 PR2: Scheduled windows for provisioned concurrency

rijojohn85:
## Summary

Second PR of RFC-0026 (Provisioned Concurrency). Builds on PR1 (#3581, merged) by adding **scheduled windows**: time-based overrides that raise or lower a function's effective provisioned-concurrency target during cron-defined windows (e.g. warm N pods 9am-5pm on weekdays), on top of the flat `Target` PR1 shipped.

Feature implementation (Stages 2-8) plus a full verification pass against the RFC's stated invariants (P1-P5, `docs/rfc/0026-provisioned-concurrency-scheduled-warming.md` lines 87-101) — property tests, DST goldens, a benchmark guardrail, and an audit of PR1's own test coverage against the same invariants.

## Feature implementation

**CRD + validation**
- `ProvisionedConcurrencyConfig.Windows []ProvisionedWindow`, each with `Start` (cron), `Duration`, `Target`, `Name`.
- Admission validates: cron syntax, duration syntax + positivity, non-negative `Target`, unique + non-empty window `Name`s.
- `Start`'s timezone semantics documented on the field: an unprefixed cron follows the executor process's local timezone; `CRON_TZ=<zone>` pins it explicitly. This is intended behavior, not a default that may change.

**Pure schedule evaluator** (`provisioner_schedule.go`)
- `windowActiveAt(window, now) (bool, error)` — is a window open at instant `now`. Works around `robfig/cron`'s forward-only `Next(t)` by walking backward from `now - duration`. Half-open interval: open at the fire instant, closed exactly at `fire + duration`. A cron denser than its own duration is capped at 100k walk iterations and treated as permanently active (`lastSched`).
- `effectiveTargetAt(cfg, now) (int, []error)` — base `Target` if no window is open, otherwise the max `Target` across open windows (an open `Target: 0` window wins over a larger base and un-warms the function). Malformed windows are skipped and reported, never fatal to the rest of the evaluation.
- `nextTransitionAt(cfg, now) (time.Time, []error)` — earliest future instant any window's state could change, used to arm precise wake-up timers.
- Both pure: no I/O, no wall-clock reads outside the `now` parameter.

**Per-function timer manager** (`provisioner.go`)
- `fnSchedule`/`perWindow` cache parsed schedules and window edges per function, invalidated on spec `Generation` change.
- `armTransition` arms a `*time.Timer` for the earliest upcoming transition; bad/capped windows fall back to the existing 30s reconcile ticker as a drift guard.
- Timer callback re-validates the function is still opted in before acting, making a timer-fire-vs-`Stop()` race benign either way.
- Wired into `reconcileFunctionLocked`, `disableProvisioningLocked`, and `forget` so timers are armed/torn down on every reconcile path.

**CLI + metrics + integration**
- `--provisioned-schedule` CLI flag for window config.
- `fission_provisioned_window_transitions_total` metric.
- Integration tests (`suites/common/provisioned_concurrency_test.go`): CLI round-trip, newdeploy rejection, window validation, `Target: 0` window unwarm/rewarm, scheduled window open/close reaping the right pods, window raising the floor, and a full lifecycle test (floor@zero-traffic, reaper liveness alongside a held floor, self-heal after pod kill, generation-refresh on target change).
- Serial suite: executor restart re-derives the provisioned floor after adopt (`TestProvisionedConcurrencyExecutorRestart`).

## Verification against the RFC's invariants (P1-P5)

The RFC lists five invariants for this feature (pure target, convergence, pacing bound, no immortal pods, no starvation) and specifies how each should be verified. This PR closes every verification gap found against that list, for both PR2's own code and a re-check of PR1's:

**Property-based tests** (`provisioner_schedule_property_test.go`, `pgregory.net/rapid`) — 7 properties over generated schedule sets, each compared against an independent oracle implementation (not a restatement of production logic) or checked for order/permutation invariance:
1. `windowActiveAt` matches a from-scratch oracle at both uniform-random and boundary-biased instants (half-open interval edges).
2. Evaluation is total — never panics or hangs, including on unsatisfiable crons and hostile instants (epoch, zero time, year 2200).
3. `effectiveTargetAt` returns exactly the max open-window target, or base if none are open — including the `Target: 0`-wins-over-base rule.
4. Result is invariant under window-list permutation and duplicate insertion (catches "first match wins" / accumulation bugs).
5. Result is deterministic and timezone-invariant for `CRON_TZ`-prefixed specs (and *not* asserted invariant for unprefixed specs, since those intentionally follow the running machine's timezone — documented in a comment on the test).
6. The `lastSched` 100k-iteration give-up path returns `true` across a swept range of cron densities.
7. A single malformed window (bad cron / bad duration / non-positive duration) is isolated — doesn't change the target, appears as exactly one named error.

Each property was verified capable of failing by temporarily breaking the corresponding production code and confirming the test caught it, then reverting.

**DST golden tests** (`provisioner_schedule_dst_test.go`) — 5 cases against real `America/New_York` transitions plus an `Asia/Kolkata` (no-DST, UTC+5:30) control:
1. Spring-forward day: a window whose start (02:30) doesn't exist that day never opens — `robfig/cron` skips the fire entirely (traced and confirmed against the actual library, not assumed).
2. A window spanning the spring-forward gap stays open continuously through the missing hour, closing an hour later on the wall clock than a naive reading suggests (duration is elapsed time, not wall-clock span).
3. Fall-back day: a window whose start (01:30) occurs twice fires twice; with a short duration the two windows abut with zero gap, giving a full extra hour of warm pods once a year (reviewed and accepted, not a bug).
4. A window spanning the fall-back repeated hour closes an hour *earlier* on the wall clock, for the same elapsed-time reason as (2).
5. The Kolkata control confirms the same window shapes behave plainly with no DST involved, and exercises the half-hour UTC offset.

All 5 pass independent of the running machine's `TZ` (swept against `America/Los_Angeles`, `Asia/Kolkata`, `UTC`).

**P5 starvation-guardrail benchmark** (`test/benchmark/pkg/scenario/provisionedwarm.go`, new `provisioned-warm-starvation` scenario, weekly-full-suite only, not gated per-PR) — measures function B's cold-start latency while function A eagerly warms a provisioned-concurrency burst on the same pool: a no-contention baseline phase, then a treatment phase run concurrently with the burst, gated on the p50 ratio between them at a fixed 2.0x ceiling.

A local run against a kind cluster (burst target 20, pool 30, 10 iterations/phase) produced real results:

| metric | value |
|---|---|
| baseline p50 / p95 | 39.7ms / 48.6ms |
| contended p50 / p95 | 73.2ms / 186.8ms |
| **contended_ratio (p50, gated at 2.0)** | **1.84x — passes** |
| contended_ratio_p95 (informational only) | **3.85x** |
| overlapped_samples / samples | 9/10 |

The p50 ratio clears the gate, but the p95 ratio doesn't — B's worst-case cold starts during A's burst run nearly 4x its own baseline. Traced to real contention on `GenericPool`'s shared `readyPodQueue` (`pkg/executor/executortype/poolmgr/gp_pod.go`) between A's concurrent eager specializations and B's on-demand request — genuine interference, not a benchmark artifact, and exactly the class of thing P5 exists to catch. Filed as **#3623** for the team to decide on a p95 gate/mitigation; deliberately **no gate added in this PR**.

**PR1 verification audit** — re-checked PR1's own tests (already merged) against P2/P3/P4:
- P3 (pacing bound) and most of P4 (no immortal pods) are solidly covered already; nothing to add.
- P2 (convergence) is proven correct but via real-time `require.Eventually` rather than the RFC's suggested `synctest` virtual-clock style — a verification-style gap, not a bug, and left as-is here.
- P4's fuller lifecycle table (RFC line 99) wants coverage for a spec generation bump and router-side quarantine replacement. Neither exists in the code yet: generation bump is RFC rollout phase 3 (a future PR); quarantine is a separate, already-existing router feature that RFC-0026 deliberately does not hook the provisioner into (RFC line 68) — permanently out of scope by design, not a future gap.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `golangci-lint run ./...` clean on every commit
- [x] `go test -race ./pkg/executor/executortype/poolmgr/...` — unit + property + DST tests, all passing
- [x] `go test -tags=integration ./test/integration/suites/common/...` — scheduled-window integration coverage
- [x] `go test -tags=integration -p 1 ./test/integration/suites/serial/...` — executor-restart floor re-derivation
- [x] `go build`/`go vet`/`go test` clean in the `test/benchmark` module
- [x] Manual run of `provisioned-warm-starvation` against a live kind cluster (real numbers above)
- [ ] Follow-up: #3623 (P5 p95 tail-latency gate/mitigation decision)
- [ ] Follow-up: generation-bump lifecycle test, once RFC-0026 phase 3 lands


pkg/executor/executortype/poolmgr/provisioner_schedule.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		t2 := sched.Next(t)
+		if t2.After(now) {
+			break
+		}
+		t = t2
+		iterations++
+	}
+	return t, false
+}
```
sanketsudake, line 158: **[High] `lastSched` treats robfig/cron's zero-time sentinel as a past fire and restarts the walk from year 1.**

`Schedule.Next` returns the zero time when no fire exists inside its 5-year search horizon. Zero is not `After(now)`, so `t = t2` sets `t` to year 1 and the walk restarts — only the 100k cap ends it, and the cap returns `capped=true`.

Verified locally: `Next(1796-02-29)` on `0 0 29 2 *` returns `0001-01-01` (1800 is not a leap year). An admission-valid window `{start: "0 0 29 2 *", duration: "1200000h"}` burns **706ms of executor CPU per evaluation**, on the per-function reconcile path. With the fix below plus the `windowActiveAt` short-circuit, the same call drops to **285µs**.

The blast radius is CPU, not a wrong target: in `windowActiveAt` the answer is `true` either way (see the next comment), and in `reBuildWindows` `capped=true` just drops the precise timer and falls back to the 30s ticker. But it is the strongest per-window amplifier for the unbounded-`Windows` issue flagged in `validation.go`.

This also folds in the missing `maxScheduleIterations` constant — the doc comment on `windowActiveAt` (line 112) already refers to it by name, but no such identifier exists; the code has a bare `100_000`.

I verified this patch builds, is gofmt-clean, and keeps `go test -race ./pkg/executor/executortype/poolmgr/` green (property + DST tests included).

```suggestion
// maxScheduleIterations bounds lastSched's forward walk. A cron denser than
// its window duration would otherwise walk unboundedly; hitting the cap is
// reported as capped=true and treated as permanently active.
const maxScheduleIterations = 100_000

// lastSched walks forward from t and returns the last fire <= now.
// capped=true means the walk hit maxScheduleIterations without converging.
func lastSched(sched cron.Schedule, t time.Time, now time.Time) (time.Time, bool) {
	for range maxScheduleIterations {
		next := sched.Next(t)
		// robfig/cron returns the zero time when no fire exists within its
		// 5-year search horizon (e.g. "0 0 29 2 *" crossing the non-leap
		// year 1800). Zero is not After(now), so without this guard the walk
		// restarts from year 1 and burns the entire iteration budget.
		if next.IsZero() || next.After(now) {
			return t, false
		}
		t = next
	}
	return time.Time{}, true
}
```

pkg/executor/executortype/poolmgr/provisioner_schedule.go
```diff
@@ -0,0 +1,154 @@
...
+	if t.IsZero() || t.After(now) {
+		return false, nil
+	} else {
+		var capped bool
+		t, capped = lastSched(sched, t, now)
+		if capped {
+			return true, nil
+		}
+		return now.Sub(t) < dur, nil
+	}
```
sanketsudake, line 134: **[High] The `lastSched` walk here is provably dead, and it is what makes this function expensive.**

`sched.Next(now.Add(-dur))` returns the first fire *strictly after* `now-dur`. So reaching the `else` branch means `now-dur < t <= now`, which already implies `now.Sub(t) < dur`. `lastSched` only moves `t` closer to `now`, so the final `return now.Sub(t) < dur` is **always `true`** — the walk cannot change the answer.

I verified this empirically across `* * * * *`/24h, `0 9 * * *`/10h, `*/5 * * * *`/1h and `@every 1s`/2000h: the short-circuit matches the full walk in every case.

What the walk *does* cost is one `sched.Next` per fire inside the duration — 1440 evaluations for `* * * * *` + `24h`, capped at 100k — paid per window, per function, on every 30s `effectiveTarget` reconcile.

`lastSched` is still needed by `reBuildWindows`, so only this caller changes. Verified: builds, gofmt-clean, `go test -race` green.

```suggestion
	// sched.Next returns the first fire strictly after now-dur, so t <= now
	// implies now-dur < t <= now: the window opened at t and closes at t+dur,
	// which is after now. Any fire in (now-dur, now] therefore means the
	// window is open — no need to walk forward to the most recent fire.
	t := sched.Next(now.Add(-dur))
	if t.IsZero() || t.After(now) {
		return false, nil
	}
	return true, nil
}
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -508,4 +610,11 @@ func (p *Provisioner) disableProvisioningLocked(ctx context.Context, fn *fv1.Fun
 func (p *Provisioner) forget(fnUID types.UID) {
 	p.reconcileLocks.Delete(fnUID)
 	p.inflight.Delete(fnUID)
+	if v, ok := p.timers.Load(fnUID); ok {
+		sched := v.(*fnSchedule)
+		if sched.timer != nil {
+			sched.timer.Stop()
+		}
+	}
```
sanketsudake, line 682: **[High] `forget` reads and stops `sched.timer` without holding `sched.mu` — a genuine `-race` failure.**

`armTransition` writes `sched.timer` at lines 168/185 **under `sched.mu`**. `forget` is called from `GenericPoolManager.DeleteFunction` (`reconciler.go:172`), which holds neither `sched.mu` nor `lockFor(fn.UID)` — it *deletes* the reconcile lock without acquiring it. Deleting a function while a `reconcileAll` pass arms that function's timer is an unsynchronized read/write of the `*time.Timer` field.

Secondary effect: `forget` can observe `sched.timer == nil`, skip `Stop()`, delete the map entry, and leave an orphaned timer to fire against a deleted function. Impact there is bounded (the callback's `Get` returns NotFound and logs), but the race itself is the blocker — this repo gates on `go test -race`.

All three review passes flagged this independently.

```suggestion
	if v, ok := p.timers.Load(fnUID); ok {
		sched := v.(*fnSchedule)
		sched.mu.Lock()
		if sched.timer != nil {
			sched.timer.Stop()
		}
		sched.mu.Unlock()
	}
```

sanketsudake (commented):
## Review: three passes, deduplicated

Ran three independent review passes over the full diff — a general code review, a security review, and a code-quality/maintainability audit — then deduplicated and **verified every claim against the source before reporting it**. All work was done in an isolated worktree at `dae3af6`.

Worth saying up front: the verification story in this PR (property tests against independent oracles, DST goldens traced against the real library rather than assumed, each property confirmed capable of failing) is unusually strong, and the P5 benchmark honestly reporting a p95 ratio that *doesn't* pass rather than quietly tuning the gate is exactly right. The findings below are mostly in the timer/lifecycle plumbing around the evaluator rather than in the evaluator's logic itself.

CI is green on all 17 checks, so none of this is a build or test break — every item is latent.

### Where the three passes disagreed

Two passes reached **opposite conclusions** about `lastSched`, so I settled it empirically rather than reporting both.

- One argued a zero-time bug pins windows permanently open. The mechanism is real — `Next(1796-02-29)` on `0 0 29 2 *` does return `0001-01-01`, and the walk does restart from year 1 — but the **correctness** conclusion is wrong.
- The other argued the walk in `windowActiveAt` is dead code. That one is right, and it explains why: the function returns `true` either way, and `true` is the *correct* answer.

So the real defect is **CPU cost, not a wrong target**. I measured it: 706ms per evaluation on an admission-valid window, dropping to 285µs with both fixes applied (~2500x). I've filed them as two separate comments since they fix different call sites and compose.

I applied both patches locally and confirmed they build, are gofmt-clean, and keep `go test -race ./pkg/executor/executortype/poolmgr/` green — property and DST tests included.

### Summary of findings

| # | Severity | Area | Finding |
|---|---|---|---|
| 1 | High | `provisioner_schedule.go` | `lastSched` misses cron's zero-time sentinel → 706ms/eval |
| 2 | High | `provisioner_schedule.go` | `windowActiveAt`'s walk is provably dead and is the cost |
| 3 | High | `provisioner.go` | `sched.timer` data race in `forget` (all three passes) |
| 4 | High | `provisioner.go` | Same race in `disableProvisioningLocked` |
| 5 | High | `provisioner.go` | `MarkSpecializationFailure` releases a reservation never taken |
| 6 | High | `validation.go` | `Windows` unbounded → cross-tenant reconcile starvation |
| 7 | Medium | `provisioner.go` | `armTransition` nil-derefs when PC is nil |
| 8 | Medium | `provisioner.go` | No delay floor → self-sustaining ~1Hz reconcile loop |
| 9 | Medium | `provisioner.go` | Timer closure pins the first `*Function`, and shadows it |
| 10 | Medium | `update.go` | `fn update --provisioned-concurrency` wipes all windows |
| 11 | Medium | `create.go` | `--provisioned-concurrency 0` silently drops the schedule |
| 12 | Medium | `provisioner_test.go` | `assert.Nil(t, sched.timer)` can never fail |
| 13 | Low | `validation.go` | Bad duration emits two stacked errors |
| 14 | Low | `provisioner.go` | `logger.Error(e, e.Error())` defeats log grouping |
| 15 | Low | `idle.go` | Comment and function name contradict the map's shape |
| 16 | Low | `idle.go` | Two of three exemption guards are dead |

Suggestions are attached inline wherever a clean single-hunk fix exists.

### Structural notes (no inline anchor)

**The "is this window open?" question has two independent implementations.** `windowActiveAt`/`effectiveTargetAt` re-parse the cron and re-walk on every reconcile, while `fnSchedule.reBuildWindows` computes the same predicate again and materializes it as `perWindow{activeUntil, nextOpen, capped}`. They even read **different clocks** — `effectiveTarget` passes `time.Now()` in, `reBuildWindows` calls `time.Now()` itself per window — so the target computation and the timer meant to fire at that target's next change point can derive from different instants. `perWindow` already holds everything both questions need. This is the root cause behind findings 1, 2 and 8, and collapsing it would retire them structurally.

**The cron parser option set is hand-built in five places.** `cron.NewParser(cron.SecondOptional | ... | cron.Descriptor)` appears at `provisioner_schedule.go:37` and `:115`, inside `IsValidCronSpec` (`validation.go:180`), in `fission-cli/cmd/timetrigger/create.go:140`, and again in `provisioner_schedule_property_test.go:30` — where the test's own comment admits it exists only to keep the copies in sync. I checked and they currently agree, so there's no live bug, but an exported `fv1.ParseCronSpec` would make it impossible for admission and the executor to diverge later.

**The CLI re-implements the whole window ruleset.** `getProvisioinedWindow` plus the dup-name check re-encode every rule in `ProvisionedConcurrencyConfig.Validate()` — cron validity, duration parse, duration > 0, target >= 0, unique names — and the wording has already drifted (`"target is negative"` vs `"window target must be >= 0"`). The parser could do nothing but turn `key=value;...` into a `ProvisionedWindow` and hand the assembled config to `Validate()`.

**Failed windows are stored as `nil` map values.** `reBuildWindows` writes `f.windows[name] = nil` on all three failure paths, which forces both loops in `armTransition` to open with `if pw == nil || pw.capped`. Skipping the entry instead removes the sentinel and both guards; the `len(sched.windows) == 0` rebuild trigger already covers the all-bad case.

**Smaller items:**
- `perWindow.name` duplicates the map key it's stored under, and `fnSchedule.name`/`.namespace` are written by `scheduleFor` but never read — every consumer uses `fn.Name`/`fn.Namespace`.
- `TestProvisionerScheduleTickerTimer`'s "proof they were serialized by `lockFor`" assertion proves nothing: both reconciles call List exactly once each under *every* interleaving, so `listCalls2 == 2` holds even with the mutex removed. Pinning serialization needs a max-concurrent-in-List counter.
- `getProvisioinedWindow` is misspelled, and the typo has already propagated into `TestGetProvisioinedWindow` and two failure messages — cheapest to fix now while it's a three-site rename.
- The blank line + `// RFC-0026` comment at `flag.go:142` broke the var block's alignment group, so gofmt rewrote 50 unrelated flag lines. Real change is 2 lines; `git blame` noise is 50.
- `provisioner_test.go` grows 1016 → 1537 lines while the PR simultaneously demonstrates it knows how to split (three new `provisioner_schedule_*_test.go` files). The four new schedule/timer tests would sit more naturally with their siblings.

### One correction to the PR description

The description documents a `nextTransitionAt(cfg, now) (time.Time, []error)` evaluator function. No such symbol exists anywhere on the branch — `armTransition` computes the next edge inline from cached `perWindow` values instead. Worth fixing so the description doesn't describe an API that reviewers then go looking for.

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -496,6 +589,15 @@ func (p *Provisioner) disableProvisioning(ctx context.Context, fn *fv1.Function)
 func (p *Provisioner) disableProvisioningLocked(ctx context.Context, fn *fv1.Function) {
 	p.clearProvisionedLabels(ctx, fn, -1)
 	p.inflight.Delete(fn.UID)
+	if fn.Spec.ProvisionedConcurrency == nil {
+		if v, ok := p.timers.Load(fn.UID); ok {
+			sched := v.(*fnSchedule)
+			if sched.timer != nil {
+				sched.timer.Stop()
+			}
+		}
```
sanketsudake, line 660: **[High] Same unsynchronized `sched.timer` access as in `forget`.**

The `lockFor(fn.UID)` this path holds is a *different* mutex from `sched.mu`, which is what `armTransition` uses to guard the `sched.timer` write. It does not exclude a concurrent `armTransition` for the same function reached via the timer callback.

The two blocks are also verbatim copy-paste — worth extracting a `func (f *fnSchedule) stop()` method on `fnSchedule` and calling it from both sites instead of applying this fix twice.

```suggestion
		if v, ok := p.timers.Load(fn.UID); ok {
			sched := v.(*fnSchedule)
			sched.mu.Lock()
			if sched.timer != nil {
				sched.timer.Stop()
			}
			sched.mu.Unlock()
		}
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -362,8 +439,16 @@ func (p *Provisioner) eagerSpecialize(ctx context.Context, fn *fv1.Function) err
 	funSvc, err := p.gpm.GetFuncSvc(ctx, fn)
 	if err != nil {
 		metrics.RecordEagerSpecialization(ctx, fn.Name, fn.Namespace, "error")
+		p.gpm.MarkSpecializationFailure(ctx, &fn.ObjectMeta)
 		return err
```
sanketsudake, line 499: **[High] `eagerSpecialize` releases a `PoolCache` specialization reservation it never took.**

`PoolCache.MarkSpecializationFailure` decrements `grp.svcWaiting`. That counter is incremented in exactly two places — `getSvcValue` (the `GetFuncSvcFromCache` path) and `ReserveCapacity` — and `ReserveCapacity`'s own doc spells out the invariant: *"The reservation must be released exactly once: by setValue on a successful specialization, or MarkSpecializationFailure on a failed one."*

The provisioner calls `gpm.GetFuncSvc` directly, which takes **no** reservation. So on a failed eager specialization this decrement steals a *concurrent router cold start's* reservation (the guard `svcWaiting > queue.Len()` only fires when someone else's surplus exists). `concurrencyUsed = len(svcs) + (svcWaiting - queue.Len())` then under-counts in-flight specializations, letting requests over-admit past the function's concurrency cap and past the `maxPending` saturation guard.

This is the two-disjoint-accounting-modes trap: router-admitted entries release via the `Release` closure, executor-resolved ones via UnTap. The `UnTapService` call just below is correct and matches `api.go`'s `cleanUp`; only this line is unpaired.

```suggestion
		metrics.RecordEagerSpecialization(ctx, fn.Name, fn.Namespace, "error")
		return err
```

pkg/apis/core/v1/validation.go
```diff
@@ -646,22 +646,38 @@ func (sc *StateConfig) Validate() error {
...
 	var errs error
 	if pc.Target < 1 {
 		errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Target", pc.Target, "must be >= 1"))
 	}
 
-	if len(pc.Windows) > 0 {
-		errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", len(pc.Windows),
-			"scheduled windows are not yet supported (RFC-0026 PR 2)"))
+	windows := make(map[string]struct{})
+	for i, window := range pc.Windows {
```
sanketsudake, line 675: **[High] `Windows` has no count bound in the CRD schema, in CEL, or here.**

I confirmed the generated `crds/v1/fission.io_functions.yaml` has no `maxItems` on the `windows` array, and `Validate()` does not bound `len(pc.Windows)`.

Each window costs an independent cron parse plus a schedule walk in `effectiveTargetAt`, and `armTransition` → `reBuildWindows` repeats that work in the same pass. Measured: 500 windows of `{start: "* * * * *", duration: "2000h"}` — every one individually admission-valid — cost **31.8s of CPU per `effectiveTargetAt` call**, all under the per-function reconcile lock. `reconcileAll` caps parallelism at `maxConcurrentReconciles = 10`, so ten such Functions from a single namespace-scoped tenant wedge provisioned-concurrency reconciliation cluster-wide.

The two `provisioner_schedule.go` fixes cut the per-window constant a lot, but a bound is still the right backstop for a tenant-supplied list. Worth adding `// +kubebuilder:validation:MaxItems=32` to the `Windows` field in `types.go` and regenerating the CRD, so the apiserver rejects it even when the webhook is down.

```suggestion
	// Bound the window count: each window costs an independent cron parse plus
	// a schedule walk on every executor reconcile pass, held under the
	// per-function reconcile lock inside a 10-slot semaphore, so an unbounded
	// list is a cross-tenant starvation vector for the provisioner.
	const maxProvisionedWindows = 32
	if len(pc.Windows) > maxProvisionedWindows {
		errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", len(pc.Windows),
			fmt.Sprintf("must not exceed %d windows", maxProvisionedWindows)))
	}

	windows := make(map[string]struct{})
	for i, window := range pc.Windows {
```

pkg/apis/core/v1/validation.go
```diff
@@ -646,22 +646,38 @@ func (sc *StateConfig) Validate() error {
...
+		if err := IsValidCronSpec(window.Start); err != nil {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "start is invalid: "+err.Error(), "window name: "+window.Name))
+		}
+		d, err := time.ParseDuration(window.Duration)
+		if err != nil {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration is invalid: "+err.Error(), "window name: "+window.Name))
+		}
+		if d <= 0 {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration must be > 0", "window name: "+window.Name))
+		}
```
sanketsudake: **[Low] A malformed duration produces two stacked validation errors.**

`time.ParseDuration` returns `d == 0` on error, so the independent `d <= 0` check fires as well. `duration: "abc"` rejects with both `duration is invalid: time: invalid duration "abc"` **and** `duration must be > 0`, which reads as two distinct problems.

The existing tests assert on substrings only, so chaining these keeps them green.

```suggestion
		if d, err := time.ParseDuration(window.Duration); err != nil {
			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration is invalid: "+err.Error(), "window name: "+window.Name))
		} else if d <= 0 {
			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration must be > 0", "window name: "+window.Name))
		}
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -111,6 +115,77 @@ func (p *Provisioner) lockFor(fnUID types.UID) *sync.Mutex {
 	return lock.(*sync.Mutex)
 }
 
+func (p *Provisioner) scheduleFor(fn *fv1.Function) *fnSchedule {
+	sched, _ := p.timers.LoadOrStore(fn.UID, &fnSchedule{name: fn.Name, namespace: fn.Namespace})
+	return sched.(*fnSchedule)
+}
+
+func (p *Provisioner) armTransition(fn *fv1.Function) {
+	sched := p.scheduleFor(fn)
```
sanketsudake, line 140: **[Medium] `armTransition` nil-derefs when `ProvisionedConcurrency` is nil.**

`reconcileFunctionLocked` calls `armTransition` *before* its `target == 0` early return, and `armTransition` → `reBuildWindows` dereferences `fn.Spec.ProvisionedConcurrency.Windows` unguarded. Every other read of that field in the function is protected by that early return.

Today's callers (`reconciler.go:151`, the timer callback, `filterOptedFunctions`) all pre-check for nil, so this is latent rather than live — but `provisioner_test.go:1253` already calls `reconcileFunction` with `ProvisionedConcurrency` cleared and survives only because the fake clientset does not bump `metadata.generation` and the cached window edges had not yet passed. With a real generation bump this panics and takes the executor process down. Cheap to make unconditional:

```suggestion
func (p *Provisioner) armTransition(fn *fv1.Function) {
	if fn.Spec.ProvisionedConcurrency == nil {
		return
	}
	sched := p.scheduleFor(fn)
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -111,6 +115,77 @@ func (p *Provisioner) lockFor(fnUID types.UID) *sync.Mutex {
...
+			candidate = pw.activeUntil
+		}
+		if next.IsZero() || candidate.Before(next) {
+			next = candidate
+		}
+	}
+	if next.IsZero() || !next.After(now) {
+		return
+	}
+	delay := time.Until(next)
```
sanketsudake: **[Medium] No floor on the timer delay: a sub-minute window becomes a self-sustaining ~1Hz reconcile loop.**

A window with `{start: "* * * * * *", duration: "2s"}` (or `@every 1s`) passes `IsValidCronSpec` and the duration check, and yields `activeUntil`/`nextOpen` edges ~1s apart — so `delay` is ~1s, the callback runs `reconcileFunction`, which re-arms. Each firing does a pod LIST and an unconditional `UpdateStatus` (`updateFunctionStatus` writes every pass, changed or not).

These `AfterFunc` goroutines are **not** gated by the `maxConcurrentReconciles` semaphore or by `ReconcileInterval`, so N such Functions drive N reconciles/second of sustained etcd writes from the cluster-privileged executor service account — triggered by an ordinary namespace-scoped Function write.

Caveat: pick the constant with the synctest-based timer tests in mind; a 10s floor may need those windows widened.

```suggestion
	// Floor the delay: a window whose cron fires about as often as its duration
	// (e.g. start "* * * * * *", duration "2s") otherwise re-arms roughly every
	// second, and each firing runs a full reconcileFunction — a pod LIST plus an
	// unconditional status write — outside both ReconcileInterval and the
	// maxConcurrentReconciles semaphore.
	const minTransitionDelay = 10 * time.Second
	delay := max(time.Until(next), minTransitionDelay)
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -111,6 +115,77 @@ func (p *Provisioner) lockFor(fnUID types.UID) *sync.Mutex {
...
+			if fn.Spec.ProvisionedConcurrency == nil || (executorType != fv1.ExecutorTypePoolmgr && executorType != "") {
+				p.logger.Info("concurrency not available for this function", "function", name, "namespace", namespace, "concurrency", fn.Spec.ProvisionedConcurrency, "execution_strategy", fn.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType)
+				return
+			}
+			p.reconcileFunction(ctx, fn)
+			metrics.RecordProvisionedWindowTransition(ctx, name, namespace)
+		})
+	} else {
+		sched.timer.Reset(delay)
+	}
```
sanketsudake, line 208: **[Medium] The timer closure pins the first `*fv1.Function` forever, and shadows it.**

The `AfterFunc` closure is built once and thereafter only `Reset`, so it keeps the `*fv1.Function` from the **first** arm alive for the provisioner's lifetime — full spec, package refs and all — even though it needs only name/namespace and immediately re-`Get`s a fresh object. And `fn, err := ...` on line 171 shadows the captured `fn`, so the same identifier means two different objects within eight lines.

Capturing the two strings and naming the fetched object distinctly fixes both:

```suggestion
	if sched.timer == nil {
		name, namespace := fn.Name, fn.Namespace
		sched.timer = time.AfterFunc(delay, func() {
			ctx := context.Background()
			latest, err := p.fissionClient.CoreV1().Functions(namespace).Get(ctx, name, metav1.GetOptions{})
			if err != nil {
				p.logger.Error(err, "error fetching latest function", "function", name, "namespace", namespace)
				return
			}
			executorType := latest.Spec.InvokeStrategy.ExecutionStrategy.ExecutorType
			if latest.Spec.ProvisionedConcurrency == nil || (executorType != fv1.ExecutorTypePoolmgr && executorType != "") {
				p.logger.Info("provisioned concurrency no longer applies to this function", "function", name, "namespace", namespace, "executorType", executorType)
				return
			}
			p.reconcileFunction(ctx, latest)
			metrics.RecordProvisionedWindowTransition(ctx, name, namespace)
		})
	} else {
		sched.timer.Reset(delay)
	}
```

pkg/executor/executortype/poolmgr/provisioner.go
```diff
@@ -472,10 +557,18 @@ func (p *Provisioner) updateFunctionStatus(ctx context.Context, fn *fv1.Function
...
+// logs any windows that failed to evaluate, and clamps to the namespace-wide
+// MaxPerFunction cap.
 func (p *Provisioner) effectiveTarget(fn *fv1.Function) int {
-	return min(fn.Spec.ProvisionedConcurrency.Target, p.config.MaxPerFunction)
+	target, errs := effectiveTargetAt(fn.Spec.ProvisionedConcurrency, time.Now())
+	if len(errs) > 0 {
+		for _, e := range errs {
+			p.logger.Error(e, e.Error())
+		}
+	}
```
sanketsudake, line 630: **[Low] `logger.Error(e, e.Error())` duplicates the error into the message field.**

logr's second argument is a *static* message used for grouping and searching; passing `e.Error()` makes every bad-window line unique, so the class can't be grepped or alerted on. The function/namespace also aren't attached, so a bad window can't be traced back to its Function.

```suggestion
	if len(errs) > 0 {
		for _, e := range errs {
			p.logger.Error(e, "provisioned concurrency window evaluation failed", "function", fn.Name, "namespace", fn.Namespace)
		}
	}
```

pkg/executor/reaper/idle/idle.go
```diff
@@ -199,6 +199,9 @@ type PoolDeleteStrategy struct {
 	// (wait.UntilWithContext), and Reap goroutines only read these maps.
 	envUIDs map[k8sTypes.UID]struct{}
 	fnByUID map[k8sTypes.UID]fv1.Function
+
+	// provisionedPods is a map of pod names to struct{}, keyed by namespace which have the fission.io/provisioned label.
+	provisionedPods map[string]struct{}
```
sanketsudake: **[Low] The field comment and the function name both contradict what this map is.**

The comment says "a map of pod names to struct{}, keyed by namespace", and the builder is called `listPodsbyNamespace` — but the code builds a single **flat** set keyed by `"<namespace>/<name>"`. It is neither keyed by namespace nor per-namespace, so a reader will reach for `m[ns]` and find nothing.

Worth renaming `listPodsbyNamespace` → `listProvisionedPodKeys` at its definition and call site too.

```suggestion
	// provisionedPods is the set of "<namespace>/<name>" keys for pods carrying
	// the fission.io/provisioned label, refreshed on every Prepare.
	provisionedPods map[string]struct{}
```

pkg/executor/reaper/idle/idle.go
```diff
@@ -231,39 +234,57 @@ func (s *PoolDeleteStrategy) Prepare(ctx context.Context) error {
...
+	// carrying fission.io/provisioned=true are exempt, so overflow pods
+	// (target drop, window close) are reaped normally.
+	if fn, ok := s.fnByUID[fsvc.Function.UID]; ok && fn.Spec.ProvisionedConcurrency != nil {
+		key := podKeyFromFsvc(fsvc)
+		if key != "" {
+			if _, provisioned := s.provisionedPods[key]; provisioned {
+				return nil
+			}
 		}
 	}
```
sanketsudake: **[Low] Two of the three guards here are dead now that the exemption is per-pod.**

- `fn.Spec.ProvisionedConcurrency != nil` adds nothing: a pod only carries the label because the provisioner put it there, and the provisioner clears it when PC is removed. It also duplicates the `s.fnByUID[fsvc.Function.UID]` lookup done again 11 lines below at line 290.
- `key != ""` is unreachable as a *positive* match: `""` can never be a key in a map whose keys always contain a `/`.

Three nested conditions collapse to one lookup:

```suggestion
	if _, provisioned := s.provisionedPods[podKeyFromFsvc(fsvc)]; provisioned {
		return nil
	}
```

pkg/fission-cli/cmd/function/update.go
```diff
@@ -154,8 +154,12 @@ func (opts *UpdateSubCommand) complete(input cli.Input) error {
...
 
-	if input.IsSet(flagkey.FnProvisionedConcurrency) {
-		function.Spec.ProvisionedConcurrency = getProvisionedConcurrencyConfig(input)
+	if input.IsSet(flagkey.FnProvisionedConcurrency) || input.IsSet(flagkey.FnProvisionedSchedule) {
+		provisionedConcurrency, err := getProvisionedConcurrencyConfig(input)
+		if err != nil {
+			return err
+		}
+		function.Spec.ProvisionedConcurrency = provisionedConcurrency
 	}
```
sanketsudake, line 177: **[Medium] `fission fn update --provisioned-concurrency N` silently deletes all existing schedule windows.**

`getProvisionedConcurrencyConfig` rebuilds the config from flags only, so when `--provisioned-schedule` is absent it returns `Windows: nil`, and this line overwrites the whole spec field. A user with three configured windows who runs `fission fn update --name f --provisioned-concurrency 10` loses all three, with no warning.

The inverse is blocked too: passing only `--provisioned-schedule` always fails with *"provisioned concurrency window is set but provisioned concurrency is not"* (create.go:55-57), so windows cannot be edited in isolation either. Between the two, there is no way to change the base target without restating the entire schedule.

```suggestion
	if input.IsSet(flagkey.FnProvisionedConcurrency) || input.IsSet(flagkey.FnProvisionedSchedule) {
		existing := function.Spec.ProvisionedConcurrency
		provisionedConcurrency, err := getProvisionedConcurrencyConfig(input)
		if err != nil {
			return err
		}
		// Preserve configured windows when only the base target was given, so
		// bumping --provisioned-concurrency does not silently drop the schedule.
		if provisionedConcurrency != nil && existing != nil && !input.IsSet(flagkey.FnProvisionedSchedule) {
			provisionedConcurrency.Windows = existing.Windows
		}
		function.Spec.ProvisionedConcurrency = provisionedConcurrency
	}
```

pkg/fission-cli/cmd/function/create.go
```diff
@@ -46,18 +49,103 @@ func Create(input cli.Input) error {
...
+			return nil, fmt.Errorf("provisioned concurrency window is set but provisioned concurrency is not")
+		}
+		return nil, nil
 	}
 	target := input.Int(flagkey.FnProvisionedConcurrency)
 	if target <= 0 {
 		// 0 or negative = off switch (update path clears the config)
-		return nil
+		return nil, nil
+	}
```
sanketsudake, line 67: **[Medium] `--provisioned-concurrency 0` silently discards `--provisioned-schedule` instead of erroring.**

Four lines above, a schedule without the concurrency flag is a hard error. Here, `--provisioned-concurrency 0 --provisioned-schedule "name=w1;..."` returns `(nil, nil)` and the parsed windows vanish with no diagnostic at all — the user gets a function with no schedule and no message explaining why.

Note the test case `"valid schedule, invalid base target"` in `function_test.go` currently pins the silent drop as intended behaviour, so it needs updating alongside this.

```suggestion
	target := input.Int(flagkey.FnProvisionedConcurrency)
	if target <= 0 {
		// 0 or negative = off switch (update path clears the config)
		if input.IsSet(flagkey.FnProvisionedSchedule) {
			return nil, fmt.Errorf("--%s requires --%s >= 1, got %d", flagkey.FnProvisionedSchedule, flagkey.FnProvisionedConcurrency, target)
		}
		return nil, nil
	}
```

pkg/executor/executortype/poolmgr/provisioner_test.go
```diff
@@ -1014,3 +1096,442 @@ func metaFindCondition(st fv1.FunctionStatus, ct string) *metav1.Condition {
...
+
+		// remove provisionedconcurrency from function
+		fn.Spec.ProvisionedConcurrency = nil
+		fn, err = p.fissionClient.CoreV1().Functions(fn.Namespace).Update(t.Context(), fn, metav1.UpdateOptions{})
+		require.NoError(t, err)
+		p.reconcileFunction(t.Context(), fn)
+		synctest.Wait()
+		assert.Nil(t, fn.Spec.ProvisionedConcurrency, "expected provisionedconcurrency to be nil")
+		sched = p.scheduleFor(fn)
+		assert.Nil(t, sched.timer)
```
sanketsudake: **[Medium] `assert.Nil(t, sched.timer)` can never fail — it inspects a freshly created `fnSchedule`.**

`disableProvisioningLocked` **deletes** the `p.timers` entry, so `p.scheduleFor(fn)` on the preceding line `LoadOrStore`s a brand-new `&fnSchedule{}` whose `timer` field is always nil — regardless of whether `Stop()` ever ran. Deleting the `sched.timer.Stop()` call from `disableProvisioningLocked` entirely leaves this assertion passing.

The sibling `TestProvisionerScheduleStopProvisionerDeleteFunction` gets this right by checking `p.timers.Load`. Same approach here:

```suggestion
		_, hasSched := p.timers.Load(fn.UID)
		assert.False(t, hasSched, "expected disableProvisioning to drop the function's schedule entry")
```

pkg/apis/core/v1/validation.go
```diff
@@ -654,22 +654,46 @@ func (sc *StateConfig) Validate() error {
...
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "name is required"))
+		}
+		if err := IsValidCronSpec(window.Start); err != nil {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "start is invalid: "+err.Error(), "window name: "+window.Name))
+		}
+		if d, err := time.ParseDuration(window.Duration); err != nil {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration is invalid: "+err.Error(), "window name: "+window.Name))
+		} else if d <= 0 {
+			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration must be > 0", "window name: "+window.Name))
+		}
```
sanketsudake, line 686: **[Medium] Admission still accepts window durations the executor cannot honor.**

With `minTransitionDelay = 10 * time.Second` in `armTransition`, a window shorter than 10s has both its edges inside one timer tick, so the timer can never land inside it — only the 30s drift ticker can. Measured over 60 phase offsets on a per-minute window: a `5s` window is observed open in **16.7%** of its instances, `1s` in **3.3%**. `duration: "5s"` passes admission today and then silently does nothing most of the time.

This needs a companion const in the `fv1` package (verified: builds, and `go test ./pkg/apis/core/v1/...` stays green):

```go
// minProvisionedWindowDuration is the shortest schedule window the executor
// can honor: poolmgr'''s armTransition floors its transition timer at the same
// value (minTransitionDelay), so a shorter window'''s edges fall inside one
// timer tick and the window is usually skipped entirely rather than warming.
const minProvisionedWindowDuration = 10 * time.Second
```

Best end state is one exported constant that both this check and `armTransition`'''s floor consume, so the validator and the timer cannot drift apart.

```suggestion
		} else if d <= 0 {
			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i, "duration must be > 0", "window name: "+window.Name))
		} else if d < minProvisionedWindowDuration {
			errs = errors.Join(errs, MakeValidationErr(ErrorInvalidValue, "FunctionSpec.ProvisionedConcurrency.Windows", i,
				fmt.Sprintf("duration must be >= %s (the executor arms window timers with a %s floor, so shorter windows are usually skipped entirely)",
					minProvisionedWindowDuration, minProvisionedWindowDuration), "window name: "+window.Name))
		}
```

sanketsudake (commented):
## Re-review of the fix round

Verified all 16 inline findings against the new head `07a3649` in an isolated worktree. **All 16 are correctly fixed** — and several went further than what I suggested. `go build ./...`, `go vet`, and `go test -race` on `poolmgr`, `apis/core/v1`, `fission-cli/cmd/function` and `reaper/...` are all green locally; CI is green on every leg (FOSSA pending as always).

I re-derived the two highest-stakes fixes from scratch rather than trusting the diff:

- **`MarkSpecializationFailure` removal** — re-traced on the new head. `svcWaiting++` occurs only in `getSvcValue` and `ReserveCapacity`; `GetFuncSvc` → `pool.getFuncSvc` reaches neither. Removal is correct, and the replacement comment in `TestProvisioner_eagerSpecialize` now states the right reason rather than the inverted one.
- **`lastSched` + `windowActiveAt`** — both applied as suggested, and `lastSched` is still correctly used by `reBuildWindows`, where the new zero-time guard now returns the real last fire instead of `capped=true`. That's a quiet bonus: windows crossing cron's 5-year horizon now get a precise timer instead of silently falling back to the 30s ticker.

Nice catch going beyond the suggestion on the window bound: adding `+kubebuilder:validation:MaxItems=32` to `types.go` and regenerating means the apiserver rejects oversized lists even when the webhook is down, which is the failure mode that actually mattered. I confirmed the regeneration correctly propagated to `fission.io_functionversions.yaml` too, since it embeds `FunctionSpec`.

### One new finding, and it is a consequence of my own suggestion

I recommended the `minTransitionDelay` floor and flagged only that it "may need those windows widened". That undersold it — the floor has a user-visible correctness edge, and the test change from `duration=15s` → `duration=60s` in `TestProvisionedScheduledWindowCloseReapsPods` is the symptom.

Once a window is shorter than the 10s floor, its open and close edges both fall inside a single timer tick, so the timer can no longer land inside the window at all. The only thing left that can observe it open is the 30s drift ticker. I measured how often that happens, sweeping 60 phase offsets over a per-minute window:

| window duration | window instances the drift ticker ever observes open |
|---|---|
| `1s` | **3.3%** |
| `5s` | **16.7%** |
| `9s` | **30.0%** |
| `30s` and longer | 100% |

So a user can write `duration: "5s"`, pass admission cleanly, and get a window that silently does nothing ~83% of the time — no error, no log, no status condition. Validation currently only enforces `d > 0`.

The fix is to make admission reject what the executor cannot honor, tying the bound to the same value as the timer floor. Suggestion attached inline; I verified it builds and keeps `pkg/apis/core/v1` tests green.

Worth pairing with a one-line note on the `Duration` field doc in `types.go`, and ideally exporting a single constant that both `validation.go` and `armTransition` consume so the two can never drift apart.

### Still open from the previous summary (all non-blocking)

None of these are correctness issues and none block merge — listing them only so they aren't lost:

- `getProvisioinedWindow` typo, now at 7 sites across `create.go` and `function_test.go`.
- `reBuildWindows` still stores failed windows as `nil` map values, so both `armTransition` loops still need `if pw == nil || pw.capped`.
- The cron parser option set is still hand-built in 5 places.
- `TestProvisionerScheduleTickerTimer`'s `assert.Equal(t, int64(2), listCalls2.Load())` still holds under every interleaving, so it does not actually pin serialization by `lockFor`.
- The PR description still documents `nextTransitionAt(cfg, now)`, which does not exist on the branch (it was removed in a later stage commit).

Everything that was actually broken is fixed. From my side this is good to merge once you've decided how you want to handle the short-window bound — that one is a judgment call about whether to reject at admission or just document the granularity.

pkg/apis/core/v1/validation.go
sanketsudake: Please see if fix required, you can piggyback in another PR

Issue #3628, opened 2026-07-31
Epic: Per-function environment variables and flexible Secret/ConfigMap mapping


sanketsudake:
Epic for giving a Function first-class, per-function configuration: **literal env vars, key-level Secret/ConfigMap references, whole-object projection, and a configurable mount path — with identical semantics on every executor.**

## Why this epic

Today the only per-function configuration channel is a pair of name-only references (`FunctionSpec.Secrets`/`ConfigMaps`) whose meaning *differs by executor*: poolmgr and newdeploy run a fetcher that writes them as files under a fixed `/secrets/<ns>/<name>` path, while the container executor injects them as whole-object `envFrom`. There is no way to set a plain `DATABASE_URL=...` at all without owning the Environment's podspec — which is per-*environment*, so values leak across every function sharing it.

Four long-standing issues ask for pieces of this:

- #1545 — *Set environment variables from configmaps and secrets* — the core 12-factor ask; every mainstream FaaS (Lambda, Cloud Functions, OpenFaaS) ships per-function env vars.
- #1556 — *Flexible secret/configmap mapping* — key selection, renaming, prefixing instead of all-or-nothing object references.
- #2407 — *Custom secret/configMap file path* — the file-projection path is hardcoded; code written for any other platform's layout must change.
- #3241 — *Insecure `secretKeyRef` exposure patterns* — informs the security section (reserved-name denylist, authenticated specialize channel, no resolved values in pod args).

The per-executor divergence is worse than a gap — it is a trap: moving a function between the container executor and poolmgr/newdeploy silently changes its configuration from env vars to files.

## What "done" looks like

1. **`FunctionSpec` carries `env` (`corev1.EnvVar`, literals + `secretKeyRef`/`configMapKeyRef`) and `envFrom` (`corev1.EnvFromSource`, optional prefix)** — upstream types, so `kubectl explain`, GitOps schemas, and user intuition align; webhook narrows them.
2. **Identical observable semantics on poolmgr, newdeploy, and container executors** — native injection on per-function-pod executors; specialize-time injection (references-only on the wire, resolved pod-locally by the fetcher) behind a runtime capability gate on poolmgr.
3. **Rotation propagates**: rotating a referenced Secret/ConfigMap recycles exactly the referencing functions, including references made only via the new fields.
4. **`MountPath` on the existing references** (#2407): redirects the file projection, constrained to the `/secrets` and `/configs` roots for cross-executor consistency.
[...]

Issue #2407, opened 2022-04-08
ability to change secret/configMap file path


edokan:
**Is your feature request related to a problem? Please describe.**
I have multiple environments with different namespaces, and it makes it very hard to set configuration externally since secrets/configMaps are mapped using their namespaces too, forcing the function to know which namespace it should be and the name of the secret too. 

**Describe the solution you'd like**
It would be very nice to be able to map a secret or configMap to a folder I can specify. something like

`fission fn create --name brevity --secret shared/connectionString:/etc/myapp/settings`

so I can map  `shared/connectionString` to `/etc/myapp/settings`

**Describe alternatives you've considered**
I can read namespace (https://stackoverflow.com/a/42449618/1000939), then read the secret file, but I still need to know the name of the secret. I cannot use a shared secret with different names in different places. 

**Additional context**
none

sanketsudake:
Implemented and merged in #3638 (RFC-0030 phase 3), on `main` as of v1.28 development.

You can now set a per-reference mount path:

```yaml
secrets:
  - name: prod-db-creds
    mountPath: db          # -> /secrets/db/<key>
configmaps:
  - name: app-config
    mountPath: conf        # -> /configs/conf/<key>
```

An empty `mountPath` keeps the historical `<namespace>/<name>` layout byte-for-byte, so nothing changes for existing functions.

**One deviation from the literal request, stated plainly:** the path is relative to the `/secrets` and `/configs` roots, not the container filesystem root — so `/etc/myapp/settings` is refused at admission with that reason rather than silently reinterpreted. The reason is poolmgr: generic pool pods share a fixed set of emptyDirs frozen at pool creation, so an arbitrary absolute path is not materialisable there at all. Honouring absolute paths would have made this work on newdeploy/container and silently fail on the default execution backend. The RFC chose identical semantics across all three executors instead.

That still solves the problem you actually described — the namespace and object name no longer leak into your function code, so the same function works unchanged across dev/staging/prod namespaces and with a differently-named secret in each.

Works on all three executors: the fetcher honours it for poolmgr/newdeploy, and the container executor reproduces the same layout with native projected volumes.

Issue #1556, opened 2020-03-17
feature request: flexible secret/configmap mapping


shaunc:
I have two secrets in another namespace, which I'd like to use to control how a function connects to a database: in namespace `foo` (say) `database-config` and `legacy-database-config`.

I would like to do the following: `fission fn create bar --secret db-config=foo/database-config`

which would map the secret keys to (say):
`/mappedsecrets/bar/db-config/...`

Here `bar` is included in the path to disambiguate over different functions.

vishal-biyani:
Hi @shaunc I am not sure I understand the use case fully. Are you trying to use two separate secrets for same function based on which environment it is invoked in, can you please elaborate?

sanketsudake:
This has been inactive for 2207 days and is being marked stale. It will be closed in a future scrub if there's no further activity. Comment to keep it open.

sanketsudake:
Status update — **partially implemented**, and the remaining half is the one that matters for most users, so this stays open.

RFC-0030 (epic #3628) phase 1 shipped on `main`: `FunctionSpec` now carries first-class configuration —

- literal environment variables,
- key-level references into Secrets/ConfigMaps (`valueFrom`),
- whole-object projection (`envFrom`, with optional prefix).

**These work today on the `newdeploy` and `container` executors only.** They are rejected at admission for `poolmgr`, which is the default executor — so for most installs this is not yet usable.

The reason is not an oversight. Poolmgr serves functions from a **pre-warmed generic pool**: the pod is already running, with its environment already set, before anyone knows which function it will serve. Injecting per-function env vars requires the language runtime to accept them on the specialize request and apply them at specialization time. That is a coordinated change in `fission/environments`, per language.

The contract is designed and additive: the executor adds `envVars` to the specialize request, and the runtime **acknowledges what it applied** in the response. The acknowledgment is the load-bearing part — `/v2/specialize` returns 200 with an empty body on every runtime today, so its absence unambiguously means "old runtime", and a function declaring env vars against one is **refused, naming the environment**. Silently yielding an empty env var would be a valid-looking value and exactly the wrong-config failure this RFC exists to prevent.

Rollout is per-language and incremental (python and node first); no runtime is forced, and every existing function keeps working unchanged.

**Explicitly rejected:** keying warm pools on `(environment, env-set)`. It would deliver this on poolmgr with zero runtime changes, but it fragments the pool and costs a cold start per distinct env set.

[...]

Issue #1545, opened 2020-02-26
Function environment variables from configmaps/secrets


soamvasani:
Today, we supply ConfigMaps and Secrets to functions only through files.  Supplying them to the function through environment variables instead has been a popular user request.  So far, we've avoided this because:

 1. Poolmgr starts the container before being associated with a function, so at process start time we don't actually know what env vars a user wants.
 2. Newdeploy could do it trivially (just a change to the podspec), but then we'd have inconsistent features across executors, something we've managed to avoid so far.

However, it's possible to implement this feature.  Options I can think of:

 1. Accept inconsistent features across executors, and enable newdeploy functions to have env vars.  So a user can choose either fast cold starts or env vars. 
 1. Proposal #1536, make poolmgr more like newdeploy and start the function process late enough to get env vars.  So all functions can use env vars, but all cold starts become slower. 
 1. Modify every environment runtime to support env var modification at runtime.  We could do this by creating a new env version, adding env vars to the specializeRequest, and interpreting that in the language runtime (i.e. one implementation in each of python, node, go, java, ...).  It's relatively more work, but you get no latency regression.

shaunc:
Hmm... would another possibility be #4 optionally run functions in separate /bin/sh instance? This adds a little overhead per run, but not so much, perhaps.

alextes:
I feel like 1. would force use of env vars from files anyway. Fast cold starts are a pretty killer feature.

Some form of 3. sounds most reasonable to me. I mean, that is what I will implement in userland. Read the files on start, then make these key-value pairs available the way they would have been available if they had been in the environment.

justinfx:
FWIW I would expect an approach proposed in your first suggestion, where newdeploy needs to diverge in features from poolmgr in order to become more useful and flexible, because it also aligns with my request in #1820 to extend newdeploy functionality to override all aspects of the Environment spec. This would be inconsistent with the poolmgr as well because of how pods are precreated and reused.

sanketsudake:
Status update — **partially implemented**, and the remaining half is the one that matters for most users, so this stays open.

RFC-0030 (epic #3628) phase 1 shipped on `main`: `FunctionSpec` now carries first-class configuration —

- literal environment variables,
- key-level references into Secrets/ConfigMaps (`valueFrom`),
- whole-object projection (`envFrom`, with optional prefix).

[...]

PR #3631, opened 2026-07-31
RFC-0030 phase 1: per-function environment variables (newdeploy + container)

sanketsudake:
## TL;DR

RFC-0030 ([#3628](https://github.com/fission/fission/issues/3628)) phase 1: functions get first-class per-function configuration — literal env vars, key-level Secret/ConfigMap references, and whole-object projection — on the newdeploy and container executors, with rotation working from day one.

Closes the core of #1545 and #1556 for those two executors. Poolmgr (#3628 phase 2) and `MountPath` (#2407, phase 3) are **gated with an explicit rejection**, not silently ignored.

## Why this shape

Today the only per-function configuration channel is a pair of name-only references whose meaning *differs by executor* — poolmgr/newdeploy write files, the container executor injects env vars — and there is no way to set a plain `DATABASE_URL` without owning the Environment's podspec (which leaks across every function sharing it).

## Review commit-by-commit

1. **`feat(crd)`** — `FunctionSpec.env`/`.envFrom` (upstream `corev1` types, so `kubectl explain` and GitOps schemas work) plus `MountPath` on the reference types; validation; the RFC-0025 classifier extension (its coverage test forced the decision, as designed).
2. **`feat(executor)`** — `ApplyFunctionEnv` on newdeploy/container; the container executor's legacy synthesis skip; **rotation propagation** (cms watcher index + `ReferencedResourcesRVSum` extended to env-referenced objects).
3. **`feat(cli)`** — `--env-var`/`-e`, `--env-from-secret`, `--env-from-configmap` on create/update and run-container/update-container.
4. **`fix(rfc-0030)`** ×2 — the review-battery fixes (below).

## The three things worth reviewing hardest

**1. Ordering, not validation, is the security guarantee.** `EnvFrom`-expanded names are the referenced object's *data keys* — unknowable at admission and mutable afterwards — so a denylist alone cannot protect platform names. The executor therefore emits platform vars **last** on the container, and kubelet last-wins makes `FISSION_STATE_URL` and friends unconditionally authoritative. The denylist at admission is fast, friendly feedback on literals.

**2. Rotation ships with the fields, not later.** Kubernetes never refreshes env in a running container. If the watcher index and RVSum did not cover `valueFrom`/`envFrom` references, rotating a Secret referenced only via the new fields would propagate to *nothing* — silently stale credentials, worse than the gap being closed.

**3. Two phase-1 limits are documented rather than papered over.** Native injection expands `envFrom` before container `env`, so an environment-podspec literal still beats an `EnvFrom` key of the same name, and reserved names appearing as data keys are not filtered on this path. Both land with the poolmgr phase, which resolves values itself. The code comments say what phase 1 does, not what the end state will.

## What the review battery caught (all fixed in this branch)

Round 1 — two reviewers independently found that **none of the validation ran at admission**: the rules were in `FunctionSpec.Validate()`, but the webhook calls `ValidateForAdmission()`, so every rule was enforced on the CLI path only and bypassed by `kubectl`/GitOps. Also: newdeploy applied env *before* the environment-podspec merge (inverting precedence, and hard-failing on a same-named var) and targeted the wrong container name when the environment declares a custom runtime container.

Round 2 — an **env-only update never reached the pods**: `updateFunction`'s change detection didn't watch the new fields, so `fission fn update --env-var …` updated the CR while the container kept the old value. Fixed by extracting a `podTemplateChanged` seam with a table test that fails if a pod-template field is ever added without being listed.

Also: split-namespace installs (`functionNamespace` set) now refuse env *object references*, because the kubelet would resolve them in the pod namespace while rotation tracking keys on the function namespace — literals still work there; container-executor `envFrom` alongside legacy `secrets`/`configmaps` is rejected rather than silently dropping the legacy projection; a container-name mismatch is a loud error rather than a silent drop.

## Backward compatibility

All fields optional. With none set, pod specs are byte-identical on both executors (`ApplyFunctionEnv` early-returns; the newdeploy `platformEnv` change is a pure refactor), and the container executor's legacy synthesis is preserved byte-for-byte. `--env` keeps meaning the Environment name; `-e` was free on create/update. The RVSum's new dedup cannot change any existing function's value (`listType=map` on both reference lists makes in-list duplicates impossible).

## Testing

`make code-checks`, `make license-check`, and `go test -race` over `pkg/apis/core/v1`, `pkg/executor/...`, `pkg/fission-cli/...`, `pkg/versioning`, `pkg/webhook/...` all green. `make codegen` + `make generate-crds` re-run (the applyconfiguration builders were missing in the first pass — server-side-apply clients could not have set the new fields; `provisionedwindow.go` in that diff is pre-existing generator drift on main being swept in).

New tests cover: the reserved-name list, every validation rule **through `ValidateForAdmission`** (the path the webhook uses), injection ordering including a user literal shadowing a platform name, the cross-namespace guard, RVSum with env references and cross-channel dedup, the classifier golden table, the CLI ref parser, and `podTemplateChanged`.


sanketsudake:
Good catch — added integration coverage in `test/integration/suites/common/function_env_vars_test.go`, plus the framework flags so they read like the rest of the suite.

They assert on what the **runtime actually received**, not on the pod spec: a pod-spec assertion would pass even if the kubelet resolved a value differently than intended, and what reaches the process is the whole point of the feature. A small echo-env fixture joins the vendored python testdata.

- **`TestFunctionEnvVars`** — E1 narrowed to the phase-1 executors: a literal var, a whole-object Secret projection, and key-level Secret/ConfigMap references both by their own key name and renamed.
- **`TestFunctionEnvVarRotation`** — E4, and the reason the watcher-index and RVSum extensions had to ship in phase 1 rather than later. The Secret is referenced **only** through the env field (the function carries no `--secret`), so nothing but the RFC-0030 index can notice the rotation; the value is changed in place with no function edit and no CLI update.
- **`TestFunctionEnvVarsPoolmgrRejected`** — the phase gate must reject rather than deploy a function whose `DATABASE_URL` would be silently empty.
- **`TestFunctionEnvVarsReservedNameRejected`** — the admission half of the denylist, over `FISSION_STATE_URL` (redirecting it would exfiltrate the RFC-0023 state token), `LD_PRELOAD` and `HTTPS_PROXY`.

Also extended `framework.FunctionOptions` with `EnvVars` / `EnvFromSecrets` / `EnvFromConfigMaps`, so future tests get the flags for free.

Issue #3626, opened 2026-07-31
Epic: GitOps-native install & declarative management


sanketsudake:
Epic for making Fission a first-class citizen of declarative, pipeline-driven platform management: **install it, upgrade it, and manage every function on it from Git, with no imperative side-channels.**

## Why this epic

This is the largest cluster of reactions in the open backlog — a set of issues that are all the same complaint from different angles: *Fission assumes a human with a CLI, but 2026 platform teams deploy through Argo/Flux/Terraform and expect everything to be a manifest.*

- #2382 — *Include CRDs in Helm Chart* — **12 reactions.** CRDs install out-of-band (`kubectl create -k crds/v1` / `make create-crds`), which breaks single-shot Helm/Argo/Flux installs and version-skews CRDs against the chart.
- #1907 — *Terraform template support* — **12 reactions.** No IaC story for Fission objects.
- #2383 — *Allow file contents inline in values.yaml* (8 reactions) — Kafka TLS certs can't be supplied without unpacking the chart.
- #1687 — *Kubectl Apply for Fission Specs* (7 reactions) — users want plain `kubectl apply`/GitOps for what `fission spec apply` does.
- #2906 — *Use .Release.Name as prefix for resources* / #2835 — *multiple instances per cluster*.
- #2742 — *Fission Operator* — manage Fission itself via CRDs.
- #2928 — *Package build trigger mechanism (esp. with GitOps)* — when specs are applied from Git, rebuilds must trigger from content change, not CLI invocation.
- #740 (closed → folded here) — the surviving half is `spec apply --set` templating.

Plus the long tail of `fission spec` correctness/UX issues that erode pipeline trust: #543 (rename handling), #551 (duplicate archive names), #555 (`--watch` vs builder conflicts), #623 (empty ArchiveUploadSpec validation), #1037/#1574 (scaffolding), #1491 (misleading duplicate-resource error), #3230 (spec-file defaulting diverges from CLI), #3296 (false-positive duplicate detection on reapply).

## The strategic observation

The hard blocker for "just `kubectl apply` everything" was always the **package archive**: a Function references a Package whose archive lives in storagesvc, uploaded imperatively by the CLI — un-GitOps-able by construction. That blocker has quietly dissolved:

- **RFC-0001/0012 (shipped, default cold-start path)**: a package can be an **OCI image reference** — content-addressed, built in the user's own CI, pushed to their registry. No storagesvc upload step, nothing imperative left.
[...]

Issue #2382, opened 2022-03-14
Include CRDs in Helm Chart


torbendury:
**Is your feature request related to a problem? Please describe.**
When rolling out `fission` to our k8s clusters, it's always an extra step to install the CRDs from GitHub. Also, this breaks our CI/CD when we're not able to have the Chart + the CRDs in one space with the same version.

**Describe the solution you'd like**
Helm enables creators to include the needed CRDs directly in the Helm Chart: [Helm Best Practices - Custom Resource Definitions](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)

**Describe alternatives you've considered**
none

**Additional context**
This enables `fission` users to keep the CRDs (plus their versions) in sync with the Chart itself.

blackfly19:
Hi @torbendury. The issue in using this way of creating crds is that helm uses the following command to create crds:

`kubectl apply -k "github.com/fission/fission/crds/v1?ref=v1.15.1"`

If you try using it normally you get a memory limit issue:

`metadata.annotations: Too long: must have at most 262144 bytes`

Therefore it's not possible to create crds this way for fission.

torbendury:
Hi @blackfly19 and thanks for your answer! I'm aware of this issue - maybe [this workaround](https://github.com/argoproj/argo-cd/issues/820#issuecomment-563709649) is valid for fission?

janosroden:
@blackfly19 I'm not sure this is an issue anymore, I was able to install crds with helm:
```
helm pull fission-charts/fission-all --untar --version v1.15.1
mkdir fission-all/crds
kubectl create --dry-run=client -o yaml -k "github.com/fission/fission/crds/v1?ref=v1.15.1" > fission-all/crds/crds.yaml
helm install fission fission-all -n fission
```

ammarlakis:
That would be quite nice to have, but helm doesn't support updating CRDs. https://github.com/helm/helm/issues/7735

sanketsudake:
Implemented as part of the RFC-0029 GitOps work (epic #3626), on `main`.

The chart now ships the CRDs and installs them itself. `crds.mode` selects the delivery mechanism:

- **`hook` (default)** — the pre-install/pre-upgrade checks Job server-side-applies the CRD bundle embedded in its image before the release manifests roll. `helm install fission` is one-shot on a fresh cluster, and CRDs are upgraded with the chart.
[...]

PR #3632, opened 2026-07-31
RFC-0029 phase 1: deliver CRDs with the chart (crds.mode hook)

sanketsudake:
## TL;DR

RFC-0029 ([#3626](https://github.com/fission/fission/issues/3626)) phase 1, first slice: **`helm install fission` on a fresh cluster no longer needs `kubectl create -k crds/v1` first**, and CRDs upgrade with the chart instead of skewing against it.

Closes the core of [#2382](https://github.com/fission/fission/issues/2382) (12 reactions, the highest-reacted issue in this cluster).

## Design

`crds.mode` selects the delivery mechanism, **single-writer by design** — exactly one mechanism ever writes the CRD objects, because two server-side appliers under different field managers is an ownership fight nobody wins.

| mode | behavior |
|---|---|
| `hook` (default) | the existing pre-upgrade-checks Job gains `pre-install` and server-side-applies the CRD bundle **embedded in its own image**, before the release manifests roll |
| `none` | bring your own — exactly today's behavior |

The bundle is `go:embed`-ed into the checker binary (new `crds` package) because the hook image is a single-binary `chainguard/static` image with no filesystem to carry YAML alongside it — and it guarantees the binary and the schemas ship as one versioned artifact. Applying in the hook also delivers RFC-0028's CRDs-before-controllers ordering for free.

The RFC's third mode (`manifests`, CRDs templated for Argo diffing) needs the generated YAML packaged **inside** the chart — Helm's `.Files` cannot read outside the chart directory — so it fails render with that explanation rather than being half-shipped.

## Three things worth reviewing hardest

**1. Server-side apply is mandatory, not a preference.** Four of the generated CRDs exceed the 262,144-byte client-side-apply annotation limit (`environments` ~1.4 MB, `functionversions` ~696 KB, `functions` ~657 KB, `messagequeuetriggers` ~573 KB), so client-side apply cannot round-trip them at all.

**2. The RBAC grant.** This is the first time any Fission component gets CRD **write** — every existing clusterrole is `get/list/watch`. It is one rule, `{create, patch, update}` fenced by `resourceNames` to Fission's 13 CRDs, no delete, no wildcards. A `ValidatingAdmissionPolicy` bound to the hook's ServiceAccount adds what RBAC cannot express: the hook may not set a conversion strategy on a CRD it legitimately owns (which would interpose on every read and write of that type cluster-wide).

> The RFC specified an **unscoped** create rule with the policy as the only fence, on the premise that "resourceNames cannot restrict create". Review testing against a live apiserver showed that premise is wrong for how this hook writes: SSA is a PATCH to a *named* resource, so `resourceNames` does fence the create authorization SSA needs when the target is absent. The grant is correspondingly narrower than the RFC asked for, and the policy is now genuine defence-in-depth rather than load-bearing.

**3. `forceConflicts` (default true).** CRDs from the documented `kubectl create -k crds/v1` flow carry managedFields owned by `kubectl-create` as an *Update*, so an apply from a new field manager conflicts on exactly the fields a schema-changing upgrade must change. Forcing adopts them — and, honestly documented, also silently wins against any other writer, so `crds.forceConflicts=false` turns that into a hard error once an install is past adoption.

## What the review battery caught

Two reviewers ran against a live apiserver (envtest) rather than reading:

- **Fresh install failed deterministically.** The preupgrade *namespaced* Role/RoleBinding — the only grant of `list` on fission.io types — never got `pre-install`. With CRDs now applied by the hook, the checker no longer short-circuits on a fresh cluster, so it listed Functions and got a 403 **after** applying the CRDs.
- **The install raced its own CRDs**: a CRD is not servable for ~100 ms after the apply returns, and the existing retry loop burns all five attempts in 3 ms with no backoff. `ApplyCRDs` now waits for `Established`.
- **`endsWith("fission.io")` in the policy admits `evilfission.io`** — now exact equality, matching the applier's own check.
- The hand-maintained chart CRD-name list is now guarded by a test comparing it against the embedded bundle (mutation-tested — removing a name fails it). Without it, adding a CRD and forgetting the helper fails on users' upgrades, not in CI.
- `hook-failed` added to the hook RBAC delete policy, so a failed hook does not leave cluster-scoped CRD write behind indefinitely.

## Backward compatibility

`crds.mode=none` preserves today's flow exactly; `make create-crds` / `kubectl create -k crds/v1` are retained. The default `hook` mode applies *over* hand-applied CRDs with no adoption step and never deletes a CRD, so no user data can be lost by an install-mechanism choice. The new grant is named in values.yaml and avoidable via `crds.mode=none` for orgs that restrict CRD write.

## Testing

`make code-checks`, `make license-check`, `helm lint`, and `go test -race` green. Rendered and diffed every mode path: default, `crds.mode=none` (no policy, no write RBAC), `crds.mode=manifests` (fails with explanation), `crds.mode=bogus` (fails), and `hook` + `preUpgradeChecks.enabled=false` (fails — the Job *is* the delivery mechanism, so a silent no-op would ship no schemas).


sanketsudake:
Heads-up on scope: I built the RFC-0029 §3 auth-material half (`internalAuth.existingSecret` + `autoGenerate` + a pre-install hook Job that mints the HMAC master only if absent) onto this branch and then **reverted it** — it is not in the PR as it stands.

Two confirmed cluster-breaking defects, and this PR's own sibling (the new upgrade-under-load leg from #3630) caught the first one in CI:

1. **`helm upgrade` prunes the existing master.** Every existing install has a chart-rendered, release-owned `fission-internal-auth`. Moving generation out of the template means Helm sees a removed release resource and deletes it. `internalAuth.envs` mounts that key with a **required** `secretKeyRef`, so the control plane wedges in `CreateContainerConfigError` — the upgrade leg failed precisely this way on `buildermgr`. Re-minting is not a recovery; it rotates every derived key with no dual-accept window.
2. **Fresh installs lose the per-namespace fan-out** the old template did deliberately (kubelet cannot resolve a cross-namespace `secretKeyRef`), which breaks fetcher sidecars → unsigned storagesvc calls → 401 on every archive fetch.

A correct version needs a `pre-upgrade` migration that stamps `resource-policy: keep` onto the live Secret before Helm computes its deletion set, plus a decision on namespace fan-out vs. the generator's create-only RBAC. The same migration is needed by `router/secret.yaml` and the webhook cert (RFC-0028 §2 defers it here), so it is worth building once as its own PR.

**This PR is therefore the CRD-delivery half only**, which is self-contained and green.

Issue #3625, opened 2026-07-31
Epic: zero-downtime upgrades & a predictable release lifecycle


sanketsudake:
Epic for making Fission's operational lifecycle trustworthy: **upgrading a running cluster must not drop traffic, and releases must be predictable enough to plan around.**

## Why this epic

The backlog says this is the adoption blocker, louder than any feature ask:

- #1856 — *Downtime during fission upgrade* — **15 reactions, the highest-reacted open issue in the repo**, filed 2020 and never addressed. Fifteen people reacted and none commented: that is the signature of a dealbreaker people walk away from, not a feature they wait for.
- #3245 — *Is there going to be a release soon?* — users cannot plan deployments because they cannot predict releases.
- #2777 — *Could we provide one or two LTS versions?* — teams want a supported line that receives fixes without forced feature upgrades.
- #776 — *Builder pod does not use latest fetcher container after upgrade* — a concrete upgrade-correctness bug: builder pods in `fission-builder` keep the old fetcher image after `helm upgrade`, so the fleet is silently version-skewed until someone recreates environments.
- #517 — *Need a release/build server for more reliable release* — the 2018 ancestor of this ask; goreleaser + GitHub Actions have since replaced laptop builds, but the release *process* (cadence, gates, upgrade testing) is still unspecified.

Everything shipped in the RFC-0021…0027 wave is depth for people already running Fission. This epic is about the people who evaluated it, asked "what happens when I upgrade?", and left.

## What "done" looks like

1. **Zero-downtime upgrade as a tested guarantee, not a hope.** A CI gate that installs release N−1, drives continuous load, `helm upgrade`s to HEAD, and asserts zero non-user 5xx. The building blocks exist and are recent:
   - `AdoptExistingResources` — the executor adopts function pods in place across a restart (same UID, instanceId flips), and the serial integration suite already exercises it.
   - The RFC-0002 EndpointSlice data plane — warm traffic is served from a slice-fed index per router replica, with the executor **off the hot path**, so an executor restart during upgrade need not touch warm invocations at all.
   - The RFC-0013 route table — routers rebuild routing state from CRDs on start without a thundering rebuild.
   - `cmd/preupgradechecks` — already runs as a pre-install/pre-upgrade hook; extend it rather than invent a new gate.
   What is missing is the *choreography*: upgrade ordering, webhook availability during CRD/webhook rollout, router drain/PDB posture, version-skew rules between components, and the CI proof.
[...]

PR #3633, opened 2026-07-31
RFC-0028 phase 2: rollout posture for router and webhook

sanketsudake:
## TL;DR

RFC-0028 ([#3625](https://github.com/fission/fission/issues/3625)) phase 2, the rollout-posture half: make a rolling upgrade survivable for the two components whose absence is immediately user-visible. Defaults flip; every old posture stays reachable through values.

The webhook's **serving-cert stability is deliberately not here** — see the bottom.

## Router

| Change | Why |
|---|---|
| `replicas` 1 → **2** | the router *is* the data plane; one replica makes every rollout, drain and eviction an outage |
| surge strategy (`maxSurge: 1`, `maxUnavailable: 0`) | explicit rather than the Kubernetes 25% defaults, which allow `maxUnavailable > 0` once scaled past 3 — exactly when a dip hurts |
| `terminationGracePeriodSeconds: 90` **+ new `gracefulShutdownTimeout: 75s`** | the grace period alone does nothing: the drain is capped at 30s in code unless `GRACEFUL_SHUTDOWN_TIMEOUT` is set, and no template set it |
| **new `preStopSleep: 5`** | the listener closed the instant SIGTERM arrived while kube-proxy still had the pod in its backend set → connection-refused, which the phase-1 gate counts as platform-attributed precisely because it carries no response and no attribution header |
| **topology spread** at `replicas >= 2` | otherwise both replicas can land on one node and a single drain takes out the whole data plane — the outage the flip exists to prevent |
| PDB on by default, behind a guard | see below |

## The PDB guard is the subtle part

A `minAvailable: 1` PDB over a genuinely single-replica Deployment blocks node drains **forever**. So `replicas: 1` correctly drops the PDB rather than deadlocking the cluster.

But an HPA/KEDA-scaled router keeps `replicas: 1` in its spec while running more, so an autoscaler counts as evidence too — **only when its own `minReplicas >= 2`**. Both autoscalers default to `minReplicas: 1`, so "an autoscaler is enabled" alone is not evidence the router ever runs more than one pod, and accepting it would have reproduced the exact deadlock the guard exists to prevent. (Review caught this; it renders correctly for all four combinations now.)

## Webhook

It had none of this surface: a `replicas` knob (default 2), a surge strategy, and a PDB behind the same guard. It is fail-closed on every configuration, so zero Ready endpoints means **every Fission CR write errors**.

It also had **no readinessProbe**, which made `maxUnavailable: 0` vacuous — the kubelet marked a pod Ready the moment the process started, so a surge pod joined the Service before it had bound `:9443`. Added a `tcpSocket` probe on the serving port.

## Router readiness now gates on the endpoint index

`RegisterInformer` returns the registration's `HasSynced` instead of discarding it. The Manager's cache sync says the *store* is populated, which is not the same as this handler having received the replay. Without it a Ready replica serves warm traffic through the executor fallback — correct, but a latency cliff precisely during the rolling upgrade this PR exists to smooth.

Verified this cannot wedge readiness: `HasSynced` is a monotone latch, and the first mux build already blocks on the same informer's cache sync, so anything that stalls this gate already stalls `ready`.

## Not here, on purpose

**Webhook serving-cert stability.** The default path mints a fresh CA per render, so the `caBundle` races the still-serving cert. Fixing it needs the same Helm-ownership migration as the internal-auth master — a `pre-upgrade` hook stamping `resource-policy: keep` onto the live object, which I verified on kind is the working mechanism (Helm reads that annotation from the *live* object at delete time). Three secrets need that migration; it belongs in one PR, not three partial ones.

## Testing

`make code-checks`, `make license-check`, `helm lint`, `go test -race ./pkg/router/...` green. Every guard path rendered and diffed: default, `replicas=1`, `replicas=1 + autoscaling`, `replicas=1 + autoscaling minReplicas=2`, webhook `replicas=1`. The readiness test now covers slice-plane-off, index-synced, and index-still-filling.

The benchmark leg is pinned to `router.replicas=1`: its pprof capture port-forwards to `items[0]`, and every recorded regression bar was measured single-replica.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -243,6 +243,12 @@ jobs:
           docker exec kind-control-plane sh -c \
             'mkdir -p /fission-coverage && chown 10001:10001 /fission-coverage && chmod 0700 /fission-coverage'
           SKAFFOLD_PROFILE=kind-ci make skaffold-deploy
+          # Pin a single router for the benchmark only. The chart default is 2
+          # (RFC-0028), but this leg's pprof capture port-forwards to
+          # items[0] — one replica's profile against the whole offered load —
+          # and every recorded regression bar was measured single-replica.
+          kubectl -n fission scale deploy/router --replicas=1
+          kubectl -n fission rollout status deploy/router --timeout=3m
 
       - name: Install released Fission (helm)
         if: ${{ matrix.fission_version != 'HEAD' }}
@@ -265,6 +271,7 @@ jobs:
           helm install fission fission-charts/fission-all \
             --version "${ver#v}" -n fission --skip-crds --wait --timeout 8m \
             --set routerServiceType=NodePort \
+            --set router.replicas=1 \
             --set pprof.enabled=true \
             --set prometheus.serviceEndpoint=http://prometheus-operated.monitoring.svc.cluster.local:9090 \
             --set serviceMonitor.enabled=true \
--- a/charts/fission-all/templates/router/deployment.yaml
+++ b/charts/fission-all/templates/router/deployment.yaml
@@ -38,6 +38,18 @@ spec:
       containers:
       - name: router
         image: {{ include "fission-bundleImage" . | quote }}
+        {{- if gt (int (.Values.router.preStopSleep | default 0)) 0 }}
+        lifecycle:
+          preStop:
+            # Delay the drain so this pod's removal from the Service
+            # EndpointSlice reaches every kube-proxy BEFORE the listener
+            # closes. Without it a terminating router still receives
+            # connections it can no longer accept: connection-refused, which
+            # the RFC-0028 §1 gate counts as a platform failure precisely
+            # because it produces no response and no attribution header.
+            sleep:
+              seconds: {{ int .Values.router.preStopSleep }}
+        {{- end }}
         imagePullPolicy: {{ .Values.pullPolicy }}
         command: ["/fission-bundle"]
         args:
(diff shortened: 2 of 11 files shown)


sanketsudake:
CI note: the first run failed the v1.34.8 integration leg with builds stuck in `running` and executor RPC deadlines across several unrelated tests — cluster-wide contention rather than a specific test breaking.

That was this PR's own doing: neither the router nor the webhook Deployment sets resource requests, so the extra replicas are best-effort and simply take CPU on a ~4-vCPU kind runner. Pinned both to 1 in `skaffold.yaml`'s `setValues`, which covers local dev and the integration legs.

The division that leaves is the right one: the integration suite proves functional correctness on an unstarved cluster, and the **upgrade leg** — which installs the HEAD chart with its real defaults, and passed on this PR — is where the two-replica rollout posture actually gets exercised. That is the leg whose entire purpose is the rollout, so it is the honest place for it.

sanketsudake:
CI status after the contention fix: **v1.34.8 now passes** (it failed before the single-replica pin), v1.32.11 is running, and v1.36.1 fails with 9 unrelated tests collapsing in one window — alias create, websocket dial to `router-internal`, workflow engine, env drift, retention sweep.

That is the documented v1.36.1 cluster-collapse signature, not this change: the same leg fails on `main` itself (run 30653111405), and a deterministic chart change would fail every leg identically rather than one.

Worth flagging separately: v1.36.1 is currently red across `main` and multiple unrelated branches while the other two legs pass on identical commits. It is costing a rerun cycle on nearly every PR and deserves its own investigation (runner sizing, or the `kindest/node` v1.36.1 image) rather than per-PR reruns.

PR #3648, opened 2026-08-02
test: fix two CI flakes blocking main (workflow delivery counts, run --watch registration race)

sanketsudake:
Two unrelated unit-test flakes are turning `lint` red on main and on everything branched from it (#3639 hit the first, #3640 the second). Both are test bugs — neither product path is at fault — and each is fixed in its own commit.

---

## 1. `TestEngineLinearPipeline` — asserting a guarantee the engine does not make

```
engine_test.go:189:
    expected: map[string]int{"fn-a":1, "fn-b":1}
    actual  : map[string]int{"fn-a":1, "fn-b":2}
```

That line is the **only** failing assertion. The run is `Succeeded`, the recorded output is `{"fn":"fn-b","call":1}` — the *first* call's response — and `assertInvariants` passes: one `EvStepScheduled` and one result per attempt.

That combination is the engine working. `execute()` sets the idempotency headers deliberately:

```go
// Idempotency contract: functions dedup on (run, attempt).
req.Header.Set("X-Fission-Workflow-Run", iv.runUID)
req.Header.Set("X-Fission-Workflow-Attempt", strconv.Itoa(int(iv.attempt)))
```

Delivery is **at-least-once**. What is exactly-once is the *recorded outcome* (W1/W2), not the side effect. A redelivered attempt re-runs the function and its append then loses the result guard — leaving the log exactly right and the harness counter one high, which is precisely the observed state.

So the assertion tested a property the design deliberately does not offer, and could only fail under the contention that makes redelivery likely: CI, not a laptop. It does not reproduce locally in 100+ runs across `-race`, `-cpu=1`, and `-cpu=2`.

`engine_test.go:364` already had the right shape (`GreaterOrEqual(..., 1, "at-least-once execution")`) for the crash-and-resume case; four other sites did not.

**Fix.** Exact assertions move onto the log, where the guarantee lives, via a new `attempts(log, state)` helper counting recorded `EvStepScheduled` per state. The harness counter becomes a lower bound. The two sites where the exact number *is* the contract keep exact expectations, now against `attempts`:

- `TestEngineRetryThenCatch` — `attempts(log, "a") == 2`, "retried to budget"
- `TestEnginePermanentErrorFailsRun` — `attempts(log, "a") == 1`, "4xx never retries"

Both still fail if the retry policy regresses (mutation-verified: `2`→`3`, `1`→`9` both compile and fail, then pass on revert).

Also routes the counter through `callCount()` under the harness mutex. It is written from the httptest handler goroutine, and the very redelivery above can still be in flight at terminal — the unsynchronised map read was a real race.

---

## 2. `TestRunLocal*Watch*` — writing before the watcher exists

```
run_test.go:507: Condition never satisfied
  editing a file inside the source dir should trigger a re-specialize
```

Fails on the **second** wait only, for the full 3s: the initial specialize is observed, the edit that should re-specialize never lands.

The tests wait for the first specialize, then immediately write the source file. But `runLocal` specializes inside `launch()` and only *then* calls `serveAndWatch` — which is where `addWatchTree`/`watcher.Add` happen. Observing the first specialize does not mean the watch exists yet, and a write landing in that gap is lost for good: inotify has no replay. That is why the symptom is a full-timeout miss rather than a slow pass. On a laptop the gap is microseconds; on a loaded runner the goroutine can be descheduled straight across it.

**Fix.** Rewrite the file on every tick instead of once, via `requireReSpecialize`. The tick stays above `serveAndWatch`'s 250ms debounce — a faster loop would keep restarting `drainBurst`'s quiet period so no reload would ever fire, trading one flake for a worse one.

Both watch tests had the identical pattern; the single-file one had simply not lost the race yet, so both are converted. Mutation-verified that they still catch a broken watcher (`watchTriggers` stubbed to `false`: both fail; pass again on revert).

---

## Testing

`go test -race -count=20 ./pkg/workflow/` and `-count=8 -run TestRunLocalWatch ./pkg/fission-cli/cmd/function/` both clean. `make code-checks` and `make license-check` exit 0.

### DIFF
--- a/pkg/fission-cli/cmd/function/run_test.go
+++ b/pkg/fission-cli/cmd/function/run_test.go
@@ -421,6 +421,33 @@ func TestRunLocalPropagatesEnvAndPorts(t *testing.T) {
 	assert.Equal(t, 5005, f.lastSpec.Ports[1].Host)
 }
 
+// requireReSpecialize waits for an edit to path to trigger a re-specialize,
+// rewriting the file on every tick rather than once.
+//
+// serveAndWatch registers the fsnotify watch AFTER the initial specialize —
+// runLocal specializes inside launch(), then calls serveAndWatch, which is
+// where addWatchTree/watcher.Add happen. A caller that writes as soon as it
+// observes that first specialize can therefore land its write in the gap
+// before the watch exists, and inotify has no replay: the event is lost for
+// good and the reload never comes. Rewriting each tick closes the gap.
+//
+// The tick must stay above serveAndWatch's 250ms debounce. A faster rewrite
+// loop would keep restarting drainBurst's quiet period, so the burst would
+// never drain and no reload would fire at all.
+func requireReSpecialize(t *testing.T, f *fakeRuntime, path, content, msg string) {
+	t.Helper()
+	require.Eventually(t, func() bool {
+		if f.getSpecializePath() == "/v2/specialize" {
+			return true
+		}
+		// Errors are not asserted here: this runs on Eventually's goroutine,
+		// where FailNow is invalid. A failing write simply leaves the
+		// condition unsatisfied and fails with msg below.
+		_ = os.WriteFile(path, []byte(content), 0o644)
+		return false
+	}, 12*time.Second, 500*time.Millisecond, msg)
+}
+
 func TestRunLocalWatchReloadsOnChange(t *testing.T) {
 	f := &fakeRuntime{echo: "ok"}
 	code := writeTempCode(t, "v1")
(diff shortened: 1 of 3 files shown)


sanketsudake:
Superseded by #3649, which consolidates this branch with the other outstanding RFC-0028/0029 work.

All commits from this branch are preserved there, so nothing is lost — the merge was done with `--no-ff` per branch. Keeping these PRs parallel meant re-resolving the same append/append conflicts in `test/helm/portdrift_test.go` and the same chart templates on every rebase, which was generating more review noise than signal.

Review continues on #3649.

PR #3653, opened 2026-08-03
fix(router): re-scope EndpointSlice cache for runtime-onboarded tenants (#3647)

### LINKED ISSUE #3647: router.endpointSliceCache.mode=on: EndpointSlice informer not re-scoped for dynamically onboarded tenants (tenancy.mode=dynamic) — router needs restart per tenant
## Summary

In `tenancy.mode: dynamic` with `router.endpointSliceCache.mode: on` (both the RFC-0002 default and our deployment), the router's EndpointSlice informer is scoped at **startup time** to the namespaces the `NamespaceResolver` already knows. Namespaces onboarded at runtime via `fission tenant enable` / the `fission.io/enabled` label are added to the resolver later by the resolver-sync controller, but the informer's cache scope is never re-widened — so the router never sees the new tenant's EndpointSlices without a manual env change + rollout restart.

## Impact

For any namespace onboarded **after** the router started:

- `fission_router_endpointcache_size` stays 0 for that tenant's function Services (verified: `endpointcache_size 0`, `hits 0`, `misses N` in v1.27.0).
- Every request falls through to the executor-RPC cold path (`resolver_fallback.go` / `resolver_executor.go`).
- Cold path specialized a **new pod per request** in our setup; combined with the quota, this accumulated specialized pods and exhausted `limits.memory` (13 pods × 256Mi > 4Gi), wedging the pool deployment at `0/2` and timing out all requests.

So the "dynamic tenancy = no control-plane restart" promise does not hold for the EndpointSlice data plane.

## Root cause

`pkg/router/router.go:103` `routerCacheOptions` builds the cache's `ByObject.Namespaces` for `EndpointSlice` once, at manager construction, from `sliceWatchNamespaces()` → `FunctionNamespaces()` → the resolver's live set **at that moment** (in our case only `default`, from `FISSION_RESOURCE_NAMESPACES=default`).

`pkg/tenant/resolversync.go` `SyncResolverFromTenants` later calls `resolver.SetTenants(...)` in every data-plane process, so the *resolver* knows the new namespace — but the already-constructed controller-runtime cache is never re-scoped, and no informer re-registration happens. CRD caches are cluster-wide in dynamic mode (`crmanager.FissionCacheOptions`), which is why Functions/Triggers/Environments work for the new tenant but EndpointSlices do not.

## Second, related gap: router data-plane RBAC is not provisioned for dynamic tenants

- The tenant controller (`pkg/tenant/provision.go` `namespaceRBACObjects`) creates per-namespace RoleBindings for fetcher/builder/executor/buildermgr but **not** for the router's data-plane Role.
- The chart's `charts/fission-all/templates/router/role-dataplane.yaml` renders the `*-router-dataplane` Role+RoleBinding only for `additionalFissionNamespaces` (static mode). In dynamic mode these are empty, so a freshly onboarded tenant namespace has no `endpointslices`/`services` grant for `fission-router`:
  - `kubectl auth can-i list endpointslices -n <tenant> --as=system:serviceaccount:fission:fission-router` → `no`
- `checkSliceWatchRBAC` (`pkg/router/router.go:139`) only validates namespaces known at startup, so it does not catch this for later-onboarded tenants.

## Workarounds applied (operational, no fork)

1. Manually create a `fission-router-dataplane` Role + RoleBinding (endpointslices get/list/watch + services get/list/watch) in each tenant namespace, bound to SA `fission-router` in the release namespace.
2. Add the tenant namespace to `FISSION_RESOURCE_NAMESPACES` on the **router and executor** deployments, then `rollout restart` both (the executor also needs the namespace to reap/adopt poolmgr pods there).

## Expected behavior / suggested fix

- On tenant onboarding (resolver-sync `SetTenants`), the router should re-scope (or re-register) the EndpointSlice informer to the newly added function namespaces and re-run `checkSliceWatchRBAC` — restoring the "onboard without restart" guarantee for the data plane.
- The tenant controller should provision the router data-plane RBAC for dynamic tenants (mirroring `namespaceRBACObjects`), or `role-dataplane.yaml` should render for FissionTenant namespaces, not just `additionalFissionNamespaces`.

## Environment

- Fission `v1.27.0` (`fission-bundle` `ghcr.io/fission/fission-bundle:v1.27.0`, GitCommit `1e1401cd`)
- `FISSION_TENANCY_MODE=dynamic`, `ROUTER_ENDPOINTSLICE_CACHE_MODE=on`
- Helm chart defaults otherwise; tenant namespaces created at runtime via a FissionTenant CR

rijojohn85:
Fixes https://github.com/fission/fission/issues/3647

## Problem

With `tenancy.mode=dynamic`, onboarding a tenant namespace at runtime worked for routing/invocation, but the router's RFC-0002 warm path never saw the new tenant's EndpointSlices:

1. **Informer scope gap** — the router's EndpointSlice informer was scoped once at startup (`routerCacheOptions`). A namespace onboarded later was never watched, so every request to its functions fell back to the executor RPC forever.
2. **RBAC gap** — the router's ServiceAccount had no RoleBinding in a runtime-onboarded namespace, so even a re-scoped informer would have gotten `forbidden` LISTs.

### A. RBAC — grant the router SA per-tenant dataplane read
- New `fission-router-dataplane-tenant` ClusterRole (dynamic-mode chart partial, single-sourced from the same `router-dataplane-kuberules` the static Role uses — the two paths cannot drift).
- `fission-router` added to the ValidatingAdmissionPolicy SA allowlist.
- The tenant controller binds it per namespace at onboarding (`fission-router-dataplane` RoleBinding) and deletes it at offboard.

### B. Data plane — per-namespace informers that re-scope at runtime
- New `NamespaceInformers` (`pkg/router/endpointcache/nsinformers.go`): one client-go EndpointSlice informer per tenant namespace, started/stopped as tenants are onboarded/offboarded. In dynamic mode this **replaces** the manager-cache informer entirely (static/cluster modes keep the existing path).
- The tenant resolver-sync reconciler gained optional hooks; the router registers one that re-scopes the informer set on every FissionTenant change.
- **Teardown fence** (PRD §11.1): `Sync` cancels the informer, waits for its goroutines to fully drain (`factory.Start` is non-blocking — `factory.Shutdown` is the real fence), and only then sweeps the index. The whole transition is serialized under one lock, so an offboard/re-onboard overlap cannot (a) resurrect stale endpoints after the sweep or (b) wipe a replacement informer's fresh entries.
- The shared Add/Update/Delete handlers (including tombstone unwrapping) are extracted so the manager-cache path and per-namespace path cannot drift.

### C. Readiness — exclusion instead of wedging
- A namespace that fails the EndpointSlice `SelfSubjectAccessReview` is **excluded** from the informer set (its functions fall back to the executor RPC, byte-identical to the legacy path) and re-checked on every reconcile. This matters twice: the tenant controller's RoleBinding lands asynchronously after the FissionTenant event, and an orphaned FissionTenant (namespace deleted) can never pass — before this, either case would wedge `/readyz` for every tenant.
- Hooks report `pending` when any namespace is still waiting on RBAC; the reconciler requeues after 2s, since no FissionTenant event fires when the missing piece is an asynchronously-created RoleBinding.

## Reviewer guide

| Area | Files |
|---|---|
| RBAC (chart) | `charts/fission-all/templates/tenant-controller/_tenant-workload-roles.tpl`, `admission-policy.yaml`, `rbac.yaml` |
| RBAC (provisioning) | `pkg/tenant/provision.go`, `pkg/apis/core/v1/const.go` |
| Per-namespace informers | `pkg/router/endpointcache/nsinformers.go`, `handlers.go`, `index.go` (`RemoveNamespace`) |
| Re-scope wiring | `pkg/router/router.go`, `pkg/tenant/resolversync.go` |
| Tests | `nsinformers_test.go`, `index_test.go`, `resolversync_test.go`, `provision_test.go`, `test/integration/suites/serial/dynamic_tenant_test.go` |

Suggested order: `nsinformers.go` (the fence + lock) → `router.go` dynamic block (seed → RBAC filter → hook) → `resolversync.go` (pending-requeue) → `provision.go` + chart partials → tests.

### 1. Chart/RBAC render checks

```bash
# ClusterRole renders in dynamic mode with the same rules as the static dataplane Role
helm template fission charts/fission-all --set tenancy.mode=dynamic | grep -A10 'fission-router-dataplane-tenant'

# Router SA can read EndpointSlices + Services in a provisioned tenant namespace
kubectl auth can-i list endpointslices -n <tenant-ns> --as system:serviceaccount:fission:fission-router   # yes
kubectl auth can-i watch endpointslices -n <tenant-ns> --as system:serviceaccount:fission:fission-router   # yes
```

### 2. End-to-end onboard → warm path (fission CLI)

```bash
kubectl create ns manual-reonboard
fission tenant enable --namespace manual-reonboard
fission env create --name python-man --image ghcr.io/fission/python-env --poolsize 1 --namespace manual-reonboard
fission fn create --name hello-man --env python-man --code hello.py --namespace manual-reonboard
fission route create --name route-hello-man --url /hello-man --method GET --function hello-man --namespace manual-reonboard
```

Observed, step by step:

```bash
# Router logs: informer started for the new tenant — no restart of any control-plane pod
"started endpointslice informer for namespace"  namespace="manual-reonboard"

# RoleBinding provisioned by the tenant controller
kubectl get rolebindings -n manual-reonboard
# fission-router-dataplane   ClusterRole/fission-router-dataplane-tenant

# First invocations (cold start → executor RPC):
curl localhost:8888/hello-man   # Hello, world!
# metrics: fission_router_endpointcache_misses_total +1, hits unchanged

# Executor patched served=true onto the specialized pod; the EndpointSlice
# controller mirrored it seconds later:
kubectl get pod -n manual-reonboard <pod> -o jsonpath='{.metadata.labels.fission\.io/served}'   # true
kubectl get endpointslice -n manual-reonboard -o yaml | grep -A2 'endpoints:'                  # populated, ready: true

# Next invocation — warm path, no executor RPC:
curl localhost:8888/hello-man   # Hello, world!
# metrics: fission_router_endpointcache_hits_total incremented (4 -> 5)
```

### 3. Offboard → re-onboard the SAME namespace

```bash
# drain functions, then:
fission tenant disable --namespace manual-reonboard
# router logs: "stopped endpointslice informer for namespace" — after the drain fence
fission tenant enable --namespace manual-reonboard
# re-create env/fn/route, invoke twice: cold start, then warm hit again
# (hits_total incremented again; no stale endpoints from the first generation)
```

### 4. Orphaned-tenant readiness (regression found during manual testing)

A leftover FissionTenant whose namespace had been deleted wedged `/readyz` at 503 permanently (informer LIST `forbidden` forever). With the exclusion fix:

```bash
kubectl get fissiontenants   # orphan present, namespace gone
kubectl logs deploy/router | grep 'excluded from the warm path'
# "endpointslice RBAC not yet granted ... namespace excluded from the warm path until it lands"
kubectl get pods -l svc=router   # 1/1 Running — readiness no longer wedges
# (orphan then removed: kubectl patch fissiontenant <name> --type json \
#    -p '[{"op":"remove","path":"/metadata/finalizers"}]' && kubectl delete fissiontenant <name>)
```

## Automated test plan

- [x] `go test -race ./pkg/router/endpointcache ./pkg/tenant` — incl. new concurrent-Sync hammer (4 goroutines × 20 Syncs under `-race`), offboard→re-onboard lifecycle, hook pending-requeue
- [x] `TestDynamicTenantLifecycle` (serial integration, kind + dynamic-tenancy profile) — passes (~35s): onboard → warm-path hit → no control-plane restarts → offboard → **re-onboard same namespace** → warm-path hit → final offboard → `go_goroutines` returns to baseline (informer leak guard)
- [ ] CI legs (static + cluster modes) — unchanged paths, expected green

## Known follow-ups (not in scope)

- The tenant controller leaves an orphaned FissionTenant when a namespace is deleted without `tenant disable` (finalizer stuck) — the router now tolerates this (exclusion), but the controller should prune orphans.

### DIFF
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -24,8 +24,7 @@ import (
 // whose SAR reactor consults the allowedNS map. A namespace present with value
 // true → SAR Allowed=true; absent or false → Allowed=false. The map is shared
 // so tests can flip a namespace's RBAC state between calls (the re-onboard
-// scenario). Returns the cache, the resolver (so tests can onboard/offboard),
-// and the clientset (so tests can mutate allowedNS via the reactor closure).
+// scenario). Returns the cache and the resolver (so tests can onboard/offboard).
 func newDynamicCacheTest(t *testing.T, allowedNS map[string]bool) (*dynamicEndpointCache, *utils.NamespaceResolver) {
 	t.Helper()
 	kubeClient := fake.NewSimpleClientset()
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -5,6 +5,7 @@
 package router
 
 import (
+	"context"
 	"sync"
 	"testing"
 
@@ -166,9 +167,6 @@ func TestDynamicCacheReOnboardAfterRBACLands(t *testing.T) {
 // event fires to re-run the check.
 func TestDynamicCacheSARErrorReportsPending(t *testing.T) {
 	t.Parallel()
-	// SAR reactor that always errors — the retry loop in sliceWatchSAR
-	// exhausts 3 attempts (~4s), then sliceWatchAllowedForNamespace returns
-	// (true, error).
 	kubeClient := fake.NewSimpleClientset()
 	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
 		return true, nil, assert.AnError
@@ -184,7 +182,10 @@ func TestDynamicCacheSARErrorReportsPending(t *testing.T) {
 	}
 	resolver.SetTenants(map[string]string{"team-a": "team-a"})
 
-	pending := dyn.syncInformers(t.Context())
+	errCtx, cancel := context.WithCancel(t.Context())
+	cancel()
+
+	pending := dyn.syncInformers(errCtx)
 
 	assert.True(t, pending, "SAR error must report pending so reconciler requeues")
 	_, cached := dyn.rbacChecked.Load("team-a")
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"sync/atomic"
 	"testing"
+	"testing/synctest"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
@@ -212,29 +214,103 @@ func TestDynamicCacheReOnboardAfterRBACLands(t *testing.T) {
 // is still set true. Without pending, a genuinely-unauthorized namespace whose
 // SAR consistently errors keeps an unsyncable informer forever — no FissionTenant
 // event fires to re-run the check.
+//
+// The test drives a genuine API error (assert.AnError from the reactor), NOT a
+// cancelled context. A cancelled context makes sliceWatchSAR return ctx.Err()
+// (context.Canceled), which syncInformers now distinguishes from a real API
+// failure and skips — so the cancellation path can no longer stand in for a
+// flake. Runs under testing/synctest so the 2 × 2s SAR retry delays elapse
+// instantly under virtual time instead of sleeping 4s real.
 func TestDynamicCacheSARErrorReportsPending(t *testing.T) {
 	t.Parallel()
-	kubeClient := fake.NewSimpleClientset()
-	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
-		return true, nil, assert.AnError
+	synctest.Test(t, func(t *testing.T) {
+		kubeClient := fake.NewSimpleClientset()
+		kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
+			return true, nil, assert.AnError
+		})
+		resolver := &utils.NamespaceResolver{}
+		nsi := endpointcache.NewNamespaceInformers(kubeClient, endpointcache.NewIndex(), logr.Discard())
+		dyn := &dynamicEndpointCache{
+			kubeClient:  kubeClient,
+			resolver:    resolver,
+			nsInformers: nsi,
+			logger:      logr.Discard(),
+		}
+		resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+		// syncInformers blocks on the SAR retry delays (time.After inside
+		// sliceWatchSAR). Under synctest those are fake timers, so run the
+		// call in a goroutine and advance virtual time to fire them.
+		var pending bool
+		done := make(chan struct{})
+		go func() {
+			pending = dyn.syncInformers(t.Context())
+			close(done)
+		}()
+		synctest.Wait()
+		<-done
+
+		assert.True(t, pending, "SAR error must report pending so reconciler requeues")
+		_, cached := dyn.rbacChecked.Load("team-a")
+		assert.False(t, cached, "namespace with SAR error must not be cached")
+
+		// Drain informer goroutines before the bubble closes.
+		go func() { nsi.Close() }()
+		synctest.Wait()
 	})
-	resolver := &utils.NamespaceResolver{}
-	nsi := endpointcache.NewNamespaceInformers(kubeClient, endpointcache.NewIndex(), logr.Discard())
-	t.Cleanup(nsi.Close)
-	dyn := &dynamicEndpointCache{
-		kubeClient:  kubeClient,
-		resolver:    resolver,
-		nsInformers: nsi,
-		logger:      logr.Discard(),
+}
+
+// TestDynamicCacheCancellationNotCountedAsSARError verifies that a cancelled
+// context (router shutting down) does NOT bump sarErrors or log at Error for
+// every namespace. sliceWatchSAR returns ctx.Err() (context.Canceled or
+// context.DeadlineExceeded), which syncInformers distinguishes from a genuine
+// API failure and skips. Without this guard a shutdown would count every
+// namespace as an SAR flake.
+func TestDynamicCacheCancellationNotCountedAsSARError(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name string
+		ctx  func(t *testing.T) context.Context
+	}{
+		{"canceled", func(t *testing.T) context.Context {
+			ctx, cancel := context.WithCancel(t.Context())
+			cancel()
+			return ctx
+		}},
+		{"deadline exceeded", func(t *testing.T) context.Context {
+			ctx, cancel := context.WithDeadline(t.Context(), time.Now().Add(-time.Second))
+			t.Cleanup(cancel)
+			return ctx
+		}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			kubeClient := fake.NewSimpleClientset()
+			kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
+				return true, nil, assert.AnError
+			})
+			resolver := &utils.NamespaceResolver{}
+			nsi := endpointcache.NewNamespaceInformers(kubeClient, endpointcache.NewIndex(), logr.Discard())
+			t.Cleanup(nsi.Close)
+			dyn := &dynamicEndpointCache{
+				kubeClient:  kubeClient,
+				resolver:    resolver,
+				nsInformers: nsi,
+				logger:      logr.Discard(),
+			}
+			resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+			// sliceWatchSAR returns ctx.Err() immediately on the first
+			// attempt (the cancelled context short-circuits the retry loop),
+			// so syncInformers returns without the retry delay.
+			pending := dyn.syncInformers(tc.ctx(t))
+
+			assert.False(t, pending, "cancelled context must not report pending")
+			_, counted := dyn.sarErrors["team-a"]
+			assert.False(t, counted, "cancelled context must not bump sarErrors")
+			_, cached := dyn.rbacChecked.Load("team-a")
+			assert.False(t, cached, "cancelled context must not cache the namespace")
+		})
 	}
-	resolver.SetTenants(map[string]string{"team-a": "team-a"})
-
-	errCtx, cancel := context.WithCancel(t.Context())
-	cancel()
-
-	pending := dyn.syncInformers(errCtx)
-
-	assert.True(t, pending, "SAR error must report pending so reconciler requeues")
-	_, cached := dyn.rbacChecked.Load("team-a")
-	assert.False(t, cached, "namespace with SAR error must not be cached")
 }
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -7,6 +7,7 @@ package router
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/go-logr/logr"
@@ -16,10 +17,60 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
 	k8stesting "k8s.io/client-go/testing"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/fission/fission/pkg/router/endpointcache"
 	"github.com/fission/fission/pkg/utils"
 )
+
+// addOnlyManager lets setupDynamicEndpointCache register its startup runnable
+// without starting a full controller-runtime manager. No other Manager method
+// is used while setupDynamicEndpointCache is being called.
+type addOnlyManager struct {
+	ctrl.Manager
+}
+
+func (*addOnlyManager) Add(manager.Runnable) error {
+	return nil
+}
+
+// TestSetupDynamicEndpointCacheReusesPreflightRBAC verifies that namespaces
+// approved by the startup preflight are reused by the initial dynamic-cache
+// sync instead of issuing the same SelfSubjectAccessReviews again.
+func TestSetupDynamicEndpointCacheReusesPreflightRBAC(t *testing.T) {
+	kubeClient := fake.NewSimpleClientset()
+	var sarCalls atomic.Int32
+	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		sarCalls.Add(1)
+		sar := action.(k8stesting.CreateAction).GetObject().(*authorizationv1.SelfSubjectAccessReview)
+		return true, &authorizationv1.SelfSubjectAccessReview{
+			Spec:   sar.Spec,
+			Status: authorizationv1.SubjectAccessReviewStatus{Allowed: true},
+		}, nil
+	})
+
+	precheckedNamespaces := sliceWatchNamespaces()
+	require.NotEmpty(t, precheckedNamespaces)
+
+	ctx, cancel := context.WithCancel(t.Context())
+	t.Cleanup(cancel)
+	_, hooks, err := setupDynamicEndpointCache(
+		ctx,
+		kubeClient,
+		endpointcache.NewIndex(),
+		&addOnlyManager{},
+		nil,
+		logr.Discard(),
+		precheckedNamespaces,
+	)
+	require.NoError(t, err)
+	require.Len(t, hooks, 1)
+
+	pending := hooks[0]()
+	assert.False(t, pending)
+	assert.Zero(t, sarCalls.Load(), "preflight-approved namespaces must not repeat SARs during dynamic-cache setup")
+}
 
 // newDynamicCacheTest builds a dynamicEndpointCache backed by a fake clientset
 // whose SAR reactor consults the allowedNS map. A namespace present with value
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -6,7 +6,6 @@ package router
 
 import (
 	"context"
-	"sync"
 	"sync/atomic"
 	"testing"
 
@@ -80,12 +79,9 @@ func TestSetupDynamicEndpointCacheReusesPreflightRBAC(t *testing.T) {
 func newDynamicCacheTest(t *testing.T, allowedNS map[string]bool) (*dynamicEndpointCache, *utils.NamespaceResolver) {
 	t.Helper()
 	kubeClient := fake.NewSimpleClientset()
-	var mu sync.Mutex
 	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
 		sar := action.(k8stesting.CreateAction).GetObject().(*authorizationv1.SelfSubjectAccessReview)
 		ns := sar.Spec.ResourceAttributes.Namespace
-		mu.Lock()
-		defer mu.Unlock()
 		return true, &authorizationv1.SelfSubjectAccessReview{
 			Spec:   sar.Spec,
 			Status: authorizationv1.SubjectAccessReviewStatus{Allowed: allowedNS[ns]},
--- a/pkg/router/dynamic_endpoint_cache_test.go
+++ b/pkg/router/dynamic_endpoint_cache_test.go
@@ -1,0 +1,193 @@
+// SPDX-FileCopyrightText: The Fission Authors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package router
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	authorizationv1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+
+	"github.com/fission/fission/pkg/router/endpointcache"
+	"github.com/fission/fission/pkg/utils"
+)
+
+// newDynamicCacheTest builds a dynamicEndpointCache backed by a fake clientset
+// whose SAR reactor consults the allowedNS map. A namespace present with value
+// true → SAR Allowed=true; absent or false → Allowed=false. The map is shared
+// so tests can flip a namespace's RBAC state between calls (the re-onboard
+// scenario). Returns the cache, the resolver (so tests can onboard/offboard),
+// and the clientset (so tests can mutate allowedNS via the reactor closure).
+func newDynamicCacheTest(t *testing.T, allowedNS map[string]bool) (*dynamicEndpointCache, *utils.NamespaceResolver) {
+	t.Helper()
+	kubeClient := fake.NewSimpleClientset()
+	var mu sync.Mutex
+	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		sar := action.(k8stesting.CreateAction).GetObject().(*authorizationv1.SelfSubjectAccessReview)
+		ns := sar.Spec.ResourceAttributes.Namespace
+		mu.Lock()
+		defer mu.Unlock()
+		return true, &authorizationv1.SelfSubjectAccessReview{
+			Spec:   sar.Spec,
+			Status: authorizationv1.SubjectAccessReviewStatus{Allowed: allowedNS[ns]},
+		}, nil
+	})
+	resolver := &utils.NamespaceResolver{}
+	nsi := endpointcache.NewNamespaceInformers(kubeClient, endpointcache.NewIndex(), logr.Discard())
+	t.Cleanup(nsi.Close)
+	dynCache := &dynamicEndpointCache{
+		kubeClient:  kubeClient,
+		resolver:    resolver,
+		nsInformers: nsi,
+		logger:      logr.Discard(),
+	}
+	return dynCache, resolver
+}
+
+// TestDynamicCacheAllowedCachesAndIncludes verifies the happy path: a namespace
+// that passes SAR is cached in rbacChecked, included in the informer set, and
+// does not report pending.
+func TestDynamicCacheAllowedCachesAndIncludes(t *testing.T) {
+	t.Parallel()
+	allowedNS := map[string]bool{"team-a": true}
+	dyn, resolver := newDynamicCacheTest(t, allowedNS)
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+	pending := dyn.syncInformers(t.Context())
+	assert.False(t, pending, "allowed namespace must not report pending")
+	_, cached := dyn.rbacChecked.Load("team-a")
+	assert.True(t, cached, "allowed namespace must be cached in rbacChecked")
+}
+
+// TestDynamicCacheDeniedExcludesAndReportsPending verifies the #3647 exclusion:
+// a namespace whose SAR returns Allowed=false is EXCLUDED from the informer set
+// (not cached) and reports pending=true so the reconciler requeues.
+func TestDynamicCacheDeniedExcludesAndReportsPending(t *testing.T) {
+	t.Parallel()
+	allowedNS := map[string]bool{"team-a": false}
+	dyn, resolver := newDynamicCacheTest(t, allowedNS)
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+	pending := dyn.syncInformers(t.Context())
+	assert.True(t, pending, "denied namespace must report pending")
+	_, cached := dyn.rbacChecked.Load("team-a")
+	assert.False(t, cached, "denied namespace must not be cached")
+}
+
+// TestDynamicCachePrunesOnOffboard verifies fix #1: when a namespace leaves the
+// live set (offboard), its cached SAR pass is pruned from rbacChecked. Without
+// pruning, the stale entry would make a re-onboard skip the SAR.
+func TestDynamicCachePrunesOnOffboard(t *testing.T) {
+	t.Parallel()
+	allowedNS := map[string]bool{"team-a": true}
+	dyn, resolver := newDynamicCacheTest(t, allowedNS)
+
+	// Onboard team-a → SAR passes → cached.
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+	dyn.syncInformers(t.Context())
+	_, cached := dyn.rbacChecked.Load("team-a")
+	require.True(t, cached, "team-a must be cached after onboarding")
+
+	// Offboard team-a → cache must be pruned.
+	resolver.SetTenants(map[string]string{})
+	dyn.syncInformers(t.Context())
+	_, cached = dyn.rbacChecked.Load("team-a")
+	assert.False(t, cached, "team-a must be pruned from rbacChecked after offboard")
+}
+
+// TestDynamicCacheReOnboardReRunsSAR is the core regression test for fix #1:
+// onboard → offboard → re-onboard with SAR now DENIED. Without the cache-prune,
+// the stale rbacChecked entry would make syncInformers skip the SAR, include the
+// namespace, and return pending=false — starting an informer whose LIST stays
+// forbidden, wedging /readyz fleet-wide. With the prune, the SAR re-runs and
+// the namespace is correctly excluded + pending.
+func TestDynamicCacheReOnboardReRunsSAR(t *testing.T) {
+	t.Parallel()
+	allowedNS := map[string]bool{"team-a": true}
+	dyn, resolver := newDynamicCacheTest(t, allowedNS)
+
+	// Onboard team-a → SAR passes → cached.
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+	dyn.syncInformers(t.Context())
+	_, cached := dyn.rbacChecked.Load("team-a")
+	require.True(t, cached, "team-a must be cached after first onboarding")
+
+	// Offboard team-a → cache pruned.
+	resolver.SetTenants(map[string]string{})
+	dyn.syncInformers(t.Context())
+	_, cached = dyn.rbacChecked.Load("team-a")
+	require.False(t, cached, "team-a must be pruned after offboard")
+
+	// Re-onboard team-a, but now SAR denies (RoleBinding not yet recreated).
+	allowedNS["team-a"] = false
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+	pending := dyn.syncInformers(t.Context())
+
+	assert.True(t, pending, "re-onboarded namespace with denied SAR must report pending")
+	_, cached = dyn.rbacChecked.Load("team-a")
+	assert.False(t, cached, "re-onboarded namespace with denied SAR must not be cached")
+}
+
+// TestDynamicCacheReOnboardAfterRBACLands verifies the full lifecycle: onboard
+// denied → requeue (pending) → RBAC lands → re-onboard allowed → cached. This
+// is the normal onboarding race (tenant controller hasn't created the binding
+// yet) resolving correctly.
+func TestDynamicCacheReOnboardAfterRBACLands(t *testing.T) {
+	t.Parallel()
+	allowedNS := map[string]bool{"team-a": false}
+	dyn, resolver := newDynamicCacheTest(t, allowedNS)
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+	// First sync: SAR denied → pending, not cached.
+	pending := dyn.syncInformers(t.Context())
+	require.True(t, pending, "denied on first check")
+	_, cached := dyn.rbacChecked.Load("team-a")
+	require.False(t, cached, "not cached when denied")
+
+	// RBAC lands: SAR now allowed.
+	allowedNS["team-a"] = true
+	pending = dyn.syncInformers(t.Context())
+	assert.False(t, pending, "no longer pending once SAR passes")
+	_, cached = dyn.rbacChecked.Load("team-a")
+	assert.True(t, cached, "cached once SAR passes")
+}
+
+// TestDynamicCacheSARErrorReportsPending verifies fix #2: when the SAR itself
+// errors (apiserver flake), the namespace is optimistically included but pending
+// is still set true. Without pending, a genuinely-unauthorized namespace whose
+// SAR consistently errors keeps an unsyncable informer forever — no FissionTenant
+// event fires to re-run the check.
+func TestDynamicCacheSARErrorReportsPending(t *testing.T) {
+	t.Parallel()
+	// SAR reactor that always errors — the retry loop in sliceWatchSAR
+	// exhausts 3 attempts (~4s), then sliceWatchAllowedForNamespace returns
+	// (true, error).
+	kubeClient := fake.NewSimpleClientset()
+	kubeClient.PrependReactor("create", "selfsubjectaccessreviews", func(action k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, assert.AnError
+	})
+	resolver := &utils.NamespaceResolver{}
+	nsi := endpointcache.NewNamespaceInformers(kubeClient, endpointcache.NewIndex(), logr.Discard())
+	t.Cleanup(nsi.Close)
+	dyn := &dynamicEndpointCache{
+		kubeClient:  kubeClient,
+		resolver:    resolver,
+		nsInformers: nsi,
+		logger:      logr.Discard(),
+	}
+	resolver.SetTenants(map[string]string{"team-a": "team-a"})
+
+	pending := dyn.syncInformers(t.Context())
+
+	assert.True(t, pending, "SAR error must report pending so reconciler requeues")
+	_, cached := dyn.rbacChecked.Load("team-a")
+	assert.False(t, cached, "namespace with SAR error must not be cached")
+}
--- a/pkg/router/endpointcache/nsinformers.go
+++ b/pkg/router/endpointcache/nsinformers.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/go-logr/logr"
 
@@ -38,11 +39,13 @@ type nsInformer struct {
 // Close stops every running informer and waits for their goroutines to exit;
 // tests must call it (typically via t.Cleanup) to avoid leaking goroutines.
 type NamespaceInformers struct {
-	kubeClient kubernetes.Interface
-	index      *Index
-	logger     logr.Logger
-	mu         sync.Mutex
-	informers  map[string]*nsInformer
+	kubeClient    kubernetes.Interface
+	index         *Index
+	logger        logr.Logger
+	mu            sync.Mutex
+	informers     map[string]*nsInformer
+	closed        bool         // set by Close; Sync returns immediately when true
+	informerCount atomic.Int64 // running per-namespace informer count; exposed as metric
 }
 
 // NewNamespaceInformers creates an empty informer manager. The caller owns
@@ -69,6 +72,9 @@ func NewNamespaceInformers(kubeClient kubernetes.Interface, index *Index, logger
 func (n *NamespaceInformers) Sync(namespaces []string) {
 	n.mu.Lock()
 	defer n.mu.Unlock()
+	if n.closed {
+		return
+	}
 	// Build a set of desired namespaces.
 	desired := make(map[string]struct{}, len(namespaces))
 	for _, ns := range namespaces {
@@ -84,6 +90,8 @@ func (n *NamespaceInformers) Sync(namespaces []string) {
 			<-ni.done
 			n.index.RemoveNamespace(ns)
 			delete(n.informers, ns)
+			n.informerCount.Add(-1)
+			RecordInformersCount(n.informerCount.Load())
 			n.logger.Info("stopped endpointslice informer for namespace", "namespace", ns)
 		}
 	}
@@ -100,6 +108,8 @@ func (n *NamespaceInformers) Sync(namespaces []string) {
 			continue
 		}
 		n.informers[ns] = &nsInformer{informer: informer, cancel: cancel, done: done}
+		n.informerCount.Add(1)
+		RecordInformersCount(n.informerCount.Load())
 		n.logger.Info("started endpointslice informer for namespace", "namespace", ns)
 	}
 }
@@ -177,6 +187,7 @@ func (n *NamespaceInformers) WaitForCacheSync(ctx context.Context) bool {
 // context is cancelled; tests should register it via t.Cleanup.
 func (n *NamespaceInformers) Close() {
 	n.mu.Lock()
+	n.closed = true
 	stopped := n.informers
 	n.informers = make(map[string]*nsInformer)
 	n.mu.Unlock()
@@ -184,4 +195,6 @@ func (n *NamespaceInformers) Close() {
 		ni.cancel()
 		<-ni.done
 	}
+	n.informerCount.Store(0)
+	RecordInformersCount(0)
 }
--- a/pkg/router/endpointcache/nsinformers.go
+++ b/pkg/router/endpointcache/nsinformers.go
@@ -1,0 +1,185 @@
+// SPDX-FileCopyrightText: The Fission Authors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package endpointcache
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	fv1 "github.com/fission/fission/pkg/apis/core/v1"
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+// nsInformer pairs one namespace's EndpointSlice informer with the cancel
+// function that stops it and a done channel closed only after the informer's
+// goroutines have fully drained (factory.Start + factory.Shutdown). Stored in
+// NamespaceInformers.informers, keyed by namespace. Waiting on done before
+// sweeping the index is the teardown fence (PRD §11.1): it guarantees no queued
+// or in-flight ApplySlice can repopulate a namespace after offboard.
+type nsInformer struct {
+	informer cache.SharedIndexInformer
+	cancel   context.CancelFunc
+	done     chan struct{}
+}
+
+// NamespaceInformers manages one EndpointSlice informer per tenant namespace.
+// Unlike the controller-runtime Manager's cache (scoped once at startup),
+// this can start/stop informers as tenants are onboarded/offboarded at runtime.
+// Close stops every running informer and waits for their goroutines to exit;
+// tests must call it (typically via t.Cleanup) to avoid leaking goroutines.
+type NamespaceInformers struct {
+	kubeClient kubernetes.Interface
+	index      *Index
+	logger     logr.Logger
+	mu         sync.Mutex
+	informers  map[string]*nsInformer
+}
+
+// NewNamespaceInformers creates an empty informer manager. The caller owns
+// the lifecycle: call Close when done (production ties it to the router's
+// context; tests register it via t.Cleanup).
+func NewNamespaceInformers(kubeClient kubernetes.Interface, index *Index, logger logr.Logger) *NamespaceInformers {
+	return &NamespaceInformers{
+		kubeClient: kubeClient,
+		index:      index,
+		logger:     logger,
+		informers:  make(map[string]*nsInformer),
+	}
+}
+
+// Sync reconciles the running informer set to match namespaces:
+// starts informers for new namespaces, stops informers for removed namespaces.
+// The whole transition is serialized under n.mu: cancel → wait for the
+// informer's goroutines to drain (factory.Shutdown via the done channel) →
+// sweep the index. Holding the lock across the wait prevents a concurrent
+// Sync from installing a replacement informer whose fresh entries the earlier
+// call would then delete (offboard/re-onboard overlap). The drain never takes
+// n.mu (handlers only touch the Index's own locks), so this cannot deadlock —
+// other callers just wait a moment.
+func (n *NamespaceInformers) Sync(namespaces []string) {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	// Build a set of desired namespaces.
+	desired := make(map[string]struct{}, len(namespaces))
+	for _, ns := range namespaces {
+		desired[ns] = struct{}{}
+	}
+	// Stop informers for namespaces no longer in the desired set. Waiting for
+	// the drain BEFORE sweeping is the teardown fence (PRD §11.1): no queued
+	// ApplySlice from the dying informer can repopulate the namespace after
+	// RemoveNamespace, resurrecting stale endpoints during offboard.
+	for ns, ni := range n.informers {
+		if _, ok := desired[ns]; !ok {
+			ni.cancel()
+			<-ni.done
+			n.index.RemoveNamespace(ns)
+			delete(n.informers, ns)
+			n.logger.Info("stopped endpointslice informer for namespace", "namespace", ns)
+		}
+	}
+	// Start informers for new namespaces.
+	for ns := range desired {
+		if _, ok := n.informers[ns]; ok {
+			continue // already running
+		}
+		ctx, cancel := context.WithCancel(context.Background())
+		informer, done, err := n.start(ctx, ns)
+		if err != nil {
+			cancel()
+			n.logger.Error(err, "failed to start endpointslice informer", "namespace", ns)
+			continue
+		}
+		n.informers[ns] = &nsInformer{informer: informer, cancel: cancel, done: done}
+		n.logger.Info("started endpointslice informer for namespace", "namespace", ns)
+	}
+}
+
+// start builds and runs a single-namespace EndpointSlice informer that feeds
+// slice add/update/delete events into the shared Index. The informer runs in a
+// background goroutine via factory.Start; cancelling ctx stops it. Returns the
+// informer (for HasSynced queries) and a done channel closed when the factory
+// goroutine has fully exited.
+func (n *NamespaceInformers) start(ctx context.Context, ns string) (cache.SharedIndexInformer, chan struct{}, error) {
+	// Label filter: only Fission-managed slices (same selector as routerCacheOptions).
+	labelSelector := labels.SelectorFromSet(labels.Set{
+		fv1.MANAGED_BY_LABEL: fv1.MANAGED_BY_VALUE,
+	}).String()
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		n.kubeClient,
+		0, // no resync — the index is the source of truth
+		informers.WithNamespace(ns),
+		informers.WithTweakListOptions(func(opts *metav1.ListOptions) {
+			opts.LabelSelector = labelSelector
+		}))
+	informer := factory.Discovery().V1().EndpointSlices().Informer()
+
+	// Shared handlers (same as the manager-cache informer path).
+	_, err := informer.AddEventHandler(endpointSliceHandlers(n.index, n.logger))
+	if err != nil {
+		return nil, nil, fmt.Errorf("error adding endpointslice event handler: %w", err)
+	}
+	// Run the factory in the background; cancelling ctx stops it. Start is
+	// non-blocking (it only launches informer goroutines), so done must wait on
+	// Shutdown, which blocks until every informer's Run has returned — and the
+	// informer's processor drains all queued handler notifications before Run
+	// returns. Closing done after Shutdown is therefore a real teardown fence:
+	// once done fires, no handler for this namespace can ever run again.
+	done := make(chan struct{})
+	go func() {
+		factory.Start(ctx.Done())
+		factory.Shutdown()
+		close(done)
+	}()
+	return informer, done, nil
+}
+
+// HasSynced reports whether all running informers have completed their initial
+// list. Satisfies cache.InformerSynced (func() bool) so the router can gate
+// readiness on the index being populated, same as the manager-cache path.
+// Returns true when no informers are running (vacuously synced).
+func (n *NamespaceInformers) HasSynced() bool {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	for _, ni := range n.informers {
+		if !ni.informer.HasSynced() {
+			return false
+		}
+	}
+	return true
+}
+
+// WaitForCacheSync blocks until all running informers have completed their
+// initial list or ctx is cancelled. Returns true if all synced, false on
+// timeout/cancel. This is the deterministic synchronization point for tests
+// (replacing polling) and for the router's readiness gate.
+func (n *NamespaceInformers) WaitForCacheSync(ctx context.Context) bool {
+	n.mu.Lock()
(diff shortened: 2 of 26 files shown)


pkg/router/router.go
```diff
@@ -441,41 +480,139 @@ func Start(ctx context.Context, clientGen crd.ClientGeneratorInterface, logger l
...
+			var rbacChecked sync.Map
+			// syncInformers re-scopes the informer set to the resolver's live
+			// namespaces, filtered by EndpointSlice RBAC. Shared by the startup
+			// seed, the post-cache-sync re-seed, and the resolver-sync hook so
+			// all three apply the same exclusion semantics. Returns pending=true
+			// when any namespace is still waiting on its RBAC — the reconciler
+			// requeues until every namespace passes (no tenant event fires for
+			// a RoleBinding the tenant controller creates asynchronously).
+			syncInformers := func() (pending bool) {
+				namespaces := resolver.FunctionNamespaces()
```
sanketsudake: **critical — `rbacChecked` is never invalidated, so re-onboarding bypasses the SAR gate this PR exists to install.**

`rbacChecked.Store(ns, ...)` (L548) has no removal path. Offboarding deletes the tenant's `fission-router-dataplane` RoleBinding (`DeleteNamespaceRBAC` label-sweeps all managed RoleBindings), but the namespace stays cached — so a **re-onboard of the same namespace** takes the `rbacChecked.Load(ns)` fast path at L529, skips the SAR entirely, and starts the informer while the tenant controller is still (asynchronously) recreating the binding.

That is precisely the race the exclusion logic was built for, now unguarded on the second onboarding. It matters because it feeds the readiness gate: `endpointsSynced = nsInformers.HasSynced` (L581) and `HasSynced` is false while **any** informer hasn't completed its initial LIST. A forbidden LIST never latches, and `routerReadinessHandler` (`pkg/router/httpTriggers.go:217`) re-evaluates it on **every** probe — not once at startup.

Failure scenario: onboard `team-a` → offboard → re-onboard while the tenant controller is backlogged or mid-rollout. Every router replica starts an un-syncable informer within seconds of the same FissionTenant event (no leader election). `/readyz` returns 503 on all of them; at `periodSeconds: 1, failureThreshold: 30` the whole fleet goes NotReady after ~30s, the router Service drains to zero endpoints, and **all tenants** lose all traffic — the exact wedge the comment above claims to prevent ("Exclusion is what keeps one bad namespace from wedging /readyz for every tenant").

Pruning the cache when a namespace leaves the set closes it:

```suggestion
				namespaces := resolver.FunctionNamespaces()
				live := make(map[string]struct{}, len(namespaces))
				for _, ns := range namespaces {
					live[ns] = struct{}{}
				}
				// Drop cached passes for namespaces that have left the set:
				// offboarding deletes the tenant's fission-router-dataplane
				// RoleBinding, so a later re-onboard MUST re-run the SAR rather
				// than start an informer whose LIST stays forbidden until the
				// tenant controller recreates the binding — an informer that
				// never syncs holds HasSynced (and so /readyz) false fleet-wide.
				rbacChecked.Range(func(k, _ any) bool {
					if _, ok := live[k.(string)]; !ok {
						rbacChecked.Delete(k)
					}
					return true
				})
```

Worth considering alongside this: gating readiness on informers *added after* the replica first went Ready couples one new tenant's LIST to every existing tenant's traffic. Dropping a namespace that fails to sync within a deadline out of the gate (rather than out of the fleet) would match the exclusion philosophy the rest of this block follows.

pkg/router/router.go
```diff
@@ -441,41 +480,139 @@ func Start(ctx context.Context, clientGen crd.ClientGeneratorInterface, logger l
...
+						continue
+					}
+					ok, err := sliceWatchAllowedForNamespace(ctx, kubeClient, ns)
+					if err != nil {
+						// Transient SAR failure — optimistic include; the
+						// reflector's own retry absorbs apiserver flakes.
+						logger.Error(err, "endpointslice RBAC check failed transiently; keeping informer", "namespace", ns)
+						allowed = append(allowed, ns)
+						continue
+					}
```
sanketsudake: **warning — the optimistic-include path starts a possibly-unsyncable informer and schedules no re-check.**

On a persistent SAR failure `sliceWatchAllowedForNamespace` returns `(true, err)`, so the namespace is included — but `pending` is left false, so the reconciler does **not** requeue. If the namespace genuinely lacks the RoleBinding, the informer's LIST is forbidden forever, `HasSynced` never latches, and `/readyz` stays 503 on every replica (see the readiness coupling noted on L526). Nothing re-runs the check until unrelated tenant churn fires another FissionTenant event.

Setting `pending` here keeps the optimism (the informer still starts, absorbing genuine flakes via the reflector's own retry) while guaranteeing the decision is revisited:

```suggestion
					if err != nil {
						// Transient SAR failure — optimistic include; the
						// reflector's own retry absorbs apiserver flakes.
						// Still mark pending: if the namespace genuinely lacks
						// RBAC the informer can never sync, and no FissionTenant
						// event will fire to re-run this check.
						logger.Error(err, "endpointslice RBAC check failed transiently; keeping informer", "namespace", ns)
						allowed = append(allowed, ns)
						pending = true
						continue
					}
```

Related: with the router ctx already cancelled, `sliceWatchSAR` fails immediately with `ctx.Err()`, so this branch marks *every* namespace allowed during shutdown — see the note on `Close` in `nsinformers.go`.

pkg/router/endpointcache/nsinformers.go
```diff
@@ -0,0 +1,187 @@
...
+func (n *NamespaceInformers) Close() {
+	n.mu.Lock()
+	stopped := n.informers
+	n.informers = make(map[string]*nsInformer)
+	n.mu.Unlock()
+	for _, ni := range stopped {
+		ni.cancel()
+		<-ni.done
+	}
+}
```
sanketsudake, line 193: **warning — `Close` is not terminal, so a `Sync` racing shutdown resurrects informers that nothing can ever stop.**

`Close` swaps the map under `n.mu` and drains outside it, but nothing records that the manager is closed — a later `Sync` simply repopulates `n.informers`. That later `Sync` is reachable in production: the `go func() { <-ctx.Done(); nsInformers.Close() }()` wiring in `router.go` races controller-runtime's graceful drain of an in-flight `Reconcile` (30s default), whose hook calls `syncInformers`. Because `start` parents each informer on `context.WithCancel(context.Background())` rather than the router ctx, informers created after `Close` are unreachable by any cancel path.

Compounding it: with the ctx cancelled, every SAR fails instantly and the optimistic-include branch marks *all* namespaces allowed, so the resurrected set is the full one. A second informer can also be started for a namespace `Close` is still draining outside the lock — two feeders for the same slices, which `ApplySlice`'s single-feeder assumption doesn't expect.

In the shipped binary the leak is bounded by process exit, so this is a correctness/hygiene issue rather than a live outage — but it makes `Close` unsafe for any embedder or long-lived test. A `closed` flag checked by `Sync` fixes both:

```suggestion
func (n *NamespaceInformers) Close() {
	n.mu.Lock()
	n.closed = true
	stopped := n.informers
	n.informers = make(map[string]*nsInformer)
	n.mu.Unlock()
	for _, ni := range stopped {
		ni.cancel()
		<-ni.done
	}
}
```

with a `closed bool` field on the struct and an early `if n.closed { return }` at the top of `Sync` (after taking `n.mu`).

test/integration/suites/serial/dynamic_tenant_test.go
```diff
@@ -148,5 +170,120 @@ func TestDynamicTenantLifecycle(t *testing.T) {
...
+		hitsAfter := scrapeRouterMetric(t, ctx, f, "fission_router_endpointcache_hits_total")
+		assert.Greaterf(c, hitsAfter, hitsBefore,
+			"endpointcache_hits_total must increase on a warm invocation — proves the warm path served the tenant's function")
+	}, 2*time.Minute, 5*time.Second, append([]any{msg}, args...)...)
+}
+
+// scrapeRouterMetric scrapes a single Prometheus metric from the router pods
+// (via pod proxy, no port-forward needed) and returns its sum across replicas.
+// Works for both gauges and counters (sums the latest sample per pod).
+func scrapeRouterMetric(t *testing.T, ctx context.Context, f *framework.Framework, metricName string) float64 {
```
sanketsudake: **warning — `require` against the outer `t` inside `EventuallyWithT` closures turns a retryable blip into a hard failure.**

`scrapeRouterMetric` asserts with `require.NoErrorf(t, ...)` on the **outer** `*testing.T`, but it is called from inside `EventuallyWithT` closures (L216 here and L239 in `assertWarmPathHit`). testify runs each tick in its own goroutine; `require` there calls `t.FailNow()` → `runtime.Goexit()`, so the tick never reports back, the poll loop stops retrying, and the test hangs until the full 2-minute timeout before failing. A single apiserver-proxy hiccup during the poll — exactly what the retry loop exists to absorb — becomes a hard failure.

`*assert.CollectT` implements `Errorf` + `FailNow`, so it satisfies `require.TestingT` and its `FailNow` is recovered by `EventuallyWithT` as a normal failed tick. Widening the parameter type is the minimal fix (drop `t.Helper()`, which `CollectT` doesn't have):

```suggestion
func scrapeRouterMetric(t require.TestingT, ctx context.Context, f *framework.Framework, metricName string) float64 {
```

Callers outside a poll keep passing `t`; the two inside the closures should pass `c`.

sanketsudake (commented):
## Review battery — PR #3653

Reviewed in an isolated worktree at merge-base `bae3c6ab`. Passes run: code review, security review, deep correctness/concurrency, test-coverage, simplify/deslop, and PR-reviewability. Verification in the worktree: `go build ./...` clean, `go vet` clean, `golangci-lint` 0 issues, `go test -race ./pkg/router/endpointcache/... ./pkg/tenant/...` green (new tests ~2s), integration suite compiles under `-tags=integration`, and `helm template --set tenancy.mode=dynamic` renders `fission-router-dataplane-tenant` with read-only `get/list/watch` on endpointslices + services.

This is a well-built change. The teardown fence (`factory.Shutdown` before the index sweep), the shared-handler extraction, the single-sourced chart partial, and the commit message are all above the bar, and the manual-verification write-up in the description is genuinely useful. One issue needs changes before merge; the rest are follow-ups.

### Verdict: needs changes — one critical finding

---

### critical

**1. `rbacChecked` is never invalidated → re-onboard bypasses the SAR gate → fleet-wide `/readyz` wedge.** `pkg/router/router.go:548` stores passes permanently. Offboarding deletes the tenant's `fission-router-dataplane` RoleBinding, so a re-onboard of the same namespace skips the SAR (L529) and starts an informer while the binding is still being recreated. Because `HasSynced` is false while *any* informer hasn't completed its initial LIST, and `routerReadinessHandler` re-evaluates it on **every** probe, a forbidden LIST holds `/readyz` at 503 on every replica simultaneously — after ~30s (`periodSeconds: 1`, `failureThreshold: 30`) the router Service drains to zero endpoints and all tenants lose traffic. That is the exact failure mode the exclusion logic was written to prevent. Inline comment has a suggested cache-prune. Flagged independently by two review passes.

### warning

**2. Optimistic-include on SAR error schedules no re-check** (`router.go:534-540`) — `pending` stays false, so the reconciler never requeues; a genuinely-unauthorized namespace keeps an unsyncable informer indefinitely. Suggestion inline.

**3. `Close` is not terminal** (`endpointcache/nsinformers.go:178`) — a `Sync` racing shutdown repopulates the map, and `start` parents informers on `context.Background()`, so those are unreachable by any cancel. Bounded by process exit today. Suggestion inline.

**4. `syncInformers` read→SAR→`Sync` is a TOCTOU window** (`router.go:525`) — the seed runnable can spend ~8-12s in SAR retries, then apply a stale namespace set that stops a tenant onboarded meanwhile; `pending` is false so nothing requeues and the tenant silently falls back to executor RPC until unrelated churn. Serializing `syncInformers` under its own mutex and re-reading the set inside it makes last-writer-wins converge.

**5. `require` on the outer `t` inside `EventuallyWithT` closures** (`dynamic_tenant_test.go:248`, used at L216/L239) — `FailNow` from a poll goroutine kills the retry loop; one proxy hiccup becomes a 2-minute hard failure. Suggestion inline.

**6. Goroutine leak guard's `+10` slop exceeds the leak** (`dynamic_tenant_test.go:217`) — a single leaked informer is ~5-7 goroutines and passes. Inline comment suggests an informer-count gauge instead.

### test-coverage

**7. The RBAC-exclusion → `pending` branch is untested at every level.** This is the actual #3647 race fix. Deleting `pending = true` (L545), or turning the exclusion `continue` into an append, survives the entire suite; `TestResolverSyncReconcilerPendingRequeues` uses a stub hook returning `true`, so it pins the requeue plumbing but not the predicate. `sliceWatchAllowedForNamespace` is directly testable with a fake-clientset SAR reactor.

**8. The `<-ni.done` teardown fence — the file's most-commented invariant — is not verified by any test.** Every offboard in the suite happens after the index has settled, so nothing is in flight to resurrect; in `TestNamespaceInformersConcurrentSync` a fence-less late `ApplySlice` resurrects the very entry the assertions expect, so end-state checks can't distinguish. A churn test (background slice updates during offboard, then assert the index *stays* empty) would pin it.

**9. Only the initial-LIST replay is exercised; the extracted `handlers.go` Update/Delete path is untested.** Deleting `DeleteFunc` or breaking the tombstone unwrap passes the suite — the offboard sweep goes through `RemoveNamespace`, not `DeleteSlice`. Since the extraction's stated purpose is keeping the two paths from drifting, a Create→Update→Delete test through the fake client is worth adding.

**10. Minor:** `HasSynced` is only tested in its vacuous-empty case (mutating it to `return true` survives), and the new `fission-router` entry in the ValidatingAdmissionPolicy SA allowlist isn't covered — `tenant_admission_policy_test.go` only exercises `fission-fetcher`.

### maintainability

**11. `parseGaugeValue`'s doc contradicts the code** (`dynamic_tenant_test.go:265`) — it returns the **first** parseable match (early `return v`), not "the last sample's value"; `scrapeRouterMetric`'s "sums the latest sample per pod" is likewise inaccurate. Correct today only because both scraped metrics are unlabeled (`hits.Add` records with no attributes), so first == last == only. It would silently under-report on any labeled metric.

**12. Third copy of the same metrics scraper.** `scrapeRouterMetric`/`parseGaugeValue` closely duplicate `scrapeCounterSum`/`sumMetricLines` in `suites/common/alias_routing_test.go:332-380`, with a third variant in `memory_soak_test.go`. Different packages, so the real fix is promoting one into `test/integration/framework/` — `sumMetricLines`' sum-all-samples semantics are the correct ones and would resolve #11 for free.

**13. The hook's `map[string]string` argument is dead.** `resolversync.go` passes `FissionResourceNamespaces()` (CR namespaces), but the only hook ever registered ignores it and re-derives `FunctionNamespaces()` (workload namespaces, `GetFunctionNS`-mapped and deduped) — i.e. the passed set is the *wrong* one for informer scoping. `func() (pending bool)` would be honest about the contract before a second caller trusts the argument.

**14. Consider extracting the ~90-line dynamic block out of `Start`**, which this PR takes to ~430 lines. The PR already set the precedent by pulling `checkSliceWatchRBACForNamespace`/`sliceWatchSAR` to top level; a `setupDynamicEndpointCache(...)` helper would also make `syncInformers` testable, which finding #7 needs.

**15. Comment duplication.** The teardown-fence rationale appears ~4x (`nsinformers.go` struct doc, inside `Sync`, inside `start`, echoed in `RemoveNamespace`) and the hooks/pending contract ~4x (`router.go` var decl + `syncInformers`, `resolversync.go` struct field + `AddResolverSync` godoc). Keeping the mechanism once where it lives (`start`'s done-channel contract; the exported `AddResolverSync` godoc) and pointing at it elsewhere would age better.

**16.** `fixed linting error` is a fixup worth squashing into the main commit.

### security

No exploitable vulnerability introduced by this PR. The new ClusterRole is read-only (`get/list/watch` on endpointslices + services), bound via a namespaced RoleBinding, and the tenant controller's `bind` verb stays `resourceNames`-scoped through the same single-sourced list, so it cannot escalate. Adding `fission-router` to the admission-policy allowlist opens no new path — only the controller can create those bindings, and it binds fixed (SA, ClusterRole) pairs in code. No prompt-injection or backdoor-shaped content in the diff.

One thing worth tracking separately: `fnKeyForSlice` keys the endpoint index purely on the slice's `fission.io/function-name` / `function-namespace` **labels**, never checking them against the slice's own namespace, and the built-in EndpointSlice controller mirrors arbitrary Service labels — so a principal with `services create` in any watched namespace can merge endpoints into another namespace's function entry. This is **pre-existing** (identical on `bae3c6ab`: cluster mode already watched slices cluster-wide with the same label-only selector, and static multi-namespace installs watch several namespaces), so it is not a blocker here — but this PR does widen the watched set to onboarded tenant namespaces in dynamic mode, which grows the blast radius. Worth a separate hardening issue: validate `es.Namespace == GetFunctionNS(labels[FUNCTION_NAMESPACE])`, or cross-check the `fission.io/function-uid` label against the Function CR, which would close it for cluster, static-multi-ns, and dynamic modes at once.

test/integration/suites/serial/dynamic_tenant_test.go
```diff
@@ -148,5 +170,120 @@ func TestDynamicTenantLifecycle(t *testing.T) {
...
+	require.NoError(t, fc.Environments(tenantNS).Delete(ctx, envName2, metav1.DeleteOptions{}))
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		_, err := fc.Functions(tenantNS).Get(ctx, fnName2, metav1.GetOptions{})
+		assert.Truef(c, apierrors.IsNotFound(err), "function %q should clear before offboarding (err=%v)", fnName2, err)
+	}, time.Minute, time.Second)
+	ns.DisableTenant(t, ctx)
+	f.WaitForTenantOffboarded(t, ctx, tenantNS)
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		g := scrapeRouterMetric(t, ctx, f, "go_goroutines")
+		assert.LessOrEqualf(c, g, goroutinesBefore+10,
```
sanketsudake: **warning — a `+10` goroutine tolerance is wider than the leak it is guarding against.**

One leaked per-namespace informer pins roughly 5-7 goroutines (reflector, `processLoop`, the processor listener's run+pop pair, the watch). A partial teardown bug — say the processor leaks but the reflector exits — leaks ~4-6 and passes comfortably under `+10`; even a total teardown failure across the two offboards lands right at the boundary. So the assertion can only catch a leak substantially worse than the one the fence in `nsinformers.go` was written to prevent.

The false-fail direction is fine (the 2-minute window outlives idle-connection churn, and `AssertNoControlPlaneRestart` rules out pod restarts), so this is about detection power, not flakiness.

A deterministic signal would be strictly better than a goroutine delta: exporting a `fission_router_endpointcache_informers` gauge (count of running per-namespace informers) and asserting it returns to 0 after offboard turns a statistical guard into an exact one, and doubles as production observability for the same leak.

pkg/router/router.go
```diff
@@ -145,49 +148,228 @@ func checkSliceWatchRBAC(ctx context.Context, kubeClient kubernetes.Interface) e
...
+		if err != nil {
+			// Transient SAR failure — optimistic include; the
+			// reflector's own retry absorbs apiserver flakes.
+			// Still mark pending: if the namespace genuinely lacks
+			// RBAC the informer can never sync, and no FissionTenant
+			// event will fire to re-run this check.
+			d.logger.Error(err, "endpointslice RBAC check failed transiently; keeping informer", "namespace", ns)
+			allowed = append(allowed, ns)
+			pending = true
+			continue
```
sanketsudake, line 346: **warning — a *persistent* SAR error still leaves the fleet-wide `/readyz` wedge unbounded.**

Fix #2 (setting `pending`) closes the case where the SAR later succeeds and returns a verdict: the requeue re-runs the check, `!ok` drops the namespace from `allowed`, and `nsInformers.Sync` cancels + drains the informer, so `HasSynced` recovers. I verified that path — the wedge is bounded to ~2-6s there. Thanks for that.

What's left is the branch that never reaches a verdict. `sliceWatchAllowedForNamespace` returns `(true, err)` on error, so this branch **starts an informer** for a namespace that may have no RBAC, and there is no attempt counter, deadline, or escalation — every 2s requeue re-executes the same optimistic include.

Failure scenario: SAR creates fail persistently (authorization API degraded, an admission webhook rejecting `selfsubjectaccessreviews` creates, sustained apiserver 5xx) while `tenant-a` genuinely lacks its `fission-router-dataplane` RoleBinding. The informer's LIST 403s, the reflector retries forever, `informer.HasSynced()` never latches, so `NamespaceInformers.HasSynced` stays false and — after ~31s of probe failures — every replica leaves `svc/router` at once. `pending = true` buys a retry but never an exit.

Bounding it needs a give-up rule rather than a different default. Something like tracking consecutive SAR errors per namespace and treating the namespace as denied (exclude + stop the informer) after K failures keeps the optimism for real flakes while guaranteeing the fleet can't stay wedged on a namespace that never resolves.

pkg/tenant/resolversync.go
```diff
@@ -49,12 +50,29 @@ func SyncResolverFromTenants(ctx context.Context, c client.Client, resolver *uti
...
 	resolver *utils.NamespaceResolver
+	// hooks are called after every successful SetTenants. See AddResolverSync
+	// for the hook/pending contract.
+	hooks []func() (pending bool)
 }
 
+// hookRetryDelay is how soon the reconciler re-runs when a hook reports
+// pending work. Short: the common case is RBAC landing a second or two after
+// the FissionTenant event that triggered the reconcile.
+const hookRetryDelay = 2 * time.Second
```
sanketsudake: **warning — `pending` drives a fixed-interval requeue with no backoff, so a permanently-denied namespace loops forever at 2s.**

`ctrl.Result{RequeueAfter: hookRetryDelay}` goes through `AddAfter`, which bypasses the workqueue's exponential rate limiter, and `hookRetryDelay` is a constant 2s. For the transient case the comment describes (RBAC landing a second or two later) that's exactly right. The problem is the cases that never resolve.

Two reachable ones: a FissionTenant orphaned because its namespace was deleted (which the PR description itself lists as a known follow-up), and a tenant onboarded while `releaseNamespace == ""`, where `pkg/tenant/provision.go` silently skips the router RoleBinding altogether. In both the SAR returns `Allowed=false` forever, so `pending` is permanently true.

The result is every router replica reconciling every 2s indefinitely — roughly 43k reconciles and 86k SAR calls per replica per day — each cycle emitting the `endpointslice RBAC not yet granted for tenant namespace…` line at Info. On a cluster with a couple of orphaned tenants that is a lot of steady-state noise in the router logs and on the authorization API.

Escalating the delay on repeated pending (2s → capped at ~30s-1m) and dropping the repeat log to `V(1)` after the first occurrence per namespace would keep the fast path fast and make the stuck state cheap.

pkg/router/dynamic_endpoint_cache_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/fission/fission/pkg/router/endpointcache"
+	"github.com/fission/fission/pkg/utils"
+)
+
+// newDynamicCacheTest builds a dynamicEndpointCache backed by a fake clientset
+// whose SAR reactor consults the allowedNS map. A namespace present with value
+// true → SAR Allowed=true; absent or false → Allowed=false. The map is shared
+// so tests can flip a namespace's RBAC state between calls (the re-onboard
+// scenario). Returns the cache and the resolver (so tests can onboard/offboard).
+func newDynamicCacheTest(t *testing.T, allowedNS map[string]bool) (*dynamicEndpointCache, *utils.NamespaceResolver) {
```
sanketsudake, line 81: **warning — this file tests the bookkeeping but never the informer set, so the actual #3647 behavior is unpinned.**

The five tests here are a real improvement and they do pin `rbacChecked` and `pending` — the re-onboard regression test in particular is exactly right, and I confirmed it fails when the prune is removed.

But the fixture discards the `Index` (`endpointcache.NewIndex()` is passed inline and never kept) and the fake client has no EndpointSlices, so there is no observable for what `syncInformers` actually *does* with the namespaces. Two mutations were applied and both left the whole package green:

- `d.nsInformers.Sync(allowed)` → `d.nsInformers.Sync(namespaces)` — the RBAC filter is ignored entirely, re-introducing the forbidden-LIST wedge this PR exists to prevent.
- `d.nsInformers.Sync(allowed)` → `d.nsInformers.Sync(nil)` — no informer is ever started, i.e. the warm path never works.

Both ship green today. The doc comments assert the behavior that isn't checked: line 56 says "included in the informer set" and line 72 says "EXCLUDED from the informer set", but neither is asserted anywhere.

Keeping the index and seeding a slice makes it observable:

```suggestion
func newDynamicCacheTest(t *testing.T, allowedNS map[string]bool) (*dynamicEndpointCache, *utils.NamespaceResolver, *endpointcache.Index, *endpointcache.NamespaceInformers) {
```

with `fake.NewSimpleClientset(makeSlice("s1", "team-a", "fn-a", "10.0.0.1"))` for the client and the `Index` kept in a local. Each test can then assert, after `syncInformers`, that `nsi.WaitForCacheSync(t.Context())` returns and `index.Lookup("team-a", "fn-a", "")` is non-empty on the allowed path and empty on the denied path — which is what kills both mutations.

sanketsudake (commented):
## Re-review — PR #3653 (fix round)

Re-reviewed the 9 fix commits in a fresh worktree at merge-base `9c716f90`, running the same battery as before (code review, security, correctness/concurrency, test-coverage, simplify/deslop, reviewability) plus **executed mutation testing** on the new guards.

Local verification, all green: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `make license-check`, `make verify-gomod`, `go test -race -count=1 ./pkg/router/... ./pkg/tenant/...`, and the integration suite compiles under `-tags=integration`. `helm template --set tenancy.mode=dynamic` still renders `fission-router-dataplane-tenant` read-only.

### Verdict: merge-ready with comments — the critical is fixed, no critical survives

**Every one of the 16 findings from the first round is addressed**, and the fixes are correct — I traced each one rather than taking the commit messages at their word. Four of them I verified by applying the mutation and confirming a test fails:

| Prior finding | Fix | Verified |
|---|---|---|
| 1 (critical) `rbacChecked` never invalidated | pruned via `Range` each sync | mutation caught by `TestDynamicCacheReOnboardReRunsSAR` + `TestDynamicCachePrunesOnOffboard` |
| 2 optimistic include didn't requeue | `pending = true` | mutation caught by `TestDynamicCacheSARErrorReportsPending` |
| 3 `Close` not terminal | `closed` flag checked under the same lock | mutation caught (narrowly — see inline) |
| 4 `syncInformers` TOCTOU | `d.mu` + re-read inside the lock | traced; no `n.mu → d.mu` inversion, no deadlock |
| 5 `require` on outer `t` in poll closures | `require.TestingT` | correct — `CollectT.FailNow` is `Goexit`, retries properly |
| 6 goroutine `+10` slop | `fission_router_endpointcache_informers` gauge | exact guard; baseline read is sound |
| 7-10 test gaps | new `dynamic_endpoint_cache_test.go`, fence test, handler/tombstone tests, `HasSynced` state machine, VAP router SA | mostly closed — see below |
| 11-15 maintainability | `framework.SumMetricLines`, `func() bool` hook, `setupDynamicEndpointCache` extraction, comment dedup | all done; `Start` is 430 → 356 lines |

On the original critical specifically: I confirmed the bounded-recovery path end to end. When a namespace is dropped from `allowed`, `nsInformers.Sync` cancels, drains and sweeps its informer, so `HasSynced` recovers and the fleet-wide 503 clears in ~2-6s instead of persisting. The `HasSynced` false→true test using a blocking list reactor is the best-built test in the set.

Two residual correctness issues and two test-efficacy gaps are worth a look before merge; none blocks.

### warning

**1. A *persistent* SAR error still leaves the wedge unbounded.** The transient branch starts an informer for a possibly-unauthorized namespace with no attempt counter or deadline, so if SAR creates never return a verdict the fleet-wide `/readyz` 503 has no exit. Inline, with a suggested give-up rule.

**2. `pending` requeues at a flat 2s with no backoff, forever.** `RequeueAfter` bypasses the rate limiter, so an orphaned FissionTenant (or one onboarded with an empty release namespace, where `provision.go` silently skips the router binding) makes every replica reconcile every 2s indefinitely — ~43k reconciles and ~86k SAR calls per replica per day, each logging the "RBAC not yet granted" line at Info. Inline.

**3. `dynamic_endpoint_cache_test.go` never asserts the informer set.** It pins `rbacChecked` and `pending` well, but the fixture discards the `Index` and seeds no slices, so there's no observable for what `syncInformers` does with the namespaces. Two mutations ship green: passing the *unfiltered* namespace list to `Sync` (re-introducing the exact forbidden-LIST wedge), and passing `nil` (warm path never works). Inline, with a fixture change that kills both.

**4. The fence test pins the fence's shape, not its ordering.** It correctly catches removing `<-ni.done`, but because it injects a hand-made `done` channel it never exercises the real one — moving `close(done)` before `factory.Shutdown()` in `start`, which is the resurrection bug itself, stays green at `-count=5`. Inline.

### suggestion

**5.** The `closed`-guard test's detection is a race it usually wins — only the `HasSynced` assert catches the mutation (50/50), the index asserts catch it 0/50. A blocking list reactor (the pattern already at line 400) makes it deterministic. Inline.

**6.** `TestDynamicCacheSARErrorReportsPending` sleeps 4.00s of real wall clock. A pre-cancelled context exercises the same branch in 0.00s — verified. Inline.

**7. `setupDynamicEndpointCache`'s error path leaks informers.** The seed `syncInformers` starts informers parented on `context.Background()` before the `<-ctx.Done() → Close()` goroutine is registered, so the `crMgr.Add` error return leaves them unstoppable. Practically unreachable and the process exits anyway, but registering the cleanup goroutine before the first `Sync` is free.

**8.** The doc on `setupDynamicEndpointCache` says "the returned nsInformers is closed by the ctx-cancel goroutine" — `nsInformers` isn't among the three return values. Similarly `WaitForCacheSync`'s doc still claims it's used "for the router's readiness gate"; production uses `HasSynced`, this is test-only.

**9. The informer gauge is a package-level synchronous gauge over per-instance state.** `RecordInformersCount` is global while `informerCount` is per-`NamespaceInformers`, so a second instance clobbers the reading (the parallel unit tests create several; only production's single instance saves it). `RegisterSizeGauge` one function up already uses `Int64ObservableGauge` over an `atomic.Pointer` — same pattern here would also make the series always-present instead of appearing after the first transition.

**10. Startup issues the same SARs twice** — `checkSliceWatchRBAC` reviews list+watch for `FunctionNamespaces()`, then the seed `syncInformers` re-reviews the same set with `rbacChecked` empty. Pre-seeding the cache from the preflight would halve startup SAR calls and retry latency on the critical path.

**11. Latent key-space mismatch in `RemoveNamespace`.** Informers are keyed by `FunctionNamespaces()` (i.e. `GetFunctionNS(ns)`, where the Service lives) but `FnKey.Namespace` comes from the slice's `functionNamespace` label (the Function CR's namespace). These diverge when `FISSION_FUNCTION_NAMESPACE` is set and the CR namespace is `default`, in which case the sweep silently matches nothing. Unreachable today since `default` never leaves the live set, but it's a trap for whoever makes the tenant set fully dynamic.

**12. The new warm-path assertions don't gate on the EndpointSlice mode**, unlike `serial/executor_down_warm_test.go` which skips when the gates are off. If the router fail-soft-degrades `mode=on→off` (documented at `router.go:138-141`), `assertWarmPathHit` burns 2 minutes and fails instead of skipping. Latent — today's dynamic leg has the gates on.

**13. Minor:** `alias_routing_test.go:329` still references `memory_soak_test.go`'s `parseMetric`, which this PR deleted. The fence test's injected `cancel` isn't idempotent (`sync.OnceFunc` would remove the double-close trap) and its `informer: nil` would nil-panic if anything consults readiness while the entry is in the map. The SAR fixture's mutex is one-sided — the reactor reads under it, the tests write bare; harmless while `syncInformers` is synchronous, but it advertises a safety property the fixture lacks. And `hits_total` is process-global, so `assertWarmPathHit` proves *some* warm hit, not this tenant's — sound under `-p 1`, but the comment overclaims.

**14.** Eleven commits including `fixed linting error` and three `address … review findings` — worth squashing into one or two before merge.

### security

Unchanged from the first round and still clean: the chart RBAC files are byte-identical to what I reviewed, the ClusterRole renders read-only, and the tenant controller's `bind` stays `resourceNames`-scoped. The pre-existing label-trust issue in `fnKeyForSlice` remains out of scope for this PR — still worth a separate hardening issue, not a blocker here.

Nice work on this round — the re-onboard regression test and the `HasSynced` blocking-reactor test are the kind of thing that keeps a fix like this from silently rotting.

pkg/router/endpointcache/nsinformers_test.go
```diff
@@ -0,0 +1,417 @@
...
+
+// TestNamespaceInformersFencePreventsResurrection verifies the teardown fence
+// deterministically: Sync must block on <-ni.done before RemoveNamespace sweeps
+// the index, and a slice applied while fenced must be swept too (no resurrection).
+//
+// Synchronization: the injected cancel closes entered when Sync calls it,
+// proving the goroutine has reached the fence. After <-entered the goroutine is
+// guaranteed to block on <-ni.done (the next statement), so asserting syncDone
+// is still open is deterministic — no sleeps, no scheduler dependence.
+func TestNamespaceInformersFencePreventsResurrection(t *testing.T) {
```
sanketsudake, line 310: **warning — the fence test pins the fence's shape, not its substance.**

This is a genuinely good deterministic construction — the injected `cancel` closing `entered` proves Sync reached the fence with no sleeps, and I confirmed it catches removing `<-ni.done` cleanly (`t.Fatal` at "fence missing", no hang).

What it can't catch is whether `done` fires at the *right moment*. Because the test injects a hand-made `done` channel, it never exercises the real one built in `start()`, where the whole guarantee lives — `close(done)` must happen after `factory.Shutdown()`, since Shutdown is what blocks until each informer's processor has drained its queued handler notifications. Moving `close(done)` to before `factory.Shutdown()` in `start` — precisely the "a queued ApplySlice resurrects the namespace after the sweep" bug the fence exists to prevent — leaves this test green (verified at `-count=5`).

Catching that needs a real informer whose handler is held mid-notification: start an informer for a namespace, block inside the event handler (or use a slow handler), deliver a slice event, then offboard and assert the index is empty after `Sync` returns. That's more machinery than the shape test, so it may be worth deciding it's out of scope — but the current test shouldn't be read as covering the ordering.

pkg/router/endpointcache/nsinformers_test.go
```diff
@@ -0,0 +1,417 @@
...
+	assert.Equal(t, 1, index.Size(), "Close must not sweep the index")
+}
+
+// TestNamespaceInformersSyncAfterCloseIsNoop verifies the closed guard: after
+// Close, a racing Sync must not repopulate informers. Without the guard, Close
+// drains the map outside the lock, a racing Sync repopulates it, and those new
+// informers — parented on context.Background() — are unreachable by any cancel
+// and leak until process exit. With the guard, Sync returns immediately and the
+// informer map stays empty.
+func TestNamespaceInformersSyncAfterCloseIsNoop(t *testing.T) {
```
sanketsudake, line 192: **suggestion — the `closed`-guard detection is timing-dependent, and there's a deterministic version.**

The mutation is caught today, but narrowly: with the guard removed, `TestNamespaceInformersSyncAfterCloseIsNoop` fails only via the `HasSynced` assertion at line 180 (50/50 runs). The index assertions at lines 176-179 caught it **0/50** — with the guard gone the ns-b informer does start, but its Add notification hasn't propagated by the time they run, so they pass either way.

That leaves detection resting on the microsecond window between `Sync` returning and `HasSynced` being called. Under heavy parallel CI load the ns-b informer can complete its LIST inside that window and latch `HasSynced` true, and the mutation survives.

The file already has the fix pattern at line 400 — register a blocking `list` reactor for `endpointslices` before the post-`Close` `Sync`, so an informer that *does* get started can never latch `HasSynced`, and the assertion becomes deterministic rather than a race the test usually wins.

pkg/router/dynamic_endpoint_cache_test.go
```diff
@@ -0,0 +1,192 @@
...
+	_, cached = dyn.rbacChecked.Load("team-a")
+	assert.True(t, cached, "cached once SAR passes")
+}
+
+// TestDynamicCacheSARErrorReportsPending verifies fix #2: when the SAR itself
+// errors (apiserver flake), the namespace is optimistically included but pending
+// is still set true. Without pending, a genuinely-unauthorized namespace whose
+// SAR consistently errors keeps an unsyncable informer forever — no FissionTenant
+// event fires to re-run the check.
+func TestDynamicCacheSARErrorReportsPending(t *testing.T) {
```
sanketsudake, line 224: **suggestion — this test spends 4 real seconds in `time.After`; a pre-cancelled context makes it instant.**

The test drives `sliceWatchSAR`'s retry loop to exhaustion (3 attempts, 2s backoff between them) with a live `t.Context()`, so it sleeps ~4.00s of wall clock. That's a bare timed wait in a unit test, which `.claude/resources/test-writing-guidelines.md` steers away from, and the repo already has seven files using `testing/synctest` for time-dependent code.

A `synctest` bubble is a poor fit here — `syncInformers` finishes by starting real client-go informer goroutines, which won't go durably blocked — but the retry loop's `select` also watches `ctx.Done()`, so a pre-cancelled context short-circuits the backoff on attempt 2 while still exercising the same branch: `sliceWatchAllowedForNamespace` returns `(true, err)` and the SAR-error path runs identically.

I tried it against this test and it passes in **0.00s** instead of 4.00s, with the same assertions. Worth doing given this runs on every CI leg.

pkg/tenant/resolversync.go
```diff
@@ -49,12 +49,24 @@ func SyncResolverFromTenants(ctx context.Context, c client.Client, resolver *uti
...
 		return ctrl.Result{}, err
 	}
+	pending := false
+	for _, h := range r.hooks {
+		if h() {
+			pending = true
+		}
+	}
+	if pending {
+		return ctrl.Result{Requeue: true}, nil
```
sanketsudake: **warning — this trades the old unbounded-2s-loop for a worst-case convergence regression, and does it through a deprecated field.**

I raised the flat-2s-forever loop last round, so this is my fault for not naming the target precisely. The rate-limited requeue does bound the tail, but it overshoots in the case that matters.

`Requeue: true` routes to `AddWithOpts(RateLimited: true)`, and — unlike the `RequeueAfter` branch — that path does **not** call `Queue.Forget(req)` (controller.go:504-507). So while any hook stays pending, the per-key delay ratchets through the default `ItemExponentialFailureRateLimiter(5ms, 1000s)` and only resets when a reconcile returns a clean `ctrl.Result{}`.

Failure scenario: the tenant controller is down for ten minutes while `team-a`'s FissionTenant exists. Each pending reconcile ratchets the delay: 5ms → 10ms → … → 655s → 1000s. When the RoleBinding finally lands, nothing enqueues `team-a` again — the binding is not a FissionTenant event, which is the whole premise for the hook — so the warm path stays unadmitted until the next scheduled retry, **up to ~16.7 minutes**. The old code converged in 2s. It also front-loads a burst: ~9 reconciles, each issuing 2 SARs per unchecked namespace, in the first ~2.5s.

There's a second hazard worth weighing: `Result.Requeue` is deprecated in the pinned controller-runtime v0.24.1, and its deprecation note describes this exact situation — "When waiting for an external event to happen, either the duration until it is supposed to happen or an appropriate poll interval should be used, rather than an interval emitted by a ratelimiter whose purpose it is to control retry on error." When upstream removes the field, `ctrl.Result{Requeue: true}` silently degrades to `ctrl.Result{}` with no compile error, and the #3647 wedge returns permanently.

The simplest non-deprecated fix is an explicit poll interval — bounded convergence, no ratchet, and ~2.9k reconciles/day instead of 43k:

```suggestion
	if pending {
		// Waiting on an external event (the tenant controller's RoleBinding),
		// not retrying a failure — so poll at a fixed interval rather than
		// borrowing the workqueue's error rate limiter. Long enough that a
		// permanently-denied namespace (orphaned FissionTenant) costs little,
		// short enough that the normal onboarding race converges promptly.
		return ctrl.Result{RequeueAfter: hookRetryPollInterval}, nil
	}
```

with `const hookRetryPollInterval = 15 * time.Second`. If you'd rather keep fast first-retry behaviour, an escalating `RequeueAfter` capped at ~30-60s (counter on the reconciler, reset when `pending` is false) gets both — reconciles are serialized at `MaxConcurrentReconciles: 1`, so a plain int field is safe enough for a backoff heuristic.

Two follow-ons either way: `AddResolverSync`'s doc at line 86 still says "the reconciler then requeues after `hookRetryDelay`", but that constant was deleted in 3589355c; and `resolversync_test.go:85-86` asserts `res.Requeue` + zero `RequeueAfter`, so it pins the mechanism rather than the behaviour and will need rewriting with this.

test/integration/suites/serial/dynamic_tenant_test.go
```diff
@@ -42,12 +42,21 @@ func TestDynamicTenantLifecycle(t *testing.T) {
...
+	}
 	pyImage := f.Images().RequirePython(t)
 
 	// Let any rollout from a prior serial test (e.g. the adopt test's un-waited
 	// executor restore) settle, then snapshot the control plane so we can prove
 	// onboarding restarts nothing — the headline promise of dynamic tenancy.
 	f.WaitForControlPlaneStable(t, ctx, 3*time.Minute)
+	if !f.RouterEndpointSliceCacheActive(t, ctx) {
+		t.Skip("router EndpointSlice cache fell back to off; dynamic warm-path assertions cannot run")
+	}
```
sanketsudake: **warning — this skip fires exactly when the bug class this PR fixes has occurred, converting a regression into a green run.**

Adding the gates was my suggestion last round, and the first two are right: line 45 skips when the *requested* mode (`ROUTER_ENDPOINTSLICE_CACHE_MODE` on the Deployment) isn't `on`, and the executor-gate skip is the same shape as `serial/executor_down_warm_test.go`. Those legitimately mean "this leg doesn't run this data plane".

This third one is different. It reads the **effective** runtime mode from `fission_router_endpointcache_mode`, and by the time control reaches it we already know the requested mode is `on`. So the only way it can be false is that the router's startup RBAC preflight failed and it silently degraded to the executor-RPC plane — a real defect, and precisely the failure this PR's RBAC work exists to prevent.

Failure scenario: a chart regression drops the router-dataplane Role (or the ClusterRole rename drifts) on the kind-ci dynamic leg. `checkSliceWatchRBAC` fails, the router degrades to `mode=off`, and `TestDynamicTenantLifecycle` reports **SKIP** — the informer-count leak guard, both warm-path assertions, and the offboard sweep all silently vanish from CI while the run stays green.

Once requested mode is `on`, a degraded effective mode is an assertion, not a precondition:

```suggestion
	require.Truef(t, f.RouterEndpointSliceCacheActive(t, ctx),
		"router requests endpointSliceCache.mode=on but degraded to the executor-RPC plane; "+
			"the startup RBAC preflight failed — the dynamic warm path cannot work (see router-dataplane Role/ClusterRole)")
```

sanketsudake (commented):
## Third-round review — PR #3653

Re-reviewed the 11 new fix commits in a fresh worktree at merge-base `b40201a2`, same battery as before, with **executed mutation testing** on the two guards I said were weakly covered last round.

Local verification, all green: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `make license-check`, `make verify-gomod`, `go test -race -count=1 ./pkg/router/... ./pkg/tenant/...`, `go test ./test/integration/framework/`, and the integration suite compiles under `-tags=integration`.

### Verdict: merge-ready with comments — no critical, and every round-2 finding is closed

All 14 round-2 findings are addressed, and the two I was least sure about are now genuinely pinned rather than nominally covered. I re-ran the exact mutations that survived last time:

| Round-2 finding | Fix | Verification |
|---|---|---|
| W1 unbounded wedge on persistent SAR error | `sarErrors` counter, excluded after 3 | `TestDynamicCachePersistentSARErrorsEventuallyExcludeAndRecover` walks errors 1→5 then recovery, asserting index size, counter and cache at each step |
| W2 flat-2s requeue forever | rate-limited requeue | bounded, but see warning 1 below |
| W3 informer set never asserted | new `router_test.go` with a real EndpointSlice + `index.Size()` per step | the `Sync(nil)` / unfiltered-`Sync` mutations no longer ship green |
| W4 fence pinned shape, not ordering | rewritten with a real informer whose watch `Stop()` blocks | **I applied the exact previously-surviving mutation** (`close(done)` before `factory.Shutdown()`): now fails 3/3 with "Sync completed before the watcher fully stopped" |
| S5 `closed`-guard detection timing-dependent | blocking `list` reactor for ns-b | **mutation now fails 5/5** (was 0/50 on the index asserts) |
| S6 4s real sleep | pre-cancelled context | `TestDynamicCacheSARErrorReportsPending` 4.00s → **0.00s**; all 8 `TestDynamicCache*` now 1.4s total |
| S7 cleanup goroutine registered late | moved above `crMgr.Add` | closes the real leak on the error return |
| S8 stale docs | `setupDynamicEndpointCache` + `WaitForCacheSync` docs corrected | verified gone |
| S9 package-level gauge over per-instance state | `RegisterInformersGauge` + `Int64ObservableGauge`, symmetric with `RegisterSizeGauge` | series is emitted at 0, so the integration baseline still works |
| S10 duplicate startup SARs | preflight returns its namespace list, pre-seeds `rbacChecked` | `TestSetupDynamicEndpointCacheReusesPreflightRBAC` asserts zero SARs |
| S11 `RemoveNamespace` key-space mismatch | rewritten to sweep by the slice's own namespace | three subtests including a genuinely-divergent `FISSION_FUNCTION_NAMESPACE` case |
| S12 missing gate skips | three gates added | two are right; the third is warning 2 below |
| S13 minor cleanups | nil-informer injection gone, one-sided mutex removed, `hits_total` comment corrected, stale `parseMetric` reference dropped | all confirmed |

`RemoveNamespace` deserves a specific note: this wasn't a hypothetical. On a `FISSION_FUNCTION_NAMESPACE` install the old `key.Namespace != ns` match swept **nothing** on offboard, because the informer set is keyed by the Service's namespace while `FnKey.Namespace` comes from the `functionNamespace` label. The rewrite fixes a real stale-endpoint bug, the lock order is still shard→entry, `ix.size` accounting stays exact under concurrent `ApplySlice`/`DeleteSlice`/`RemoveNamespace`, and `strings.Cut` is safe since both halves are DNS-1123 names.

`MetricHasLabels` using `expfmt` with its own table test — including the malformed-input case — is a nicer result than the hand-rolled parser I suggested.

Two things worth changing before merge, neither blocking.

### warning

**1. The requeue change overshoots and uses a deprecated field.** `Requeue: true` never calls `Forget`, so a persistently-pending namespace ratchets to the rate limiter's 1000s cap — when the RoleBinding finally lands, the warm path can stay unadmitted for ~16.7 minutes where the old code converged in 2s. `Result.Requeue` is also deprecated in the pinned v0.24.1, and its note recommends a poll interval for exactly this "waiting for an external event" case; when it's removed, the call silently becomes `ctrl.Result{}` and the wedge returns with no compile error. I see the `//nolint:staticcheck` in the test, so this was deliberate — the substantive argument is the convergence regression, not the lint. Inline, with a fixed-poll suggestion and an escalating-cap alternative. My round-2 wording ("rate-limit") pointed at the wrong mechanism; that's on me.

**2. `RouterEndpointSliceCacheActive` should be an assertion, not a skip.** Reaching it means the requested mode is already `on`, so a false result means the router degraded after an RBAC preflight failure — the exact defect this PR guards against. As written, a chart regression that drops the router-dataplane Role turns the whole dynamic warm-path suite into a green SKIP. Inline, with a `require.Truef` suggestion. The other two gates are correct.

### suggestion

**3. `maxOptimisticSARErrors = 3` is undocumented** (router.go:230) in a file where every other constant carries a rationale, and `sliceWatchAllowedForNamespace`'s doc still says a transient error "is reported as allowed … so a wobble never tears down working watches" — true of the function, no longer true of the caller, which now overrides it after three tries. Worth a line on the type doc or the constant. The inline comment at router.go:307-308 ("we check for SAR errors optimistically") is also lowercase/first-person, out of step with the surrounding prose.

**4. Cancellation counts as an SAR error.** `sliceWatchSAR` returns `ctx.Err()`, so a shutting-down router increments `sarErrors` and logs at `Error` for every namespace. Cosmetic, but it's also the path the new tests drive (pre-cancelled context) rather than a genuine apiserver flake — worth distinguishing `ctx.Err()` from a real API failure if you touch this again.

**5. `.gitignore`** gained `/.agents/`, `/.continue/`, `/.devin/`, `/.windsurf/`, `/skills-lock.json`, bundled inside `fix(router): register informer cleanup early`. Unrelated agent-tooling ignores in a router bug-fix PR, and folding them into a behaviour commit makes that commit non-atomic to revert or bisect. Better split out or dropped.

**6. Two notes on the `RemoveNamespace` rewrite** (index.go:340-364), neither a bug: a *partially* removed entry now has `quarantined`/`strikes` cleared for the surviving namespace's addresses, which the old code never did for surviving entries — defensible under the "any slice event lifts quarantines" rule, but it's a new side effect the doc doesn't mention; and `s.mu.Lock()` is now held across a full shard scan with one `e.mu.Lock()` per entry, so on a large index offboard is a measurable stall against `Lookup`/`Admit`. Fine for a rare operation, worth a word in the comment.

**7. `NewNamespaceInformers` now has a process-global side effect** (`RegisterInformersGauge`). Production is safe with its single call site, but CLAUDE.md asks library constructors to stay deterministic, and a future second instance would silently mis-attribute the gauge and pin the older instance alive. Registering from `setupDynamicEndpointCache`, next to the sibling `RegisterSizeGauge(index)` call, would keep the constructor clean and the two gauges symmetric at the call site too.

**8.** 21 commits now, including `fixed linting error` and four `address … review findings`. Worth squashing to one or two before merge.

### security

Unchanged and still clean. The chart RBAC files are byte-identical to what I cleared in round one, the ClusterRole renders read-only, and the tenant controller's `bind` stays `resourceNames`-scoped. The pre-existing `fnKeyForSlice` label-trust issue remains out of scope for this PR and still deserves its own hardening issue.

This round is a real step up — the fence test now pins the ordering it claims to, and the SAR escalation table walks the whole state machine including recovery. Nice work.

pkg/router/endpointcache/nsinformers_test.go
```diff
@@ -0,0 +1,476 @@
...
+	t.Cleanup(nsi.Close)
+	assert.True(t, nsi.HasSynced(), "empty informer set is vacuously synced")
+}
+
+// waitForIndexSize waits for the index to reach size n. WaitForCacheSync only
+// guarantees the reflector's initial LIST is stored; the Add notifications
+// still travel through the informer's async processor before ApplySlice runs,
+// so index state is genuinely eventual-convergence after a sync (bounded here
+// — never a bare sleep).
+func waitForIndexSize(t *testing.T, index *Index, n int) {
```
sanketsudake, line 126: **warning — `waitForIndexSize` is not a sufficient sync point for the address assertions that follow it, and it flakes.**

I hit a real failure running `go test -race ./pkg/router/... ./pkg/tenant/...`:

```
--- FAIL: TestNamespaceInformersCreateUpdateDeleteThroughInformer
    ([]string) {
    }
```

Root cause is a window in `ApplySlice` (index.go): the entry is created and `ix.size.Add(1)` happens under the **shard** lock, which is then released; `e.slices` is populated and `rebuildLocked()` runs afterwards under the **entry** lock. So between those two sections `Size()` already reports the entry while `Lookup` still returns nothing.

I confirmed the window exists rather than inferring it — a throwaway probe that races `ApplySlice` against a poller observed `Size()==1` with an empty `Lookup()` directly (then deleted; worktree left clean).

That makes every `waitForIndexSize(...)` immediately followed by an `ElementsMatch` on addresses a race the test usually wins. It's used that way at lines 70, 99, 161, 204, 241, 250, 286, 322 and 389. It is genuinely rare — one failure in about nine multi-package runs, and 0/40 for the individual tests without CPU contention — so it needs the load of a parallel package run to surface, which is exactly what CI does.

Waiting on the thing you're about to assert removes it entirely:

```suggestion
// waitForIndexSize waits for the index to reach size n. Note that Size() alone
// is NOT a sufficient sync point for asserting on endpoints: ApplySlice bumps
// ix.size under the shard lock and only fills e.slices/rebuilds under the entry
// lock afterwards, so a poller can observe the entry before its endpoints exist.
// Assert addresses with waitForEndpoints instead.
func waitForIndexSize(t *testing.T, index *Index, n int) {
```

plus a companion the address sites can call:

```go
// waitForEndpoints waits until the function's endpoint set matches want,
// closing the size-before-endpoints window in ApplySlice.
func waitForEndpoints(t *testing.T, index *Index, ns, fn string, want ...string) {
	t.Helper()
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		assert.ElementsMatch(c, want, addrs(index.Lookup(ns, fn, "")))
	}, 10*time.Second, 10*time.Millisecond, "endpoints for %s/%s never converged to %v", ns, fn, want)
}
```

The update step in this same test already has the right shape (`Eventually` on `len(got) == 2` before asserting), so this is just making the other sites match it.

pkg/router/router.go
```diff
@@ -145,47 +151,286 @@ func checkSliceWatchRBAC(ctx context.Context, kubeClient kubernetes.Interface) e
...
 			}
+			d.logger.Error(err, "endpointslice RBAC check continues to fail; excluding namespace", "namespace", ns, "consecutive_errors", d.sarErrors[ns])
+			continue
 		}
+		delete(d.sarErrors, ns)
+		if !ok {
+			d.logger.Info("endpointslice RBAC not yet granted for tenant namespace; "+
+				"the tenant controller provisions fission-router-dataplane at onboarding — "+
+				"namespace excluded from the warm path until it lands", "namespace", ns,
+			)
```
sanketsudake, line 353: **suggestion — with the poll back at a flat 2s, this line is the steady-state noise I raised two rounds ago.**

Reverting to `RequeueAfter` is the right call and the rationale comment on `hookRetryDelay` is excellent — it captures the ratchet problem precisely. But the interval landed at 2s rather than the 15s I suggested, and this branch has no rate limiting of its own, so for a namespace that is *permanently* denied the pair is back to firing forever.

The reachable case is the one your own PR description lists as a known follow-up: a FissionTenant orphaned when its namespace is deleted without `tenant disable`. Its SAR returns `Allowed=false` on every pass, so `pending` is permanently true and each cycle costs one SAR plus this Info line — roughly 43k reconciles, 43k SARs and 43k log lines per replica per day, multiplied by replica count. (The `sarErrors` cap doesn't help here: that governs the *error* branch, and this is a clean deny.)

Two cheap ways out, either fine:

- Escalate the interval on consecutive pending rounds and cap it — `2s` doubling to `30s` gets fast convergence on the normal onboarding race *and* a quiet steady state, without going near the deprecated `Requeue` or its 1000s ratchet.
- Or keep 2s and make the repeat cheap: log this at `Info` the first time per namespace and `V(1)` thereafter, tracked next to `sarErrors`.

Not blocking — an orphaned tenant is already a known follow-up, and 2s-with-noise is strictly better than the 16.7-minute convergence hole it replaced.

sanketsudake (commented):
## Fourth-round review — PR #3653

Re-reviewed commit `d980d898` in a fresh worktree at merge-base `b40201a2`.

Local verification, all green except one intermittent test (below): `go build ./...`, `go vet`, `golangci-lint` (0 issues), `make license-check`, `make verify-gomod`, `go test ./test/integration/framework/`, integration suite compiles under `-tags=integration`.

### Verdict: merge-ready with comments — every round-3 finding is closed

| Round-3 finding | Resolution |
|---|---|
| **W1** `Requeue: true` deprecated + ~16.7min convergence hole | `RequeueAfter: hookRetryDelay` restored, with a `hookRetryDelay` doc that spells out both the ratchet and the silent-no-op-on-removal hazard — see the one caveat below |
| **W2** effective-mode gate skipped instead of asserting | now `require.Truef` with a comment explaining why a degraded router must fail rather than green-skip |
| **S3** `maxOptimisticSARErrors` undocumented, stale `sliceWatchAllowedForNamespace` doc, off-style comment | all three fixed; `sarRetryDelay` extracted as a named constant too, and the "3 attempts / 2 delays" off-by-one in the retry docs corrected |
| **S4** cancellation counted as an SAR error | `errors.Is(context.Canceled/DeadlineExceeded)` short-circuits before the error accounting, with a new table test covering both |
| **S6** `RemoveNamespace` undocumented side effects | both documented — the quarantine-clearing on partially-removed entries and the shard-lock stall on large indexes |
| **S7** constructor global side effect | `RegisterInformersGauge` moved out of `NewNamespaceInformers` into `setupDynamicEndpointCache`, after setup succeeds |

I checked the one thing that worried me about the S4 fix: the early `return pending` skips `nsInformers.Sync(allowed)` entirely, so if `DeadlineExceeded` could arise from something other than shutdown it would silently stall a tenant with no requeue. It can't here — no REST client timeout is configured and the hook is passed the router context, which carries no deadline, so the branch is shutdown-only in practice. Worth remembering if a client timeout is ever added; `ctx.Err() != nil` would be the precise test now that the tests no longer depend on `errors.Is` (they were rewritten to use `synctest` with a genuine API error).

Moving `TestDynamicCacheSARErrorReportsPending` into a `synctest` bubble is a better answer than the pre-cancelled context I suggested — it exercises the real retry path at virtual time and matches the repo's existing convention.

### warning

**1. New tests flake: `waitForIndexSize` isn't a valid sync point for the address assertions that follow it.** I hit a real `--- FAIL` on `TestNamespaceInformersCreateUpdateDeleteThroughInformer` during `go test -race ./pkg/router/... ./pkg/tenant/...`. `ApplySlice` bumps `ix.size` under the shard lock and only fills `e.slices`/rebuilds under the entry lock afterwards, so `Size()` can report an entry whose `Lookup` is still empty — I demonstrated that window directly with a throwaway probe rather than inferring it. Nine `waitForIndexSize` → `ElementsMatch` sites share the pattern. Rare (1 in ~9 multi-package runs, 0/40 without CPU contention), so it needs parallel-package load to surface — which is what CI does. Inline, with a `waitForEndpoints` helper that closes it. This predates `d980d898`; I'm only catching it now because I ran the packages together.

### suggestion

**2. The poll interval landed at 2s rather than 15s, so the steady-state noise from round 2 is back.** For a permanently-denied namespace — an orphaned FissionTenant, which your description already lists as a follow-up — `pending` stays true forever and each 2s cycle costs a SAR plus an Info line: ~43k of each per replica per day. Escalating 2s→30s on consecutive pending rounds, or demoting the repeat log to `V(1)` after the first occurrence per namespace, resolves it without going near `Requeue`. Inline. Strictly better than the 16.7-minute hole it replaced, so this is a polish item, not a regression.

**3. `.gitignore`** still carries `/.agents/`, `/.continue/`, `/.devin/`, `/.windsurf/`, `/skills-lock.json`, bundled inside `fix(router): register informer cleanup early`. Unrelated agent-tooling ignores in a router bug-fix PR, and folding them into a behaviour commit makes it non-atomic to revert or bisect.

**4.** 22 commits, including `fixed linting error` and five `address … review findings`. Worth squashing to one or two before merge.

### security

Unchanged and still clean — chart RBAC byte-identical to round one, ClusterRole read-only, tenant-controller `bind` still `resourceNames`-scoped. The pre-existing `fnKeyForSlice` label-trust issue remains out of scope and still deserves its own issue.

Four rounds in, the substance is done: the critical is fixed and verified by mutation, the fence and closed-guard tests pin what they claim, and the SAR escalation walks its whole state machine including recovery. What's left is one flaky-assertion pattern and three housekeeping items.

sanketsudake:
Pushed three commits to this branch (maintainer edit) addressing the outstanding items from the last review — please review them like any other change and revert anything you disagree with.

**`08773e9b` — test: wait on endpoints, not index size, before asserting addresses.** This is the flake I hit. `Index.Size()` isn't a valid sync point for an endpoint assertion: `ApplySlice` bumps `ix.size` under the *shard* lock, releases it, and only then fills `e.slices`/rebuilds under the *entry* lock — so a poller can see `Size()==n` while `Lookup` still returns nothing. Added `waitForEndpoints`, which polls the value being asserted, and used it at the four sites that asserted addresses straight after `waitForIndexSize`. The other `waitForIndexSize` calls only assert size and are unchanged; its doc now describes the window so the pattern doesn't come back.

**`6866e810` — fix: escalate the resolver-sync pending poll interval.** 2s → 4s → 8s → 16s → 30s, holding at 30s, reset on the first clean reconcile. Keeps the normal onboarding race converging at 2s while an orphaned FissionTenant settles into a cheap poll instead of ~43k reconciles/SARs/log lines per replica per day. Kept your `RequeueAfter` decision and the reasoning comment — this only adds the escalation on top.

**`a6abe72d` — chore: drop the agent-tooling `.gitignore` entries.** Unrelated to the router change and folded into a behaviour commit. If you want them, a separate PR (or a personal global gitignore) keeps this one bisectable.

**Not done: squashing.** That needs a force-push over your branch, and I won't rewrite published history on someone else's fork without you asking. GitHub's squash-merge will collapse it at merge time anyway.

Verified locally before pushing: `go build ./...`, `go vet`, `golangci-lint` (0 issues), `make license-check`, `make verify-gomod`, `go test -race -count=1 ./pkg/router/... ./pkg/tenant/...` (4 consecutive clean multi-package runs — that's the load pattern that surfaced the flake), and the integration suite compiles under `-tags=integration`.

PR #3662, opened 2026-08-04
executor: create-routed reconciles must converge the deployment spec — fixes the swallowed-update staleness behind the TestFunctionNameUpdate flake

sanketsudake:
## TL;DR

A `fission fn update` can be **consumed without effect** — the CR updated, the reconcile "successful", the pods never rolled, and no event left to re-fire — whenever a create-routed reconcile branch carries the update.
Measured in the wild as `TestFunctionNameUpdate` serving the old entrypoint's body (`alpha`, status 200) for its entire 180s window.

## Forensics

The flake's failed run had two red attempts with **different** root causes (attempt-scoped log API; the rerun had overwritten the artifacts):
- **Attempt 1**: `router GET → 500: Post http://executor.fission/...: lookup executor.fission: i/o timeout` — the DNS cascade, already fixed at the source by #3659 (that exact short name is now an absolute FQDN).
- **Attempt 2**: **zero** DNS lines, status 200, `alpha` for 180+s — this bug.

## Mechanism

`reconcileNewdeployFunc` / `reconcileContainerFunc` route to `createFunction` on first sight and on drift (`resourcesExist` false).
`createFunction` only **adopts/scales** an existing Deployment — it never rewrites the pod spec.
So when a create-routed branch fires on the reconcile that carries a spec change — a create+update coalesced into one first reconcile, or `resourcesExist` **false-alarming from Manager-cache lag** on the update's own reconcile — the old-spec Deployment is adopted, the reconcile succeeds, and `lastReconciled` stores the **new** spec.
The update is now swallowed: `GenerationChangedPredicate` filters resyncs, so nothing ever re-fires.

newdeploy already carried the cure for exactly this on its first-sight branch: `reconcileDeploymentSpec` — annotation-keyed (`FUNCTION_RESOURCE_VERSION`), a no-op when converged, and a template-identical push doesn't roll pods.
Its drift branch lacked it; **container lacked the helper entirely**, on both branches.

## Fix

Chase `createFunction` with `reconcileDeploymentSpec` on **every** create-routed branch, in both executors; container gains the mirrored helper.
The routing is now genuinely level-triggered: any reconcile that routed through create converges the deployment to the current spec.

## Tests

The three routing cases (container first-sight, container drift, newdeploy drift) now pin the chaser with failure messages that explain the swallow; removing any chaser fails exactly the matching case (mutation-verified, mutants compiled).
Existing behavior pins untouched: plain updates still go through the diff path only.

### DIFF
--- a/pkg/executor/executortype/container/containermgr.go
+++ b/pkg/executor/executortype/container/containermgr.go
@@ -606,6 +606,42 @@ func (caaf *Container) updateFunction(ctx context.Context, oldFn *fv1.Function, 
 	return nil
 }
 
+// reconcileDeploymentSpec brings an already-existing deployment up to the
+// function's current spec when it lags — the container-type mirror of
+// newdeploy's helper, and the level-trigger half of reconcileContainerFunc:
+// createFunction only adopts/scales an existing deployment, so a reconcile
+// that both routed through createFunction AND carried a spec change (a
+// coalesced create+update, or a drift false-alarm on the update's reconcile)
+// would otherwise leave the deployment on the old spec with nothing left to
+// re-fire. The deployment carries the function's ResourceVersion as an
+// annotation (getDeployAnnotations); compare it and push the current spec
+// when stale. A no-op when already current.
+func (caaf *Container) reconcileDeploymentSpec(ctx context.Context, fn *fv1.Function) error {
+	// Live entry only: versioned projections sharing the UID pin immutable
+	// snapshots that must not be rewritten to the live spec.
+	fsvc, err := caaf.fsCache.GetLiveByFunctionUID(fn.UID)
+	if err != nil {
+		// Not specialized yet — no deployment to reconcile; the on-demand path
+		// creates it from the current spec on first invocation.
+		return nil
+	}
+	ns := caaf.nsResolver.GetFunctionNS(fn.Namespace)
+	existingDepl, err := caaf.kubernetesClient.AppsV1().Deployments(ns).Get(ctx, fsvc.Name, metav1.GetOptions{})
+	if err != nil {
+		if k8sErrs.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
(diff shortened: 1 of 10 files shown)


sanketsudake:
Review battery (code-review + thermo-nuclear quality audit + simplify, three independent agents on an isolated worktree) ran against `7464c50f`; everything actionable is applied in `b16ea77b`.

**Two behavioral gaps found and closed:**
1. **The HPA half of the swallow** (code-review, confirmed): a create-routed reconcile carrying an *InvokeStrategy* update was still consumed — the deployment chaser ran, but `CreateOrGetHpa`'s adopt branch reconciled only `Metrics`, so min/max/behavior stayed stale forever while the deployment scaled up and the stale HPA dragged it back. The adopt branch now converges min/max/behavior (the ExecutionStrategy is their sole writer — verified no other writers exist). A pre-existing idempotency test had been silently leaning on that drift being ignored; its seed is fixed to mean what its name says.
2. **Container selector wedge** (code-review, confirmed): container's `updateFuncDeployment` lacked newdeploy's immutable-selector guard, and the new chaser made the rejected in-place Update reachable without a pod-spec change — a permanent requeue-forever. Guard ported.

**Structure (all three lenses converged):** the duplicated create-then-chase blocks are now a named `createAndConverge` per package — the single documented home of the swallow story (six other retellings became pointers); the comments the first commit falsified (`podTemplateChanged`'s "only on the old == nil branch", the "on first sight" log/docs) now tell the truth. Cross-package hoisting was considered and rejected with a concrete reason: the manager interfaces are built from unexported methods, so sharing would force exporting the executor's internals — the structurally-identical mirror is deliberate.

**Test hardening:** ordered routing pins (create *then* converge), createFunction-failure short-circuit, direct branch tests for `reconcileDeploymentSpec` in both packages (RV-equal writes nothing; stale converges and restamps), plus pins for the HPA convergence and the selector guard. All new guards mutation-verified with compiling mutants.

PR #3669, opened 2026-08-05
fix(spec): list cluster Packages once per apply, not twice

### LINKED ISSUE #3664: spec apply: enumerate cluster-wide Packages once, not twice
### Summary

A single `fission spec apply` enumerates **every `Package` in every namespace twice**, from two different functions in `pkg/fission-cli/cmd/spec/apply.go`. The two lists serve genuinely different purposes and both are correct — this is an efficiency / RBAC-surface cleanup, **not** a correctness or security bug, and it is **not release-blocking**.

### Where

Both calls use `metav1.NamespaceAll` and both run unconditionally in every apply:

1. **`applyArchives`** — [`apply.go:383`](https://github.com/fission/fission/blob/main/pkg/fission-cli/cmd/spec/apply.go#L383)
   ```go
   pkgs, err := fclient.FissionClientSet.CoreV1().Packages(metav1.NamespaceAll).List(input.Context(), metav1.ListOptions{})
   ```
   Lists all packages cluster-wide to build a content-indexed map (`sha256 → archive URL`) so it can skip re-uploading an archive whose bytes already exist somewhere in the cluster.

2. **`applyPackages`** — [`apply.go:730`](https://github.com/fission/fission/blob/main/pkg/fission-cli/cmd/spec/apply.go#L730)
   ```go
   l, err := packages(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
   ```
   Lists all packages cluster-wide *again* to diff the spec's desired packages against what is live (the create/update/delete reconcile).

### Why it's worth fixing

- One command does two full cluster-wide `Package` enumerations. On a small install (tens of packages) it is negligible; on a large multi-tenant cluster with thousands of packages — which carry archive metadata and build logs in `.status`, so they are not tiny — it is two full listings of extra apiserver load and latency per apply.
- The second list is the redundant one: the full `Package` list `applyArchives` already fetched is exactly the input `applyPackages` needs.
- `NamespaceAll` also requires cluster-wide list permission on packages — a broader RBAC ask than a namespaced apply arguably should need (secondary; pre-existing to this specific redundancy).

### Why it's a "decide", not a trivial dedup

The two lists are taken at **different points in time**: `applyArchives` runs **before** archives are uploaded, `applyPackages` **after** (packages reference the just-uploaded archives). Reusing the first snapshot for the second asserts that nothing relevant changed in between. Within a single apply that is defensible (the CLI is the only writer), but it is a real assumption that should be made **explicit in a comment** rather than assumed away.

### Proposed shape

Fetch the cluster-wide `Package` list once, thread it to both consumers (`applyArchives` builds its `sha256 → URL` map from it; `applyPackages` diffs against it), with a comment pinning the single-writer assumption that makes the shared snapshot safe. Add a small regression test that asserts a single apply issues one `Packages.List` (e.g. via a fake clientset reactor counting the verb).

### Not release-blocking

Correctness is unaffected today; both lists are correct. Filing as tracked tech-debt / a good-first-issue-sized cleanup.

akshita317:
## What

`fission spec apply` enumerated every Package cluster-wide **twice** — `applyArchives` (to build the sha256→archive-URL dedup map) and `applyPackages` (to diff desired vs live) each issued their own `Packages.List(NamespaceAll)`.

This fetches the list **once** in `applyResources` (new `listClusterPackages` helper) and threads the same snapshot to both consumers. `destroy`'s `forceDeleteResources` fetches-and-passes it too. A comment pins the single-writer assumption that makes sharing one snapshot safe: within an apply the fission CLI is the only writer of Packages, and the archive uploads that happen between the two consumers touch storage, not Package objects.

## Why

Negligible on small installs; on large multi-tenant clusters (thousands of Packages that carry archive metadata and build logs in `.status`, so they are not tiny) it is two full cluster-wide `Package` listings of apiserver load and latency per apply. This halves it. `NamespaceAll` also only needs to be requested from one call site now.

## Testing

Added `TestSpecApplyListsPackagesOnce`, which drives a dry-run apply against a fake clientset with a reactor counting the `list` verb on `packages` and asserts it fires exactly once (it was twice before). Full `spec` package suite and `go vet` pass locally.

Fixes #3664

### DIFF
--- a/pkg/fission-cli/cmd/spec/apply.go
+++ b/pkg/fission-cli/cmd/spec/apply.go
@@ -355,13 +355,27 @@ func pluralize(num int, word string) string {
 	return word + "s"
 }
 
+// listClusterPackages returns every Package across all namespaces. A single
+// `spec apply` needs this list twice — to build the archive dedup map
+// (applyArchives) and to diff desired vs live packages (applyPackages) — so the
+// caller fetches it once and threads the same snapshot to both, halving the
+// apiserver load of an apply on large multi-tenant clusters where Packages carry
+// archive metadata and build logs in .status and are far from tiny.
+func listClusterPackages(ctx context.Context, fclient cmd.Client) ([]fv1.Package, error) {
+	l, err := fclient.FissionClientSet.CoreV1().Packages(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return l.Items, nil
+}
+
 // applyArchives figures out the set of archives that need to be uploaded, and uploads them.
 // Under dryRun the read-only work still runs — local archives are built/checksummed
 // and matched against archives already on the cluster — so the resolved Package
 // specs (and therefore the diff) are accurate for unchanged archives; only the
 // actual upload of a new/changed archive is skipped (such a Package legitimately
 // shows as a would-create/update).
-func applyArchives(input cli.Input, fclient cmd.Client, specDir string, fr *FissionResources, dryRun bool) error {
+func applyArchives(input cli.Input, fclient cmd.Client, specDir string, fr *FissionResources, dryRun bool, clusterPkgs []fv1.Package) error {
 	// archive:// URL -> archive map.
 	archiveFiles := make(map[string]fv1.Archive)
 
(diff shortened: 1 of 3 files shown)


sanketsudake:
Closing — issue #3664 received multiple PRs for the same fix, so we are consolidating on a single one (#3668). Thanks for the contribution.

PR #3670, opened 2026-08-06
support: capture KEDA ScaledObjects and TriggerAuthentications in the dump

### LINKED ISSUE #1914: Add support to capture keda CRDs in support dump
If a user is using fission mqt of type keda, several keda CRDs are also created. However, these CRDs aren't captured by support dump. 

Capability to capture keda CRDs using support dump command.

sushantlokhande14:
Closes #1914.

## What

A `MessageQueueTrigger` of type `keda` causes the scaler manager to create a `ScaledObject` and, for secret-bearing triggers, a `TriggerAuthentication` (`pkg/mqtrigger/scalermanager.go`). Neither appeared in `fission support dump`, so a scaling problem on a keda mqt had to be diagnosed without the objects that actually drive the scaling.

Adds a `KedaDumper` for both kinds, registered as `fission-keda-scaledobjects` and `fission-keda-triggerauthentications`.

## Three decisions worth reviewing

**1. Ownership filter, not "dump everything."** KEDA is a cluster-wide component that other workloads use too, so listing every `ScaledObject` would put an unrelated team's scaling config into a fission support bundle. Only objects whose `ownerReferences` contain a `MessageQueueTrigger` are dumped.

This is also why it can't reuse the existing label-selector-based `KubernetesObjectDumper`: `getScaledObject`/`getAuthTriggerSpec` set **no labels** on what they create — only `newOwnerReference(...)` — so ownership is the only selector available.

**2. A cluster without KEDA is not an error.** That's the normal case for anyone not using keda mqts. `NoKindMatch`/`NotFound` from the list is reported via `console.Verbose(2, ...)` rather than `console.Warn`, so a warning doesn't appear in every bundle from every non-KEDA install. A `Forbidden` or any other error still warns — there's a table test pinning that distinction.

**3. The Vault token is masked.** `TriggerAuthenticationSpec.HashiCorpVault.Credential.Token` is a **plaintext token**, and it is the one field in the spec that carries a credential inline — `secretTargetRef`, `configMapTargetRef`, `env` and even `azureKeyVault`'s `clientSecret.valueFrom` all *name* a Secret or ConfigMap rather than holding its value, so those are safe to dump as-is.

Masked with `"-"`, matching what `pkgClean` already does for archive literals. Fission's own scaler only ever sets `secretTargetRef`, so in practice this never fires — but a support dump is sent to a third party, and nothing stops an operator editing the object the scaler created. The masking deep-copies first, since the token sits behind a pointer shared with the List result.

## Tests

`keda_test.go`, 5 tests + 4 subtests, all passing:

- **Ownership** — of three ScaledObjects (MQT-owned, Deployment-owned, unowned), only the MQT-owned one is written.
- **Masking** — the token string never appears in the dumped file, and the field is present as `token: '-'` so it is clear it *was* set rather than absent.
- **No mutation** — `triggerAuthClean` leaves the input object's token intact.
- **Unknown kind** — writes nothing, doesn't panic.
- **`kedaAbsent`** — table test: `NoKindMatchError` and `NotFound` → quiet skip; `Forbidden` and an arbitrary error → real warning.

## Verification

```
go build ./...                              clean
go vet ./pkg/fission-cli/cmd/support/...    clean
go test ./pkg/fission-cli/cmd/support/...   ok  (5 tests, 4 subtests)
gofmt                                       clean
```

`go.mod`/`go.sum` are untouched — `github.com/kedacore/keda/v2` is already a direct dependency, and both the typed and fake clientsets come from that module. No codegen, no API or CRD change.

## Note

I used `console.Verbose(2, ...)` for the KEDA-absent path to keep bundles from non-KEDA clusters clean. If you'd rather that be `Warn` for discoverability, it's a one-line change — say the word.

### DIFF
--- a/pkg/fission-cli/cmd/support/dump.go
+++ b/pkg/fission-cli/cmd/support/dump.go
@@ -98,6 +98,11 @@ func (opts *DumpSubCommand) do(input cli.Input) error {
 		"fission-crd-mqtriggers":    resources.NewCrdDumper(opts.Client(), resources.CrdMessageQueueTrigger),
 		"fission-crd-timetriggers":  resources.NewCrdDumper(opts.Client(), resources.CrdTimeTrigger),
 		"fission-crd-canaryconfigs": resources.NewCrdDumper(opts.Client(), resources.CrdCanaryConfig),
+
+		// KEDA objects created by the mqt-keda scaler for MessageQueueTriggers.
+		// Skipped quietly when KEDA is not installed.
+		"fission-keda-scaledobjects":          resources.NewKedaDumper(opts.Client(), resources.KedaScaledObject),
+		"fission-keda-triggerauthentications": resources.NewKedaDumper(opts.Client(), resources.KedaTriggerAuthentication),
 	}
 
 	dumpName := fmt.Sprintf("%v_%v", DUMP_ARCHIVE_PREFIX, time.Now().Unix())
--- a/pkg/fission-cli/cmd/support/resources/keda.go
+++ b/pkg/fission-cli/cmd/support/resources/keda.go
@@ -1,0 +1,143 @@
+// SPDX-FileCopyrightText: The Fission Authors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package resources
+
+import (
+	"context"
+	"fmt"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	kedaClient "github.com/kedacore/keda/v2/pkg/generated/clientset/versioned"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/fission/fission/pkg/fission-cli/cmd"
+	"github.com/fission/fission/pkg/fission-cli/console"
+)
+
+const (
+	KedaScaledObject          = "ScaledObject"
+	KedaTriggerAuthentication = "TriggerAuthentication"
+)
+
+// mqtKind mirrors mqtrigger.MqtKind, the Kind the mqt-keda scaler stamps into
+// the OwnerReferences of every KEDA object it creates. Duplicated rather than
+// imported because pkg/mqtrigger would pull the controller-runtime scaler
+// machinery into the CLI binary.
+const mqtKind = "MessageQueueTrigger"
+
+// KedaDumper dumps the KEDA objects that fission's mqt-keda scaler manager
+// creates for MessageQueueTriggers (see pkg/mqtrigger/scalermanager.go).
+//
+// Only objects owned by a MessageQueueTrigger are dumped. KEDA is a
+// cluster-wide component that other workloads use too, and a support bundle
+// should carry fission's own resources, not an unrelated team's ScaledObjects.
+// The scaler sets no labels on what it creates, so ownership is the only
+// selector available — which is why this cannot reuse the label-selector-based
+// KubernetesObjectDumper.
+type KedaDumper struct {
+	client   kedaClient.Interface
+	kedaType string
+	// clientErr defers a client-construction failure to Dump, so one
+	// unavailable client cannot abort the whole support bundle.
+	clientErr error
+}
+
+func NewKedaDumper(client cmd.Client, kedaType string) Resource {
+	kc, err := kedaClient.NewForConfig(client.RestConfig)
+	return KedaDumper{client: kc, kedaType: kedaType, clientErr: err}
+}
+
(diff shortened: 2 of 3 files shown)


sanketsudake:
Reviewed this closely. All three decisions in your description hold up under checking, and I verified each rather than taking them on trust:

- **Ownership is the only available selector.** `scalermanager.go:353` and `:490` both stamp `newOwnerReference(mqt.Name, mqt.UID)`, and the only `Labels:` in that file is on the consumer Deployment — so `KubernetesObjectDumper` genuinely cannot be reused.
- **The masking claim is accurate for the pinned KEDA v2.20.1.** I walked the whole `TriggerAuthenticationSpec`: `HashiCorpVault.Credential.Token` really is the only inline-string credential. `AzureKeyVault.Credentials.ClientSecret`, `GCPSecretManager`, `AwsSecretManager` and `OAuth2.ClientSecret` all go through `ValueFrom` → `SecretKeyRef`, `Env` carries variable names only, and `VaultSecret.Value` is tagged `json:"-"` so the JSON-based `sigs.k8s.io/yaml` drops it.
- **DeepCopy-before-mask is necessary, not defensive noise** — `Credential` is a pointer shared with `items.Items[i]`. `TestKedaDumperLeavesTheListedObjectUnmutated` is the standout test here; shared-pointer mutation is exactly the class that survives review.

I pushed one commit to your branch (`4e7f2833`), tests only, no production change.

### What I changed and why

Two assertions passed regardless of the behaviour they named:

1. **The ownership filter is applied at two call sites, but only the ScaledObject one was covered.** Replacing the `TriggerAuthentication` check with `if false && ...` left the whole suite green. That is the branch most worth covering, since it is what keeps another team's Vault address and secret names out of a bundle sent to a third party. Added an unowned `TriggerAuthentication` alongside the owned one.
2. **`TestKedaDumperUnknownTypeWritesNothing` seeded an empty clientset**, so "wrote nothing" held however the `default` branch behaved — making it dump every ScaledObject still passed. Seeded both kinds so a fallthrough produces a file.

Verified both mutations now fail.

### On your open question

`console.Verbose(2, ...)` is right — it matches the level all 22 other `Verbose` calls in `pkg/fission-cli` use, and a warning in every bundle from every non-KEDA install would be noise. One thing worth considering as a follow-up: at default verbosity nothing is printed **and** nothing lands in the dump directory, so an engineer reading the tarball later cannot tell "no KEDA installed" from "the list failed". A one-line `keda-not-installed.txt` marker in the dump dir would carry the reason in-band.

### Not blocking, worth a follow-up

- **`keda.go:26-30` — the duplication rationale is not correct.** The comment says importing `pkg/mqtrigger` would pull the scaler machinery into the CLI binary; it is already there. `go list -deps ./cmd/fission-cli` contains `github.com/fission/fission/pkg/mqtrigger`, via the blank imports at `cmd/fission-cli/app/app.go:36-37`, and `go list -deps ./pkg/mqtrigger` contains zero `pkg/fission-cli` packages, so there is no cycle either. The constant itself is harmless — the wrong reason is the problem, because the next reader will re-duplicate on it. Either use `mqtrigger.MqtKind`, or replace the rationale with a "keep in sync" note.
- **`keda.go:56-63` — the ownership match ignores `APIVersion`.** Any CRD whose Kind is `MessageQueueTrigger`, in any group, satisfies a filter whose stated purpose is excluding other teams' objects. `newOwnerReference` stamps both fields from `mqtrigger.MqtKind` / `MqtAPIVersion`, so comparing both composes with the point above.
- **`keda.go:70` — the `IsNoMatchError` branch is unreachable through a typed clientset.** `NoKindMatchError` comes from a RESTMapper; the generated KEDA clientset goes straight to REST and gets a 404, which `IsNotFound` handles. Harmless as defensive breadth, but the test comment `// What a cluster without the KEDA CRDs actually returns` sits on the NoMatch case and enshrines the misconception. Worth moving onto the `NewNotFound` case.
- **`keda.go:83`/`:102` — `kedaAbsent`'s wiring is untested.** Inverting the branch, so real errors go quiet and missing CRDs warn, fails nothing. `PrependReactor("list", "scaledobjects", ...)` is the seam.
- **`keda.go:50` — a nil `RestConfig` panics rather than taking the `clientErr` path.** `kedaClient.NewForConfig` opens with `configShallowCopy := *c`. Because `NewKedaDumper` is called inside the `ress` map literal, that panic would abort the whole dump before any dumper runs — exactly what `clientErr` exists to prevent. Not reachable through the shipped CLI today, so latent rather than live, but `cmd.Client{...}` literals without `RestConfig` are common across the CLI test tree.

One note that is **not** on this PR: `ScaledObject.spec.triggers[].metadata` is dumped unmasked, and it is literally `mqt.Spec.Metadata`. That is pre-existing — `fission-crd-mqtriggers` already writes the same map unmasked — so it is a separate issue, not a reason to hold this.

Issue #1873, opened 2020-12-03
Pod describe output in fission support dump


therahulbhati:
**Is your feature request related to a problem? Please describe.**
Sometimes in pod logs and CRDs fail give a complete picture of the problem.

**Describe the solution you'd like**
To have pod describe output in fission support dump.

**Describe alternatives you've considered**

**Additional context**

sanketsudake:
This has been inactive for 1653 days and is being marked stale. It will be closed in a future scrub if there's no further activity. Comment to keep it open.

sanketsudake:
Still open. Related work in flight: #3671 — "support: dump kubernetes events for fission pods".

Note the scope difference so this issue is not closed prematurely. On current `main`, `pkg/fission-cli/cmd/support/resources/kubernetes.go` dumps raw Service/Deployment/Pod/HPA/DaemonSet/Node objects as JSON plus pod logs. #3671 adds the *events* half of `kubectl describe`, which is the most diagnostically useful part. The remaining gap after it merges is the rendered describe-style narrative (conditions, container statuses, volume/tolerations summary) rather than raw JSON.

PR #3671, opened 2026-08-06
support: dump kubernetes events for fission pods

sushantlokhande14:
Addresses #1873.

## What

The issue asks for `kubectl describe pod` output in the support dump, because "pod logs and CRDs fail to give a complete picture."

I want to be upfront that this delivers the **information** describe carries rather than describe's text format, and say why — happy to change course if you'd rather have the literal thing.

The dump already writes the whole `Pod` object, `.status` included, via `NewKubernetesObjectDumper(..., KubernetesPod, ...)`. Almost everything `describe` prints is derived from that object, which is already in the bundle. The one section it adds that is genuinely missing is **Events** — `FailedScheduling`, `ImagePullBackOff`, `Unhealthy`, `OOMKilled` — and those live on `Event` objects, not on the pod. That is the part that explains a pod which never became ready.

Reproducing describe's exact rendering would mean taking a dependency on `k8s.io/kubectl` (for `pkg/describe`), which the CLI does not currently have. That felt like a lot of dependency for text formatting when the underlying data is what's diagnostic — so this adds `KubernetesPodEventDumper` instead, registered for all three existing pod groups:

- `fission-components-pod-events`
- `fission-builder-pod-events`
- `fission-function-pod-events`

using the same label selectors as the matching `-pod-spec` / `-pod-log` dumpers.

## Implementation notes

**One event list, not one query per pod.** Events are fetched with a single cluster-wide `List` and indexed by `involvedObject.UID`, rather than a per-pod field-selector query. A busy install has many pods and many short-lived events, and a request per pod would hit an apiserver that may already be struggling — which is usually *why* someone is collecting a support bundle.

**No empty files.** A pod with no events (or whose events have aged out of etcd) is skipped, so the bundle isn't padded with empty files.

**Oldest first.** Events come back in no guaranteed order; they're sorted by `LastTimestamp` so each file reads as that pod's history.

**Kind is checked, not just UID.** Only events whose `involvedObject.Kind` is `Pod` are attributed to a pod.

## Tests

`podevents_test.go`, 3 tests, all passing:

- Selector-matching pod gets a file containing its events; a non-matching pod's events are excluded; ordering is oldest-first.
- A pod with no events produces no file.
- An event whose `involvedObject` is a `Deployment` is not attributed to a pod, even when the UID collides.

## Verification

```
go build ./...                              clean
go vet ./pkg/fission-cli/cmd/support/...    clean
go test ./pkg/fission-cli/cmd/support/...   ok  (3 tests)
gofmt                                       clean
```

No new dependencies — `go.mod`/`go.sum` untouched. No API, CRD, or codegen change.

## Open question

If you do want literal `describe` output, the options are (a) depend on `k8s.io/kubectl/pkg/describe`, or (b) hand-render the describe layout in fission. I didn't want to make that call unilaterally, since (a) is a meaningful dependency and (b) is code to maintain that drifts from upstream. Tell me which you'd prefer and I'll follow up.

Note this touches `dump.go`'s resource map, as does #3670 (KEDA objects, issue #1914) — whichever merges second will need a trivial rebase.

### DIFF
--- a/pkg/fission-cli/cmd/support/dump.go
+++ b/pkg/fission-cli/cmd/support/dump.go
@@ -76,18 +76,22 @@ func (opts *DumpSubCommand) do(input cli.Input) error {
 			"svc in (buildermgr, executor, kubewatcher, mqtrigger, router, storagesvc, timer)"),
 		"fission-components-pod-log": resources.NewKubernetesPodLogDumper(k8sClient,
 			"svc in (buildermgr, executor, kubewatcher, mqtrigger, router, storagesvc, timer)"),
+		"fission-components-pod-events": resources.NewKubernetesPodEventDumper(k8sClient,
+			"svc in (buildermgr, executor, kubewatcher, mqtrigger, router, storagesvc, timer)"),
 
 		// fission builder logs & spec
 		"fission-builder-svc-spec":        resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesService, "owner=buildermgr"),
 		"fission-builder-deployment-spec": resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesDeployment, "owner=buildermgr"),
 		"fission-builder-pod-spec":        resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesPod, "owner=buildermgr"),
 		"fission-builder-pod-log":         resources.NewKubernetesPodLogDumper(k8sClient, "owner=buildermgr"),
+		"fission-builder-pod-events":      resources.NewKubernetesPodEventDumper(k8sClient, "owner=buildermgr"),
 
 		// fission function logs & spec
 		"fission-function-svc-spec":        resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesService, "executorType=newdeploy"),
 		"fission-function-deployment-spec": resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesDeployment, "executorType in (poolmgr, newdeploy)"),
 		"fission-function-pod-spec":        resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesPod, "executorType in (poolmgr, newdeploy)"),
 		"fission-function-pod-log":         resources.NewKubernetesPodLogDumper(k8sClient, "executorType in (poolmgr, newdeploy)"),
+		"fission-function-pod-events":      resources.NewKubernetesPodEventDumper(k8sClient, "executorType in (poolmgr, newdeploy)"),
 
 		// CRD resources
 		"fission-crd-packages":      resources.NewCrdDumper(opts.Client(), resources.CrdPackage),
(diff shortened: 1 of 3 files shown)


sanketsudake:
Thanks — the reasoning here is right, and I agree events are the part that was genuinely missing. I reviewed it and pushed one commit to your branch (`7ef21754`) rather than round-tripping, since it is small and I wanted CI to run on it.

### What I changed and why

`sort.Slice` on `LastTimestamp` alone mis-orders the events this dumper exists to capture.

An event written through `events.k8s.io/v1` carries only `EventTime`. Read back through the core/v1 endpoint its `LastTimestamp` is zero, so it sorts ahead of every event that has one. `k8s.io/client-go/tools/events/event_recorder.go` sets `EventTime` and never the deprecated fields, while `tools/record/event.go` sets `First`/`LastTimestamp`. kube-scheduler has used the newer recorder since 1.19 and kubelet still uses the legacy one — so `FailedScheduling`, the first example in your own doc comment, was written first regardless of when it actually happened. In a file of only such events, `Before` is false for every pair, `sort.Slice` becomes a no-op, and the output keeps whatever order the apiserver returned.

The fix sorts on a coalesced effective time (`Series.LastObservedTime` → `LastTimestamp` → `EventTime` → `FirstTimestamp`), the same fallback `kubectl` uses, and switches to `SliceStable`.

The ordering assertion could not have caught this, for two reasons worth knowing:

1. Every event `podEvent` built was legacy, so the events.k8s.io case was never constructed.
2. The fake clientset returns list results **name-sorted**, and `router-abc.FailedScheduling` < `router-abc.Pulled` already matched the asserted order — so deleting `sort.Slice` outright left the test green.

So the test now names reasons such that alphabetical order is the reverse of chronological, and puts an `EventTime`-only event in the middle. I verified both mutations fail against it.

### On your open question

Events instead of literal `describe` is the right call, and your reasoning for it is the reason — everything else `describe` prints is derived from the Pod object the `-pod-spec` dumpers already write. No `k8s.io/kubectl` dependency, please.

### Not blocking, worth a follow-up

- **Three concurrent cluster-wide Event LISTs.** The dumper is registered three times and `dump.go` runs each in its own goroutine, so one `fission support dump` fires three unfiltered full-cluster LISTs and holds three copies. That inverts the rationale in the comment at lines 52-56. Cheapest improvement is `FieldSelector: "involvedObject.kind=Pod"` — keep the client-side `Kind` check, because `k8sfake` ignores field selectors and only applies label selectors. The real fix is to build the index once in `dump.go` and share it across the three registrations. If you add `Limit`, add the `Continue` loop with it — `Limit` alone truncates silently.
- **UID matching is unguarded.** Keying the index on `InvolvedObject.Name` instead of UID passes every test. UID is the correct choice precisely because pool and function pods get recreated under recycled names; a test with two same-name different-UID events would pin it.
- **Events for pods that no longer exist are dropped.** Iteration is pod-first, so an OOMKilled or evicted pod contributes no file — often the most diagnostic case in a bundle. Either collect unmatched Pod-kind events, or note the limitation in the type comment.

Nice work — the empty-file skip, the `involvedObject.Kind` check, and the single-list-over-N-queries instinct are all the right calls.

sanketsudake (approved):

PR #3675, opened 2026-08-09
support: close out the dump follow-ups, and two gaps they uncovered

sanketsudake:
Closes out every follow-up left by the reviews of #3670 and #3671, plus two gaps those reviews uncovered in the existing dump.
All changes are confined to `pkg/fission-cli/cmd/support/`.
No API, CRD, codegen or runtime change; `go.mod` untouched.

## Two things that were already broken

**Eight components were missing from every support bundle.**
The component label selector was written out inline at each of the five places that needed it, and it drifted.
It listed 7 `svc` values while the chart ships 15, so `canaryconfig`, `mcp`, `mqtrigger-keda`, `statestore`, `statesvc`, `tenantcontroller`, `webhook-service` and `workflow` had no spec, no pods and no logs — silently, because a selector that matches nothing looks exactly like a component that is not deployed.
`mqt-keda` being invisible is the sharpest of these: it is the component behind every keda MessageQueueTrigger.

The function selectors had the same problem from the other side.
`executorType in (poolmgr, newdeploy)` omits `container`, so container-executor functions were absent too, and the service selector `executorType=newdeploy` missed them as well — that executor does create a Service per function.

All four selectors are now named constants with one place to change, and a test walks the chart's deployment templates and fails if any shipped `svc` value is unmatched, naming the file.

**Function HPAs were never collected.**
`KubernetesObjectDumper` has handled `KubernetesHPA` since it was written, but nothing ever registered it.
The HPA is what decides whether a function scales, which makes it the first object to look at when one does not.

## The deferred items

**Events for pods that no longer exist.**
The pod event dumpers iterate pods first, so an event reached the bundle only if its pod was still listed — and the pod that was OOMKilled, evicted or replaced by a rolling update is exactly the one being investigated.

The naive definition of orphan (any Pod event whose UID matches no live pod) is not acceptable here.
Fission functions commonly share a namespace with a user's own workloads and a bundle goes to a third party, so that would sweep in unrelated teams' deleted pods — the same leak the KEDA ownership filter exists to prevent.
Scope is by derivation instead: every pod a Deployment owns is named `<deployment>-<rs hash>-<random>`, which is checkable without the pod existing, so an orphan counts only when its name and namespace match a fission-owned Deployment.

Knowingly out of scope, and recorded on the type: pods from the chart's Helm hook Jobs, which are Job-owned rather than Deployment-owned.
Fission creates no bare Pods or StatefulSets at runtime, so that is the whole exception list.

**The workload a keda mqt scales.**
The scaler creates a Deployment per trigger alongside the ScaledObject and no selector matched it, so the bundle carried the scaling policy but not the thing being scaled.

This cannot be a label-selector dumper and the scaler cannot simply be given a label: `Spec.Selector` is immutable on an existing Deployment, and putting one on the pod template would change the template hash and force a rollout of every consumer on upgrade.
It selects on the MQT owner reference instead, the same signal `KedaDumper` uses — no runtime change.

Pods need a second filter beyond the Deployment's own selector.
`app` is the most generic label key in kubernetes and `<mqt name>` is user-chosen, so an unrelated pod sharing it would otherwise have its spec and logs written into the bundle; the `<deployment>-` name prefix is required too.

**Credentials in trigger metadata.**
A MessageQueueTrigger's `Spec.Metadata` is copied verbatim into `ScaledObject.spec.triggers[].metadata`, and for several KEDA scalers that map is the documented place for an inline connection string (`amqp://user:pass@host`).
Fission's own path routes credentials through `Spec.Secret`, so this does not fire on a stock setup — but nothing stops an operator putting a password there.

Masked when the key names a credential or the value is a URL carrying a password in its userinfo.
Keys are always preserved: which settings were set is diagnostic, their values are not.
Both call sites are fixed — masking only the ScaledObject would leave the same values in the bundle via `fission-crd-mqtriggers`, which dumps the source map.

Two deliberate choices in the key list: `sasl` alone is **not** a credential marker, because KEDA's Kafka scaler uses `sasl`/`saslType` for the mechanism name, so masking those would hide diagnostics and no secret; and an empty value is left alone, so "set but blank" stays distinguishable from "masked".

**A marker when KEDA is not installed.**
The absent-CRD branch returned after a `console.Verbose(2, ...)` that prints nothing at default verbosity, so "no KEDA" and "the list failed" both produced an empty directory.
A `keda-not-installed.txt` now names the kind and the error, on that branch only — a real failure is warned about on the console and deliberately leaves no marker.
This also closed a test gap: the two branches previously differed only in console output, which this package cannot capture without swapping a global that `t.Parallel` would race.

**Event list pagination.**
`Limit` alone would be worse than the unbounded read it replaces — it truncates to the first page silently, so a short file reads as "no more events".
The loop feeds the continue token back, refuses to spin on a repeated token, and keeps the pages already collected when a watch cache ages out mid-pagination (410 Gone), logging that the history is partial.

## Correctness fixes carried along

- **Aliasing.** `ScaledObject.Spec.Triggers` is a slice, so masking through it would edit the backing array shared with the List result — that path DeepCopies, and only when something needs masking. In `crd.go` the range variable is a value copy and the masker returns a fresh map, so no copy is needed there; the code and comments reflect the difference.
- **`append(Containers, InitContainers...)`** in the log dumper writes into `Containers`' spare capacity when it has any, editing the pod object it was handed. Now `slices.Concat`.
- The event and deployment lists are each fetched **once** and shared, rather than once per dumper.

## Tests

Every new test was mutation-verified: the behaviour was removed and the test confirmed to fail, then restored. Twenty mutations, twenty caught.

The ones worth knowing about, because the obvious test would not have worked:

| What | Why the obvious test fails |
|---|---|
| Component selector coverage | Walks the chart templates; a hand-written list would drift the same way the selector did |
| Event pagination | `k8sfake`'s tracker ignores `Limit`/`Continue`, so seeding N events and counting requests passes against an unpaged implementation — the tests read the continue token off the concrete `ListActionImpl` |
| Server-side field filtering | `k8sfake` applies label selectors but ignores field selectors, so this inspects the `ListAction`'s restrictions instead |
| Shared index once-semantics | Runs its dumpers concurrently as `dump.go` does, so under `-race` it also guards the read-only-slices claim |
| Orphan privacy guard | An unrelated deleted pod **in the same namespace** must not be collected |
| Consumer pod attribution | An unrelated pod carrying the same `app` label must not be collected |

## Verification

```
go build ./...                                   clean
gofmt -l pkg/fission-cli/                         clean
golangci-lint run pkg/fission-cli/...             0 issues
make license-check                                clean
make verify-gomod                                 OK
go test -race ./pkg/fission-cli/cmd/support/...   ok
```

### DIFF
--- a/pkg/fission-cli/cmd/support/dump.go
+++ b/pkg/fission-cli/cmd/support/dump.go
@@ -32,7 +32,7 @@ const (
 	// with the svc: labels in charts/fission-all/templates/*/deployment.yaml —
 	// a component absent here has no spec and no logs in the bundle at all.
 	componentSelector = "svc in (buildermgr, canaryconfig, executor, kubewatcher, mcp, mqtrigger, " +
-		"mqtrigger-keda, router, statestore, statesvc, storagesvc, tenantcontroller, timer, " +
+		"mqtrigger-keda, router, router-internal, statestore, statesvc, storagesvc, tenantcontroller, timer, " +
 		"webhook-service, workflow)"
 
 	builderSelector = "owner=buildermgr"
--- a/pkg/fission-cli/cmd/support/dump.go
+++ b/pkg/fission-cli/cmd/support/dump.go
@@ -125,6 +125,9 @@ func (opts *DumpSubCommand) do(input cli.Input) error {
 		"fission-function-pod-spec":        resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesPod, functionSelector),
 		"fission-function-pod-log":         resources.NewKubernetesPodLogDumper(k8sClient, functionSelector),
 		"fission-function-pod-events":      resources.NewKubernetesPodEventDumper(k8sClient, functionSelector, podEvents),
+		// The HPA is what decides whether a function scales at all, and the
+		// dumper for it existed but was never registered.
+		"fission-function-hpa": resources.NewKubernetesObjectDumper(k8sClient, resources.KubernetesHPA, functionSelector),
 
 		// Events of fission pods that no longer exist — the OOMKilled or evicted
 		// pod is usually the one being investigated.
--- a/pkg/fission-cli/cmd/support/dump.go
+++ b/pkg/fission-cli/cmd/support/dump.go
@@ -57,6 +57,11 @@ func (opts *DumpSubCommand) do(input cli.Input) error {
 
 	k8sClient := opts.Client().KubernetesClient
 
+	// Every pod-event dumper shares one index. The dumpers below run
+	// concurrently, so a list each would mean three full-cluster Event LISTs in
+	// flight at once.
+	podEvents := resources.NewPodEventIndex(k8sClient)
+
 	ress := map[string]resources.Resource{
 		// kubernetes info
 		"kubernetes-version": resources.NewKubernetesVersion(k8sClient),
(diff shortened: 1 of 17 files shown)


sanketsudake:
Ran the full review battery on this PR — three independent adversarial lenses plus my own verification. It found a **critical defect I introduced in this very PR**, so it was worth doing. Six commits pushed (`45af1233..f137b134`).

## The critical one

Collecting the mqt consumer workload reopened the leak this PR had just closed.

The scaler writes every `mqt.Spec.Metadata` value as a **literal env var** on the consumer Deployment (`scalermanager.go:145-151`). So after masking that map on the MessageQueueTrigger and on the ScaledObject, the new dumper wrote it out a third way: `metadata.password: hunter2` landed in the bundle as `PASSWORD: hunter2`. The Pod carries the resolved env too, so both modes leaked — and this was *new* surface, because the consumer workload was not collected at all before.

Container env is now masked by the same rule, deep-copied first since both objects come from the shared index. `ValueFrom` entries carry no value and are untouched.

## Five more, all confirmed rather than taken on trust

**Orphan attribution collected other workloads' pods.** A bare name-prefix test let a function named `api` vouch for a stranger's `api-gateway-<hash>-<rand>` in the same namespace. Deployment names here are user-chosen, so this is plausible, not theoretical — I reproduced it against the PR's own test helper. Requiring the ReplicaSet suffix shape rejects it without ever rejecting a real pod.

**A short pod list made live pods look deleted.** `listAllPages` returned partial results with a nil error, so a truncated pod list meant running pods were written into a directory whose entire meaning is "this pod is gone". Inventing a failure is worse for a diagnostic bundle than missing one. `partial` is now separate from `err`, and each caller decides — a short event list is usable and noted, a short pod list stops the orphan dumper outright.

**Paging stopped on an empty page.** The API contract allows a short or empty page with a live continue token, and the event walk uses a field selector, which is where that happens. The repeated-token check already covers the non-advancing server that clause was for.

**One failing container abandoned the pod's remaining logs.** Pre-existing, but this PR gave it a second caller. The usual failure is a container still `PodInitializing`, and regular containers stream before init containers — so the usual outcome was a stuck pod with *no* log files, indistinguishable from a pod that logged nothing. Also fixes a leaked stream on every error path, a double `Close` on the success path, and a discarded buffer on mid-stream read errors.

**Two masking gaps, both directions.** `url.Parse` reads a schemeless `user:pass@host` as scheme `user`, so it was not masked; it also fails outright on an unencoded character in a password. Conversely KEDA's `passwordFromEnv` holds an env var *name*, and masking it destroyed the main diagnostic for a trigger that will not authenticate.

**`svc/router-internal` was missing** from the component selector — and my guard test structurally could not catch it, because it globbed only `*/deployment.yaml` and that component is a Service with no Deployment. It now walks every template with a floor on labels found, so a glob that stops matching fails loudly instead of passing vacuously.

## A correction to my own claim

I said "20 mutations, 20 caught". That was true but self-selected — the mutations came from paths the tests already covered. An independent run was **51 mutations, 25 survived**. The tests that existed did discriminate; coverage had real holes.

Those are now closed: the log dumper had no tests at all, `DeploymentIndex` had no error path, the Deployment/Pod paging wiring was unpinned, and the multi-selector, unknown-mode and `eventTime` Series branches were untested. 13 new mutations, all caught — with a corrected harness, after my first run gave false passes by globbing two packages and reading the wrong result line.

## Verification

```
go build ./...                                   clean
gofmt -l pkg/fission-cli/                         clean
golangci-lint run pkg/fission-cli/...             0 issues
make license-check / verify-gomod                 clean / OK
go test -race ./pkg/fission-cli/cmd/support/...   ok
```

PR #3696, opened 2026-08-22
[CI only] json/v2 migration full-battery run

sanketsudake:
Temporary draft to run the full CI battery (integration legs, upgrade tests,
pinned-legacy-data-plane) against the json/v2 migration before #3694 lands —
the stacked #3695 only triggers the lightweight checks.
Will be closed once CI reports; the real review target is #3695.


sanketsudake:
Full battery ran twice: round 1 flagged two leg failures root-caused as environmental (etcd timeout, go-builder slowness; workflow test passed 4/4 locally against the migrated build), round 2 on the final tree is fully green except the known repo-wide CodeQL default-setup conflict. Purpose served; review target is #3695 (stacked on #3694).

PR #3699, opened 2026-08-24
Agent runtime slice 1: FunctionSpec.Agent config block and CLI surface

sanketsudake:
This PR adds the CRD and CLI surface for declaring a function as an agent (agent-runtime slice 1).

`FunctionSpec` gains an optional `Agent *AgentConfig` block —
presence is the on switch, like `Streaming`/`Tool`/`State` —
declaring session identity (`session.source`/`session.name`, defaulting to the `X-Fission-Session` header),
idle/archive lifecycle windows (`idleAfter`, `archiveAfter`),
and a live-session cap (`maxSessions`).

Admission validation runs on the webhook path (`validateForAdmission`),
so kubectl/GitOps writers are bound, not just the CLI:
source enum, HTTP-token session-name charset (max 128),
non-negative durations, `archiveAfter >= idleAfter`, and non-negative `maxSessions`.
Deepcopy, applyconfigurations, and both Function/FunctionVersion CRD schemas are regenerated.

The CLI grows six `--agent*` flags on `fn create`/`fn update`
with the same toggle semantics as `--state`
(explicit `--agent=false` clears the block; other `--agent-*` flags merge onto the existing config),
and a new `fission agent list` command that lists agent-declared functions
with their session source and lifecycle settings.

Everything is additive and backward compatible —
a nil `agent` block means exactly today's behavior.
The `pkg/agentruntime` subsystem that consumes this config
(dispatcher, session store, sweeper) follows in a separate PR built on this branch.


sanketsudake:
Superseded by #3701, which consolidates agent-runtime slices 1–4 into a single PR against `main` (this PR's commits are included in that branch's history). Closing in favor of #3701.

PR #3700, opened 2026-08-24
Agent runtime slice 2: pkg/agentruntime session dispatcher, registry, and sweeper

sanketsudake:
This PR ships slice 2 of the distributed agent runtime:
the `pkg/agentruntime` subsystem (`fission-bundle --agentPort`, port 8894).
Builds on #3699 (`FunctionSpec.Agent`).

It dispatches session-scoped turns at `POST /agents/{namespace}/{name}`
to agent-enabled Functions over the router internal listener,
with the same ServiceRouterInternal HMAC signing the other publishers use,
streaming responses with per-write flush.

Session records are statestore KV entries, never CRDs:
create-only CAS writes with an atomic `SetCounted` MaxSessions quota,
a sibling archived keyspace so retention never consumes quota,
and orphan self-expiry TTLs.
A CAS-idempotent sweeper ages sessions active → idle → archived on every replica
with no leader election —
a lost CAS means another replica already applied the transition.

AuthZ mirrors the MCP server's fail-closed JWT contract
(`allowed_namespaces`, 403 on out-of-scope namespaces;
chart render fails unless authentication is enabled or `agentRuntime.allowInsecure` is explicit).
The Helm chart adds the deployment/service/SA/RBAC
plus both NetworkPolicy allowlists (router-internal 8889 and statestore 8891),
and an integration test covers mint-and-reuse session dispatch end-to-end,
gated in CI by an explicit `deploy/agentruntime` rollout wait
so a profile regression can never turn the coverage into a green skip.

Per-session meters (turns, errors, active-millis, token/cost placeholders)
are recorded in the session stats block from day one;
metrics export is deliberately deferred to the metering slice,
matching how the MCP subsystem shipped.

In v1 only Content-Type and the session headers cross the dispatcher in either direction
(plus `X-Fission-Agent-Yield` on the way back) —
custom caller headers are intentionally not proxied yet.

### DIFF
--- a/.github/workflows/push_pr.yaml
+++ b/.github/workflows/push_pr.yaml
@@ -196,6 +196,13 @@ jobs:
           # would silently turn all MCP coverage into a green skip — the
           # loop above only waits for deployments that exist.
           kubectl -n fission rollout status deploy/mcp --timeout=3m
+          # deploy/agentruntime must exist and be ready (kind-ci enables
+          # agentRuntime.enabled): the agent-runtime integration test skips
+          # itself when the endpoint is unreachable, so without this
+          # explicit gate a profile regression would silently turn all
+          # agent-runtime coverage into a green skip — the loop above only
+          # waits for deployments that exist.
+          kubectl -n fission rollout status deploy/agentruntime --timeout=3m
 
       # Pin the v1.34 leg back to the legacy data plane (RFC-0002 gates off):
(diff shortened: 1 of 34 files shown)


sanketsudake:
Superseded by #3701, which consolidates agent-runtime slices 1–4 into a single PR against `main` (this PR's commits are included in that branch's history). Closing in favor of #3701.

PR #3701, opened 2026-08-25
Agent runtime for Fission (slices 1–4): sessions, continuation, introspection, boardroom demo

sanketsudake:
## Agent runtime for Fission (slices 1–4)

Adds a first-class **agent runtime** to Fission: session-scoped, durable, self-continuing function invocations with a live introspection surface — the machinery for running managed AI agents on the existing serverless data plane, no new stateful infrastructure.

This is a consolidated PR of four slices (developed as a stack; **supersedes #3699 and #3700**, which can be closed).

### Slice 1 — `FunctionSpec.Agent` config block + CLI
An optional-pointer `Agent *AgentConfig` on FunctionSpec (presence-is-on-switch, like `Tool`/`State`/`Streaming`): session identity, idle/archive lifecycle, and a live-session cap. Admission-validated on the webhook path; deepcopy/applyconfig/CRDs regenerated; classified runtime-affecting for RFC-0025 auto-publish. CLI: six `--agent*` flags on `fn create/update` (State-parity toggle semantics) and a `fission agent list`.

### Slice 2 — `pkg/agentruntime` subsystem (`--agentPort`, 8894)
`POST /agents/{namespace}/{name}` dispatches a session turn to the agent-enabled Function over the router internal listener, HMAC-signed like every other publisher, streaming the response with per-write flush. Session records live in the statestore (KV, CAS on version, `SetCounted` quota, a sibling archived keyspace so retention never consumes quota, orphan self-expiry TTLs) — **never CRDs**. A CAS-idempotent sweeper ages sessions active→idle→archived on every replica with **no leader election**. Fail-closed JWT authz mirroring the MCP server; Helm chart with both NetworkPolicy allowlists and render gates.

### Slice 3 — `yield: continue` self-continuation
A turn returning `X-Fission-Agent-Yield: continue` re-invokes the same session via an `agent-wake` statestore queue (at-least-once, dedup-while-unsettled), consumed with the asyncinvoke lease→classify→backoff→ack/nack/kill discipline: precise DLQ reasons, delivery timeout bounded below the lease, would-expire pre-check, and a per-session continuation cap (`AGENT_MAX_CONTINUATIONS`, default 1000).

### Slice 4 — introspection, SSE feed, pool view, CLI, boardroom demo
Read-only observability: `GET /registry/agents`, `.../sessions[/{id}]`, and `/registry/pool` (Pod + EndpointSlice informers, RBAC-preflighted so a missing grant degrades to 503 instead of wedging dispatch); sessions gain a best-effort `currentPod` anchor via a pure `endpointcache.PredictSticky` (rendezvous ranking, zero admission side effects, UI truth only). A poll-diff SSE feed (`GET /registry/events`, single poller, slow-consumer drop, browser `?token=` transport). A `fission agent sessions` CLI, an embedded single-file boardroom UI (`/ui`), and `demo/agent-boardroom/` (fixtures, specs, stdlib loadgen).

### Review
Every task went through per-task review with fix loops, each slice a whole-branch review, then a **6-lens review battery** over the full diff (silent-failure, security, correctness, type-design, test-coverage, conventions). The battery found and this PR fixes: a wake continue-chain fork/amplification bug, `MaxSessions` quota failing open on store degradation, an idle-longer-than-archive lifecycle bug, unbounded session creation, a poller doing statestore work with zero subscribers, statestore-close vs server-drain ordering, and an observability/RBAC-hardening cluster — plus the battery's high-value tests (claim-decode fuzz, dispatch-route auth composition, store failure-injection, chart auth gates). Security review found no cross-namespace leak and a coherent authz mesh.

### Deferred follow-ups (tracked, not blocking)
Short-TTL single-use SSE ticket exchange (vs reusing the dispatch JWT in the URL); per-session-metrics OTel counters; SSE write-watchdog + subscriber ceiling; `agentruntime` role-dataplane analog for `functionNamespace` installs and dynamic-tenancy pool-cache staleness; a shared registry-route mount helper; `maxContinuations` per-chain-vs-lifetime semantics. The `demo/` directory location (in-tree vs the `../examples` sibling) is a maintainer decision for this PR.

Everything is additive and backward compatible — a nil `agent` block is exactly today's behavior. The subsystem is off by default (`agentRuntime.enabled: false`).


sanketsudake:

PR #3703, opened 2026-08-25
agentruntime: session history + checkpoints (G13 compaction-as-projection, phase 1)

sanketsudake:
Stacked on #3701 (base: feat/agentruntime-introspection). Implements the G13 fact-log + checkpoint + projection contract for agent sessions.

## What

- **statesvc**: scoped EventLog routes for function pods — `POST /v1/eventlog/append|read|head` under the existing bearer/keyspace auth. CAS-only (`expectedSeq >= 0`; AppendAny unreachable); CAS conflict → 412 with the current `head` in the error envelope for client resync. Caps: ≤64 events/append, per-event payload ≤ keyspace `MaxValueBytes`, event `type` ≤128 bytes. statesvc now requires the EventLog capability at startup (all shipped drivers provide it).
- **agentruntime**: `HistoryStore` reading the same streams/checkpoints via direct store handles; two authed registry endpoints — `GET .../sessions/{id}/history` (head + events; 404 unknown session; graceful `{"head":0,"events":[]}` for agents without `State`) and `GET .../sessions/{id}/checkpoint`.
- **Contract**: history stream `agentlog/<sessionID>` and checkpoint KV key `agentckpt.<sessionID>` (reserved prefix, session ids may not start with it) live in the **function-state scope** — the pod's existing keyspace token is the credential; checkpoints ride the shipped KV CAS route, zero new surface. Internal stream naming single-sourced in `stateapi.StreamName`.
- **Turn tags**: events carry the persisted turn count from `X-Fission-Agent-Turns` (set pre-increment; `Stats.Turns` settles post-response), so all redeliveries of one logical turn tag identically — CAS append conflicts are the duplicate-attempt detector (step-level at-least-once replay).
- **Sweeper owns all trims**: archive-time cleanup (head captured before the archive CAS; trim bounded to the pre-archive head; checkpoint CAS-deleted at its pre-archive version so a re-created session's fresh checkpoint survives) and the new `AgentConfig.historyTrimBelowCheckpoint` knob (admission-rejected without `State`; CLI `--agent-history-trim`) trimming live sessions to the committed coverage boundary, clamped to head.
- **Demo + UI**: support-desk fixture writes real per-turn events + a checkpoint every 5th turn with 412-resync replay; boardroom UI gains a click-to-open session transcript panel (turn-grouped events, coverage marker, tail-windowed fetch).
- **Integration**: `TestAgentRuntimeHistory` covers the composed wire path (pod → statesvc → store → registry) plus the no-State negative path.

## Notes

- Growth story: no stream quota in v1 — the body/value caps bound each append; a never-idle session with the knob off grows until archive (documented; stream quota deferred).
- Trim failures after a committed archive leave orphan residue by design (logged; same family as the deferred RFC-0027 orphan-stream age sweep).
- Archived-retention of history is deferred (trim-at-archive in v1; needs statestore stream TTLs).
- Field-level spill for oversized checkpoint payloads deferred with the caps.

Process: 7 planned tasks + plan review, per-task reviews (all PASS), whole-branch final review (SHIP, 0 blocker/major), one fix round + scoped re-review.


sanketsudake:

PR #3704, opened 2026-08-25
agentruntime: session parentage + graph-over-sessions (slice 5)

sanketsudake:
Stacked on the G13 branch (#3703 ← #3701). Implements qualified session parentage and the multi-agent graph substrate: Workflow = the schedule, sessions-with-parent = the operators, statestore CAS admission = the exactly-once identity.

## What

- **Record**: sessions gain an immutable qualified `parent {namespace, agent, id}` + `depth` (set only at creation; SSE/sweeper/diff untouched by construction — immutable fields can't be why a record differs).
- **Dispatcher**: `X-Fission-Agent-Parent` applies only when a turn creates a session. Depth is **never wire-supplied** — it derives from the claimed parent's own persisted record (live keyspace, then archived), capped by `AGENT_MAX_SPAWN_DEPTH` (default 3; explicit 0 = no spawning; 400 past the cap, rejected before any record/quota is minted). A parent unreadable in both keyspaces degrades to root admission (lineage is descriptive, not referential — no dangling-ref policing, no archive cascade).
- **Security (final-review finding, fixed)**: parentage is **same-namespace-only**. The dispatcher resolves depth with its privileged store handle, so honoring a foreign-namespace ref would have made root-degrade-vs-inheritance a cross-tenant session-existence + depth oracle. A foreign ref is now byte-indistinguishable from an unparseable one (guard before any store read; pinned by a pairwise-equality test). Cross-agent parentage within a namespace — the Handoff/Magentic shape — is unaffected.
- **Exactly-once spawn**: `SpawnSessionID(parentRef, stepKey)` = `"c-" + hex(sha256(ref+"/"+stepKey))[:32]` — same parent + stepKey ⇒ same id, so the shipped `Create(IfVersion=0)` is the admission row (`ErrSessionExists` = already admitted = resume). Byte-identical JS twin in the fixtures; the integration test computes the id in Go and fetches exactly that session spawned by the node fixture — cross-language agreement is asserted, not assumed.
- **Surfaces**: `fission agent sessions list --tree` (paginates to exhaustion, explicit `(unknown: <ref>)` nodes, deterministic order); boardroom family tree (children indent under parents, depth badges) + a real bug fixed en route: the polling fallback fetched one page and deleted absent sessions — past 500 sessions it deleted live cards; it now pages to exhaustion with a ceiling, and never runs the delete pass on a partial set. Cycle rendering hardened in both renderers (printed-exactly-once).
- **Demo**: `architect` fixture spawns 3 expert child sessions per turn (best-effort, never fails its own turn); README swarm-beat walkthrough.
- **Integration**: spawn parentage end to end — cross-language id agreement, idempotent re-spawn, root-degrade, live depth-cap 400.

## Notes

- Wake-recreate loses lineage by design (wake carries no parent; recreated session is a root) — the RFC's wake-envelope open item closes as "no". Transient parent-read failures fragment lineage into fresh capped chains rather than extending one; benign for the exhaustion case the cap guards.
- Handler-driven spawn presents the dispatch credential: works under `AGENT_ALLOW_INSECURE` (dev/demo); G16 (per-agent identity) is the gate for secured installs.
- CI note: workflows are main-base-scoped, so this stacked PR gets meta-checks only; the full battery runs when the stack retargets main.

Process: reviewed plan (APPROVE + 3 folded warnings), 5 SDD tasks all PASS/PASS, whole-branch final review (FIX-FIRST: 1 major security + 3 minors), fix round + CLEAN scoped re-review.

### DIFF
--- a/demo/agent-boardroom/README.md
+++ b/demo/agent-boardroom/README.md
@@ -26,10 +26,19 @@ sub-second resume, live UI.**
   a real agent would call it mid-conversation,
   but the load generator below drives `support-desk` directly,
   so this stays decoupled from the density/teleport story.
+- `fixtures/architect.js` —
+  the swarm-beat agent function.
+  On every turn it spawns 3 deterministically-derived "expert" child sessions on `support-desk`
+  (see `pkg/agentruntime/session.go`'s `SpawnSessionID` / `ParentRef`),
+  carrying `X-Fission-Agent-Parent` on each spawn so the platform records real parentage —
+  the boardroom UI's family tree (below) renders the result.
+  Workers ARE `support-desk` sessions;
+  no separate worker function exists.
 - `specs/` —
   `fission spec`-format YAML: one shared `node` Environment plus a Package + Function per fixture.
   `support-desk`'s Function carries `spec.agent` (session lifecycle) and `spec.state` (keyed-state keyspace, with `sticky` routing on the session header so the router's real request routing lands on the same pod the dispatcher's `CurrentPod` prediction shows in the UI);
-  `order-lookup`'s Function carries `spec.tool`.
+  `order-lookup`'s Function carries `spec.tool`;
+  `architect`'s Function carries `spec.agent` (its own session lifecycle) and a minimal `spec.state` (used only so the fetcher writes the state token file `architect.js` reads its own namespace from).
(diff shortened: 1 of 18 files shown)


sanketsudake:

PR #3705, opened 2026-08-25
agentruntime: per-agent identity — HKDF bearer for spawn auth (G16 phase 1)

sanketsudake:
Stack position 5 (#3701 ← #3703 ← #3704 ← this). Gives agent function pods a real per-agent credential so handler-driven spawns authenticate on secured installs — the G16 gate on spawning is lifted.

## Design (amended from the gap sketch: HKDF, not JWT)

No `allowed_namespaces`-JWT minting infrastructure exists in-repo (those tokens are operator-provisioned), so the credential mirrors the shipped statesvc bearer end to end — no minting service, no key distribution, stateless verification, rotation via the existing active/old masters:

- **Derivation**: `DeriveAgentIdentityKey(master, ns, agent)` — new `agentident` link in the HKDF chain, collision-free by tag.
- **Injection**: the fetcher writes `.fission-agent-token` (`{namespace, agent, token}`) at specialize time, sibling of the state token (same `os.Remove`/0444 dance, same `dev-unauthenticated` placeholder without a master; deliberately NOT the state file — identity must not require `State`). The executor injects `FISSION_AGENT_TOKEN_PATH` + `FISSION_AGENT_RUNTIME_URL` (chart adds `AGENT_RUNTIME_URL` on the executor, fixing the fixtures' hardcoded install-namespace URL). Container executortype: no fetcher → no file (documented boundary).
- **Verification (dispatch route ONLY, fail-closed)**: identity claims headers select the identity path exclusively — no JWT fallback. Constant-time compare against both masters → live-view revocation check (the `requireKnownKeyspace` mirror: a deleted agent's token dies with it) → **inline authorization: claimed namespace must equal the dispatched path namespace** (403 otherwise; the claimed agent names the caller, not the target — same-namespace cross-agent dispatch is the spawn shape). Full truth table unit-tested through the real middleware stack; allowInsecure remains a bytewise pass-through.
- **Boundaries**: per-AGENT, not per-session (Q9 open); tool grants stay server-side from the live view, never in the token; scope composes with the same-namespace parentage rule below this PR — a spawning agent can never reach further than parentage already allows.

## Verification

- Task 3's middleware survived a dedicated adversarial review (timing, oracles, header-predicate abuse, mount scope) with zero blockers/majors.
- Integration proves the injection chain end to end without a secured install: the fixture echoes sha256(token), the test re-derives from `FISSION_INTERNAL_AUTH_SECRET` and asserts equality (placeholder path pinned too). Secured-mode e2e (env-flip serial test) is a named follow-up.
- Final whole-branch review: SHIP (1 minor — a JSON.parse-snippet token-leak-to-logs class in fixture error paths, fixed across all token-file reads).

## Follow-ups (named, not in this PR)
mcp `ctx.tools` scoping (rides tool consumption); mcp/agentruntime authz shared-package extraction; cross-namespace parentage relaxation via `Scope.Allows`; per-session tokens (Q9); secured-mode serial e2e.

CI note: workflows are main-base-scoped; full battery runs when the stack retargets main.

Process: reviewed plan (REVISE→fixed: 1 blocker — the identity path bypasses the JWT middleware's inline namespace check, so the wrapper enforces it itself), 6 SDD tasks all PASS/PASS, adversarial authz review, final SHIP, fix round + re-review (which itself caught an incomplete fix, closed inline).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -29,6 +29,7 @@ linters:
             - "!**/pkg/executor/executortype/poolmgr/gp_pod.go"
             - "!**/pkg/executor/client/client.go"
             - "!**/pkg/executor/executortype/poolmgr/oci_specialize.go"
+            - "!**/pkg/fetcher/agenttoken.go"
             - "!**/pkg/fetcher/config/config.go"
             - "!**/pkg/fetcher/fetcher.go"
             - "!**/pkg/fetcher/statetoken.go"
--- a/charts/fission-all/templates/executor/deployment.yaml
+++ b/charts/fission-all/templates/executor/deployment.yaml
@@ -77,6 +77,13 @@ spec:
         - name: STATESVC_URL
           value: "http://statesvc.{{ .Release.Namespace }}:{{ include "fission.statesvcPort" . }}"
         {{- end }}
+        {{- if .Values.agentRuntime.enabled }}
+        # The executor stamps this URL into agent-identity-bearing function
+        # pods as AGENT_RUNTIME_URL (the dispatch endpoint spawned agents
+        # call home to); unset (feature off) keeps pods byte-identical.
+        - name: AGENT_RUNTIME_URL
+          value: "http://agentruntime.{{ .Release.Namespace }}:{{ include "fission.agentRuntimePort" . }}"
+        {{- end }}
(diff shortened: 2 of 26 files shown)


sanketsudake:

PR #3706, opened 2026-08-25
agentruntime,mcp: OTel GenAI spans — invoke_agent turns, wake-stitched traces, MCP _meta (G15)

sanketsudake:
Stack position 6 (#3701 ← #3703 ← #3704 ← #3705 ← this). Agent-native tracing on infrastructure that already existed — the branch adds spans and context carriage, zero new dependencies (go.mod untouched).

## What

- **`invoke_agent` spans** (GenAI semconv v1.37.0, pre-stable, version-pinned): one span per agent turn in the dispatcher — started before the background-context derivation (load-bearing: that's what lets the wake enqueue inherit it), defer-ended across every return path, `gen_ai.*` + `fission.*` attrs, yield + status on completions.
- **Wake-stitched traces**: the wake envelope carries an optional `trace` field (injected at enqueue, extracted at delivery; lenient decode = no version bump, dedup key untouched, rolling-deploy safe in both directions) — so a whole `yield: continue` self-continuation chain renders as ONE trace, including the post-Ack revival enqueue (threaded from the delivery's context). The stitch test drives yield=continue end-to-end and captures the real enqueued envelope; its non-vacuity was verified by deleting the propagator install and watching four tests fail.
- **Server spans**: agentruntime mux wrapped (filters: `/registry/events`, `/ui`, `/healthz`, `/readyz` — no hour-long SSE spans, no kubelet-probe noise; otel outside auth so 401s get spans, router-public precedent) and the `/mcp` handler.
- **MCP `_meta` trace-context** (2026-07-28 spec, keys verified against the spec text): `traceparent`/`tracestate`/`baggage` extracted from `tools/call` `_meta` into an `execute_tool` span. **Additive-only**: `_meta` absent → byte-identical behavior; malformed values extract to nothing and the call proceeds (untrusted input); the SDK's ctx-detach was verified value-preserving (empirically) so parenting claims are tested against reality.
- **Collectorless propagation proof** (integration): a synthetic traceparent sent on dispatch must reach the function pod with the SAME trace-id and a DIFFERENT span-id (no sampled-flag assertion — NeverSample strips it by design).
- **Docs**: boardroom README observability appendix (span inventory, kind-opentelemetry BYO-collector steps, Langfuse/Phoenix as reference backends).

## Visible behavior note (final-review corrected framing)

Function pods already received a `traceparent` header on every proxied call before this branch (otelhttp injects even parentless). The actual delta: that header's trace-id is now **continuous with the caller's trace**, and caller-supplied `tracestate`/`baggage` now flow through to pods and wake envelopes — standard W3C propagation, HMAC-safe (the canonical signs method+URI+body only).

## Boundaries
- **No meters** — `fission_agent_*` metrics remain the parked G19-export follow-up.
- SDK-side `execute_tool` children from inside function code are the env-SDK track, not this PR.

Process: reviewed plan (REVISE→fixed: 1 blocker — span placement vs the background-context derivation would have silently broken every stitch), 4 SDD tasks all PASS/PASS (one implementer recovered mid-task from an API error), final review SHIP with an empirical framing correction, fix round inline.

### DIFF
--- a/demo/agent-boardroom/README.md
+++ b/demo/agent-boardroom/README.md
@@ -190,6 +190,80 @@ Mapped from the boardroom demo script; check these off live against the
       (`density (sessions:pods): NNx`) climbs as `--sessions` grows for a
       fixed pool size.
 
+## Observability
+
+The agent runtime and `pkg/mcp` emit OTel traces for the boardroom demo's
+request paths — no meters yet (that stays a follow-up; see the observability
+plan's Global Constraints).
+This section is a reference, not a walkthrough:
+the spans exist whether or not a collector is running,
+so nothing here is required to run the demo itself.
+
+### What spans exist
+
+- **`invoke_agent <agent>`** —
+  one span per agent turn, started in `pkg/agentruntime/dispatcher.go`'s `DispatchTurn`,
+  a child of the inbound HTTP request's server span.
+  Attributes: `gen_ai.operation.name=invoke_agent`, `gen_ai.agent.name`,
+  plus `fission.agent.namespace`, `fission.session.id`, `fission.turn`, `fission.turn.source` (`http` or `wake`),
+  and on completion `fission.yield` + a span status (`Error` on transport failure, `Ok` otherwise).
+  **Wake-stitched:** when a turn yields `continue`, the dispatcher injects its span's trace context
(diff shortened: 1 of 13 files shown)


sanketsudake:

PR #3707, opened 2026-08-26
agentruntime,storagesvc: session artifact workspace (G12/G18a)

sanketsudake:
Stack position 7 (#3701 ← #3703 ← #3704 ← #3705 ← #3706 ← this). Session-scoped file storage for agent functions — artifacts too big for the statestore's 256KiB KV cap — with lifecycle tied to session archive.

## Architecture

**agentruntime is the workspace API** (proxy): pods authenticate with the G16 identity bearer, which only agentruntime can verify *with revocation* (the live AgentView); it signs through (master `ServiceStoragesvc` HMAC) to a new internal-only storagesvc surface. A bearer path on storagesvc would have minted a third authz-code copy and honored tokens for dead agents.

- **storagesvc `/v1/workspace`** (PUT/GET/DELETE/list/prefix-delete, raw bodies, streamed, master-signer-only): caller-chosen ids under a reserved `_workspace_` root — **pruner-excluded** (the archive pruner deletes anything matching no Package URL within ~an hour; exclusion is segment-based because local candidate ids are absolute paths and S3 keys may be subDir-prefixed) and **sealed off from the archive routes** (the legacy grandfather would otherwise have served/deleted workspace objects to tenant signers). Backend keys join `STORAGE_S3_SUB_DIR` (shared-bucket isolation); wire ids don't.
- **agentruntime `/workspace/{ns}/{name}/{session}/{path...}`**: identity bearer with **own-workspace** authorization (ns AND agent equality — a deliberate, documented widening of G16's dispatch-only identity surface); JWT callers keep namespace scope. **A live session record is required (404 otherwise) — it is the GC anchor**: the sweeper purges the prefix at archive, so a recordless workspace file would be collected by nothing, ever.
- **Quotas on the session record** (storagesvc has none): 32MiB/artifact (over-read-detected 413 — never a silently truncated store), 256MiB + 1000 artifacts/session (pre-check → write → CAS-settle; delta-accounted via an existence probe; **fail-closed on every PUT-side race** including the shrink race the first fix round introduced and the second removed). One documented fail-open exception: racing same-path DELETEs can double-decrement (self-inflicted, bounded, named in code + README — exact settle would need per-path serialization).
- **Sweeper purge** at archive (between the archive CAS and the state-keyspace early return, so state-less agents purge too); log-don't-retry; orphan residue for TTL'd-out unswept records joins the named orphan-sweep follow-up.
- **NetPol**: storagesvc's allowlist gains the agentruntime pod (final-review blocker — without it, netpol-enabled installs silently drop every workspace call and orphan every archived workspace).
- **Demo + integration**: support-desk artifact beat; put→get sha equality, list, no-session 404 in-cluster; the test skips explicitly on secured installs (cross-agent 403 is unit-owned until the named secured-mode serial e2e).

## Accepted residues (documented in code)
DELETE-side double-decrement (above); concurrent same-new-path phantom `ArtifactCount` (+N-1 bound); probe-failure treated as create; in-flight-PUT-vs-purge orphan; shrink reclaim = DELETE+PUT.

Process: reviewed plan (REVISE: 2 blockers — the objectStore seam misname and the archive-route grandfather hole), 4 SDD tasks (Task 2 adversarially reviewed, 2 fix rounds; Task 4 spec-FAIL closed by the explicit-skip ruling), final review FIX-FIRST (the netpol blocker) → fix round → CLEAN re-review. Three machine-sleep interrupts recovered mid-run.

### DIFF
--- a/charts/fission-all/templates/agentruntime/deployment.yaml
+++ b/charts/fission-all/templates/agentruntime/deployment.yaml
@@ -82,6 +82,22 @@ spec:
           value: {{ .Values.agentRuntime.maxContinuations | quote }}
         - name: AGENT_MAX_SESSIONS_DEFAULT
           value: {{ .Values.agentRuntime.maxSessions | quote }}
+        # Upstream for the session workspace routes (pkg/agentruntime/
+        # workspace.go): mirrors buildermgr's own --storageSvcUrl VALUE
+        # (charts/fission-all/templates/buildermgr/deployment.yaml) as an ENV
+        # instead of a container arg. Unset means the workspace routes answer
+        # 503 "workspace disabled" rather than agentruntime guessing a URL —
+        # here it is always set, since storagesvc is a core, always-deployed
+        # component (no storagesvc.enabled gate exists to condition on).
+        - name: STORAGESVC_URL
+          value: "http://storagesvc.{{ .Release.Namespace }}"
+        # AGENT_MAX_ARTIFACT_BYTES / AGENT_MAX_SESSION_ARTIFACT_BYTES /
+        # AGENT_MAX_SESSION_ARTIFACTS are deliberately left unset here: the
+        # code's own defaults (32MiB per artifact, 256MiB and 1000 artifacts
+        # per session — pkg/agentruntime/main.go's defaultMaxArtifactBytes/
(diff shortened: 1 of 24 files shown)


sanketsudake:

PR #3708, opened 2026-08-26
cli: fission agent create — one-command agent scaffold (G17 in-repo half)

sanketsudake:
Stack position 8 (#3701 ← #3703 ← #3704 ← #3705 ← #3706 ← #3707 ← this). The DX slice: `fission agent create --name X --env node` scaffolds a working stateful agent — handler + Package + Function (Agent+State blocks) — in one command. Cluster → running agent in well under 10 minutes.

## What

- **Embedded handler templates** (node + python), distilled from the turn-loop skeleton + the state read/write essentials, heavily commented as the tutorial: session id from `X-Fission-Session`, turn count from `X-Fission-Agent-Turns`, state load/save via `FISSION_STATE_URL` + the token file (with an in-memory fallback so the scaffold works before statestore is wired), JSON reply, yield `waiting` (never `continue` — a continue default would self-re-invoke forever on a fresh install; the chain semantics live in a commented block). Python contract verified against the env repo docs, not guessed from node.
- **`fission agent create`**: writes the template, then composes `fn create`'s *exact* shipped pipeline (`CreatePackage`/`CreateArchive`, the shared `getStateConfig`/`getAgentConfig` — export-renamed in place, no forked copy) into `--spec` (auto-init if needed, 3-doc append) or direct (clientset) modes; conditional next-steps output carries the <10-min path (env-missing → leads with the `fission env create` line).
- **Contract-drift tripwires** (the slice's discriminating tests): the templates are pinned against *platform sources* — dispatcher header consts (imported test-only), env-var names via `util.StateAPIEnvVars` under `t.Setenv`, `fetcher.StateCredentials` json tags — so a future header/env rename breaks THIS build instead of shipping a silently dead scaffold.
- **Integration**: `ns.WithCWD` → `agent create --spec` → `spec apply` → dispatch a turn → assert the scaffolded handler's `{session, turn}` echo + `yield: waiting`.

## Security note (review-caught)
The first template cut interpolated `--name` unescaped into JS backtick / Python format strings — an adversarial name injected live code into the scaffolded file. Fixed: `Render` validates the name via `fv1.ValidateKubeName` before any interpolation or disk write; every name-derived path uses the validated name. Also fixed (in new code) a latent `ferror.Error`-vs-`StatusError` type-assertion bug inherited from `fn create`'s env-missing path; the identical original is flagged as a follow-up, untouched.

## Deferred
statesvc `/v1/state/<id>` path lacks a drift tripwire (memory-fallback keeps the scaffold working regardless; a `stateapi` path const closes it later); the LangGraph checkpointer + examples gallery are the sibling-repo (`fission/examples`) half of G17, a separate arc. No fission.io quickstart page (release-cut-gated; `--help` + the README pointer for now).

Process: reviewed plan (REVISE: 2 blockers — env-var names are literals not consts so tests assert via the util helpers; `EnvName` collided with `--name`, the right symbol is `FnEnvName`), 3 SDD tasks all PASS (Task 1's injection hole caught by its own review before Task 2 could wire the name through), final SHIP.

### DIFF
--- a/demo/agent-boardroom/README.md
+++ b/demo/agent-boardroom/README.md
@@ -110,6 +110,11 @@ It stays here for now so the whole stack (fixtures, specs, load driver) is self-
    list (`--agents-url`, `--namespace`, `--agent`, `--environment`,
    `--concurrency`, `--timeout`).
 
+**Scaffold your own agent:** `fission agent create --name my-agent --env node` generates a working handler + Package + Function in one command
+(add `--spec` to write it under `specs/` instead of applying directly).
+Run `fission agent create --help` for the full flag reference —
+a fission.io quickstart page lands at release cut.
+
 ## Swarm beat
 
 `architect` (`fixtures/architect.js` / `specs/architect.yaml`) is a second agent function:
--- a/pkg/fission-cli/cmd/agent/command.go
+++ b/pkg/fission-cli/cmd/agent/command.go
@@ -2,12 +2,14 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-// Package agent implements the `fission agent` command group: discover and
-// inspect agent-declared functions (FunctionSpec.Agent != nil), and inspect
-// their sessions via `fission agent sessions` (sessions.go), which talks to
-// the agentruntime's RegistryAPI introspection endpoints (slice 4).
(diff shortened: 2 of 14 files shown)


sanketsudake:

PR #3709, opened 2026-08-26
env: opt-in RuntimeClassName for syscall-isolated pods (G18 tier B)

sanketsudake:
**Off `main`, independent of the agent-runtime PR stack** — this touches only EnvironmentSpec + the executor/buildermgr pod builders + the env CLI, with zero agentruntime dependency, so it merges on its own timeline.

## What

`fission env create --name node --image ghcr.io/fission/node-env --runtime-class gvisor` runs that environment's pods under a syscall-isolating RuntimeClass (gVisor/Kata) — host isolation for code-running agents, with **zero change to the warm-pool density model**.

- **`EnvironmentSpec.RuntimeClassName *string`** (opt-in), applied **fill-if-nil after `MergePodSpec`** at all three env-aware pod-spec sites — poolmgr runtime pools, newdeploy, **and buildermgr builder pods** (build commands are user code too, so isolation is env-wide). A shared `util.ApplyEnvRuntimeClass` keeps the three sites from drifting.
- **`Runtime.PodSpec.runtimeClassName` still wins** if both are set — the full-PodSpec override is the documented escape hatch; a simple knob silently beating a manual override would violate specific-beats-general (the merge layer's own principle).
- **CLI**: `--runtimeclass` (house convention) with `--runtime-class` alias; `--runtime-class ""` on update clears the field (the un-opt path).
- **Validation**: struct-level CEL on EnvironmentSpec (DNS-1123 label, ≤253) mirrored by a Go check in `Validate()` (byte-identical regex, verified) so the CLI rejects bad values pre-flight and kubectl/GitOps writers hit the same rule at admission.
- **Container-executor functions are NOT covered** (no Environment in their pod-spec path) — stated in the field godoc and the flag usage string. The irony: container functions most want isolation; that's Tier C's job (K8s Agent Sandbox, on trigger).

## Why this shape

`runtimeClassName` *already* flowed through today via `env.Spec.Runtime.PodSpec.RuntimeClassName` (MergePodSpec forwards it, CEL denylists permit it) — but you'd have to hand-build a full PodSpec override, with a redundant `Containers[]` entry, to reach one field nobody knows is settable. This PR makes it the discoverable, validated, CLI-flagged, documented knob G18 tier B specified. Propagation being free made it cheaper, not less worth doing.

## Tests / caveat
Controller-level pod-spec tests at all three sites (including the both-set precedence case and non-vacuous field-only cases — the field-only test was mutation-checked to confirm it catches inside-the-guard placement). Integration: a **CEL round-trip** test (valid admits, invalid rejected by the apiserver — the real backstop that the CEL shipped, since there's no CRD-drift CI gate) + a pod-template-reach test. **CI can't assert scheduling** — kind has no gVisor/Kata RuntimeClass, so the tests read the created objects via the API and never wait for a pod to run (that would hang forever); operators need the RuntimeClass object + node runtime handler installed. Docs: `docs/environment-runtime-class.md`.

Process: seam-mapped (found the already-works-but-undiscoverable fact), plan-reviewed (APPROVE + 3 load-bearing notes folded: unconditional fill-if-nil or the common field-only case silently drops; CEL must be struct-level; verify the CRD regenerated), 3 SDD tasks all PASS, final review SHIP.


sanketsudake:

PR #3711, opened 2026-08-28
cli: fission agent create --template interpreter — exec-verb scaffold

sanketsudake:
## What

A blessed interpreter scaffold: `fission agent create --name pyexec --env python --template interpreter` generates an agent handler implementing the exec verb — `{"code"|"command", "timeoutSeconds"}` in, `{stdout, stderr, exitCode, durationMs, truncated}` out — running through the normal agent dispatch path. Zero new platform Go, zero new images: the interpreter is handler logic shipped exactly like the existing scaffold templates.

## Contract

- Exactly one of `code` (run via the language runtime) or `command` (via `/bin/sh -c`).
- `timeoutSeconds` default 60, ceiling 300; on expiry the WHOLE process group is killed (both templates verified against the backgrounded-grandchild case: `sh -c 'daemon &'` cannot hold the turn open past the deadline).
- stdout/stderr each capped at 256 KiB (pinned to `fv1.DefaultStateMaxValueBytes`, not a re-typed literal); overflow sets `truncated: true`, never errors.
- Subprocess env is an allow-list (PATH/HOME/LANG/LC_ALL/TZ) — `FISSION_*` vars never pass through.
- `--expose-as-mcp` + `--tool-*` flags (newly registered on `agent create`; `function.GetToolConfig` exported in place) set `FunctionSpec.Tool` so the interpreter doubles as an MCP `code_exec` tool; a tools/call without a session header is stateless exec.

## Trust boundary (documented in the scaffold's next-steps output)

Env-scrubbing prevents accidental credential leakage; it is not a sandbox — exec'd code runs at the agent's own trust level (same container, token files readable). Right for run-your-own-code tools; adversarial multi-tenant code needs per-session pods (future executor work). The scaffold recommends pairing with `fission env create ... --runtime-class gvisor`.

## Tests

- Template tripwires anchored on wire forms (request-key reads + response-literal keys), header/state-header constants, timeout ceiling, env-allow-list, non-finite timeout guards, name-injection rejection (exploit-shaped cases), FunctionTimeout↔template-ceiling pin.
- Integration (`agent_exec_test.go`): basic exec, timeout group-kill with exitCode -1 + stderr note, 256 KiB truncation, env-scrub — first live run happens in this PR's CI.

## Notes for reviewers

- The templates assume `/bin/sh`, `sleep`, `echo` exist in the runtime image (same assumption as existing shell fixtures).
- Follow-up (not in this PR): a python-lang integration leg skip-gated on `PYTHON_RUNTIME_IMAGE` — the python template's timeout path is currently pinned by direct verification and static tripwires; a live leg would pin it in CI.
- Follow-up: per-turn unique subdir under the session scratch dir to remove a concurrent same-session direct-invocation race (robustness only).

### DIFF
--- a/pkg/fission-cli/cmd/agent/command.go
+++ b/pkg/fission-cli/cmd/agent/command.go
@@ -49,13 +49,19 @@ func Commands() *cobra.Command {
 	}, Create, flag.FlagSet{
 		Required: []flag.Flag{flag.FnName},
 		Optional: []flag.Flag{
-			flag.FnEnvName, flag.FnAgentLang, flag.PkgCode,
+			flag.FnEnvName, flag.FnAgentLang, flag.FnAgentTemplate, flag.PkgCode,
 			agentCreateStateFlag, flag.FnStateKeyspace, flag.FnStateMaxKeys,
 			flag.FnStateMaxValueBytes, flag.FnStateTTL,
 			flag.FnStateStickySource, flag.FnStateStickyName,
 			agentCreateAgentFlag, flag.FnAgentSessionSource, flag.FnAgentSessionName,
 			flag.FnAgentIdleAfter, flag.FnAgentArchiveAfter, flag.FnAgentMaxSessions,
 			flag.FnAgentHistoryTrim,
+			// MCP exposure (PR-1): --expose-as-mcp lets the scaffolded handler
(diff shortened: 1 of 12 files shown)


sanketsudake:

PR #3712, opened 2026-08-28
templates: session workspace hydration for scaffolded agents

sanketsudake:
## What

Stacked on #3711. The scaffolded templates gain a hydrate/sync helper against the session artifact workspace: at turn start the exec scratch dir is lazily hydrated from the session's workspace (one LIST, downloads only for missing/size-mismatched files); at turn end new/changed files are PUT back (size+mtime compare, content hash only on ambiguity, manifest-keyed so a warm pod re-syncs nothing unchanged). Template/SDK-side only — no new platform Go, no new AgentConfig fields.

This is what makes a file exec'd code writes survive a pod respecialize: pod-local scratch dies with the pod; the workspace copy rehydrates onto whichever pod serves the next turn.

## Semantics

- Additive-only: local deletion never DELETEs from the workspace (explicit app call stays the delete path) — a tripwire asserts no DELETE in the helper block.
- Failures are never silent: hydrate/sync errors and quota 413s land in a capped `workspaceWarnings` reply field.
- Auth: the pod's agent-identity credential — bearer token plus both claim headers — pinned by tripwires against the platform constants.
- Sync walk skips symlinks (a planted symlink at the pod's token file would otherwise upload its target's bytes); hydrate writes are guarded the same way.
- The helper block is byte-identical between interpreter.*.tmpl (wired in) and agent.*.tmpl (exported, opt-in call only) — asserted by test.

## Security note

The identity token's boundary is per-(namespace, agent): a session can read sibling-session artifacts of the same agent. Hydration is per-session by convention, not enforcement; per-session tokens remain a tracked follow-up.

## Tests

Tripwires for route shape (list URL has no trailing slash — the shipped route 400s one), identity header names, warnings key, additive-only, helper-block identity, opt-in-only wiring. Integration: write via exec → kill every specialized pod → next turn on a different pod reads identical content (hash-equal) — first live run in CI.

## Notes for reviewers

- Sessionful scratch dirs are no longer wiped between turns by design (skip-clean needs the manifest); a warm pod accumulates scratch dirs until it dies — follow-up: idle-scratch reaper.
- When `FISSION_AGENT_RUNTIME_URL` is unset (workspace feature not deployed) the helper degrades to no-remote-backing; scratch still persists for the pod's lifetime.

### DIFF
--- a/pkg/fission-cli/cmd/agent/templates/agent.js.tmpl
+++ b/pkg/fission-cli/cmd/agent/templates/agent.js.tmpl
@@ -15,7 +15,9 @@
 // contract it implements -- then replace the TODO in the middle with real
 // agent logic (call an LLM, run a tool, whatever your agent does).
 
+const crypto = require('crypto');
 const fs = require('fs');
+const path = require('path');
 
 // The dispatcher stamps two request headers on every forwarded call (see
 // pkg/agentruntime/dispatcher.go), and reads one response header back:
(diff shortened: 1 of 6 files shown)


sanketsudake:

PR #3713, opened 2026-08-28
chart: per-environment egress NetworkPolicy knob

sanketsudake:
## What

Chart-only: `environmentNetworkPolicy.{enabled, allowClusterDNS, extraEgress[]}` (default off) templates an egress NetworkPolicy selecting environment pods: internet egress allowed, RFC1918 / link-local / metadata ranges denied, cluster DNS deniable. This is the secure-by-default posture agent workloads want without any Go changes.

## Notes for reviewers

- Enabling it also blocks in-cluster control-plane flows that cross RFC1918 space from the same pods: fetcher→storagesvc (cold-start package fetch), function pod→statesvc, agent pod→agentruntime (workspace sync). The required `extraEgress` allow-list entries are documented in values.yaml and the template comment rather than carved out unilaterally — whether a default carve-out for those three flows belongs in the chart is a deliberate rollout-time call.
- Known edge: with `defaultNamespace` set to a non-default value plus `functionNamespace`, the template's namespace mapping diverges from `GetFunctionNS` (which maps only the literal `default`); noted in the template comment, follow-up if that configuration is ever exercised.

## Tests

helm lint green; 8 render-assertion subtests in `test/helm/environmentnetworkpolicy_test.go` (on/off gating, CIDR set, DNS toggle, extraEgress passthrough, namespace scoping) alongside the existing NetworkPolicy render tests.

### DIFF
--- a/charts/fission-all/templates/executor/networkpolicy-environment-egress.yaml
+++ b/charts/fission-all/templates/executor/networkpolicy-environment-egress.yaml
@@ -21,11 +21,21 @@
 # environmentNetworkPolicy.allowClusterDNS=false drops that rule for a
 # stricter numeric-IP-only posture.
 #
-# Rendered into the namespace where executor creates function pods
-# (`fission-function-ns` helper), same as functionpods-allow-internal
-# above — NetworkPolicy podSelector only matches pods in the policy's own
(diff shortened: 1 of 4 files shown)


sanketsudake:

Issue #3183, opened 2025-03-05
Container Executor Scale to Zero, Error with wrong timing


GilbertBckr:
**Fission/Kubernetes version**

<pre>
$ fission version
client:
  fission/core:
    BuildDate: "2025-01-25T15:57:43Z"
    GitCommit: 2853498a
    Version: v1.21.0
server:
  fission/core:
    BuildDate: "2025-01-25T15:57:43Z"
    GitCommit: 2853498a
    Version: v1.21.0
$ kubectl version
Client Version: v1.32.1
Kustomize Version: v5.5.0
Server Version: v1.32.0
</pre>

**Kubernetes platform (e.g. Google Kubernetes Engine)**
Kind

**Describe the bug**

Using the Container as Function execution mode and scaling to zero there is a problem when requesting the function while it is shutting down.
Depending on the timing of the request there is either a 500 error
<pre>
$ curl http://localhost:8000/main-function -i
HTTP/1.1 500 Internal Server Error
Date: Wed, 05 Mar 2025 12:32:55 GMT
Content-Length: 33
Content-Type: text/plain; charset=utf-8
</pre>
or the function just takes really long to respond about 27 seconds compared to usually around 2 seconds on cold starts.

**To Reproduce**
Function:
```yaml
apiVersion: fission.io/v1
kind: Function
metadata:
  generation: 1
  name: main-function
  namespace: default
  resourceVersion: "748"
  uid: 3efcbda3-be15-4f18-ac4b-3ae35754a84b
spec:
  InvokeStrategy:
    ExecutionStrategy:
      ExecutorType: container
      MaxScale: 1
      MinScale: 0
      SpecializationTimeout: 12
      TargetCPUPercent: 0
    StrategyType: execution
  environment:
    name: ""
    namespace: ""
  functionTimeout: 60
  idletimeout: 15
  package:
    packageref:
      name: ""
      namespace: ""
  podspec:
    containers:
    - image: nginx:latest
      name: main-function
      ports:
      - containerPort: 80
        name: http-env
        protocol: TCP
      resources: {}
    imagePullSecrets:
    - name: ""
    terminationGracePeriodSeconds: 0
  resources: {}
```

Trigger:
```yaml
apiVersion: fission.io/v1
kind: HTTPTrigger
metadata:
  generation: 1
  name: main-http-route
  namespace: default
  resourceVersion: "749"
  uid: ae08d08c-8faa-48d2-887c-f389e138cd8e
spec:
  createingress: false
[...]

PR #3725, opened 2026-09-13
container: wait for the replica after scaling a function in from zero

dualfroz:
After a scale-up from zero, readiness was decided from the pre-scale status, which can still count the pod the idle reaper is terminating, so the router got a Service with no endpoints and returned 500.
A scale-up we issue now always waits for the replica. This explains the 500, not the slower cold starts also mentioned in the issue.

Refs #3183


sanketsudake:
@mergifyio rebase

